PR #3, opened 2016-05-19
minimal pip install support

Volodymyrk:
If we want to make mcmcplotlib an (optional) dependancy for pymc3, we need to make it auto-installable with pymc3. 
Refactoring and testing are probably needed to be added, before making it a full dependancy.

As a first step, we can put a reference to:
 `pip install git+git://github.com/mcmcplotlib/mcmcplotlib.git` to pymc3 and start adding reference to mcmcplotlib in examples in pymc3.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# IDE configs
+.idea/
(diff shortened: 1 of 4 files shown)


twiecki:
Thanks @Volodymyrk, this is very useful. As for making a PyMC3 dependency, we'll hold off until other packages have adopted `mcmcplotlib` to make sure we don't add a dependency unnecessarily. So all future plotting code should still go into pymc3.

PR #13, opened 2018-03-06
begin pymc3-independence process

aloctavodia:
This is a clean (re)start. This PR removes all the old code and (re)add functions (hpd, r2_score) and plots (kdeplot, densityplot) that works without PyMC3.

In future PRs I will keep adding plots and functions migrated from PyMC3, once this process is finished I will start adding new functions not present in PyMC3.


aloctavodia:
@twiecki maybe you want to take a look at this before merging. What do you think if during this first stage I directly merge PRs?

twiecki:
@aloctavodia Thanks for taking that on, I think that's exactly the right next step. And definitely feel free to move fast on this.

PR #58, opened 2018-04-22
add arviz-darkgrid style based on seaborn-darkgrid

aloctavodia:


setup.py
```diff
@@ -1,15 +1,35 @@
...
+
+class InstallStyles(install):
+    def run(self):
+        import sys
+        import shutil
+        import os
+        from matplotlib import get_configdir
+
+        sd = os.path.join(get_configdir(), "stylelib")
+        lsd = './arviz/plots/styles'
```
ColCarroll: this looks really cool. 

 do you know if you need to do something like 

```
here = os.path.dirname(os.path.realpath(__file__))
lsd = os.path.join(here, 'arviz', 'plots', 'styles')
```

to supply an absolute path instead of a relative one?

aloctavodia: you are right! otherwise it will work only from the same directory where setup.py is located 

BTW this works for me with _python setup.py install_, but not with pip (the post-installation code is not executed). I found someone had a similar problem, his solution was to install it from github, with it is really weird to me, but I will try, just in case.

aloctavodia: OK magic does not exist. I just installed from my branch and it didn't work.

PR #61, opened 2018-04-25
Energyplot: small aesthetic change

aloctavodia:


arviz/plots/energyplot.py
```diff
@@ -41,8 +44,11 @@ def energyplot(trace, kind='kde', figsize=None, legend=True, shade=0.35, bw=4.5,
 
     energy = get_stats(trace, 'energy')
 
-    series = [('Marginal energy distribution', energy - energy.mean()),
-              ('Energy transition distribution', np.diff(energy))]
+    if color_shade is None:
+        color_shade = ('C0', 'C7')
```
ColCarroll: these can be default arguments so long as they are tuples, instead of using `None`

aloctavodia: Thank you. I was unsure about using tuples or None. I was not sure which option was considered more Pythonic.

PR #63, opened 2018-04-27
Pairplot for arviz

agustinaarroyuelo:
I did the required changes for porting PyMC3's pairplot into Arviz.

### DIFF
--- a/arviz/plots/pairplot.py
+++ b/arviz/plots/pairplot.py
@@ -1,0 +1,120 @@
+import warnings
+import matplotlib.pyplot as plt
+from matplotlib import gridspec
+from ..utils.utils import scale_text, trace_to_dataframe, get_stats
+
+
+def pairplot(trace, varnames=None, figsize=None, text_size=None,
+             gs=None, ax=None, hexbin=False, divergences=False, kwargs_divergences=None,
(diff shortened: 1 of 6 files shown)


arviz/utils/utils.py
```diff
@@ -182,3 +182,15 @@ def _create_flat_names(varname, shape):
     labels = (np.ravel(xs).tolist() for xs in np.indices(shape))
     labels = (map(str, xs) for xs in labels)
     return ['{}__{}'.format(varname, '_'.join(idxs)) for idxs in zip(*labels)]
+
```
aloctavodia: `scale_text` is already used by  `posteriorplot`. Given that you also need it for `pairplot` I suggest to move  `scale_text` inside `pltos/plot_utils.py` and from there import it into `posteriorplot`  and `pairplot`.

arviz/plots/pairplot.py
```diff
@@ -0,0 +1,121 @@
+import warnings
```
aloctavodia: You are not using this import

aloctavodia:
Thanks @agustinaarroyuelo!!!

agustinaarroyuelo:
You're welcome!

PR #69, opened 2018-05-01
add jointplot

aloctavodia:
New jointplot, it could be implemented as part of pairplot, with a _marginals_ arguments, but for now I am adding it as a standalone plot.  

```python
az.jointplot(trace, varnames=['e', 'd'])
```
scatter

```python
az.jointplot(trace, varnames=['e', 'd'], hexbin=True)
```
hexbin

This PR also adds:

* kdeplot: new `rotated` argument
* gridsize is now computed as len(trace)**0.35, instead of 1% of samples
* pairplot: small fix, use `get_varnames` to handle varnames

### DIFF
--- a/arviz/plots/jointplot.py
+++ b/arviz/plots/jointplot.py
@@ -1,0 +1,108 @@
+import numpy as np
+import matplotlib.pyplot as plt
+from matplotlib.ticker import NullFormatter
+from .kdeplot import kdeplot
+from ..utils import trace_to_dataframe
+from .plot_utils import _scale_text
+
+
+def jointplot(trace, varnames=None, figsize=None, text_size=None, hexbin=False, gridsize='auto', skip_first=0,
+              joint_kwargs=None, marginal_kwargs=None):
+    """
+    Plot a scatter or hexbin of two variables with their respective marginals distributions.
+
+    Parameters
+    ----------
+
+    trace : Pandas DataFrame or PyMC3 trace
+        Posterior samples
+    varnames : list of variable names
+        Variables to be plotted, if None all variable are plotted
+    figsize : figure size tuple
+        If None, size is (8, 8)
+    text_size: int
+        Text size for labels
+    hexbin : Boolean
+        If True draws an hexbin plot
+    gridsize : int or (int, int), optional.
+        Only works when hexbin is True.
+        The number of hexagons in the x-direction. The corresponding number of hexagons in the
+        y-direction is chosen such that the hexagons are approximately regular.
+        Alternatively, gridsize can be a tuple with two elements specifying the number of hexagons
+        in the x-direction and the y-direction.
+    skip_first : int
+        Number of first samples not shown in plots (burn-in)
+    """
+    trace = trace_to_dataframe(trace, combined=True)[skip_first:]
+
+    if figsize is None:
+        figsize = (8, 8)
+
+    if text_size is None:
+        text_size = _scale_text(figsize, text_size=text_size)
+
+    if len(varnames) > 2:
+        raise Exception('Number of variables to be plotted must 2')
+
+    if joint_kwargs is None:
+        joint_kwargs = {}
+
+    if marginal_kwargs is None:
+        marginal_kwargs = {}
+
+    plt.figure(figsize=figsize)
+
+    axjoin, axHistx, axHisty = _define_axes()
+
+    x_var_name = varnames[0]
+    y_var_name = varnames[1]
+
+    x = trace[x_var_name].values
+    y = trace[y_var_name].values
+
+    axjoin.set_xlabel(x_var_name)
+    axjoin.set_ylabel(y_var_name)
+
+    if hexbin:
+        if gridsize == 'auto':
+            gridsize = int(len(trace)**0.35)
+            print(gridsize)
+        axjoin.hexbin(x, y, mincnt=1, gridsize=gridsize, **joint_kwargs)
+        axjoin.grid(False)
+    else:
+        axjoin.scatter(x, y, **joint_kwargs)
+
+    if x.dtype.kind == 'i':
+        bins = range(x.min(), x.max() + 2)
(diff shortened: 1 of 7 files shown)


arviz/plots/jointplot.py
```diff
@@ -0,0 +1,108 @@
...
+        if gridsize == 'auto':
+            gridsize = int(len(trace)**0.35)
+            print(gridsize)
+        axjoin.hexbin(x, y, mincnt=1, gridsize=gridsize, **joint_kwargs)
+        axjoin.grid(False)
+    else:
+        axjoin.scatter(x, y, **joint_kwargs)
+
+    if x.dtype.kind == 'i':
+        bins = range(x.min(), x.max() + 2)
```
ColCarroll: `x.min() - 1` to be symmetric?  

it might be neat to also automatically do, say, `np.linspace(x.min() - 2, x.max() + 3, 100, dtype=int)` if `x.max() - x.min() > 100`.

arviz/plots/jointplot.py
```diff
@@ -0,0 +1,108 @@
...
+    else:
+        kdeplot(x, ax=axHistx, **marginal_kwargs)
+    if y.dtype.kind == 'i':
+        bins = range(y.min(), y.max() + 2)
+        axHisty.hist(y, bins=bins, orientation='horizontal', **marginal_kwargs)
+    else:
+        kdeplot(y, ax=axHisty, rotated=True, **marginal_kwargs)
+
+    axHistx.set_xlim(axjoin.get_xlim())
+    axHisty.set_ylim(axjoin.get_ylim())
```
ColCarroll: maybe return the three axes?

arviz/plots/jointplot.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    trace = trace_to_dataframe(trace, combined=True)[skip_first:]
+
+    if figsize is None:
+        figsize = (8, 8)
+
+    if text_size is None:
+        text_size = _scale_text(figsize, text_size=text_size)
+
+    if len(varnames) > 2:
```
ColCarroll: `if len(varnames) != 2:`

arviz/plots/jointplot.py
```diff
@@ -0,0 +1,108 @@
...
+        The number of hexagons in the x-direction. The corresponding number of hexagons in the
+        y-direction is chosen such that the hexagons are approximately regular.
+        Alternatively, gridsize can be a tuple with two elements specifying the number of hexagons
+        in the x-direction and the y-direction.
+    skip_first : int
+        Number of first samples not shown in plots (burn-in)
+    """
+    trace = trace_to_dataframe(trace, combined=True)[skip_first:]
+
+    if figsize is None:
```
ColCarroll: tuples can be used as default arguments.  also, thank you for allowing `figsize` as an argument!

arviz/plots/jointplot.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    Plot a scatter or hexbin of two variables with their respective marginals distributions.
+
+    Parameters
+    ----------
+
+    trace : Pandas DataFrame or PyMC3 trace
+        Posterior samples
+    varnames : list of variable names
+        Variables to be plotted, if None all variable are plotted
```
ColCarroll: this should be updated to specify that two variables are required.

ahartikainen:
Can we have option for 2D kde and/or contours? (I did not watch source, so sorry if those are already implemented)

arviz/plots/jointplot.py
aloctavodia: using `None` is consistent with other plots. For some plots figsize is not a predefined tuple of integers, but is computed taking into account the number of variables.

arviz/plots/jointplot.py
aloctavodia: This will give as many bins as unique integers there are in x, which seems to be the  _natural_ number of bins for a histogram, but you are right  having more than 100 _bars_ does not makes too much sense. Using linspace with integers could give weird histograms, because the numbers are not necessarily even spaced. Anyway, I took your idea and create a `get_bins()`

aloctavodia:
thanks @ColCarroll for all the feedback! 

@ahartikainen yes we can have 2D kde, but probably we should first implement a 2D version of `fast_kde`. One of the differences between `fast_kde` and scipy's kde, is that the former takes into account the boundaries of distributions. I will try this in a separated PR, if this is OK with you.

ahartikainen:
Sounds good. fast_kde in 2D has some quirks if one wants to set custom limits (outside the data range) so a different PR is a good idea.

ahartikainen:
Also, this is OT but is fast_kde's scott's method overestimating it due to digitize command? Have you tested this? (I think scott's method underestimates bandwidth for unimodal distributions)

aloctavodia:
`fast_kde` uses a modified version on Scott's rule. Empirically works pretty well (just my opinion), although I would like a more formal proof. Next week I have a group meeting, and I will be presenting ArviZ and probably I will be briefly talking about `fast_kde` and how to improve it. This is a figure I am preparing.

kde_entropy

Blue curves are theoretical and orange the estimated ones using ArviZ/`fast_kde` (left column) or Scipy (right column).

ahartikainen:
Looks good. You could also try multimodal with different widths.

What is the result if you increase the bw value, e.g. Silverman / scott's rule * 1.33

aloctavodia:
The third row is bimodal with different width (but probably too similar), I should also try other distributions. In the current implementation there is an argument the user can change to control the bandwidth, by default `bw=4.5` the larger this value the smoother the curve. I estimated the value of 4.5 empirically/numerically, but visually I think a lower value (between 3 and 4) could work better (except for the uniform distribution that gets too wiggly, and such lower values are not good idea when the size of the data is _too low_). I would like too put all this on a firm ground and also carry more experiments, I think is an interesting problem (I still need to read more about it).

PR #77, opened 2018-05-11
Additional diagnostics for loo

agustinaarroyuelo:
This PR adds into Arviz the features discussed in PyMC3 [issue](https://github.com/pymc-devs/pymc3/issues/2930) #2930.

### DIFF
--- a/arviz/plots/khatplot.py
+++ b/arviz/plots/khatplot.py
@@ -46,6 +46,6 @@ def khatplot(ks, figsize=None, textsize=None, ax=None, hlines_kwargs=None, **kwa
     rgba_c[:, 3] = alphas
     ax.scatter(np.arange(len(ks)), ks, c=rgba_c, marker='+', linewidth=ms)
     ax.set_xlabel('Data point', fontsize=textsize)
-    ax.set_ylabel(r'Shape parameter $\kappa$', fontsize=textsize)
+    ax.set_ylabel(r'Shape parameter κ', fontsize=textsize)
     ax.tick_params(labelsize=textsize)
     return ax
--- a/arviz/plots/khatplot.py
+++ b/arviz/plots/khatplot.py
@@ -12,7 +12,7 @@ def khatplot(ks, figsize=None, textsize=None, ax=None, hlines_kwargs=None, **kwa
     ks : array
       Paretto tail indices.
     figsize : figure size tuple
-      If None, size is (10, 6)
+      If None, size is (8, 4)
     textsize: int
       Text size for labels. If None it will be autoscaled based on figsize.
     ax: axes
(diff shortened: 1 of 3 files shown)


arviz/stats/diagnostics.py
```diff
@@ -298,3 +298,35 @@ def _get_geweke(x, first=.1, last=.5, intervals=20):
...
+    df_k = (pd.DataFrame(dict(_=['(good)', '(ok)', '(bad)', '(very bad)'],
+                              Count=kcounts,
+                              Pct=kprop))
+            .rename(index={0: '(-Inf, 0.5]',
+                           1: ' (0.5, 0.7]',
+                           2: '   (0.7, 1]',
+                           3: '   (1, Inf)'}))
+
+    if np.sum(kcounts[1:]) == 0:
+        print("All Pareto k estimates are good (k < 0.5)")
```
junpenglao: Should this be a info instead of print? 
Maybe it is good to implement the same logging system similar to what we have in PyMC3?

arviz/plots/khatplot.py
```diff
@@ -0,0 +1,51 @@
...
+    ax.hlines([0, .5, .7, 1], xmin=-1, xmax=len(ks)+1,
+              alpha=.25, **hlines_kwargs)
+
+    alphas = .5 + .5*(ks > .5)
+    rgba_c = np.zeros((len(ks), 4))
+    rgba_c[:, 2] = .8
+    rgba_c[:, 3] = alphas
+    ax.scatter(np.arange(len(ks)), ks, c=rgba_c, marker='+', linewidth=ms)
+    ax.set_xlabel('Data point', fontsize=textsize)
+    ax.set_ylabel(r'Shape parameter $\kappa$', fontsize=textsize)
```
junpenglao: is the latex label here always work?

aloctavodia: This need a working Latex installation on the user machine.

arviz/stats/diagnostics.py
aloctavodia: A warning may do the trick, at least for this PR, right? We can think of a logging system _a la PyMC3_ in the future.

arviz/plots/khatplot.py
junpenglao: Is the latex label style used in anywhere else? otherwise, it is better to remove it.

arviz/stats/diagnostics.py
junpenglao: I agree.

arviz/plots/khatplot.py
aloctavodia: probably not, removing it is the safest thing to do

aloctavodia: ohh wait Arviz only works for Python 3.X thus we can use the symbol kappa, no need to use latex

arviz/stats/diagnostics.py
ahartikainen: We are using logging also in PyStan. Could we somehow give a handle for the current log?

junpenglao: I _love_ using logging

PR #79, opened 2018-05-13
Add linting framework

ColCarroll:
Trying to help tidy up for a release.  This adds a pylint file to help enforce code style.  I fixed all the warnings and errors, which turned up a few places where arguments weren't getting passed to internal functions).  

If you think this is helpful, I am happy to finish linting, and add the linter to CI.  I tried to err on the side of leaving lots of warnings enabled, which led to a few inline annotations (`# pylint disable=...`).  Also happy to disable or enable more if you'd like!  

The remaining suggested changes are [at this gist](https://gist.github.com/ColCarroll/54ed8bdd7e47d0d383e41a226ae2a6ce).  Most are because of 1 or 2 letter variable names.

### DIFF
--- a/arviz/plots/energyplot.py
+++ b/arviz/plots/energyplot.py
@@ -5,8 +5,8 @@ from ..stats import bfmi as e_bfmi
 from ..utils import get_stats
 
 
-def energyplot(trace, kind='kde', bfmi=True, figsize=None, legend=True, shade=(1, .75),
-               color_shade=('C0', 'C5'), bw=4.5, skip_first=0, kwargs_shade=None, ax=None,
+def energyplot(trace, kind='kde', bfmi=True, figsize=None, legend=True, fill_alpha=(1, .75),
+               fill_color=('C0', 'C5'), bw=4.5, skip_first=0, kwargs_shade=None, ax=None,
                **kwargs):
     """Plot energy transition distribution and marginal energy distribution in
     order to diagnose poor exploration by HMC algorithms.
@@ -23,10 +23,10 @@ def energyplot(trace, kind='kde', bfmi=True, figsize=None, legend=True, shade=(1
         If None, size is (8 x 6)
     legend : bool
         Flag for plotting legend (defaults to True)
-    shade : tuple of floats
+    fill_alpha : tuple of floats
         Alpha blending value for the shaded area under the curve, between 0
         (no shade) and 1 (opaque). Defaults to (1, .75)
-    color_shade : tuple of valid matplotlib color
+    fill_color : tuple of valid matplotlib color
         Color for Marginal energy distribution and Energy transition distribution.
         Defaults to ('C0', 'C5')
     bw : float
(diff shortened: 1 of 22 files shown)


aloctavodia:
This is really helpful. Thanks. It will be great if you could add linter to CI.

ColCarroll:
These changes are getting pretty big, and I am starting to touch more of the code logic.  Most remaining lintings are variable names (notice I added a bunch more "allowed" 1 and 2 letter variables like `ax`, `mu`, `tr` (for trace)) 

I would suggest merging this, and I can see if I can rename more variables during the next week so that linting passes cleanly.

arviz/plots/energyplot.py
```diff
@@ -5,8 +5,8 @@
 from ..utils import get_stats
 
 
-def energyplot(trace, kind='kde', bfmi=True, figsize=None, legend=True, shade=(1, .75),
-               color_shade=('C0', 'C5'), bw=4.5, skip_first=0, kwargs_shade=None, ax=None,
+def energyplot(trace, kind='kde', bfmi=True, figsize=None, legend=True, fill_alpha=(1, .75),
```
aloctavodia: Not sure about using alpha, because for some functions alpha is related to credible intervals. Maybe we should give preference to the matplotlib...

arviz/stats/stats.py
```diff
@@ -538,7 +540,7 @@ def r2_score(y_true, y_pred, round_to=2):
 
     r2 = var_y_est / (var_y_est + var_e)
     return pd.Series([np.median(r2), np.mean(r2), np.std(r2)],
-                     index=['r2_median', 'r2_mean', 'r2_std'])
+                     index=['r2_median', 'r2_mean', 'r2_std']).round(decimals=round_to)
 
 
 def summary(trace, varnames=None, round_to=2, transform=lambda x: x, circ_varnames=None,
```
aloctavodia: I removed the transform argument from almost all functions. I missed this one.

arviz/tests/test_diagnostics.py
```diff
@@ -1,48 +1,46 @@
...
-    assert eff_n.shape == (3,)
+from ..stats import gelman_rubin, effective_n, geweke
 
 
-def test_geweke():
-    trace = fake_trace(1000)
-    gw = geweke(trace)
-    assert max(abs(gw['a'][:, 1])) < 1
-    assert max(abs(gw['a'][:, 0])) > -1
+class TestDiagnostics(object):
```
aloctavodia: What is the advantage of using classes for tests?

arviz/plots/energyplot.py
ColCarroll: That's a good point, but maybe giving preference to matplotlib is right here:  in a few of these functions, passing in `alpha` will go to the `**kwargs`, and then get passed to a matplotlib function (in particular, I had renamed this to `alpha` before, and got an error for reusing an argument).

arviz/tests/test_diagnostics.py
ColCarroll: I did this mostly to:

1. stop code from running in the global namespace:  all that code actually runs while `pytest` is _discovering_ tests, so `pytest -x` would run all the stuff outside of functions, then start running the tests and stop on the first failure.  Now it will start running the tests, only doing setup it needs to do.
2. create more granular tests: so failures would be easier to spot
3. convention: it is convention to have `test_foo.py` to test `foo.py`, and then `class TestBar` to test the `Bar` class in `foo.py`.  This makes finding where a function/method/class/module is easier.

PR #81, opened 2018-05-16
Add more linting

ColCarroll:
This leaves only `stats.stats` and `stats.diagnostics` to be linted before I can stop making these nitpicky changes.  

A few conditions I turned off can be useful when writing maintainable functions (`too-many-locals`, `too-many-statements`, `too-many-branches`), but it seemed like too big a change to take those on.


aloctavodia:
I vote for you keep doing  _nitpicky_ changes. Thanks for helping with this.

PR #82, opened 2018-05-17
Finish linting, add pylint to travis

ColCarroll:
Linting should be enabled on travis now.  I would guess the most offensive change here is `Rhat -> r_hat`.  Let me know if you'd rather add that to the whitelist.


aloctavodia:
Great! Thanks a lot!

PR #89, opened 2018-05-20
Add optional dependency class

ColCarroll:
This is a way to use methods from `pymc3` or `altair` when it is clear the user wishes to use such a method.  For example, if the user passes a `pymc3` trace, I think it would be fair to use a method from `arviz.compat.pymc3`.  

I also plan to use this to implement some methods using `arviz`.

image

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -61,7 +61,8 @@ disable=missing-docstring,
         too-many-locals,
         too-many-branches,
         too-many-statements,
-        no-self-use
+        no-self-use,
+        too-few-public-methods
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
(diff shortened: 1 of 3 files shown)


ColCarroll:
also just turned off coveralls comments.  can re enable if you find them useful.  there are no tests here because it is quite difficult to test, but I can add those if you'd like (I think pymc3 enforces that matplotlib is not required).

aloctavodia:
I like the coveralls comments turned off and I am OK about skipping the tests. I don't really understand what this new class is for, but I guess I will when you use it for something concrete, so fill free to merge this.

ColCarroll:
Sorry for not explaining better! 

My immediate use case is to implement something like `traceplot_altair`, as well as to get rid of some `pymc3` specific code: it seems like `arviz.trace_to_dataframe(trace)` should call `pymc3.trace_to_dataframe` if `trace` is a multitrace from pymc3.

PR #98, opened 2018-05-24
Update docs

ColCarroll:
Did a pretty complete rebuild of the docs, and wanted earlier input.  I added examples for 5 of the plots, and plan to get the rest later.  I used most of the styling and configuration from https://seaborn.pydata.org/ .  

I decided to not explain any of the plots yet, since pointing to PyMC3/Stan/other resources might be better (and I think you/others would do a better job).

I am not sure if travis builds docs for PRs, and this is a big change, so I `rsync`'d the docs to my own server if you'd like to look:

https://colindcarroll.com/static/arviz/index.html


ColCarroll:
This will take a little more work to get `arviz` into the python path.  Probably the best thing is to do `pip install .` in `.travis.yml`, which both tests the `setup.py`, and actually adds the library to the path.

aloctavodia:
I was talking about this with @agustinaarroyuelo and she mention something very important. The thumbnails/previews show just a part of the plot and in general this is not very clear/informative (especially for new users). As you mention this is the same style used by seaborn. An example of a different style is [Matplotlib](https://matplotlib.org/gallery.html) gallery, where the thumbnails/previews show the full plot. What do you think?

ColCarroll:
Great suggestions!  I will take a swing at all of them:

- logo/text should be easy
- darkgrid is a great idea and should be easy
- header color might be hard -- i'll look around, but might punt to an issue
- brief explanation I'll punt to others who understand this better
- full preview should be easy.  I'll take a look and pick the one I like better (and perhaps post a screenshot).  the seaborn script for making the gallery is surprisingly pleasant to use.  I think I like the hip "zoomed in" view, but agustina/you are right that it is also confusing.

I'll push another iteration, but maybe merge before adding an example of all the plots, so others can help.  Two followup thoughts:

-- In the seaborn gallery, they write a bunch of examples that produce a single plot.  I expect you will know some interesting uses of the keyword arguments that show other aspects of traces/models.  For example, I will have another example of the arviz traceplot.
-- The images on the index are manually curated, and there's space for ~4-6 of them, so you can think about your favorite plots to showcase on the front page.

aloctavodia:
Great! yeah feel free to merge this whenever you think is appropriate. I agree no need to add all the plots, neither to address all the comments right now.

ColCarroll:
Got all I said I would!  A funny thing, since the thumbnails are square but the figures are rectangular, is that I capture the largest square I can, but it is still fairly obscured.

You can play around with the two css colors that I updated -- light cyan to highlight, and "pymc3 blue" for links!

Issue #97, opened 2018-05-23
Use xarray throughout


ColCarroll:
There have been proposals to use [xarray](https://xarray.pydata.org/en/stable/) as a common language for pymc3, pystan, and pymc4.  This library might be a good place to start that by implementing utility functions for translating pymc3's Multitrace and pystan's OrderedDict into xarray objects, and then having all plotting functions work with xarrays.

ahartikainen:
PyStan fit object is a StanFir4Model cython class. The ordered dict comes from fit.extract, but ghere are otherways to interact with it.

junpenglao:
Should the `xarray` object also includes diagnostics such as rhat and effective sample size? The PyStan object has cython function that computes the rhat and effective sample size quite efficiently. Last time when I was updating the effective sample size implementation in pymc3 I found that the python implementation is much slower when the input array is large. @aseyboldt has a numba implementation which is much faster but PyMC3 does not have numba as dependence. 

My proposal is that, we make a faster implementation of rhat and effective size implementation in arviz (the numba version), and a general converter that change pymc3 trace and `StanFir4Model` into `xarray` (could be two independent one). It has the trace as the first level, and diagnostics and stats as the second level. The second level is optional but will be computed by default if not provided.

Thoughts?

ahartikainen:
@twiecki yes, that is a good idea. I'm currently writing function to transform fit object to suitable format for the current implementation (of plotting code).

But if Arviz will implement code to read specific type of xarray object we can wait for that before we release the arviz wrapped code.

@junpenglao It would be ideal to just give a fit object to arviz and let the arviz do the magic inside it.

Do you think the numba is going to be a dependency or optional?

aseyboldt:
I'm not sure computing rhat and other stats in arviz is a good idea. Some of that depends on the sampler (eg treedepth), and if we want to print warnings or a report of some kind after sampling then we need to compute it there anyway. I think it would be good if those stats could be in the xarray as well, but there is the problem of name collisions. If we only have rhat and n_effective_samples or so in the array, then that would be fine, but if we also want to put all sampler specific stats in it, then that list would suddenly be much longer and also change over time.
An alternative would be to have two xarrays, one for the samples, and one for the stats.

ahartikainen:
Yes, summary stats and samples basically live in a different spaces.

Either two xarrays or some tricks with the indexing could work.

twiecki:
[...]

PR #102, opened 2018-06-01
Add xarray trace conversion for PyMC3

ColCarroll:
This uses @aseyboldt 's code, as well as a bit of the details from #97 (including the recommendation to use `draw` over `sample`).  Example usage is [here](https://gist.github.com/ColCarroll/c607842947b08bc44d4e1588e6bef98d).  Note that it is not automated, but gives reasonably helpful error messages in certain cases.

I erred on the side of writing too many tests so that as the "spec" changes over there, it will be easier to update this code.

Definitely understand if you want to cut a release or wait for the conversation to settle before merging, but I thought an implementation might help (and I want to start converting the plotting functions to use `xarray`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ before_script:
 
 script:
   - pylint arviz/
-  - pytest -xv arviz/tests/ --cov=arviz/
+  - pytest -v arviz/tests/ --cov=arviz/
   - travis-sphinx build --nowarn --source=doc
 
 after_success:
--- a/arviz/tests/test_xarray_utils.py
+++ b/arviz/tests/test_xarray_utils.py
@@ -1,0 +1,80 @@
+import numpy as np
+import pytest
+
+from ..compat import pymc3 as pm
+from ..utils.xarray_utils import pymc3_to_xarray, default_varnames_coords_dims, verify_coords_dims
+
+
+class TestXarrayUtils(object):
+
+    @classmethod
+    def setup_class(cls):
+        # Data of the Eight Schools Model
+        cls.J = 8
+        y = np.array([28., 8., -3., 7., -1., 1., 18., 12.])
+        sigma = np.array([15., 10., 16., 11., 9., 11., 10., 18.])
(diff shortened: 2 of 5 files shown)


aloctavodia:
Having an implementation will help a lot. It is easier to discuss with concrete examples and code. By not not automated do you mean it is necessary to pass the coordinates and dimensions, right? This is not necessary if the shape is 1, right? This will be automated, right? Too many _rights_, right? :-)

Regarding the first release, given that everyone agrees on using xarray we should wait and switch to xarray before the first release. We still should have the last days of July as a deadline. @ahartikainen is this OK with you?

ColCarroll:
Yep! 1 dimensional things will go through just fine, and it gives reasonable suggestions for higher dimensions.

ahartikainen:
Sound good. I will do the `stan_to_xarray` next week. Should I wait this to be merged.

We can probably make a function that automatically recognizes format `to_array`

ColCarroll:
^^ yes, It may be a good idea to do some basic checks and decide which function to send an object.  This also allows adding `pyro`/`edward`/`turing`/other converters.

In that case it will make sense to add some tests to confirm that the same models lead to the same shape of `xarray`, independent of "inference backend", but that is getting way ahead of ourselves!

twiecki:
Shouldn't this live in pymc3? I would think arviz should stay agnostic and just accept xarray and it's up to the downstream library to provide the right inputs.

ColCarroll:
I think that long term you're right, but wanted to push here first so we could iterate a little quicker and iron out the API while implementing the visualizations. 

I am in favor of moving this into PyMC3, but I would probably keep the directory structure here: according to the [design doc linked in this discussion](https://github.com/arviz-devs/arviz/issues/97), this PR only implements `/trace`.  It may be nice for `arviz` to have a stable import path from `xarray_utils` as stan and pymc3 work though syncing the whole design.

ColCarroll:
(oh, especially with #104 I feel good moving stuff upstream)

PR #107, opened 2018-06-21
Pystan xarray

ahartikainen:
Added functions for xarray_utils to translate fit object to xarray object.

Also I replaced some `.copy` methods to `deepcopy` functions. They were done over the dictionary containing lists, so `deepcopy` was needed. 

Currently these functions work only with the current PyStan github (develop) version, but given the assumption that these functions are removed  and transported to corresponding libraries before official release, that is ok.

If we leave these functions as a backup for the users, then I can fix that step.

### DIFF
--- a/arviz/utils/xarray_utils.py
+++ b/arviz/utils/xarray_utils.py
@@ -1,8 +1,9 @@
 import re
+from copy import deepcopy as copy
+
 import numpy as np
 import xarray as xr
 
-from copy import deepcopy as copy
 from arviz.compat import pymc3 as pm
 
 
@@ -37,7 +38,9 @@ def pymc3_to_xarray(trace, coords=None, dims=None):
         vals = np.array(vals)
         dims_str = base_dims + dims[key]
         try:
-            data[key] = xr.DataArray(vals, coords={v: coords[v] for v in dims_str}, dims=dims_str)
+            data[key] = xr.DataArray(vals,
+                                     coords={v: coords[v] for v in dims_str if v in coords},
+                                     dims=dims_str)
         except KeyError as exc:
             if not verified:
                 raise TypeError(warning) from exc
@@ -80,7 +83,14 @@ def default_varnames_coords_dims(trace, coords, dims):
         dims = {}
 
     for varname in varnames:
-        dims.setdefault(varname, [])
+        if varname not in dims:
+            vals = trace.get_values(varname, combine=False, squeeze=False)
+            vals = np.array(vals)
+            shape_len = len(vals.shape)
+            if shape_len == 2:
+                dims[varname] = []
+            else:
+                dims[varname] = [f"{varname}_dim_{idx}" for idx in range(1, shape_len-2+1)]
 
     return varnames, coords, dims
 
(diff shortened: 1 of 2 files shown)


ColCarroll:
This is great - I am not sure why the test is failing (it looks like one of mine?)  I think relying on github versions of both libraries is fine for now.

It would be nice to have, for example, a unit test that verifies that the eight schools model implemented in either pystan or pymc3 ends up looking the same when converted to `xarray`, but that can go in a separate PR.  

Also, after this is merged, it would be nice to have a general function that sends `OrderedDict`s to `pystan2xarray`, and `MultiTrace`s to `pymc_to_xarray`.  

I have most of an implementation for `autocorrplot` using `xarray`, and can probably port the rest of the plotting functions fairly soon.

Looks good to merge, unless you think the test failure is due to these changes!

ahartikainen:
Hi, pylint is giving some warnings. (And probably failing due to these?)

The test that fails is probably related to a "broken" function.

default_varnames_coords_dims(self.trace, None, None)

That will fail if `dims` is None. Does that work for you locally?

It is also failing with the `pystan_varnames...`

Should we fix it in another PR?

ahartikainen:
Ok, I went and fixed some error for both PyStan and PyMC3 sides.

I think that the `pystan_verify_coords_dims` and `verify_coords_dims` should check if the "current" dims follow some possible combinations for dimensions.

Schools example:

What I mean is like --> there are currently 2 dims with size 8 (`theta` and `theta_tilde`). It could give a warning / info that the some dimensions could be same.

This, of course, won't work when some dimensions have the same number of variables but are not connected, which is the reason it should not raise an error in these cases.

ahartikainen:
@ColCarroll you can merge this PR if you think the changes are ok. I did change some pymc3 parts too, so please review them.

Currently, for the most of the plots, it does not really matter if the dims are correctly named, but for the future if we are going to exploit this feature, it is a good idea to make it easy. (e.g. automatically selecting pairs/groups for plotting)

ColCarroll:
This looks good, thanks!  Yes, the failing pymc test was meant to make sure it threw the appropriate exception, and I can take a look at that.  I am excited to have two inference libraries to work with now.  This will be very cool if model criticism becomes backend-agnostic.

ahartikainen:
That sounds good.

btw the current code can't get the NUTS parameters (energy, treedepth) so when(and if) needed I can add an option for that too.

ColCarroll:
Yes, I think most of the existing functions (all but one of the `plots`, and most of the `stats`) so far only require posterior draws, so this should be good perhaps through an initial release?

ahartikainen:
I think so too. 

It is not too hard to access NUTS parameters. The only thing needed is that we have common names for each parameter (in xarray object).

arviz/utils/xarray_utils.py
```diff
@@ -78,7 +83,14 @@ def default_varnames_coords_dims(trace, coords, dims):
...
     for varname in varnames:
-        dims.setdefault(varname, [])
+        if varname not in dims:
+            vals = trace.get_values(varname, combine=False, squeeze=False)
+            vals = np.array(vals)
+            shape_len = len(vals.shape)
+            if shape_len == 2:
+                dims[varname] = []
+            else:
+                dims[varname] = [f"{varname}_dim_{idx}" for idx in range(1, shape_len-2+1)]
```
ColCarroll: Ah, I see why things were breaking now.  This definitely makes everything "just work" smoothly, which is nice, but I think I'd rather have this throw an exception instead of silently inferring intent here.  I think the analysis we can do are much more interesting if we know which dimensions are shared between variables.

The test case I am looking at is the non-centered eight schools model.  It now works with no exception, but there is no indication that `theta` and `theta_tilde` both have a coordinate for the school.

Happy to leave it as is if you or @aloctavodia have strong feelings!

PR #108, opened 2018-06-25
Refactor xarray utilties

ColCarroll:
This does a few things to get ready to port functions to `xarray`:

- in `xarray_utils`, there is an abstract base class so that `PyMC3ToXarray` and `PyStanToXarray` have the same interface
- Adds a function `convert_to_xarray` that dispatches code to the correct converter
- Adds `pystan` to the travis build
- Tests the `PyStanToXarray` code (in the exact way the `PyMC3ToXarray` code is tested)

I have `autocorrplot` mostly ready to go after this lands.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,10 @@ addons:
   apt_packages:
     - pandoc
 
+cache:
+  directories:
+    - $HOME/.theano
+    - $HOME/miniconda
 
 install:
   - sudo apt-get update
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,18 +14,10 @@ cache:
 
 install:
   - sudo apt-get update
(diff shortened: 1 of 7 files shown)


ColCarroll:
I also just added a cache to travis, so I think it will not need to keep reinstalling everything.  I also cached `~/.theano`, which should speed up any pymc3 stuff going on.  @ahartikainen, are there external folders that `pystan` uses?

ahartikainen:
No. 

But if you pickle the `fit`, then you first need to pickle the model too. And when loading, load it first. `list` works nicely

    with open(...) as f
        pickle.dump([model, fit], f)

ColCarroll:
Will plan to merge tomorrow morning if there are no objections - I have `autocorrplot` implemented which builds on this, and works fine for either pystan or pymc3.

PR #109, opened 2018-06-26
Use xarray for autocorrplot, speed up test suite more

ColCarroll:
Looks the same as current implementation, but this one uses `numpy` to have a bit more control over what is going on.  I also added a few utility functions to decide on grid size and to iterate over the coordinates/dimensions of `xarray` in order to make plots.

Also, I pickled a trained pymc3 and pystan model so that tests run faster.  Locally, running the test suite went from ~5 minutes to ~40s.

image


aloctavodia:
I just take a quick look at the code and it looks great. Although, I do not have time right now to look at this in detail, I have a couple of deadlines around the 4th of July, so feel free to merge when you think is ready. I will play with this in a few days.

PR #113, opened 2018-06-27
change r2_score

aloctavodia:
This is an alternative way to compute r2. This version provides a much more better agreement with the one computed by `stats.linregress`. Just an example.

```python
np.random.seed(1)
N = 100
alpha_real = 2.5
beta_real = 0.9
eps_real = np.random.normal(0, 100, size=N)

x = np.random.normal(10, 1, N)
y_real = alpha_real + beta_real * x 
y = y_real + eps_real

x = np.random.normal(10, 1, N)
y_real = alpha_real + beta_real * x 
y = y_real + eps_real

with pm.Model() as model:
    α = pm.Normal('α', mu=0, sd=1000)
    β = pm.Normal('β', mu=0, sd=1000)
    ϵ = pm.HalfCauchy('ϵ', 50)

    μ = pm.Deterministic('μ', α + β * x)
    y_pred = pm.Normal('y_pred', mu=μ, sd=ϵ, observed=y)
    
    trace = pm.sample(1000)

ppc = pm.sample_ppc(trace, samples=1000, model=model)
```

With the previous version we get 0.34, with the new one 0.01 (in agreement with `stats.linregress` and common sense). I guess the intuition is that the previous version reflects the fact that even when x and y have a lower correlation the model is capable of generating _reasonable_ values of y, in the sense that they are very spread. While the new version reflects better the fact that the correlation of the data is indeed low.

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -512,8 +512,7 @@ def _gpinv(probs, kappa, sigma):
 
 def r2_score(y_true, y_pred, round_to=2):
     """
-    R-squared for Bayesian regression models. Only valid for linear models.
-    http://www.stat.columbia.edu/%7Egelman/research/unpublished/bayes_R2.pdf
+    R² for Bayesian regression models. Only valid for linear models.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


junpenglao:
Didnt manage to review it before you merge but it seems very interesting - this could warrant a small arXiv paper

aloctavodia:
ohhh sorry, I will wait more the next time. A small arXiv paper? That could be interesting. Would you like to help. I have a deadline in about a week, but then I could work on this...

junpenglao:
Sign me up ;-)

PR #116, opened 2018-07-07
Add xarray support to forestplot

ColCarroll:
Phew! This was sort of tough.  I used an object-oriented approach that leans heavily on the `VarHandler.iterator` method, which figures out where every row in this figure should go.  This makes it easy to add more facets (for example, a traceplot) to this in the future.  I added a joyplot to show it can be supported.

## Sample inputs/outputs

Just showing off a bit of what can be done

```python
# default styling
fig, ax = az.forestplot(data)
```
image

```python
# comparing models (I am sure I did not get as many stan samples as pymc3 samples here)
# note elegantly handling multiple chains, and that the centered implementation has no `theta_tilde`,
# and that coloring/order is consistent
fig, axes = az.forestplot([stan_xarray, centered_data, pymc3_xarray], 
                          model_names=['PyStan', 'Centered', 'PyMC3'], 
                          var_names=['theta', 'theta_tilde'], 
                          colors=['#AF0A22', 'C1', '#1B6989']);
```
image

```python
# Joyplots!
fig, ax = az.forestplot(data, 
                        var_names=['theta'], 
                        joyplot=True, 
                        combined=True, 
                        textsize=12, 
                        colors='white')
```
image

```python
# Just because you can doesn't mean you should
fig, axes = az.forestplot([centered_data, data], 
                          model_names=['Centered', 'Non-Centered'], 
                          var_names=['theta'], 
                          joyplot=True,
                          linewidth=1,
                          textsize=12,
                          joyplot_alpha=0.4)
```

image

## Things that might concern you
This version supports most of what the previous version did, with two notable exceptions.  Both choices were to simplify the signature of the function, and the implementation.

- No `vline` support: This implementation returns a figure and an array of axes. I also removed a bunch of options for labelling axes, since there is such direct access to them. I would rather allow a pattern like:
```python
fig, axes = az.forestplot(data)
y_lims = axes[0].get_ylim()
axes[0].vlines(my_vlines, *y_lims, **whatever_else)
```
- No `skip_first`: This just seemed like data manipulation, not plot manipulation.

I also renamed a few arguments. Mostly to follow snake_case convention of underscore between words (`var_names`, `r_hat`, `n_eff`), but also to clarify intent.  In particular, I renamed `alpha` to `credible_interval`, since `alpha` consistently means "transparency" in matplotlib.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -453,7 +453,7 @@ known-third-party=enchant
 max-args=10
 
 # Maximum number of attributes for a class (see R0902).
-max-attributes=7
+max-attributes=10
 
 # Maximum number of boolean expressions in a if statement
 max-bool-expr=5
--- a/arviz/plots/densityplot.py
+++ b/arviz/plots/densityplot.py
@@ -102,7 +102,7 @@ def densityplot(data, data_labels=None, var_names=None, alpha=0.05, point_estima
     for m_idx, plotters in enumerate(to_plot):
         for var_name, selection, values in plotters:
             label = make_label(var_name, selection)
-            _d_helper(values, label, colors[m_idx], bw, textsize, linewidth, markersize,
+            _d_helper(values.flatten(), label, colors[m_idx], bw, textsize, linewidth, markersize,
                       alpha, point_estimate, hpd_markers, outline, shade, axis_map[label])
 
     if n_data > 1:
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -1,328 +1,408 @@
+from collections import defaultdict
+from itertools import tee
+
 import numpy as np
 import matplotlib.pyplot as plt
-from matplotlib.gridspec import GridSpec
-from ..stats import hpd, gelman_rubin, effective_n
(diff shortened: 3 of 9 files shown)


aloctavodia:
This is really really great!

I agree about removing `skip_first`, and renaming `alpha` to `credible_interval`. Eventually, we should do the same with the rest of the plots. Not entirely sure about removing `vline`, but I totally see your point, may be I just need some time to let `vline` go :-)

junpenglao:
This is some beautiful stuff.

aloctavodia:
I think that instead of `joyplot=True` should be `kind = 'joyplot'` to be consistent with other plots (and maybe to allow for different kinds of plot in the future)

ahartikainen:
Looks great!

ColCarroll:
Yeah, the `vlines` thing I am pretty neutral about, but when I am doing that I will usually set the color to black, the style to "dashed", and the width to something custom. It felt like 4 more arguments was too many.

Good point on the `kind='joyplot'`: I'd been following the pattern of `n_eff` or `r_hat`, but this _replaces_ the plot instead of _adding/removing_ a plot.

PR #117, opened 2018-07-08
save_trace is under az.utils

AustinRochford:
`save_trace` is not under the top level namespace, but rather under `az.utils`


AustinRochford:
Somehow other changes made it in here, will close and fix in new PR

PR #121, opened 2018-07-10
Remove the need to use bootstrapping

aloctavodia:
Basically the same as #113, but without bootstrapping


ColCarroll:
Deleting code is the best way to lower coverage % !

PR #122, opened 2018-07-10
Update kdeplot

ColCarroll:
This did not actually have that many updates: the function already works for a big variety of objects, including xarray DataSets.  I spent some time having it automatically make a kde plot for each element in an xarray, but that ends up looking very similar to the densityplot, so I got rid of it.

Code is tidied up a bit, and improved some of the xarray utilities. I put a picture of the new example in the gallery below.

```python
ax = az.kdeplot(np.random.gumbel(size=100), label='100 gumbel samples', rug=True,
                plot_kwargs={'linewidth': 5, 'color': 'black'}, 
                rug_kwargs={'color': 'black'});
```

image

### DIFF
--- a/arviz/plots/energyplot.py
+++ b/arviz/plots/energyplot.py
@@ -6,7 +6,7 @@ from ..utils import get_stats
 
 
 def energyplot(trace, kind='kde', bfmi=True, figsize=None, legend=True, fill_alpha=(1, .75),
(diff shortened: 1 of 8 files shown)


ColCarroll:
Oops, those are legit failures -- will fix them tonight.

ColCarroll:
Ok, *now* I think it is ready, and I had to make some internal changes so you can see how the API looks.

I also added `plt.close('all')` to the test suite after each message, which reduces the number of warnings about open windows I get after running it!

PR #131, opened 2018-07-16
Update jointplot to use xarray

### LINKED ISSUE #125: Joinplot needs update to work with xarray

ColCarroll:
Fixes #125 

This keeps most of the current code.  The biggest difference is when specifying `var_names`, you also pass in `coords` for slicing.  If we find that pleasant (and I do!), we might go back and update the rest of the API to follow that.  For example, the new example in the gallery is 

```python
az.jointplot(data,
             var_names=['theta'],
             coords={'school': ['Choate', 'Phillips Andover']},
             kind='hexbin')
```
image

### DIFF
--- a/arviz/plots/jointplot.py
+++ b/arviz/plots/jointplot.py
@@ -2,36 +2,33 @@ import matplotlib.pyplot as plt
 from matplotlib.ticker import NullFormatter
 
 from .kdeplot import kdeplot
-from ..utils import trace_to_dataframe, get_varnames
-from .plot_utils import _scale_text, get_bins
+from ..utils import convert_to_xarray
+from .plot_utils import _scale_text, get_bins, xarray_var_iter, make_label
 
 
-def jointplot(trace, varnames=None, figsize=None, textsize=None, kind='scatter', gridsize='auto',
-              skip_first=0, joint_kwargs=None, marginal_kwargs=None):
+def jointplot(data, var_names=None, coords=None, figsize=None, textsize=None, kind='scatter',
+              gridsize='auto', skip_first=0, joint_kwargs=None, marginal_kwargs=None):
     """
     Plot a scatter or hexbin of two variables with their respective marginals distributions.
 
     Parameters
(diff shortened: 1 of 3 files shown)


arviz/plots/jointplot.py
```diff
@@ -45,17 +42,22 @@ def jointplot(trace, varnames=None, figsize=None, textsize=None, kind='scatter',
     ax_hist_x : matplotlib axes, x (top) distribution
     ax_hist_y : matplotlib axes, y (right) distribution
     """
-    trace = trace_to_dataframe(trace[skip_first:], combined=True)
-    varnames = get_varnames(trace, varnames)
+
+    data = convert_to_xarray(data)
+    data = data.where(data.draw >= skip_first).dropna('draw')
```
aloctavodia: We should remove `skip_first` from every plot.

arviz/plots/jointplot.py
```diff
@@ -80,26 +82,22 @@ def jointplot(trace, varnames=None, figsize=None, textsize=None, kind='scatter',
         axjoin.scatter(x, y, **joint_kwargs)
     elif kind == 'hexbin':
         if gridsize == 'auto':
-            gridsize = int(len(trace)**0.35)
+            gridsize = int(len(x)**0.5)
```
aloctavodia: did you change 0.35 to 0.5 on purpose?

arviz/plots/jointplot.py
ColCarroll: can do!

arviz/plots/jointplot.py
ColCarroll: oops, was playing around with that number and forgot to change back. let me look again to see if it *should be* on purpose

ColCarroll: Here it is with 0.35, 0.42, and 0.5, respectively.  I've got no strong feelings, so I'll change back to 0.5!

image

image

image

PR #132, opened 2018-07-17
Add DictToXarray

AustinRochford:
Convert `dict` objects of posterior samples to `xarray`.  Needs tests.  See this [notebook](https://gist.github.com/AustinRochford/3b92f5b162d33b3981c8c966ee059239) showing basic equivalence between `DictToXarray` and `PyStanToXarray`.

Needs:
- [x] Better documentation and error messages around specifying `chains` in `coords`.
- [ ] Tests

### DIFF
--- a/arviz/utils/xarray_utils.py
+++ b/arviz/utils/xarray_utils.py
@@ -88,9 +88,10 @@ class DictToXarray(Converter):
             The coordinates are those passed in and ('chain', 'draw')
         """
         if chains is None:
-            varnames = list(obj.keys())
-            val = obj[varnames[0]]
-            self.chains = val.shape[-1] if val.ndim > 1 else 1
+            try:
+                self.chains = min(val.shape[1] for val in obj.values())
+            except IndexError:
+                self.chains = 1
         else:
             self.chains = chains
 
--- a/arviz/utils/xarray_utils.py
+++ b/arviz/utils/xarray_utils.py
@@ -8,7 +8,7 @@ import xarray as xr
 from arviz.compat import pymc3 as pm
 
 
-def convert_to_xarray(obj, coords=None, dims=None):
+def convert_to_xarray(obj, coords=None, dims=None, **kwargs):
     """Convert a supported object to an xarray dataset.
 
     This function sends `obj` to the right conversion function. It is idempotent,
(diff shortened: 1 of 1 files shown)


arviz/utils/xarray_utils.py
```diff
@@ -53,15 +55,156 @@ def __init__(self, obj, coords=None, dims=None):
...
         return
 
 
+class DictToXarray(Converter):
+    def __init__(self, obj, coords=None, dims=None, chains=None):
+        """Convert a dict to an xarray dataset.
+
+        Parameters
+        ----------
+        obj : dict[str, iterable]
```
twiecki: What do people think of using type annotations?

AustinRochford: I have no strong feelings

ahartikainen:
What is the format we want to handle sample

    dict
        key1 : ndarray (draw, chain, *shape),
        key2: ndarray (draw, chain, *shape)

or

    dict
        key1 : dict
            chain1 : ndarray (draw, *shape)
            chain2 : ndarray (draw, *shape)
        key2: ndarray (draw, chain, *shape)
            chain1 : ndarray (draw, *shape)
            chain2 : ndarray (draw, *shape)

or 

    dict
        (key1, chain1) : ndarray (draw, *shape)
        (key1, chain2) : ndarray (draw, *shape)
        (key2, chain1) : ndarray (draw, *shape)
        (key2, chain2) : ndarray (draw, *shape)

Also

Could user give (chains,) draws and shape in different order?

AustinRochford:
@ahartikainen i believe PyStan gives (draw, *shape, chain), so I'll probably start there, but I'll give some thought to different orders.

AustinRochford:
That corresponds to the first format you've posted, which I'll add to the docstrings.  That corresponds to the output of `stan_fit.extract(stan_fit.model_pars, permuted=False)`, which is the first test case I've been using.

arviz/utils/xarray_utils.py
ColCarroll: yes, it can be gradually introduced, too.  i don't use pycharm, but i've heard it is quite helpful there.  i could set up a CI check if someone started adding type annotations (doesn't have to be here, though).

I'd guess it was difficult for some of these "edge" functions that accept objects that `ArviZ` does not know about (neither PyStan nor PyMC3 are requirements).

ColCarroll:
Waiting on tests, right? i'm happy to help with/add those, especially if you can give an example of using this function.

AustinRochford:
@ColCarroll i worked a bit on testing last night, got hung up on the static-ness of `default_varnames_coords_dims`.  May need to refactor a bit to enable easier testing.

AustinRochford:
I think I see a simpler way to do this, going to close this one and open a new PR.

PR #133, opened 2018-07-18
Outline markers in forestplot

ColCarroll:
Very small change.  The dots in the r-hat and effective N plots in forestplot are now outlined in black.

image


ahartikainen:
Should that outline also work outside the axis

https://stackoverflow.com/questions/9912206/how-do-i-let-my-matplotlib-plot-go-beyond-the-axes

aloctavodia:
good catch, I missed that!

PR #134, opened 2018-07-18
Add a converter from dict to xarray

AustinRochford:


arviz/utils/xarray_utils.py
```diff
@@ -8,40 +8,70 @@
...
 
     Returns
     -------
     xarray.Dataset
         The coordinates are those passed in and ('chain', 'draw')
     """
-
-    if isinstance(obj, xr.Dataset):
-        return obj
+    if isinstance(obj, dict):
```
ColCarroll: It looks like this would throw a `TypeError` if it was given an `xarray.Dataset` object.  I'm not sure if there's an elegant way to avoid that, but tests *should* fail if these functions are not idempotent.

AustinRochford:
Linting errors and idempotency should be fixed.

@ColCarroll the `PyStanToXarray` converter does some extra `dtype` guessing that I don't think we want to lose by basing it off `DictToXarray`.

arviz/utils/xarray_utils.py
AustinRochford: This should be fixed now.

ColCarroll:
:clap: thanks for checking on the pystan stuff, too.  i think you just added support to tensorflow probability, and made it easier to work with edward and pyro.

AustinRochford:
Happy to help! If you run into trouble working with those, shoot me a script and I will try to fit them into the workflow.

PR #135, opened 2018-07-19
Traceplot supports xarray

ColCarroll:
Looks quite similar to the current implementation.  Two big differences:

1. I removed the `prior` argument. It seemed hard to understand how to use, hard to implement, and I'd rather wait and accept an xarray dataset there (perhaps similar to how forestplot just accepts multiple datasets?)

2. The `lines` argument now looks like `[('theta', {'school': ['Choate', 'Phillips Andover']}, [0, 1, 8])]`.  The first two arguments say which plots the lines should go on, and the last argument says where the vlines go.

Here's the example that goes in the documentation:

image

### DIFF
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -6,7 +6,7 @@ from ..utils import convert_to_xarray
 
 
 def autocorrplot(posterior, var_names=None, max_lag=100, symmetric_plot=False, combined=False,
-                 figsize=None, textsize=None, skip_first=0):
+                 figsize=None, textsize=None):
     """
     Bar plot of the autocorrelation function for a posterior.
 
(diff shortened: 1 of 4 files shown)


ahartikainen:
Looks good. 

How does it handle multidimensional variables? Vectors, Matrices, nD Tensors?

ColCarroll:
Multidimensional RVs are handled as usual, and you can specify subsets by specifying coordinates, like 
`az.traceplot(data, var_names=['theta'], coords={'school': ['Deerfield', 'Phillips Andover']})`

image

ColCarroll:
(the above is the output of just az.traceplot(data, var_names=['theta'])`, to be clear )

ColCarroll:
(this also kills the altair traceplot, but not the _dream_ of an altair traceplot... one new library at a time :smile: )

aloctavodia:
Like the phoenix (or Terminator) Altair traceplot will come back!

PR #136, opened 2018-07-19
Fix pystan dtype

ahartikainen:
remove deprecated comments `# ...`
Ignore limits `<...>`
clearer code

### DIFF
--- a/arviz/utils/xarray_utils.py
+++ b/arviz/utils/xarray_utils.py
@@ -509,11 +509,19 @@ class PyStanToXarray(Converter):
             r'//.*?$|/\*.*?\*/|\'(?:\\.|[^\\\'])*\'|"(?:\\.|[^\\"])*"',
             re.DOTALL|re.MULTILINE
         )
+        STAN_INTEGER = r"int"
+        STAN_LIMITS = r"(?:\<[^\>]+\>)*" # <....>
+        STAN_SHAPE = r"(?:\[[^\]]+])*" # [...]
+        WS = r"\s*"
+        STAN_PARAM = r"([^;=\s]+)" # ends= ";", "=" or whitespace
         pattern_int = re.compile(
-            r"int(?:\[.*\])*\s*(.*)(?:\s*[=;]|(?:\s*<-))",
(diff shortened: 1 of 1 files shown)


arviz/utils/xarray_utils.py
```diff
@@ -509,11 +509,19 @@ def infer_dtypes(self):
...
         )
+        STAN_INTEGER = r"int"
+        STAN_LIMITS = r"(?:\<[^\>]+\>)*" # <....>
+        STAN_SHAPE = r"(?:\[[^\]]+])*" # [...]
+        WS = r"\s*"
+        STAN_PARAM = r"([^;=\s]+)" # ends= ";", "=" or whitespace
         pattern_int = re.compile(
-            r"int(?:\[.*\])*\s*(.*)(?:\s*[=;]|(?:\s*<-))",
+            "".join((STAN_INTEGER, WS, STAN_LIMITS,
+                     WS, STAN_SHAPE, WS, STAN_PARAM))
```
ColCarroll: needs comma at end of this line

arviz/utils/xarray_utils.py
```diff
@@ -509,11 +509,19 @@ def infer_dtypes(self):
             r'//.*?$|/\*.*?\*/|\'(?:\\.|[^\\\'])*\'|"(?:\\.|[^\\"])*"',
             re.DOTALL|re.MULTILINE
         )
+        STAN_INTEGER = r"int"
```
ColCarroll: i like this change, but the linter is relentless! these need to be lower snakecase (`stan_limits`, `stan_shape`, etc), and line 524 needs to be broken up (I'd probably do it carefully in a few steps).

ahartikainen:
No problem, I will change them.

ahartikainen:
I was thinking of adding a test + model to tests, but the current implementation only allows one model per library.

ColCarroll:
True - you might change the `load_cached_models` to be `load_eight_schools`, and then you could add more models?

ahartikainen:
or we save them inside a dictionary with names as keys?

ColCarroll:
That sounds great!  The number of chains/draws can be quite small, since we
aren't testing that.

PR #137, opened 2018-07-20
Use conda cython

ahartikainen:
Cython is needed for PyStan (and also for some of the mkl modules need it).


ahartikainen:
Actually requirements-dev doesn't need to be changed, pip will install it if one install PyStan.

Issue #138, opened 2018-07-21
`utils.trace_to_dataframe` fails on `pymc3.MultiTrace` input.


MFreidank:
Hello everyone,

I just noticed that `trace_to_dataframe` may fail on entirely valid `pymc3.MultiTrace` objects, this is due 
to the following code:
`if type(trace).__name__ == "MultiTrace"`. 

I get the intention, I believe, one wants to avoid a dependency on `pymc3`. 
However, if I import `pymc3` as `import pymc3 as pm` and then use `pm.backends.base.MultiTrace` to construct a trace I will get an error as:
`type(trace).__name__` will return `"pm.backends.base.MultiTrace`. 

I suggest the following approach:
```
try:
    from pymc3.backends.base import MultiTrace
except ImportError:
    pass   # this still avoids a dependency on pymc3
else:
    is_multitrace = isinstance(trace, MultiTrace) 
```
This avoids a dependency and handles arbitrary ways to import `MultiTrace`.
I have code ready for this, if you agree with me on the way forward I'm willing to provide a pull request.

Best,
MFreidank

aloctavodia:
A pull request is more than welcome! Thanks!

PR #140, opened 2018-07-22
safely check for pymc3 objects without additional dependency

MFreidank:
Without introducing a fixed dependency on `pymc3` (or any other non-standard library module), this commit introduces facilities to safely recognize `pymc3` objects (and objects with origin in other modules). This is achieved by reliance on standard library module `importlib`.
In summary, we add a function with signature `_has_type(object_, typename, module_path)` that checks if `object_` has a type given as string `typename` that can be found under a `module_path`, also given as string. If a module under `module_path` can not be imported (e.g. because it is not installed), we return `False` to cover cases where a user does not use `pymc3`. 
We use this function to refactor a few problematic string-based checks in `arviz.utils.utils`, for further information on those check issue #138.

### DIFF
--- a/arviz/utils/utils.py
+++ b/arviz/utils/utils.py
@@ -1,5 +1,6 @@
 import bz2
 import gzip
+import importlib
 import lzma
 import os
 import re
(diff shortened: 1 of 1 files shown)


MFreidank:
Travis/pylint wants me to use `isinstance(object_, type_cls)` rather than `type(object) == type_cls`. I could do that change, no problem, but the original author had `type(object_).__name__ == "MultiTrace"` and I was trying to stay closer to the original implementation. Should I make that change?

aloctavodia:
I think is fine, but @ColCarroll could have a more informed opinion.

ColCarroll:
This is very nice and quite useful!

I prefer `isinstance` since it is more explicit about intentions and avoids edge cases.

MFreidank:
I changed `type(object_) == type_cls` to `isinstance(object_, type_cls)`. From my point of view, this might be ready for merging now?

ColCarroll:
:thumbsup:  Will merge once tests pass.  Thank you!

MFreidank:
@ColCarroll Thanks a lot! :+1:

Issue #141, opened 2018-07-23
autocorr vs autocorrplot


MFreidank:
This issue is actually more of a question, it is quite possible that we can resolve this without actual code changes. `arviz` maintains `autocorr` code in `stats.diagnostics._autocorr` and `plots.autocorrplot` (both are inherited from the `pymc3` codebase). Both here and in `pymc3` the way the autocorrelation is computed differs, in particular `plots.autocorrplot` uses `np.correlate(y, y, mode=2)` whereas `stats.diagnostics.autocorr` uses `fftconvolve(y, y[::-1])`. Is there a reason why `plots.autocorrplot` does not simply use `stats.diagnostics.autocorr` under the hood?

aloctavodia:
Not sure probably historical reasons, if so It would be wise to follow your suggestion and use `stats.diagnostics.autocorr`. Maybe @junpenglao can provide further insight.

ahartikainen:
I think they have updated the correlate function in numpy. I think we should use the one which is faster.

    n = 100, 1000, 10000, 100000, 1e6

ColCarroll:
Ooh, maybe the opposite of historical reasons. I believe I was porting this to use `xarray`, and wanted to avoid `plt.acorr`, which used to be used: it does not add much to the code complexity, but allows a lot more flexibility in signature and styling.  I did not even notice `az.diagnostics._autocorr`!

Checking out benchmarks and actual values, it looks like the two functions are *not* identical, but are quite close where there is actual autocorrelation, and that `az.diagnostics._autocorr` is much faster for large inputs, and modestly slower for small ones. I attached the code below.  Trying `np.correlate` on 1e6 points scares my computer.

![image](https://user-images.githubusercontent.com/2295568/43122807-90266962-8f22-11e8-834c-7c41e8e2c7e6.png)

```python
def numpy_version(y):
    points = len(y)
    z = np.correlate(y, y, mode=2)[points - 1:]
    return z / z[0]

metrics = []
for j in range(2, 6):
    points = 10 ** j
    y = np.random.randn(points)
    for lag in (0, 1, 10, 20):
        metrics.append({'points': points, 'lag': lag})
        z = y.copy()
        if lag:
            z[:-lag] = z[:-lag] + z[lag:]
        t0 = time.time()
        a = numpy_version(z)
        metrics[-1]['numpy_version'] = time.time() - t0
        t0 = time.time()
        b = az.diagnostics._autocorr(z)
[...]

PR #143, opened 2018-07-25
Fix/unify autocorr interface

MFreidank:
Following previous discussion in issue #141 this pull request fixes inconsistencies between autocorrelation computation in `az.stats.diagnostics.autocorr` (previously `az.stats.diagnostics._autocorr`) and `az.stats.plots.autocorrplot`. The notions of autocorrelation computation differed previously, in this pull request we settle for the (recently thoroughly discussed) notion as (recently, in February 2018) implemented in `az.stats.diagnostics.autocorr`, `pymc3.stats._autocorr` and `stan::math::autocorrelation`. 

As proof of concept and to make differences between old and new interface visible, consider the following plots:

New:
autocorrplot_new

Previous:
autocorrplot_old

### DIFF
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -3,6 +3,7 @@ import numpy as np
 
 from .plot_utils import _scale_text, default_grid, make_label, xarray_var_iter
 from ..utils import convert_to_xarray
+from ..stats.diagnostics import autocorr
 
 
 def autocorrplot(posterior, var_names=None, max_lag=100, symmetric_plot=False, combined=False,
@@ -56,13 +57,10 @@ def autocorrplot(posterior, var_names=None, max_lag=100, symmetric_plot=False, c
     for (var_name, selection, x), ax in zip(plotters, axes.flatten()):
         if combined:
             x = x.flatten()
-        y = x - x.mean()
-        y = np.correlate(y, y, mode=2)
-        y = y / np.abs(y).max()
-        midpoint = len(y) // 2
+        y = autocorr(x)
         ax.vlines(x=np.arange(min_lag, max_lag),
-                  ymin=np.zeros(max_lag - min_lag),
-                  ymax=y[midpoint + min_lag:midpoint + max_lag],
+                  ymin=0,
+                  ymax=y[min_lag:max_lag],
                   lw=linewidth)
         ax.hlines(0, min_lag, max_lag, 'steelblue')
         ax.set_title(make_label(var_name, selection), fontsize=textsize)
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -46,12 +46,12 @@ def autocorrplot(posterior, var_names=None, max_lag=100, combined=False,
 
     axes = np.atleast_2d(axes)  # in case of only 1 plot
     for (var_name, selection, x), ax in zip(plotters, axes.flatten()):
-        x_ = x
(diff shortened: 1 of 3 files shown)


arviz/plots/autocorrplot.py
```diff
@@ -56,13 +57,10 @@ def autocorrplot(posterior, var_names=None, max_lag=100, symmetric_plot=False, c
     for (var_name, selection, x), ax in zip(plotters, axes.flatten()):
         if combined:
             x = x.flatten()
-        y = x - x.mean()
-        y = np.correlate(y, y, mode=2)
-        y = y / np.abs(y).max()
-        midpoint = len(y) // 2
+        y = autocorr(x)
         ax.vlines(x=np.arange(min_lag, max_lag),
-                  ymin=np.zeros(max_lag - min_lag),
```
ColCarroll: this was allowing for `symmetric_plot=True`, though I am not sure I see the use of that option any more.

ColCarroll: Specifically, I think this is fine, but either here or in a followup PR we should remove the `symmetric_plot` option.

ColCarroll:
I restarted the hpd test. it should probably be run with a fixed random seed since it fails very occasionally.

arviz/plots/autocorrplot.py
MFreidank: I removed the option. `autocorr(x)` already removes symmetry, so we would have to reintroduce it after that computation (by essentially appending a reversed copy of `y` to itself). I think this is a rare, specialized usecase that I have not seen used in actual plots too much.

MFreidank:
@ColCarroll I also introduced capping `y_lim` at `(-1, 1)` - I think that makes the most sense as these are the natural bounds of autocorrelation data. (and in the rare case where someone would need to adjust these manually, it is still possible as we return the axis after the fact)

MFreidank:
Shouldn't the last line be  `return axes` instead of `return ax`?

ColCarroll:
Wow, yeah, that code is a little nasty (it deals with the edge case that there are 0 plots, so `ax` doesn't exist after the loop).  

I might also change the end to 

```python
if axes.size > 0:
    axes[0, 0].set_xlim(...)
    axes[0, 0].set_ylim(...)
```

Note that this is ok because of the earlier `np.atleast_2d`, and we passed `sharex=True, sharey=True` to the `plt.subplots` command.

MFreidank:
@ColCarroll I refactored that as outlined by you (I think it conveys the intention slightly more clearly) and we now return all axes instead of just the last one processed in the `for` loop. 

Plots now look like this:
autocorrplot_new

arviz/tests/test_plots.py
```diff
@@ -40,7 +40,7 @@ def test_posteriorplot(self):
 
     def test_autocorrplot(self):
         for obj in (self.short_trace, self.fit):
-            assert autocorrplot(obj).get_geometry() == (6, 6, 36)
+            assert autocorrplot(obj)[0, 0].get_geometry() == (6, 6, 36)
```
ColCarroll: elsewhere we're using `assert autocorrplot(obj).shape == (6, 6)`

MFreidank: yeah, I like that better. Changed it to that now.

MFreidank:
Travis passed. Could this be merged?

ColCarroll:
Thanks again for all this careful work!

Issue #128, opened 2018-07-13
posteriorplot needs update to work with xarray


aloctavodia:

ColCarroll:
Also needs to update the documentation example to work with xarray.

ColCarroll:
dibs!

PR #145, opened 2018-07-29
[WIP] Add some sample figures in docs

canyon289:
While trying working on the pull request for prior distribution plots in pymc3, I thought it would be helpful both for me and others to have more documentation on what various plotting options, similar to how Seaborn's documentation.

To that extent I tried it out. Attached is a exported PDF of the documentation to illustrate the change.

If this is welcome I would be happy to continue with this PR until the documentation is comprehensive.  Any feedback positive or negative would be appreciated!

Seaborn Inspiration
https://seaborn.pydata.org/generated/seaborn.distplot.html

[arviz.posteriorplot — ArviZ 0.1.0 documentation_example.pdf](https://github.com/arviz-devs/arviz/files/2238547/arviz.posteriorplot.ArviZ.0.1.0.documentation_example.pdf)

### DIFF
--- a/arviz/plots/posteriorplot.py
+++ b/arviz/plots/posteriorplot.py
@@ -58,6 +58,32 @@ def posteriorplot(trace, varnames=None, figsize=None, textsize=None, alpha=0.05,
     -------
     ax : matplotlib axes
 
+    Examples
+    --------
+    Show one example plot for reference
+
+    .. plot::
+        :context: close-figs
+
+        >>> import numpy as np
+        >>> import pandas as pd
+        >>> import arviz as az
+        >>> np.random.seed(0)
+        >>> trace = pd.DataFrame({"mu":np.random.randn(100), "theta":np.random.randn(100)})
+        >>> az.posteriorplot(trace, varnames=["mu", "theta"])
+
+    Change a parameter and show another plot to show what happens
+
+    .. plot::
+        :context: close-figs
+
+        >>> import numpy as np
+        >>> import pandas as pd
+        >>> import arviz as az
+        >>> np.random.seed(0)
+        >>> trace = pd.DataFrame({"mu":np.random.randn(100), "theta":np.random.randn(100)})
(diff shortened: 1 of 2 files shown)


ahartikainen:
`posteriorplot` is not updated yet, but you could change the example to use xarray interface. See e.g. `traceplot`

canyon289:
@ahartikainen 
So if I understand correctly I should
1. Update posteriorplot to use an xarray interface
2. Finish off example plots in the documentation

ColCarroll:
I wonder if we could link these to the example gallery so we do not have to write everything twice.  See [here](http://arviz-devs.github.io/arviz/examples/index.html) for current examples.  Note there are two for a few plots. 

I plan to port posteriorplot this week to xarray, but will mark the issue (#128) when I actually start (so if you get impatient, feel free to comment on the issue and I won't work on it :D )

canyon289:
I personally like Seaborn's approach where the gallery is "Come look at at the cool stuff possible" when just browsing around for general ideas, and then the function specific documentation shows what each argument specifically does. In the end though I'm happy to do whatever you guys think is best!

@ColCarroll In terms of porting things to xarray you'll probably beat me to it. I'm way slower than the rest of you all since I'm still learning and don't want to hold things up. I would still be happy to create documentation after the interface change!

For now I'll just update my branch with plots that I believe would be useful so I can continue with the prior distribution plots ask in the pymc3 repo, and when the xarray interface is done I'll make the appropriate modifications!

Thank you guys

ColCarroll:
I think you're totally right, thanks for contributing *and* explaining! Happy to merge these as you make them.

I'll let you know when/if I get to posteriorplot... it looks like it should lean pretty heavily on `kdeplot`, I think.

ColCarroll:
*sigh* and I'm reading ari's comment now. i'll add a comment to #128 to remind us to update the docs when it uses `xarray`.

PR #148, opened 2018-07-30
Update posterior plot to work with xarray

ColCarroll:
A change I'm happy with is the addition of `load_arviz_data`, which allows you to load some pretrained datasets.  You can train some data, use `az.save_data(...)` on it, then move the file to the appropriate folder if you want to add more (I'd like to add models trained with pystan, edward, pymc4, tensorflow probability...) I made this so I could update @canyon289's documentation examples.

The actual plot definitely works but has some rough edges.  I think it is ok to merge given that there may be others who want to work on it.  As an example,
 
```python
az.posteriorplot(non_centered, var_names=('theta_tilde',), ref_val=0, rope=(-1, 1))
```
image

If we wanted followups, I think

1. The way of passing `rope` or `ref_val` arguments to individual plots is pretty nasty, but the best I could do.  Something like
```python
az.posteriorplot(non_centered, var_names=('mu', 'theta_tilde'), ref_val=1, textsize=10, 
                rope={'mu': [{'rope': (-2, 2)}], 
                      'theta_tilde': [{'school': 'Choate', 'rope': (2, 4)}]
                })
```
image

2. The `_plot_posterior_op` function is a bit of a mess.  I'd much rather have a functional approach with a bunch of small functions adding on to the axes instead of everything sharing so much state.

### DIFF
--- a/arviz/plots/posteriorplot.py
+++ b/arviz/plots/posteriorplot.py
@@ -3,22 +3,24 @@ import numpy as np
 from scipy.stats import mode
 from .kdeplot import kdeplot, fast_kde
 from ..stats import hpd
-from ..utils import trace_to_dataframe, expand_variable_names
-from .plot_utils import _scale_text, _create_axes_grid
+from ..utils import convert_to_xarray
+from .plot_utils import xarray_var_iter, _scale_text, make_label, default_grid
 
 
-def posteriorplot(trace, varnames=None, figsize=None, textsize=None, alpha=0.05, round_to=1,
-                  point_estimate='mean', rope=None, ref_val=None, kind='kde', bw=4.5, bins=None,
-                  skip_first=0, ax=None, **kwargs):
+def posteriorplot(data, var_names=None, coords=None, figsize=None, textsize=None, alpha=0.05,
+                  round_to=1, point_estimate='mean', rope=None, ref_val=None, kind='kde', bw=4.5,
+                  bins=None, **kwargs):
     """
     Plot Posterior densities in the style of John K. Kruschke's book.
 
     Parameters
     ----------
-    trace : Pandas DataFrame or PyMC3 trace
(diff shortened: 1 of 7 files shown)


arviz/plots/posteriorplot.py
```diff
@@ -47,10 +49,6 @@ def posteriorplot(trace, varnames=None, figsize=None, textsize=None, alpha=0.05,
         Controls the number of bins, accepts the same keywords `matplotlib.hist()` does. Only works
         if `kind == hist`. If None (default) it will use `auto` for continuous variables and
         `range(xmin, xmax + 1)` for discrete variables.
-    skip_first : int
```
canyon289: Out of curiosity why was this functionality removed? Not needed?

canyon289: Nevermind figured it out. Coordinates replaces it

ColCarroll: Actually, that's being removed everywhere, with the reasoning that users can filter the first elements themselves instead.

canyon289: Is there a document somewhere showing the "Roadmap" for arviz or is it just in peoples head. I'm not very opinionated, more so am just curious where you got that information from

I agree with the approach though, xarray seems easy and users can remove the burn in period themselves. Just to note it here I think somewhere there should be documentation just explaining that the user should filter the burn in period, and showing how. This is largely to the benefit of people that are newer to MCMC techniques, the argument made it obvious that a user should consider this. Its removal makes the code base simpler, just more precarious for new folks.

ColCarroll: We just sketched out a document last week! I will try to convert to markdown and post soon. 

This particular decision came out while porting one of the early functions over, but is not documented anywhere. I think the API will stabilize once we finish converting existing functions to xarray, then getting rid of some old code, and making a release on PyPI.  I'd guess that release comes in the next ~month.

Issue #149, opened 2018-07-31
Multidimensional traceplot with Pystan and xarray


Bonnevie:
I realize that this is undergoing development at the moment, but what is the intended method for constructing a simple traceplot for a Pystan fit object with 2-dimensional array variables? The following [line](https://github.com/arviz-devs/arviz/blob/c58e0b779309997da5315c2dc91df6147d891791/arviz/utils/xarray_utils.py#L458) seems to discard variables that are 2-dimensional? I get an error saying that `pars` is empty if I try to just call `traceplot` directly on the Pystan fit object:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-101-28cd15193e9c> in <module>()
----> 1 arviz.traceplot(samples)

~/anaconda3/envs/stan-dev/lib/python3.6/site-packages/arviz/plots/traceplot.py in traceplot(data, var_names, coords, figsize, textsize, lines, combined, kde_kwargs, trace_kwargs)
     37     axes : matplotlib axes
     38     """
---> 39     data = convert_to_xarray(data)
     40 
     41     if coords is None:

~/anaconda3/envs/stan-dev/lib/python3.6/site-packages/arviz/utils/xarray_utils.py in convert_to_xarray(obj, coords, dims, chains)
     73         return obj
     74     else:
---> 75         return get_converter(obj, coords=coords, dims=dims, chains=chains).to_xarray()
     76 
     77 

~/anaconda3/envs/stan-dev/lib/python3.6/site-packages/arviz/utils/xarray_utils.py in get_converter(obj, coords, dims, chains)
     37         return DictToXarray(obj, coords, dims, chains=chains)
     38     elif obj.__class__.__name__ == 'StanFit4Model':  # ugly, but doesn't make PyStan a requirement
---> 39         return PyStanToXarray(obj, coords, dims)
     40     elif obj.__class__.__name__ == 'MultiTrace':  # ugly, but doesn't make PyMC3 a requirement
     41         return PyMC3ToXarray(obj, coords, dims)

~/anaconda3/envs/stan-dev/lib/python3.6/site-packages/arviz/utils/xarray_utils.py in __init__(self, fit, coords, dims)
    389             The coordinates are those passed in and ('chain', 'draw')
    390         """
--> 391         super().__init__(fit, coords=coords, dims=dims)
    392 
    393     def to_xarray(self):

[...]

PR #151, opened 2018-07-31
fix empty pars if at least one par nonempty

ahartikainen:
This way no need to check dims for all the variables to see if they are empty.


ColCarroll:
Closes #149

PR #152, opened 2018-08-01
a couple of small fixes

aloctavodia:
* Loo was returning `loo_i` even when `pointwise=False`
* Colors for rug and fill in a kdeplot is set by default as the same color of the kde-line by default. 
* I think the fill under the line should be 0 (transparent) by default. KDEs are very useful to overlaid many distributions (contrary to filled-histtogram) having a shade by defaults reduce the usefulness of this feature.


ColCarroll:
Looks much nicer!

PR #154, opened 2018-08-03
Fix Bug in mode and add additional plots to posterior documentation

canyon289:
In writing the docs I found a bug in mode. In summary it was expecting a dataframe instead of a numpy array. Its fix is part of this pr

### DIFF
--- a/arviz/plots/posteriorplot.py
+++ b/arviz/plots/posteriorplot.py
@@ -58,31 +58,78 @@ def posteriorplot(data, var_names=None, coords=None, figsize=None, textsize=None
 
     Examples
     --------
-    Show one example plot for reference
+
+    Show a default kernel density plot following style of John Kruschke
 
     .. plot::
         :context: close-figs
 
         >>> import arviz as az
         >>> non_centered = az.load_arviz_data('non_centered_eight')
-        >>> az.posteriorplot(non_centered, var_names=('mu', 'tau'), textsize=12)
+        >>> az.posteriorplot(non_centered)
 
-    Change a parameter and show another plot to show what happens
+    Plot subset variables by specifying variable name exactly
 
     .. plot::
         :context: close-figs
 
-        >>> import arviz as az
-        >>> non_centered = az.load_arviz_data('non_centered_eight')
-        >>> az.posteriorplot(non_centered, var_names=('theta_tilde',),
-                             ref_val=0, rope=(-1, 1), textsize=11)
+        >>> az.posteriorplot(non_centered, var_names=("mu",), textsize=11)
 
+    Plot subset of variables by matching start of variable name
+
+    .. plot::
+        :context: close-figs
+
+        >>> az.posteriorplot(non_centered, var_names=("mu", "theta_tilde"))
+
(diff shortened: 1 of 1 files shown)


canyon289:
One question I have, is there a place to just ask general questions? As you may guess I have some questions about some of the implementations, but its hard to tell whether its a bug or just my misunderstanding

canyon289:
Lastly if I should write a test for this let me know. I have was having issues testing locally on my mac so I pushed this without one but if someone says I need one I'll do it

ColCarroll:
Looks great to me! Would you guess the testing problems are on arviz, or on your setup? I have been using `pytest.parametrize` to test options, but trying to to go overboard.  I wouldn't mind adding tests for mean/median/mode here.  See [this test](https://github.com/arviz-devs/arviz/blob/master/arviz/tests/test_plots.py#L30) as an example.

Let me know if that's too much for this PR, and I'm happy to merge and we can add testing later.

The issues are a good place for general questions.

canyon289:
Could you merge this one and I'll follow up with a PR with the tests if thats ok?

I'm guessing the problems are with arviz since I'm having an issue with starting the container. It just hangs on setting up pystan and I'm also hung up on how to fix it. I've attached the console log, maybe you'll see something I'm missing.
[start_container.txt](https://github.com/arviz-devs/arviz/files/2257984/start_container.txt)

canyon289:
I should have mentioned there's no obligation to read my console output especially if you're busy. Thanks for reviewing all these PRs

ahartikainen:
It takes some time to compile pystan from source. Try `python setup -v ...`

canyon289:
@ahartikainen Thanks for the tip! 

Trying the -vv flag right now. Worst case I'll just switching to my linux desktop when I need to test.

canyon289:
hi @ahartikainen 
I'm getting this error now on osx when pytest tries to load pystan_2.17.1.0_500_2.pkl

Tried googling around but I can't figure out how to recompile the model in this context. Any pointers?

------------------------------ Captured log setup ------------------------------
model.py                   369 WARNING  No module named 'stanfit4anon_model_ce9fa8665e99e480b81f7474f0230c56_314051519800918326'
model.py                   370 WARNING  Something went wrong while unpickling the StanModel. Consider recompiling.
========================== 1 error in 185.53 seconds ===========================

ahartikainen:
Is it using the precompiled models? They were probably compiled on linux and then they wont work on osx.

To recompile those pickle files, just remove them and try again the tests.

@colcarroll we should add platform and possible a 32/64 mark for windows to filenames.

ColCarroll:
LGTM. Opened #155 to address this. Pickles are nothing but trouble :frowning_face: .  There's some serialization code we recently added to pymc3 that should be more robust, I might try to add that.

ColCarroll:
(also, thanks again, @canyon289 - plots are showing up at https://arviz-devs.github.io/arviz/generated/arviz.posteriorplot.html)

canyon289:
@ColCarroll and @ahartikainen 
Makes sense why it works on my linux desktop but not my osx laptop

@ColCarroll I'll still write the tests

PR #161, opened 2018-08-08
remove extra axes

aloctavodia:
This remove the extra axes when row * cols > number of plots.


aloctavodia:
Not sure why pylint is suddenly complaining, but the suggestions totally make sense. So I address them :-)

PR #160, opened 2018-08-07
Documentation Fix 2

canyon289:
I'm using this pull request to test TravisCI. Please don't merge yet!

Update
--------
I believe this is good for a merge. Tested new strategy and it should work. Here are the build logs from the test.

https://travis-ci.org/arviz-devs/arviz/jobs/413130428
https://travis-ci.org/arviz-devs/arviz/jobs/413130427
screen shot 2018-08-07 at 7 26 45 am
screen shot 2018-08-07 at 7 26 38 am

PR #162, opened 2018-08-09
Actually use versions for tests

### LINKED ISSUE #159: Possible bug in TravisCI pipeline
I think there's a bug in the travisci pipelines regarding which python version is being used for testing

On line https://travis-ci.org/arviz-devs/arviz/jobs/411966731#L513 python 3.5 in a virtual env is correctly used, but when travisci runs pytest its using python 3.6 in all jobs
https://travis-ci.org/arviz-devs/arviz/jobs/411966731#L1430

Same condition can be seen here in the intended python 3.4 pipeline
https://travis-ci.org/arviz-devs/arviz/jobs/411966732#L1430

ColCarroll:
Should fix #159. Using a solution from #160, by reading a version by default from travis.


ColCarroll:
Wowza. This "works" in that we are now testing on the right versions, but pylint is unhappy in different ways for different versions.  I can follow this up (probably by disabling stuff in `.pylintrc`) unless someone else wants to.

aloctavodia:
Pylint is having a bad weekend, yesterday was also unhappy out of the blue.

canyon289:
We could have pylint only run on the 3.6 build, instead of running three times perhaps.

Thanks for getting this one fixed for me

ColCarroll:
Soooo pymc3 isn't safe to import with scipy < 1.0.0, and conda automatically chooses 0.18 when we install on python 3.4. I opened a pull request into pymc3 to fix that. Three possible solutions:

- Wait for the request to get merged. `create_testenv.sh` already installs pymc3 and pystan from master, and arviz doesn't depend on them anyways, so this might fix itself in a day or two.
- Don't support python 3.4 (looks like final release is scheduled for next march: https://www.python.org/dev/peps/pep-0429/)
- Use `pip install -r` for `create_testenv`.  This ignores the dependency resolution logic that conda builds in, but it works on my machine.

Thoughts?

ColCarroll:
I'm going to give this one more try before punting on python 3.4.  Current failure is because conda installs pandas 0.19.2, and `pandas.testing` was added in 0.21.0.

ahartikainen:
There is a py37 that can be installed with conda. Not all packages have updated yet and I think many py36 works with py37.

PR #163, opened 2018-08-09
change order mc_error and credible interval default value

aloctavodia:
* the mean, standard deviation and hpd intervals are estimates of the _true_ variable while mc_error, n_eff and rhat are properties only of the samples used to approximate that variable. This change the order of mc_error accordingly.

* This also changes the name of the argument `alpha` to `credible_interval` to avoid confusion with matplotlib's alpha parameter.

* The value of the credible interval is set to 0.94, just to make the point that 0.95 is completely arbitrary


aloctavodia:
I change `alpha` to `credible_interval`, in several places

PR #171, opened 2018-08-18
[WIP] make diagnostics work with xarray

aloctavodia:
* The result of any of the 3 functions is a Pandas DataFrame, I was unsure on which data-structure to choose. In some sense this is like the numerical equivalent of a plot.

* This uses `xarray_var_iter`, `selection_to_string` from `plots.plot_utils.py` and thus maybe these functions should be placed inside `utils.xarray_utils.py`

Let me know your thoughts.


aloctavodia:
btw, test are going to fail, as I did not update them

ColCarroll:
This looks good to me! I wonder if we should perhaps have a dummy `to_xarray` function that works for numpy arrays, if only to use these functions more flexibly.

aloctavodia:
I update summary to make it work with xarray too. Still unsure of what the index of the returned dataframe should be, now is just the name of the variable, thus for a multidimensional variables its name will appear repeated in the index.

ahartikainen:
Do we need to rewite these due to new functions for numpy-->xarray?

aloctavodia:
Right, this PR is totally outdated.

ColCarroll:
I'm taking a look at reviving this - trying to get rid of all the `trace_to_dataframe` calls.

Issue #166, opened 2018-08-10
Forestplot: Dimensions are ordered from bottom to top.


aloctavodia:
See this example.

![index](https://user-images.githubusercontent.com/1338958/43984801-4cc959ee-9cd9-11e8-9cd7-b361ec1d226a.png)

MFreidank:
Hey, I would be interested in doing a pr for this. 
It would be nice if you could provide the code you used to generate your example, so that I have a good starting point.

aloctavodia:
Great!

You can try with a simpler example like 
```python
with pm.Model() as model:
    a = pm.Normal('a', 0, 1, shape=2)
    trace = pm.sample()
    
az.forestplot(trace)
```

ColCarroll:
Oh, I did [a lot of reversing](https://github.com/arviz-devs/arviz/blob/master/arviz/plots/forestplot.py) here to get orders of things right, but failed to do so for `selection`.  The easiest solution might be to edit `xarray_var_iter` to add a `reverse_selections` flag to it.

aloctavodia:
I say something wrong, the order of the variables is right, the problems is with multidimensional variables as the dimensions are ordered from bottom to top.

ColCarroll:
Closed by #172

PR #172, opened 2018-08-19
fix forestplot: order dimensions in ascending order

MFreidank:
Adresses #166. Dimensions of multidimensional variables are now ordered ascending instead of descending as they were previously. I decided to go the route @ColCarroll suggested in #166. I introduced 
a flag `reverse_selections` for `xarray_var_iter` in `arviz.plots.plot_utils.py` and set it to `True` inside `arviz.plots.forestplot`. 

They say pictures say more than a thousand words, so:

Old:

old_behavior

New:
new_behavior


ColCarroll:
Beautiful, thank you!

Issue #169, opened 2018-08-12
Representing schema in xarray


ColCarroll:
In order to get to feature parity with pymc3 plotting, we need to have a way to access sampler statistics (specifically, to access divergences), and ppc samples. @aseyboldt outlined a good way to think about the [rest of the schema here](https://github.com/arviz-devs/arviz/blob/master/schema.md).  

At the same time, xarray supports groups, but it doesn't look like it does so natively (yet? see the discussion at https://github.com/pydata/xarray/issues/1092).  

I am proposing something like

```python
import xarray as xr
import netCDF4 as nc

class Trace(object):
    def __init__(self, filename):
        self.filename = filename
        self.data = nc.Dataset(filename)
        self.groups = self.data.groups
        
    def __getattr__(self, name):
        if name in self.groups:
            return xr.open_dataset(self.filename, group=name)
        raise AttributeError("informative message")
    
    def __dir__(self):
        """Allows for tab completion on netCDF group names"""
        return super(Trace, self).__dir__() + list(self.groups.keys())
```

This is a pretty light wrapper around netCDF and xarray.  Usage is something like 

```python
t = Trace('mytrace.nc')
t.posterior  # this is an xarray.Dataset
t.posterior.mu.mean()  # calculate the mean of a variable
```

I think this will have to change a little bit so that nested groups work fine.  In particular, something like

```python
t = Trace('mytrace.nc')
t.sampler_stats.divergences  # should return an xarray.Dataset
t.sampler_stats  #  I think this would tend to return an empty xarray.Dataset
```

ColCarroll:
Also pinging @shoyer in case he has advice for doing this.

shoyer:
This seems more or less reasonable to me, but note that opening a netCDF file isn't always cheap (this is also true to a lesser extent with creating an xarray.Dataset). I expect you will be happier with caching or eagerly creating Dataset objects rather than recreating them in `__getattr__`.

canyon289:
@ColCarroll Should adding a representation of a prior be considered as well? Thinking of this in regards to this issue

https://github.com/pymc-devs/pymc3/issues/3104

canyon289:
Nevermind, I see that its already mentioned in the schema document

ColCarroll:
[...]

PR #173, opened 2018-08-23
Add an `InferenceData` object

ColCarroll:
This is a pretty big change - see also #169. I am sure there are some rough edges, but this seems like a flexible way to get to feature parity with current PyMC3 plotting.

- `InferenceData` is the object that carries all the schema data that is available.  Working with it a little bit, I like it - it is just a light wrapper on a `netCDF.Dataset` that accesses `xarray.Dataset`'s.  Supports tab completion, and usage looks like:
```python
data = az.InferenceData('my_analysis.nc')
data.posterior.mu.mean()   # 42.0
data.sample_stats.diverging.sum()  # 2
data.prior.mu.mean()  # AttributeError until it gets implemented
print(data)

Inference data from "/home/my_analysis.nc" with groups:
	> posterior
	> sample_stats
```

- I added `sample_stats` to the PyMC3 extractor. It was pretty easy, and I can follow up with a similar job on PyStan (or @ahartikainen can!).  We should argue about names for those `sample_stats` as well as the required/optional stats then, and update the schema accordingly.

- A funny thing about `InferenceData` is that it is file based, *not* memory based. I did not want to require every plotting function to require a filename, and I want them to work out of the box with PyMC3 or PyStan objects, so making a plot with one of these objects will write a file to disk. It uses `tempfile` to get a unique filename, and will always write into the same directory. By default, it writes to `.arviz_data/`, but that can be updated with 
    ```python
    import arviz as az
    az.config['default_data_directory'] = 'somewhere_else'
    ```
    If anyone has a more elegant way of handling this, I'm all ears. I was thinking of at least adding a warning every once in a while about the existence of this folder along with a suggestion to clean it out.  Maybe every time the number of files in the directory is a multiple of 10, spawn a warning?

- I updated the sample data to use `InferenceData`.  `az.load_arviz_data('centered_eight')` is a nice way to start playing with this.

- I tried to update documentation and function names, and got most of the way there.


ahartikainen:
I can update PyStan stuff, I just need the normalized names.

`sample_stats == rhat, etc` ?

And then we also have `sampler_stats` or `diagnostics`?

ColCarroll:
Heh, this maybe means we need better names.  I took these straight from their names in pymc3 after a NUTS run:

```python
data.sample_stats

<xarray.Dataset>
Dimensions:           (chain: 4, draw: 500)
Coordinates:
  * chain             (chain) int64 0 1 2 3
  * draw              (draw) int64 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 ...
Data variables:
    depth             (chain, draw) int64 ...
    diverging         (chain, draw) bool False False False False False False ...
    energy            (chain, draw) float64 ...
    energy_error      (chain, draw) float64 ...
    max_energy_error  (chain, draw) float64 ...
    mean_tree_accept  (chain, draw) float64 ...
    step_size         (chain, draw) float64 ...
    step_size_bar     (chain, draw) float64 ...
    tree_size         (chain, draw) float64 ...
    tune              (chain, draw) bool ...
```

arviz/utils/xarray_utils.py
```diff
@@ -1,18 +1,21 @@
 from abc import ABC, abstractmethod, abstractstaticmethod
-import re
```
canyon289: Does it make sense to call this xarray_utils now?

ahartikainen:
I have update the PyStan code (I also did update some dim calculations).
I can't push to your branch so maybe I wait until this is merged and create another PR.

ahartikainen:
Also, I found that only `divergent__`-- >`diverging` and `energy__` --> `energy` are "same". Not sure about the rest.

arviz/utils/xarray_utils.py
ColCarroll: good point - I think we're close to getting rid of some of the original dataframe functions in `utils.py`. probably there should be a refactor then. Really, these classes to convert objects should probably live in their own module, with the `inference_data` object.

ColCarroll:
sounds good to me! most sampler stats will have to be optional (`divergent` doesn't make much sense for metropolis-hastings samples, for example), so we will also have to implement some schema checking after this.

PR #176, opened 2018-08-26
add sample_stats for PyStan

ahartikainen:
Add sample_stats for PyStan.

### DIFF
--- a/arviz/utils/xarray_utils.py
+++ b/arviz/utils/xarray_utils.py
@@ -440,6 +440,49 @@ class PyStanToNetCDF(Converter):
         """
         super().__init__(fit, filename=filename, coords=coords, dims=dims)
 
+    def sample_stats_to_xarray(self):
+        """Extract sampler statistics from PyStan fit."""
+        dims = ['chain', 'draw']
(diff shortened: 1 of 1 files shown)


ahartikainen:
I added `lp__` to here. Is this correct thing to do or should it be inside posterior?

ColCarroll:
Looks good to me -- can merge when tests pass.

Is `lp__` the log probability?  Maybe we should rename it to `log_probability` - there is an open PR to add that to PyMC3 where it is called `logp`(https://github.com/pymc-devs/pymc3/pull/3134), and I think sampler stats is the right place for it. I'll add a PR to the pymc3 converter with whatever name you prefer.

ColCarroll:
I really like including that in sampler stats. Maybe we call it `log_density` since that _technically_ covers both cases?

junpenglao:
log_density could be confusing, I dont mind keeping it as `lp__` actually - it minimized the name changes across in Stan, and we dont have it yet anyway in pymc3.

ahartikainen:
I guess we can keep it as is (I stripped __ from the end)

PR #179, opened 2018-08-27
[WIP] Rename joyplot to ridgeplot

canyon289:
ggjoy has been deprecated, and renamed joyplot to ridgeplot, for a compelling historical reason. I propose we do the same.

I did hasty find and replace job to illustrate the point. If you all agree I'll clean this up when I get home.

https://serialmentor.com/blog/2017/9/15/goodbye-joyplots


ColCarroll:
Wow, never saw that. This PR looks good - is there any work left to do?

canyon289:
Doesn't look like it. Thanks for the review

PR #184, opened 2018-08-30
Add some minor fixes

canyon289:
abstractstaticmethod is deprecated for reference

https://docs.python.org/3/library/abc.html#abc.abstractstaticmethod


ColCarroll:
Do you mind if we hold off on this until #186 gets settled? Everything looks fine, but I would rather not deal with a merge conflict.

canyon289:
I don't mind!

PR #185, opened 2018-08-30
[WIP] Fix labels for plot

canyon289:
The Example Generator was using the module name to label the thumbnails in the gallery so the ridgeplot was showing labeled as a forestplot

I propose changing this to the sphinxtag which allows for more flexibility in naming. The tradeoff though is it matches the "human" name now for the plot but not the method name, although I argue the user can just click into the example to see the code.

Before change

forestplot

After change
afterfix


ColCarroll:
Cool! Agree this is nicer.

PR #188, opened 2018-09-02
Fix bug where argument was ignored

canyon289:
While the code works, the posterior argument is ignored and to a human it looks like a second non keyword argument is actually doing something when it isnt

@ColCarroll Out of curiosity why is an api where all non keyword arguments are ignored, except the first one preferred?

> (obj, *_, group='posterior', coords=None, dims=None)

*Before change*
image

### DIFF
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -32,7 +32,7 @@ def autocorrplot(data, var_names=None, max_lag=100, combined=False,
     -------
     axes : matplotlib axes
     """
-    data = convert_to_dataset(data, 'posterior')
+    data = convert_to_dataset(data, group='posterior')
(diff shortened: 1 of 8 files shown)


ahartikainen:
I think instead of 

    (obj, *_, group='posterior', coords=None, dims=None)

it should be

    (obj, *, group='posterior', coords=None, dims=None)

which will throw a TypeError exception.

canyon289:
Fair point, not sure if the idea is to ignore extra arguments, or disallow extra arguments. The pattern is repeating in a couple other functions as well

ColCarroll:
Sorry, I meant to do what Ari was suggesting, but messed up the pattern. Can fix tomorrow, unless you've got something ready!

PR #189, opened 2018-09-02
support PyStan version < 2.18

ahartikainen:
This will first try to extract with parameters and  `permuted=False`. In older versions, it will return numpy.array and in later versions a dictionary. If a dictionary is not seen, it will extract again, but permuted=True and then goes to reorder permutation.

Also added `lp` back to sample_params, it went missing after the "rebase".

### DIFF
--- a/arviz/utils/xarray_utils.py
+++ b/arviz/utils/xarray_utils.py
@@ -264,6 +264,17 @@ class PyStanConverter:
         dtypes = self.infer_dtypes()
         data = {}
         var_dict = self.fit.extract(self._var_names, dtypes=dtypes, permuted=False)
+        if not isinstance(var_dict, dict):
+            # PyStan version < 2.18
+            var_dict = self.fit.extract(self._var_names, dtypes=dtypes, permuted=True)
+            permutation_order = self.fit.sim["permutation"]
+            original_order = []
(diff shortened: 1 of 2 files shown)


ColCarroll:
LGTM -- travis failures are:

```
************* Module arviz.utils.xarray_utils
arviz/utils/xarray_utils.py:310:0: C0301: Line too long (104/100) (line-too-long)
arviz/utils/xarray_utils.py:301:8: C0103: Variable name "lp" doesn't conform to snake_case naming style (invalid-name)
arviz/utils/xarray_utils.py:310:12: C0103: Variable name "lp" doesn't conform to snake_case naming style (invalid-name)
arviz/utils/xarray_utils.py:313:12: C0103: Variable name "lp" doesn't conform to snake_case naming style (invalid-name)
arviz/utils/xarray_utils.py:315:15: W0612: Unused variable 'sampler_dict' (unused-variable)
```

ahartikainen:
I will still test on my local machine, after that we can merge.

ahartikainen:
Tested with PyStan 2.18 and PyStan 2.17.1. Works now with both of them.

This means that we can start to use `pip install pystan` in testing.

I will add a change for that too.

ColCarroll:
Cool!

PR #190, opened 2018-09-03
Implement InferenceData with ParallelPlot

canyon289:
Looking for two pieces of feedback.
1. Is the approach palatable? 
2. This approach won't work with numpy array inputs at time of writing.numpy_to_data array doesn't have a mechanism for the user to specific sample_stats. Is that an immediate problem?

If this is headed in the right direction though I can clean this up e.g. make functions to add to plot_utils  etc

### DIFF
--- a/arviz/plots/parallelplot.py
+++ b/arviz/plots/parallelplot.py
@@ -1,22 +1,22 @@
 """Parallel coordinates plot showing posterior points with and without divergences marked."""
 import matplotlib.pyplot as plt
 import numpy as np
+from ..utils import convert_to_dataset
+from .plot_utils import _scale_text, xarray_to_nparray
 
-from ..utils import trace_to_dataframe, get_varnames, get_stats
-from .plot_utils import _scale_text
 
-
-def parallelplot(trace, varnames=None, figsize=None, textsize=None, legend=True, colornd='k',
-                 colord='C1', shadend=.025, skip_first=0, ax=None):
-    """Parallel coordinates plot showing posterior points with and without divergences marked.
+def parallelplot(data, var_names=None, figsize=None, textsize=None, legend=True, colornd='k',
+                 colord='C1', shadend=.025, ax=None):
+    """
+    A parallel coordinates plot showing posterior points with and without divergences
 
     Described by https://arxiv.org/abs/1709.01449, suggested by Ari Hartikainen
 
     Parameters
     ----------
-    trace : Pandas DataFrame or PyMC3 trace
+    data : xarray, or object that can be converted (pystan or pymc3 draws)
         Posterior samples
-    varnames : list of variable names
+    var_names : list of variable names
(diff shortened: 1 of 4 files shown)


arviz/plots/parallelplot.py
```diff
@@ -29,23 +29,32 @@ def parallelplot(trace, varnames=None, figsize=None, textsize=None, legend=True,
...
+                                                           combined=True):
+        posterior.append(_posterior.flatten())
+        _var_names.append(make_label(var_name, selection))
+    posterior = np.array(posterior)
 
-    trace = trace[varnames]
+    # Merge chains and to find divergent draws
+    divergent_data = convert_to_dataset(data, group='sample_stats')
+    divergent = xarray_var_iter(divergent_data, var_names=('diverging',), combined=True)
+    diverging_mask = next(divergent)[2].flatten()
```
ColCarroll: ` next(divergent)[2].flatten()` looks super dangerous, like it might throw either a `StopIterationError` or an `IndexError`.

ColCarroll: oh, i see now - `divergent` should have length 1 as an iterator, since the chains are combined, and the `[2]` element will be the actual array.  Maybe a comment about that, then?

examples/parallelplot.py
```diff
@@ -23,4 +23,7 @@
     obs = pm.Normal('obs', mu=theta, sd=sigma, observed=y)
     centered_eight_trace = pm.sample()
 
-az.parallelplot(centered_eight_trace, varnames=['theta', 'tau', 'mu'])
+az.parallelplot(centered_eight_trace, var_names=['theta', 'tau', 'mu'])
+
+import matplotlib.pyplot as plt
```
ColCarroll: are these last two lines supposed to be here?

arviz/plots/parallelplot.py
```diff
@@ -29,23 +29,32 @@ def parallelplot(trace, varnames=None, figsize=None, textsize=None, legend=True,
...
+    # Merge chains and variables
+    for var_name, selection, _posterior in xarray_var_iter(posterior_data, var_names=var_names,
+                                                           combined=True):
+        posterior.append(_posterior.flatten())
+        _var_names.append(make_label(var_name, selection))
+    posterior = np.array(posterior)
 
-    trace = trace[varnames]
+    # Merge chains and to find divergent draws
+    divergent_data = convert_to_dataset(data, group='sample_stats')
```
ColCarroll: we should probably check that this exists earlier in the function.

examples/parallelplot.py
canyon289: Just for WIP. I'll take it out when I replace this commit.

canyon289:
Thanks for the comments. I'll make the changes and force push a new commit tomorrow

canyon289:
Per request
parallelplot

PR #191, opened 2018-09-03
Fix extra argument behavior

canyon289:
Same question as before though! Why force people to use keyword argument?

### DIFF
--- a/arviz/plots/densityplot.py
+++ b/arviz/plots/densityplot.py
@@ -8,6 +8,7 @@ from ..utils import convert_to_dataset
 from .plot_utils import _scale_text, make_label, xarray_var_iter
 
 
+# pylint:disable-msg=too-many-function-args
 def densityplot(data, data_labels=None, var_names=None, credible_interval=0.94,
                 point_estimate='mean', colors='cycle', outline=True, hpd_markers='', shade=0.,
                 bw=4.5, figsize=None, textsize=None):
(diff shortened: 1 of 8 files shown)


ColCarroll:
Cool - looks like this change turned up a few legitimate bugs like the one you pointed out in #188: 

```
************* Module arviz.tests.test_xarray_utils
arviz/tests/test_xarray_utils.py:111:15: E1121: Too many positional arguments for function call (too-many-function-args)
************* Module arviz.plots.densityplot
arviz/plots/densityplot.py:64:20: E1121: Too many positional arguments for function call (too-many-function-args)
************* Module arviz.plots.forestplot
arviz/plots/forestplot.py:161:21: E1121: Too many positional arguments for function call (too-many-function-args)
```

ColCarroll:
Oh, and I wanted a positional argument so the library would be more explicit and readable. It is easy to remove in the future, but hard to add back in. If it gets too annoying, we can remove!

ahartikainen:
It will make usage more robust. Especially important for plotting library where users have used other libs before (e.g mpl)

ahartikainen:
especially important if we go with multiple backends

canyon289:
> Oh, and I wanted a positional argument so the library would be more explicit and readable. It is easy > to remove in the future, but hard to add back in. If it gets too annoying, we can remove!

You mean you wanted to a *keyword* argument?

Makes sense and I agree, by forcing people to use keywords itll make it way easier to make changes without breaking peoples code later.

PR #193, opened 2018-09-03
Docs now build locally, and perhaps with no warnings

ColCarroll:
I think (again) this will fix docs. I installed from github in a fresh environment, and the problem was that `setup.py` will only include files in `arviz/arviz`, unless we also include a `MANIFEST.in`. So travis (and my fresh install) could not find the data files that `load_arviz_data` uses.

I sort of like that we have no MANIFEST.in, and it appears just moving the data folder to the arviz/ directory fixes this.

As an aside it looks like we are also now able to delete the pandas scaffolding that used to be used.


ColCarroll:
Didn't work :( trying again...

canyon289:
It worked?

canyon289:
Nevermind I'm crazy I guess. I swear it just loaded for me.

Issue #196, opened 2018-09-03
Update documentation for InferenceData


canyon289:
A lot of the docstrings read 

>     data : xarray, or object that can be converted (pystan or pymc3 draws)
>       Posterior samples

but I believe this should be changed to 
>    data : xarray, or object that can be converted InferenceData

https://github.com/arviz-devs/arviz/blob/master/arviz/plots/posteriorplot.py#L18

ColCarroll:
Yeah, maybe with "see `convert_to_inference_data` for details" or similar

canyon289:
I'll wait for https://github.com/arviz-devs/arviz/pull/198 to merge before working on this!

canyon289:
Closed by https://github.com/arviz-devs/arviz/pull/257

PR #198, opened 2018-09-03
Numpydoc

### LINKED ISSUE #194: Consistent, good documentation
A good way for this to be done is 

```bash
$ pydocstyle --convention=numpy arviz | wc -l
352
```

this could be added to the travis build as well.  Note that right now each violation is 2 lines, so there are "only" 176 violations. 

I'm working on this now, but it should not effect any code.

ColCarroll:
Fixes #194 , mostly. This does not address #196, which is maybe more important. This also adds `scripts/lint.sh`, which can be used locally to make sure travis does not have style nitpicks!


ColCarroll:
Merging, but if failures due to documentation get annoying, we can turn that check off again.

PR #201, opened 2018-09-04
Xarray trouble

ColCarroll:
This stops travis-sphinx from deploying docs until we can figure out what the error is. 

Maintainers can deploy manually using

```bash
git pull origin master
cd doc
make html
ghp-import -p -n -f -r git@github.com:arviz-devs/arviz.git _build/html/
```

This is equivalent to what `travis-sphinx` is doing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,7 @@ before_script:
 script:
   - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then . ./scripts/lint.sh ; fi
   - pytest -v arviz/tests/ --cov=arviz/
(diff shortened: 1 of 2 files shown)


canyon289:
How did you test docs deploy in your own fork? If I learn how I can help troubleshoot this as well. Autodeploy docs are really nice

ColCarroll:
Yeah, agree it is very nice. I went to travis and enabled builds on my fork. I did not enable the travis-sphinx deploy there (it requires an additional key), so I have to check the travis logs. 

Note that this change keeps _building_ the docs, so we can keep an eye on if they fix themselves, it just won't deploy the docs.

PR #203, opened 2018-09-04
Open dataset in context manager

### LINKED ISSUE #202: Changing the order of plotting and file load seems to cause netcdf errors
For some reason it seems that changing the order of plotting and loading data causes failures, not entirely sure why yet. This is only occurring for me on Ubuntu 16.04 with the latest version of arviz. It does not occur on my osx laptop

Load, Load, Plot, Plot doesn't work. Stack trace attached
[stack_trace.txt](https://github.com/arviz-devs/arviz/files/2349232/stack_trace.txt)

```
import arviz as az

non_centered = az.load_arviz_data('non_centered_eight')
az.violintraceplot(non_centered, var_names=["mu", "tau"], textsize=8)
az.violintraceplot(centered, var_names=["mu", "tau"], textsize=8)
```

Load, Plot, Load, Plot works

```
import arviz as az

non_centered = az.load_arviz_data('non_centered_eight')
az.violintraceplot(non_centered, var_names=["mu", "tau"], textsize=8)

az.violintraceplot(centered, var_names=["mu", "tau"], textsize=8)
```

ColCarroll:
This fixes #202 on my machine. I will check travis logs to see if it fixes it there, too.

### DIFF
--- a/arviz/inference_data.py
+++ b/arviz/inference_data.py
@@ -51,8 +51,11 @@ class InferenceData():
         """
         groups = {}
         with nc.Dataset(filename, mode='r') as data:
-            for group in data.groups:
-                groups[group] = xr.open_dataset(filename, group=group)
+            data_groups = list(data.groups)
+
+        for group in data_groups:
+            with xr.open_dataset(filename, group=group) as data:
+                groups[group] = data
         return InferenceData(**groups)
 
     def to_netcdf(self, filename):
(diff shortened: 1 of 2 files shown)


aseyboldt:
Do we actually want to close this file here? This means that we need to load everything into memory, right? For some model that might not even fit. Maybe it would be better if we behave a bit more like xarray itself here, and keep the file open as long as the inference data object exists. It could also be a context manager, so that people can make sure the file gets closed.

aseyboldt:
It also looks to me as if we stumbled into a netcdf or xarray bug in #202. If we have a way to actually reproduce that with xarray alone, we should report that. I'm still failing to reproduce it myself though... :-)

PR #208, opened 2018-09-06
Fix dims and coords

ahartikainen:
Enable more flexible usage for dims and coords.

This PR enables that user can input names only to specific dims or coords.
If dims is None or not defined, it will use the default name, and if coord is not defined, it creates the default coord.

### DIFF
--- a/arviz/tests/test_xarray_utils.py
+++ b/arviz/tests/test_xarray_utils.py
@@ -46,6 +46,73 @@ class TestNumpyToDataArray():
         assert inference_data.foo.draw.shape == shape[1:2]
         assert inference_data.foo[var_name].shape == shape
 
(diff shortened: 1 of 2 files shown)


ahartikainen:
Only thing that we might want to do, is to check that all given coords are found in the dims and all the dims are less or equal to the dim count.

Should these raise Warning or Exception?

ColCarroll:
It looks like this breaks converting numpy arrays: `az.convert_to_dataset(np.random.randn(size=100))` now throws an exception.

Can you post an example/add a test of what this fixes?

ColCarroll:
Good tests, thanks! Looks good to merge if tests still pass.

ColCarroll:
Thanks @ahartikainen !

Issue #212, opened 2018-09-07
Pairplot example should get updated


ColCarroll:
Currently uses old API: https://arviz-devs.github.io/arviz/examples/pairplot.html

Something like this should work, but the gaps between plots is funny:

```python
import arviz as az

az.style.use('arviz-darkgrid')

centered = az.load_arviz_data('centered_eight')
az.pairplot(centered, var_names=['theta'], divergences=True)
```

![image](https://user-images.githubusercontent.com/2295568/45227638-f1679e00-b28e-11e8-906e-a45edd5926da.png)

canyon289:
Working on this

PR #214, opened 2018-09-08
Update Pairplot example and spacing

canyon289:
Fixes https://github.com/arviz-devs/arviz/issues/212

Example Plots, It's easy enough to change the spacing if someone has a strong preference.

One thing to note is that there's some strange behavior with the y axis label positions. If this is problematic I create an issue.

pairplot_1
pairplot_2
pairplot_3
pairplot_4
pairplot_5
pairplot_6
pairplot_7
pairplot_8

### DIFF
--- a/arviz/plots/pairplot.py
+++ b/arviz/plots/pairplot.py
@@ -111,7 +111,7 @@ def pairplot(data, var_names=None, coords=None, figsize=None, textsize=None, kin
 
     if gs is None and ax is None:
         plt.figure(figsize=figsize)
-        gs = gridspec.GridSpec(numvars - 1, numvars - 1)
+        gs = gridspec.GridSpec(numvars - 1, numvars - 1, wspace=0.05, hspace=0.05)
 
         axs = []
         for i in range(0, numvars - 1):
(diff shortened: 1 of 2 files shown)


ahartikainen:
Can we split varname to multiple lines?

Dim first, then coords?

Something like

`"\n".join(dim_name, coord1, coord2)`

Also, will pairplot fail if gridspec (gs) is defined? Is there indentation error?

ahartikainen:
Also, should divergent samples have greater contrast to main color?

E.g. a bit lighter color like salmon, lime, yellow?

aloctavodia:
Splitting varnames to multiple lines would be great.
Default colors should be restricted to the default cycle (C0, C1, etc), as they have been optimized to be colorblind-friendly (C4 is yellow).

aloctavodia:
One good think of hexgrid is that is really fast, we should also have 2D KDE (thus `kind` will accept 3 options 'scatter', 'hexgrid' and 'kde'). I may have a 2D-fast_kde version somewhere (not sure), I should check... Probably we should have a `fast_kde` and a separated `fast_kde_2d` version, in order to not slow down `fast_kde`.

ahartikainen:
Yes, separate functions is needed. I have one working function somewhere.

aloctavodia:
OK, don't worry I will send a PR on Monday or Tuesday.

PR #216, opened 2018-09-08
Use coordinates in pairplot

ColCarroll:
Coordinates were defined, but not used in the example!


canyon289:
Thanks for catching this!

PR #217, opened 2018-09-08
Hexplot Example and other fixes

canyon289:
- Make exception more helpful for PairPlot when people get `type` wrong
- Add Hex pairplot example
- Remove padding from hexplot (still is more spaced out than scatter for some reason)
- Fix mistake in pairplot example where coords wasn't being used
- Fix end of line inconsistencies in all examples


canyon289:
More pretty pictures 
hexpairplot_1
hexpairplot_2
hexpairplot_3
hexpairplot_4
hexpairplot_5
hexpairplot_6


hexpairplot_8

ahartikainen:
I think bin count or is it grid parameter could follow draw count?

arviz/plots/pairplot.py
```diff
@@ -129,18 +132,18 @@ def pairplot(data, var_names=None, coords=None, figsize=None, textsize=None, kin
                     if i == j == 0 and colorbar:
                         hexbin = ax.hexbin(var1, var2, mincnt=1, gridsize=gridsize, **plot_kwargs)
                         divider = make_axes_locatable(ax)
```
ColCarroll: I can do this in a followup pr, but I'm not familiar with `make_axes_locatable`: could we move the `divider = make_axes_locatable(ax)` above the `if/else` block, and the `divider.append_axes('top' , size='7%').set_axis_off()` after the block?

ColCarroll:
This is good to merge. On this one especially I agree the divergences need to "pop" more - maybe manually making them `'C4'`

arviz/plots/pairplot.py
canyon289: Honestly I don't know what this code does, I was focused on minimal changes but am happy to dive in.

What is the end result you're looking for so I can fix the problem?

ColCarroll: Just keeping code duplication to a minimum!

canyon289: Sounds good. Are you ok with merging this one and I'll follow up with your suggestions?

ColCarroll: Definitely - seems like the comments on here were all feature requests, rather than any problems with the code. We should merge when the tests pass.

canyon289: Thanks, I prone to a medical conditions where I have mild allergy to merge conflicts :)

aloctavodia:
Maybe the facecolor and edgecolor should be different. A white edgecolor could work.

PR #219, opened 2018-09-08
Add a way to extract observations from pymc3 trace

ColCarroll:
Note also that I refactored the `numpy_to_xarray` function so that consistent names would be generated. Now the `posterior_predictive` and `data` variables will have the same default dimension names.

This change will make it easier to add and test the `ppc_plot`.

### DIFF
--- a/arviz/utils/xarray_utils.py
+++ b/arviz/utils/xarray_utils.py
@@ -132,6 +132,55 @@ class requires: # pylint: disable=invalid-name
             return func(cls, *args, **kwargs)
         return wrapped
 
+def _generate_dims_coords(shape, var_name, dims=None, coords=None, default_dims=None):
+    """Generate default dimensions and coordinates for a variable.
+
+    Parameters
+    ----------
+    shape : tuple[int]
+        Shape of the variable
+    var_name : str
+        Name of the variable. Used in the default name, if necessary
+    dims : list
+        List of dimensions for the variable
+    coords : dict[str] -> list[str]
+        Map of dimensions to coordinates
+    default_dims : list[str]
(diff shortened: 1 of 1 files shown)


arviz/utils/xarray_utils.py
```diff
@@ -277,6 +336,7 @@ def to_inference_data(self):
             'sample_stats': self.sample_stats_to_xarray(),
             'posterior_predictive': self.posterior_predictive_to_xarray(),
             'prior': self.prior_to_xarray(),
+            'data': self.data_to_xarray(),
```
canyon289: The only thing that concerns me about this change is that we use the variable data throughout the library to refer to  posterior. 

Would you alright if I opened an issue and/or made a pull request to rename this in the various plots? For example traceplot

https://github.com/arviz-devs/arviz/blob/master/arviz/plots/traceplot.py

ahartikainen: I don't like it either. `data` is really arbitrary thing.

Could it be `observed`, `observed_data` or something else?

ahartikainen:
Also, should we have `observed_data` group or should we assume that `posterior_predictive` only contains data generated from `posterior`  and data used in sampling (`prior_data`/`sampling_data`/`observed_data`)?

We can deal with prior later?

arviz/utils/xarray_utils.py
ColCarroll: I think calling it `data` was a nod to stan's data section :)

I updated it to 'observed_data', and also updated `schema.md`. @canyon289 let me know if that's ok with you, too.

ColCarroll:
Merging this, but will create an issue for how/whether to store other data associated with a model.

PR #221, opened 2018-09-09
Update ppcplot for xarray

### LINKED ISSUE #181: Make `ppcplot` work with `InferenceData`
After #180 this will probably be the last plot to get to parity with pymc3 plotting.

ColCarroll:
Fixes #181 

This adds another dataset to `load_arviz_data` (the very last model at https://docs.pymc.io/notebooks/multilevel_modeling.html), and updates `ppcplot` to use xarray. I changed the colors a little, so the ppc lines would be `C4` instead of `C5`. Can change back if anyone prefers!

These two examples use the radon dataset, and are added to the example gallery:

`az.ppcplot(data, alpha=0.03, kind='cumulative', figsize=(12, 6), textsize=14)`
image

`az.ppcplot(data, alpha=0.03, figsize=(12, 6), textsize=14)`
image


ColCarroll:
Er, wow, the `radon` example is 45MB. Let's not merge this yet.

ColCarroll:
With compression, it is 15MB. I'll get rid of it for now, even though the ppc plot for the eight schools example looks somewhat funny.

arviz/plots/ppcplot.py
```diff
@@ -4,81 +4,104 @@
...
-
-    if figsize is None:
-        figsize = (7, 5)
+    for group in ('posterior_predictive', 'observed_data'):
+        if not hasattr(data, group):
+            raise TypeError(
+                '`data` argument must have the group "{group}" for ppcplot'.format(group=group))
 
-    _, ax = _create_axes_grid(len(ppc_sample), rows, cols, figsize=figsize)
+    observed = data.observed_data
```
canyon289: Why not use `convert_to_dataset` like the other plots?

arviz/tests/test_plots.py
```diff
@@ -80,8 +81,10 @@ def test_pairplot(self):
         pairplot(self.short_trace, kind='hexbin', var_names=['theta'],
                  coords={'theta_dim_0': [0, 1]}, plot_kwargs={'cmap': 'viridis'}, textsize=20)
 
-    def test_ppcplot(self):
-        ppcplot(self.data['y'], self.sample_ppc)
+    @pytest.mark.parametrize('kind', ['kde', 'cumulative'])
+    def test_ppcplot(self, kind):
+        data = pymc3_to_inference_data(trace=self.short_trace, posterior_predictive=self.sample_ppc)
+        ppcplot(data, kind=kind)
```
canyon289: Related to the question above. It seems that PPC plot presents an inconsistent interface to the other plots because the user needs to convert to an InferenceData object themselves, whereas in the other plots it happens under the hood. Is my understanding correct?

ahartikainen: My guess is that we can't infer what are the predictive values needed for comparisons.

This plot is not trivial, because user needs to do extra work to get it working.

arviz/plots/ppcplot.py
ahartikainen: convert to dataset does not implement posterior and other extra values yet.

ColCarroll:
I'm merging this for now, since it is pretty similar to the old version, but will add a note about future work for it.

arviz/plots/ppcplot.py
canyon289: To word this specifically, convert_to_dataset does not implement posterior *predictive* distributions yet. In my understanding it does supporting posterior parameters/trace conversely?

Issue #180, opened 2018-08-28
Add `posterior_predictive` to InferenceData


ColCarroll:
This needs to be done to both the pymc3 and pystan converters.

ColCarroll:
Closed by #222

PR #222, opened 2018-09-09
add prior, posterior predictive and observed data for pystan conversion

ahartikainen:
## prior 

accepts only dictionaries extracted from `fit` object.

    fit.extract(pars=pars, permuted=False) # results shape => draws, chains, *shape

works natively with PyStan 2.18+, but needs manual transforming for earlier version

## posterior_predictive

accepts a string, a list of strings or dictionary.

For dictionary, see prior rules

## observed_data

str or a list of str: extracted from fit object.

    fit.data[par]


ColCarroll:
Fixes #180 

Nice.

PR #227, opened 2018-09-10
[WIP] Break axes labels at selection

canyon289:
Per at @ahartikainen comments make the labels a little nicer.

Still not great, but for a two character change I think this is moving the labels in the right direction.

I haven't checked all the plots yet but below are examples from pairplot.  They're getting cut off a little but If they generally look good I can fix this and check more plots.

hexpairplot_1_
hexpairplot_2_
hexpairplot_3_
hexpairplot_4_
hexpairplot_5_
hexpairplot_6_
hexpairplot_7_
hexpairplot_8_

### DIFF
--- a/arviz/plots/plot_utils.py
+++ b/arviz/plots/plot_utils.py
@@ -155,7 +155,7 @@ def make_label(var_name, selection):
         A text representation of the label
     """
     if selection:
-        return '{} ({})'.format(var_name, selection_to_string(selection))
+        return '{} \n ({})'.format(var_name, selection_to_string(selection))
     return '{}'.format(var_name)
 
 
--- a/arviz/plots/plot_utils.py
+++ b/arviz/plots/plot_utils.py
@@ -135,7 +135,7 @@ def selection_to_string(selection):
     str
         key1: value1, key2: value2, ...
     """
-    return ', '.join(['{}: {}'.format(k, v) for k, v in selection.items()])
+    return ', '.join(['{}'.format(v) for _, v in selection.items()])
 
 
 def make_label(var_name, selection):


ahartikainen:
`plt.savefig(..., bbox_inches='tight')` is a quick fix, but having plots done with correct size from start is a better approach.

canyon289:
I'll use your suggestion and make examples after work today. Thanks!

canyon289:
Here's what it looks like with just the selection value, and not its key. Are two or more level hierarchies possible and/or common? Is there an example of one we should test out if so?

I still need to fix the problem with the labels getting cut off.




hexpairplot_1_
hexpairplot_2_
hexpairplot_3_
hexpairplot_4_
hexpairplot_5_
hexpairplot_6_
hexpairplot_7_
hexpairplot_8_

ahartikainen:
Looks good. I think if there are more levels, user probably needs to handle it somehow manually?

Is there a possibility to select which level to use?

canyon289:
So the user can override the labels because they the axes objects are returned.

I'm sure there's a way to select which level to use, but I don't know how. I can make a multi level example and see what happens.

ahartikainen:
Sounds good.

I hope users start to use coordinates, make these plots much nicer.

canyon289:
Will users have a choice? I think if given a pystan or pymc3 object arviz does all the coordinate and dim assignment for them under the hood

ahartikainen:
Yes, but like giving names and stuff

canyon289:
Sorry I dont follow. When you say giving names what do you mean? In the API the only mechanism I see is users selecting a subset of names and coords with the `var_names` and `coords` argument but they are  unable to specify them directly. 

Do you mean the users should specify the names in pystan or pymc3 before they use arviz?

I feel like Im asking you a bunch of questions recently, I hope you don't mind :) Thanks for the feedback

ahartikainen:
Yes, people have let's say Beta vector of 10 variables, and they have names for each of them. So instead of using only Beta0, ... they can use names for them, and plotting will consider them automagically. It is easier for users to handle more complex datasets.

ColCarroll:
This looks beautiful, way better than before! 

I think using the labels will rely on us providing good examples and docs, and on the default dimension names being so ugly (`beta_dim_0`, etc)!

Issue #231, opened 2018-09-12
Port log_post_trace to use xarray


ColCarroll:
I can not quite figure out what is going on here - it looks like the information is not currently exported as a sample_stat in PyMC3, but maybe @aloctavodia has some information on how to compute it (and how to extract it from `pystan`). 

Leaving this here for myself or others to possibly catch up on LOO:
https://www.youtube.com/watch?v=hpr8pxqkCH8&index=10&list=PLuwyh42iHquU4hUBQs20hkBsKSMrp6H0J&t=0s

![image](https://user-images.githubusercontent.com/2295568/45396242-6416a780-b608-11e8-879a-38250f9b97b7.png)

ahartikainen:
`lp` and `log_lik` are different things. `lp` is the log probability/density for joint distribution and loo needs log likelihood elementwise for the data. Currently, in PyStan user needs explicitly create a `log_lik` variable.

```
generated quantities {
    vector[N] log_lik;
    for (n in 1:N) {
        log_lik[n] = normal_lpdf(y[n] | m * x[n] + b);
    }
}
```

for a model like in https://gist.github.com/ahartikainen/8e7b1a6e4b7083920b212dcafa130015

The end result is (chains, draws, N)

canyon289:
Ah this makes more sense now to me

ColCarroll:
A temporary strategy might be to add it to `sample_stats` - I think I can do that for pymc3 using @aloctavodia's existing function.

ahartikainen:
What about PyStan?

If we assume there is only one way to calculate `log_lik` we could read it from the posterior samples and transfer to `sample_stats`. Not sure if this is going to be very robust.

ColCarroll:
Could you change

```python
def pystan_to_inference_data(*, fit=None, prior=None, posterior_predictive=None,
                             observed_data=None, coords=None, dims=None):
```

to something like

```python
def pystan_to_inference_data(*, fit=None, prior=None, posterior_predictive=None,
                             observed_data=None, likelihood_var=None, coords=None, dims=None):
```

And then use
```python
if likelihood_var is not None:
    likelihood = fit.extract(pars=[likelihood_var])[likelihood_var]
```

to attach the likelihood to the `sample_stats`?  This is a little awkward, but it seems like model comparison is something people care about, so we should work to make sure at least pystan and pymc3 are well supported. Open to other ideas, too!

ahartikainen:
Sounds good. I can do that.

[...]

PR #232, opened 2018-09-12
Convert stats and diagnostics to new API

ColCarroll:
These changes get the test suite for `diagnostics.py` to pass, with everything being xarray based. Now will need to work on `stats.py`, which will remove (most of) the pandas requirements.  I am working on that, but will probably need help for some of the functions (#231) .

Note that most functions in `diagnostics.py` work with either a dataset, or a numpy array.


ColCarroll:
I _think_ this is now ready to go. There's some missing documentation, but also this will allow us to delete most of `utils.py`.  Note that `compare` and summary could now work with pystan objects, though I took out some of the arguments to make them simpler and maybe less flexible. The summary also does not look great, but it is computationally fine - I think this should go in, and we add an issue to improve pretty-printing.

image

image

arviz/stats/stats.py
```diff
@@ -221,8 +237,9 @@ def _ic_matrix(ics, ic_i):
     for idx, val in enumerate(ics.index):
         ic = ics.loc[val][ic_i]
         if len(ic) != rows:
-            raise ValueError('The number of observations should be the same '
-                             'across all models')
+            raise ValueError(
+                "The number of observations should be the same " "across all models"
```
canyon289: Not picky, redundant parentheses here?

ColCarroll: Fixed - I actually ran `black` on the file as an experiment - this might have been part of the result.

ahartikainen:
LGTM

arviz/stats/diagnostics.py
```diff
@@ -1,29 +1,28 @@
...
-
-from ..utils import trace_to_dataframe, get_varnames
+from ..utils.xarray_utils import convert_to_dataset
 
 
 __all__ = ['effective_n', 'gelman_rubin', 'geweke']
 
 
-def effective_n(trace, varnames=None, round_to=2):
+def effective_n(data, *, var_names=None):
```
junpenglao: I didnt know you can use `*` as input - what does it do?

canyon289: It prevents people from using positional arguments

canyon289:
LGTM as well

PR #234, opened 2018-09-13
Add log_like to pymc3 sample_stats

ColCarroll:
I pulled this out of a larger PR I am working on to port diagnostics and stats to work with xarray. This takes `utils.log_post_trace`, and extracts it into the xarray object when reading a pymc3 object in.  @ahartikainen perhaps this could be added to `pystan` by having an argument giving the name of the generated quantity?

This implementation takes the coordinates for the observed variable, if the exist, and apply it to the likelihood.  For example, in the 8 schools model, the likelihood is 4 x 500 x 8 (chain x draw x school).

image

### DIFF
--- a/arviz/utils/xarray_utils.py
+++ b/arviz/utils/xarray_utils.py
@@ -271,6 +271,41 @@ class PyMC3Converter:
         self.dims = dims
 
     @requires('trace')
+    def _extract_log_likelihood(self):
+        """Compute log likelihood of each observation.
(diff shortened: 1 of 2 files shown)


ColCarroll:
Merging this since it blocks further work on porting stats.py (I am making sure `stats.loo` gives the same answer now as before) - can revert or tidy up in followups if there is a problem with it.

PR #235, opened 2018-09-13
expose fast_kde

aloctavodia:
Sometimes I find useful to have access to `fast_kde`, at the sometime it should not be of interested to most users hence the underscore.


ahartikainen:
LGTM

Issue #213, opened 2018-09-08
Make exception for invalid coords more user friendly


canyon289:
I think it'd be nicer if it suggested to the user they need to fix the coords argument which is hard to suss out from the exception 6 layers deep in the stack

```
import arviz as az
data = az.load_arviz_data('non_centered_eight')
az.traceplot(data, var_names=('theta_t'), coords={'theta_t_dim_0_t': [0, 1]})
```

<img width="1025" alt="screen shot 2018-09-07 at 6 33 02 pm" src="https://user-images.githubusercontent.com/7213793/45248990-488c6380-b2cd-11e8-8833-0560625e06b9.png">

ColCarroll:
Fixed by #236

PR #236, opened 2018-09-13
Add nicer exception message for mistake in coords argument

canyon289:
Fix for https://github.com/arviz-devs/arviz/issues/213

If you guys like how this looks I'll change all the other plots to this pattern. Did a couple as an example

```
# Sample code
import arviz as az
az.traceplot(data, var_names=('theta_t'), coords={'theta_t_dim_0': [0, 1, 100]})
az.traceplot(data, var_names=('theta_t'), coords={'theta_t_dim_0_t': [0, 1]})
```

### DIFF
--- a/arviz/plots/plot_utils.py
+++ b/arviz/plots/plot_utils.py
@@ -247,3 +247,26 @@ def xarray_to_nparray(data, *, var_names=None, combined=True):
         unpacked_var_names.append(make_label(var_name, selection))
 
     return unpacked_var_names, np.array(unpacked_data)
+
+
+def get_coords(data, coords):
+    """Subselects xarray dataset object to provided coords. Raises exception if fails
+
+    Raises
+    ------
+    ValueError
+        If coords name or dims are not available in data
+
+    Returns
+    -------
+    data: xarray
+        xarray.Dataset object
+    """
+    try:
+        return data.sel(**coords)
+
+    except ValueError as err:
+        raise ValueError("Verify coords keys. {}".format(err))
+
+    except KeyError as err:
+        raise KeyError("Verify coords values. {}".format(err))
--- a/arviz/plots/plot_utils.py
+++ b/arviz/plots/plot_utils.py
@@ -255,7 +255,10 @@ def get_coords(data, coords):
     Raises
     ------
     ValueError
-        If coords name or dims are not available in data
+        If coords name are not available in data
+
+    KeyError
+        If coords dims are not available in data
 
     Returns
     -------
--- a/arviz/plots/plot_utils.py
+++ b/arviz/plots/plot_utils.py
@@ -250,7 +250,7 @@ def xarray_to_nparray(data, *, var_names=None, combined=True):
 
 
 def get_coords(data, coords):
-    """Subselects xarray dataset object to provided coords. Raises exception if fails
+    """Subselects xarray dataset object to provided coords. Raises exception if fails.
 
     Raises
     ------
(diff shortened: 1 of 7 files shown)


arviz/plots/plot_utils.py
```diff
@@ -247,3 +247,29 @@ def xarray_to_nparray(data, *, var_names=None, combined=True):
...
+    Returns
+    -------
+    data: xarray
+        xarray.Dataset object
+    """
+    try:
+        return data.sel(**coords)
+
+    except ValueError as err:
+        raise ValueError("Verify coords keys. {}".format(err))
```
ColCarroll: can you verify that the stack trace is still informative? I think the pattern is

```
except ValueError as err:
    raise ValueError("Verify coords keys") from err
```

I might also print out more information - like "make sure {list(coords.keys()} match with the data data_vars ({list(data.data_vars)})"

arviz/plots/plot_utils.py
```diff
@@ -247,3 +247,29 @@ def xarray_to_nparray(data, *, var_names=None, combined=True):
...
+
+    except KeyError as err:
```
ColCarroll: ...are these right? A `KeyError` gets thrown for bad `.values` and a `ValueError` for bad `keys`?

canyon289: Maybe not. I'll double check. Thanks for double checking

canyon289: So yes,
Having a wrong key in coords dictionary raises a ValueError in xarray, and having a value in the coords dictionary raises a KeyError.

I wrote tests but just double checked by hand as well

Ill make the exceptions more clear

arviz/plots/plot_utils.py
canyon289: If you do from err the text from the original exception is lost. I was keeping it in because it was informative as to which coordinates were missing. 

But I agree with your second statement which means the above point will be irrelevant! I'll make the changes and post another example.

canyon289:
There's a gist below of both xarray defined ValueError and KeyError, and also two Options at the bottom to consider for arviz

Let me know which one you guys like, (or any other suggestions to try) and I'll implement throughout the library.

https://gist.github.com/canyon289/cbd3744a36d8198c6d4ac796eae60e4b

ahartikainen:
Could we have suggestion where to go to see correct coord/dim format?

For new users these things can be complicated.

Our docs + xarray docs?

canyon289:
I agree it is complicated! There's two separate issues, one is the format, but the other is just trying to select something that doesn't exist.

Maybe we should have three exceptions?
1. Format is invalid
2. Format is valid but coord name doesn't exist
3. Format is valid but dim doesn't exist

ahartikainen:
That does sound reasonable.

I think if we add any links, they should be on their own lines (or something users can click):

See.
https://linktoxarrayprops.com/...

canyon289:
Turns out there's four paths to three types of exceptions all shown in the gist.

I took the strategy of letting xarray raise the exception before handling it for two reasons
1. Xarray's exceptions have useful information as to what might be wrong so its worth capturing
2. I don't want arviz to raise exceptions in cases when xarray would not. If arviz handles the exception first we can't be certain of this.

https://gist.github.com/canyon289/51281affde8e8a8891d503e45dea0920

In regards to @ahartikainen's comment, I do agree we need an "xarray 101" type documentation. For this PR though would you guys be ok with merging this (once I update all plots) and I'll follow up with the documentation, and link in exception, in another PR?

ColCarroll:
Looks good to me - is this all the places `dataset.sel(**coords)` happens (even if it isn't, feel free to merge and we can followup with the rest).

ahartikainen:
LGTM

canyon289:
Updated to all the places with **sel. I'll merge if tests pasts. Thank you

PR #240, opened 2018-09-16
Remove a lot of old code

ColCarroll:
This removes a bunch of code that was pymc3 -> dataframe specific


canyon289:
This is a great PR

PR #244, opened 2018-09-17
Rearrange data modules

ColCarroll:
Fairly big reorganization. The major difference here is that the `utils` module is removed, in favor of a `data` module. Additionally, there is `arviz.data.convert` which contains all the functions for converting stuff into xarray and netcdf objects. This should let the conversion functions breathe a bit more (`xarray_utils` was pushing 1k lines).

Two followups might be:
- Rename plotting functions `plot_trace`, `plot_autocorr`, etc,
- Think about where the functions in `plot_utils` ought to go: `xarray_var_iter` is used everywhere, for example.


ColCarroll:
(merged so no one has to work off a branch)

PR #248, opened 2018-09-18
Rename api for plots and io

ahartikainen:
io contains now

    io_pymc3.py
        from_pymc3 --> InferenceData
    io_pystan.py
        from_pystan --> InferenceData

`convert_to_inference_data` now contains `**kwargs` and they are passed for `from_pymc3` and `from_pystan` functions.

All the plots are imported to `plots.__init__` as 

    from xyzplot import xyzplot as plot_xyz

This will keep the names of the files and everything normal.

### DIFF
--- a/arviz/data/convert/__init__.py
+++ b/arviz/data/convert/__init__.py
@@ -1,5 +1,5 @@
 """Code to convert data into xarray and netCDF format."""
 from .base import numpy_to_data_array, dict_to_dataset
 from .converters import convert_to_dataset, convert_to_inference_data
-from .from_pymc3 import pymc3_to_inference_data
-from .from_pystan import pystan_to_inference_data
+from .io_pymc3 import from_pymc3
+from .io_pystan import from_pystan
--- a/arviz/data/convert/io_pystan.py
+++ b/arviz/data/convert/io_pystan.py
@@ -277,6 +277,8 @@ def from_pystan(*, fit=None, prior=None, posterior_predictive=None,
                 observed_data=None, log_likelihood=None, coords=None, dims=None):
     """Convert pystan data into an InferenceData object.
 
+    Parameters
+    ----------
     fit : StanFit4Model
         PyStan fit object.
     prior : dict
@@ -312,7 +314,11 @@ def from_pystan(*, fit=None, prior=None, posterior_predictive=None,
         A dictionary containing the values that are used as index. The key
         is the name of the dimension, the values are the index values.
     dims : dict[str, List(str)]
-        A mapping from variables to a list of coordinate names for the variable
+        A mapping from variables to a list of coordinate names for the variable.
+
+    Returns
+    -------
+    InferenceData object
     """
     return PyStanConverter(
         fit=fit,
--- a/arviz/data/convert/io_pystan.py
+++ b/arviz/data/convert/io_pystan.py
@@ -280,7 +280,7 @@ def from_pystan(*, fit=None, prior=None, posterior_predictive=None,
             `prior_dict = prior_fit.extract(pars=prior_vars, permuted=False)`
         For PyStan 2.17 and earlier:
             `prior_dict = prior_fit.extract(pars=prior_vars)`
-            `prior_dict = {var : az.from_pystan.unpermute(vals) for var, vals in prior_dict.items()}`
+            `prior_dict = {k : az.from_pystan.unpermute(v) for k, v in prior_dict.items()}`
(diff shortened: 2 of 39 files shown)


ahartikainen:
Is there something we need to fix for the `doc`?

ColCarroll:
I wouldn't worry about that here - the API doc will pick it up automatically, but all the examples will need to change.

ColCarroll:
*maybe* a sphinx directive too, but i can wrestle with that.

arviz/plots/__init__.py
```diff
@@ -1,15 +1,15 @@
...
-from .kdeplot import kdeplot, _fast_kde, _fast_kde_2d
-from .parallelplot import parallelplot
-from .posteriorplot import posteriorplot
-from .traceplot import traceplot
-from .pairplot import pairplot
-from .jointplot import jointplot
-from .khatplot import khatplot
-from .ppcplot import ppcplot
-from .violintraceplot import violintraceplot
+from .autocorrplot import autocorrplot as plot_autocorr
```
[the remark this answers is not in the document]
ahartikainen: Yes we can.

It will affect plots/__init__ only.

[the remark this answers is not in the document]
canyon289: Can we leave the module name the same? I feel like renaming the functions should be sufficient.

arviz/plots/__init__.py
```diff
@@ -1,15 +1,15 @@
...
+from .compareplot import compareplot as plot_compare
+from .densityplot import densityplot as plot_density
+from .energyplot import energyplot as plot_energy
+from .forestplot import forestplot as plot_forest
+from .kdeplot import kdeplot as plot_kde, _fast_kde, _fast_kde_2d
```
canyon289: If were exposing `_fast_kde`, and `_fast_kde_2d` for users we should remove the underscore to make it obvious its meant for end users.

If we don't intent for users to use this method we should remove it from the import

arviz/data/convert/__init__.py
```diff
@@ -1,5 +1,5 @@
 """Code to convert data into xarray and netCDF format."""
 from .base import numpy_to_data_array, dict_to_dataset
 from .converters import convert_to_dataset, convert_to_inference_data
-from .from_pymc3 import pymc3_to_inference_data
-from .from_pystan import pystan_to_inference_data
+from .io_pymc3 import from_pymc3
```
canyon289: I think the IO naming is redundant. Made a PR with an alternative proposal

arviz/data/convert/io_pystan.py
```diff
@@ -268,9 +268,47 @@ def to_inference_data(self):
         })
 
 
-def pystan_to_inference_data(*, fit=None, prior=None, posterior_predictive=None,
-                             observed_data=None, log_likelihood=None, coords=None, dims=None):
-    """Convert pystan data into an InferenceData object."""
+def from_pystan(*, fit=None, prior=None, posterior_predictive=None,
```
canyon289: This documentation is great. Thank you!

canyon289:
Thank you for getting this renaming started!

ColCarroll:
Felt guilty for merging my changes, and then got a little caught up renaming the plot functions, and now I feel bad that #249 will need to be rebased, but I am now happy with these changes. Looking at other PR now.

canyon289:
Don't worry! I can deal with a rebase. I'll get to it when I get back home
in 1.5 hours

arviz/plots/__init__.py
```diff
@@ -1,15 +1,15 @@
...
-from .jointplot import jointplot
-from .khatplot import khatplot
-from .ppcplot import ppcplot
-from .violintraceplot import violintraceplot
+from .autocorrplot import plot_autocorr
+from .compareplot import plot_compare
+from .densityplot import plot_density
+from .energyplot import plot_energy
+from .forestplot import plot_forest
+from .kdeplot import plot_kde, _fast_kde, _fast_kde_2d
```
canyon289: Are these supposed to be private at the in __init__.py?

ahartikainen: Yes, I think they should be, because they are not meant for the average user.

I think they still propagate to main namespace?

canyon289: Yes but only because we import them into __init__.py I believe. I think if we don't import them here they won't propagate to the main namespace.

In either case I'm not too worried about it now and don't want to block this for something this minor, especially since "private" isn't even a real thing in Python.

PR #252, opened 2018-09-19
fix dict order with OrderedDict and streamline plot_utils functions

ahartikainen:
Let's go with the odict. It is easiest solution.

At least I usually assume that the order of input equals output.

I streamlined a few functions, nothing big.


ColCarroll:
Thanks, this is very nice.

ahartikainen:
Hmm, this did not the problem, so I think we should remove the odict from there.

I think the problem is that xarray sorts our variables? Or we have dict somewhere.

Maybe we need to change the test so it sorts input based on the names?

PR #255, opened 2018-09-19
Faster plot ppc

ahartikainen:
This version does kde plot in one step with all the data. Basically it does plot_kde manually.

### DIFF
--- a/arviz/plots/ppcplot.py
+++ b/arviz/plots/ppcplot.py
@@ -1,6 +1,6 @@
 """Posterior predictive plot."""
 import numpy as np
-from .kdeplot import plot_kde
+from .kdeplot import plot_kde, _fast_kde
 from .plot_utils import _scale_text, _create_axes_grid, default_grid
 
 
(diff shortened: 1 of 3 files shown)


ahartikainen:
Could someone test the integercase also for cumulative, should it be steps-pre or steps-post? What does that level mean?

Also for cumulative, should it start from 0?

ahartikainen:
Actually, the edgecase is still broken, so I will need to fix it.

arviz/plots/ppcplot.py
```diff
@@ -1,10 +1,10 @@
 """Posterior predictive plot."""
 import numpy as np
-from .kdeplot import plot_kde
+from .kdeplot import plot_kde, _fast_kde
 from .plot_utils import _scale_text, _create_axes_grid, default_grid
 
 
-def plot_ppc(data, kind='kde', alpha=0.2, mean=True, figsize=None, textsize=None, data_pairs=None):
+def plot_ppc(data, kind='density', alpha=0.2, mean=True, figsize=None, textsize=None, data_pairs=None):
```
aloctavodia: update the docstring. We should also check if the user is passing valid arguments (density or cumulative)

PR #258, opened 2018-09-20
Remove private method

canyon289:
Also name all arviz imports relative imports

This one was only was mostly due to my OCD. I hope you guys don't judge me too much

### DIFF
--- a/arviz/data/converters.py
+++ b/arviz/data/converters.py
@@ -2,7 +2,7 @@
 import numpy as np
 import xarray as xr
 
-from arviz.data.inference_data import InferenceData
+from .inference_data import InferenceData
 from .base import dict_to_dataset
 from .io_pymc3 import from_pymc3
 from .io_pystan import from_pystan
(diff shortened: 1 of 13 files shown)


arviz/plots/__init__.py
```diff
@@ -4,7 +4,7 @@
 from .densityplot import plot_density
 from .energyplot import plot_energy
 from .forestplot import plot_forest
-from .kdeplot import plot_kde, _fast_kde, _fast_kde_2d
```
aloctavodia: I introduced this on purpose #235. Sometimes is useful to have access to `_fast_kde` and `_fast_kde_2d`. Although they should be of no interest to most users so I kept the underscore.

canyon289: Ah I see, my apologies. I'll undo this change

aloctavodia: Thanks! Don't worry, lot of changes recently and is hard to follow all of them.

canyon289:
Made the updates

PR #259, opened 2018-09-20
Remove odict and fix test

ahartikainen:
This removes unneeded odict and fixes the order in tests.

### DIFF
--- a/arviz/tests/test_data.py
+++ b/arviz/tests/test_data.py
@@ -270,6 +270,32 @@ class TestPyStanNetCDFUtils(CheckNetCDFUtils):
                                 }
                            )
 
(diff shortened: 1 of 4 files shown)


arviz/tests/test_data.py
```diff
@@ -270,6 +270,32 @@ def get_inference_data(self):
                                 }
                            )
 
+    def get_inference_data2(self):
```
ColCarroll: didn't we already merge this?

ahartikainen: Yeah, artifact from rebasing. No idea why it stayed, but didn't have time to fix it (I didn't want to break this pr).

PR #260, opened 2018-09-20
Kde docs

canyon289:
Fix up documentation for KDE docs and example
Fix typo in PPC plot documentation

### DIFF
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -55,7 +55,59 @@ def plot_kde(values, values2=None, cumulative=False, rug=False, label=None, bw=4
     -------
     ax : matplotlib axes
 
+    Examples
+    --------
+
+    Plot default KDE
+
+    .. plot::
+        :context: close-figs
+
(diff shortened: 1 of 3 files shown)


arviz/plots/kdeplot.py
```diff
@@ -46,7 +46,7 @@ def plot_kde(values, values2=None, cumulative=False, rug=False, label=None, bw=4
         Keywords passed to the fill under the line (use fill_kwargs={'alpha': 0} to disable fill).
         Ignored for 2D KDE
     rug_kwargs : dict
-        Keywords passed to the rug plot. Ignored if rug=False or for 2D KDE
+        Keywords passed to the rug plot. Ignored if rug=False for 2D KDE
```
aloctavodia: the or is right in the sense that `rug_kwargs` will be ignored either if rug=False or if the plot is a 2D KDE

canyon289: My mistake, thank you for the review. Will make the change

PR #263, opened 2018-09-21
clean namespace

aloctavodia:
I noticed that it was possible to have access to numpy and pandas from ArviZ or that we have, for example  `jointplot` and `plot_joint` at the same level. This fix it. Not sure if this is the best way, but it's one way :-)

### DIFF
--- a/arviz/data/__init__.py
+++ b/arviz/data/__init__.py
@@ -1,4 +1,4 @@
-#pylint: disable=wildcard-import
+# pylint: disable=wildcard-import
 """Code for loading and manipulating data structures."""
 from .inference_data import InferenceData
 from .io_netcdf import load_data, save_data, load_arviz_data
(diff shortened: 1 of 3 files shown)


ahartikainen:
I think this is the correct solution. LGTM.

arviz/data/__init__.py
```diff
@@ -1,8 +1,13 @@
-#pylint: disable=wildcard-import
+# pylint: disable=wildcard-import
```
canyon289: Theres no wildcard import here anymore so can't we get rid of this?

aloctavodia: I guess you are right!

canyon289: Ah this is merged, No worries I can made the change and upload a PR. Will do that right now

canyon289: Haha its no problem. This is a very minor thing.

Issue #274, opened 2018-09-23
Do not use pyplot


ColCarroll:
This is considered an anti-pattern, and the axes handling can get messed up with the global namespace.

junpenglao:
Is there a page for current best practice?

canyon289:
I've found these docs/posts to be useful
http://pbpython.com/effective-matplotlib.html
https://matplotlib.org/tutorials/introductory/lifecycle.html
https://matplotlib.org/faq/usage_faq.html

PR #279, opened 2018-09-24
Convert all plots to object oriented mpl

canyon289:
Starting down the path of converting plots to object oriented matplotlib syntax per https://github.com/arviz-devs/arviz/issues/274. Please feel free to edit this branch if you have spare time and want to help out.

I plan to have all plots converted by Friday 9 pm pst at the latest.

Updated Pairplot
image

### DIFF
--- a/arviz/plots/pairplot.py
+++ b/arviz/plots/pairplot.py
@@ -118,7 +118,7 @@ def plot_pair(data, var_names=None, coords=None, figsize=None, textsize=None, ki
         ax.tick_params(labelsize=textsize)
 
     if gs is None and ax is None:
-        plt.figure(figsize=figsize)
+        fig, _ = plt.subplots(0, 0, figsize=figsize)
         gs = gridspec.GridSpec(numvars - 1, numvars - 1, wspace=0.05, hspace=0.05)
 
         axs = []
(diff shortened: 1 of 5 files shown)


arviz/plots/pairplot.py
```diff
@@ -128,7 +128,7 @@ def plot_pair(data, var_names=None, coords=None, figsize=None, textsize=None, ki
             for j in range(i, numvars - 1):
                 var2 = _posterior[j + 1]
 
-                ax = plt.subplot(gs[j, i])
+                ax = fig.add_subplot(gs[j, i])
```
junpenglao: Since the shape of the subplot is known (numvars-1, numvars-1), why not just create `fig, ax = plt.subplots(numvars-1, numvars-1, figsize=figsize)` above and index to the right one here?

canyon289: It'll then create all the other empty plots in the upper triangular

canyon289: Is this implementation what you are referring to?
https://github.com/canyon289/arviz/tree/alternative_pairplot

Example image
https://user-images.githubusercontent.com/7213793/45938966-bb5f3500-bf83-11e8-9461-eda0054c3539.png

junpenglao: I see, agree with you - thanks for the explanation!

ColCarroll:
Thanks! This looks great.

PR #280, opened 2018-09-24
[Demo] Why not use plt.subplots(numvars-1, numvars-1)  

canyon289:
@junpenglao Demonstration per your PR comment. 

> Since the shape of the subplot is known (numvars-1, numvars-1), why not just create fig, ax = plt.subplots(numvars-1, numvars-1, figsize=figsize) above and index to the right one here?

Is this implementation what you were referring to?

image


canyon289:
Er realized I didn't need to create a PR for this right after I did. Sorry for the noise

PR #282, opened 2018-09-25
Update AutoCorrPlot test

canyon289:
One question I have,
Does it make sense to have all the plot tests take a pymc3.MultiTrace and a StanFit4Anon model as an input, and instead have az.InferenceData as an input?

I ask because these objects get converted to InferenceData right at the top of most plots, and I vaguely feel like we should have `(pymc3.MultiTrace, StanFit4Anon) -> Inference Data` be its own test and decouple the object conversion from "does it plot" test?

I could make arguments about repeated computation, and more atomic tests, but if someone has strong(er) feelings let me know.

### DIFF
--- a/arviz/tests/test_plots.py
+++ b/arviz/tests/test_plots.py
@@ -43,12 +43,6 @@ class TestPlots(SetupPlots):
                               combined=combined, lines=[('mu', {}, [1, 2])])
             assert axes.shape == (2, 2)
 
-    def test_plot_autocorr(self):
-        for obj in (self.pymc3_fit, self.stan_fit):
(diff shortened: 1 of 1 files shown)


aloctavodia:
Having plotting and conversion as separated tests is cleaner, but I am afraid this could lead to miss errors. I prefer to be overly-caution and keep the tests as they are right now as plots are intended to interact with PyMC3 and Stan objects and not `Inference Data`.

canyon289:
Sounds good to me!

Thanks for the input

canyon289:
heads up https://github.com/arviz-devs/arviz/pull/283 needs to merge before this one

PR #283, opened 2018-09-25
Rename Class fixture attrs to be more descriptive

canyon289:
Makes it possible to know what object the attr is without having to run a debugger or read across modules


ColCarroll:
I'm fine with this when it passes (and will definitely improve debugging). PyMC3 examples and documentation is pretty uniform on using `trace` as the variable name (though I agree it could be changed to be better).

PR #285, opened 2018-09-26
Functionalize tests for test_plots.py

canyon289:
Converted all the tests to follow functional fixture pattern. This is a big change so please criticize or discuss as much as needed.


ColCarroll:
This looks fantastic!

PR #288, opened 2018-09-27
CmdStan support

ahartikainen:
Add support for CmdStan.

~No tests yet. Should we import a CmdStan output.csv?~
Example csv files are static. This way we don't need to install CmdStan on travis.

~Also currently I use glob for reading files, but I think I could implement regex there or is glob too much.~

~Glob fails:~

   ~`output*.csv` --> output1.csv, output2.csv output1_diagnostics.csv etc.~

One needs to be careful with glob, but in general, it is a nice feature.

     `output*.csv` --> output1.csv, output2.csv output1_diagnostics.csv

     `output[0-9].csv` --> output1.csv, output2.csv


ahartikainen:
Class has 11 init variables, apparently too much for lint. Can we change it?

ahartikainen:
What would be the best way to add tests?

CmdStan outputs csv file for each run.

I have an example model which is just testing scalar, 1D vector and 2D matrix. We could test that it will import it correctly (shapes + means/std).  https://github.com/ahartikainen/misc_bayes/tree/master/cmdstan_example

Should we implement a working CmdStan installation + example files on another PR?

canyon289:
From a visual inspection if cmdstan outputs a csv we could just add the csv to version control. The risk is that cmdstan's output could change over time but if it doesn't perhaps the above solution would be sufficient

arviz/data/io_cmdstan.py
```diff
@@ -0,0 +1,417 @@
...
+
+class CmdStanConverter:
+    """Encapsulate CmdStan specific logic."""
+
+    # pylint: disable=too-many-instance-attributes
+
+    def __init__(self, *, output=None, prior=None, posterior_predictive=None,
+                 observed_data=None, observed_data_var=None,
+                 log_likelihood=None, coords=None, dims=None):
+        self.output = glob(output) if isinstance(output, str) else output
```
canyon289: I don't complete understand what glob is doing or what the expected input for output is. Not saying its not just unsure at the moment

ColCarroll: Looks like `output` is either a glob pattern (`output/*.csv`) or a list of strings, and similar with `prior`

arviz/data/io_cmdstan.py
```diff
@@ -0,0 +1,417 @@
...
+        if isinstance(output, str) and len(self.output) > 1:
+            msg = "\n".join("{}: {}".format(i, os.path.normpath(path)) \
+                            for i, path in enumerate(self.output, 1))
+            print("glob found {} files for 'output':\n{}".format(len(self.output), msg))
```
ColCarroll: can we use a logger if we need to print?

arviz/data/io_cmdstan.py
```diff
@@ -0,0 +1,417 @@
...
+                                for i, path in enumerate(prior, 1))
+                print("glob found {} files for 'prior':\n{}".format(len(prior), msg))
+        self.prior = prior
+        self.posterior_predictive = posterior_predictive
+        self.observed_data = observed_data
+        self.observed_data_var = observed_data_var
+        self.log_likelihood = log_likelihood
+        self.coords = coords if coords is not None else {}
+        self.dims = dims if dims is not None else {}
+        self.chains = len(output)
```
ColCarroll: this variable does not look like it is used anywhere else

arviz/data/io_cmdstan.py
```diff
@@ -0,0 +1,417 @@
...
+    -------
+    Dict
+        key, values pairs. Values are formatted to shape = (nchain, ndraws, *shape)
+    """
+    col_groups = defaultdict(list)
+    columns = dfs[0].columns
+    for col in columns:
+        key, *loc = col.split('.')
+        loc = list(map(int, loc))
+        loc = tuple([i-1 for i in loc])
```
ColCarroll: I think just `loc = tuple(int(i) - 1 for i in loc)` suffices here

arviz/data/io_cmdstan.py
ahartikainen: Oh, it might be true. Also it could be wrong with future changes.

arviz/data/io_cmdstan.py
ahartikainen: Do we implement logger?

ahartikainen: I'm not going to implement logger on this PR. Let's do it in another one.

ColCarroll:
Rebased after all the test fixes to see if that fixes everything.

ColCarroll:
Wow! I am going to merge and then make sure that `pip install` will not ship the test suite!

PR #290, opened 2018-09-28
Add plot density tests

canyon289:


arviz/tests/test_plots.py
```diff
@@ -12,6 +12,9 @@
                      plot_joint, plot_ppc, plot_violin)
 
 
+np.random.seed(0)
```
ColCarroll: does this seed the whole test suite? can we add a comment or something to that?

ColCarroll:
85% coverage!

ahartikainen:
Could we use `np.random.RandomState`

PR #296, opened 2018-09-28
Fix warnings on doc build

ColCarroll:
Docs aren't building which is breaking the build. This tackles a few of those warnings - I think the main problem is the `matplotlib.sphinxext.only_directives` not being found with matplotlib 3.0. I removed that locally and everything worked, but I think sphinx caches aggressively, so going to try on travis.

### DIFF
--- a/arviz/tests/test_plots.py
+++ b/arviz/tests/test_plots.py
@@ -21,12 +21,15 @@ def models(request):
         models = load_cached_models(draws=500, chains=2)
         pymc3_model, pymc3_fit = models['pymc3']
         stan_model, stan_fit = models['pystan']
+    return Models()
 
+
+@pytest.fixture(scope='function')
+def clean_plots(request):
     def fin():
         plt.close('all')
 
     request.addfinalizer(fin)
-    return Models()
 
 
 @pytest.fixture(scope='module')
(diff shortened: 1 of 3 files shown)


canyon289:
This is great! My one suggestion is use the "autouse" feature of pytest. This way you don't have to explicitly call out the fixture in each test

Also for pytest disabling the "not used" linting error in the module is fine
http://pythontesting.net/framework/pytest/pytest-fixtures-nuts-bolts/#autouse
steroids
https://docs.pytest.org/en/latest/fixture.html#autouse-fixtures-xunit-setup-on-steroids

arviz/tests/test_plots.py
```diff
@@ -21,12 +21,15 @@ class Models:
         models = load_cached_models(draws=500, chains=2)
         pymc3_model, pymc3_fit = models['pymc3']
         stan_model, stan_fit = models['pystan']
+    return Models()
+
 
+@pytest.fixture(scope='function')
```
canyon289: Using inline comments to be more targeted; you should be able to set this to be an autouse fixture here

doc/conf.py
```diff
@@ -44,7 +44,6 @@
     'sphinx.ext.autosummary',
     'sphinx.ext.viewcode',
     'sphinx.ext.githubpages',
-    'matplotlib.sphinxext.only_directives',
```
canyon289: Just curious, what was the desired change

canyon289: Nevermind Found the comment https://mail.google.com/mail/u/1/#inbox/FMfcgxvzKkwmXtvtShWhBDkNHGJqCMXw

ahartikainen:
After this is merged we can merge CmdStan. Or we can merge before this? Tests are passed for cmdstan.

ColCarroll:
i'm closing this in favor of a new pr

PR #302, opened 2018-09-29
Add emcee support

ColCarroll:
This adds support for [emcee](http://dfm.io/emcee/current/), and adds a model to the test suite.

I ended up not adding test support to too many plots: `emcee` does not "know" the names of its variables, so it is hard in `test_plots` to reuse the machinery too much. The test coverage for `io_emcee.py` is still 94% - I will add a small test to throw the two `ValueErrors` that are missing.

Note that now arviz can mostly recreate the plots from http://dfm.io/emcee/current/user/line/ :

```python
import arviz as az
from arviz.tests import helpers

sampler = models['emcee']
data = az.from_emcee(sampler, var_names=['ln(f)', 'b', 'm'])


az.plot_trace(data, trace_kwargs={'color': 'k'})
```
image

```python
az.plot_pair(data, plot_kwargs={'c': 'k', 'alpha': 0.05})
```
image

### DIFF
--- a/arviz/tests/test_data.py
+++ b/arviz/tests/test_data.py
@@ -8,6 +8,7 @@ from arviz import (
     convert_to_dataset,
     from_pymc3,
     from_pystan,
+    from_emcee,
 )
 from .helpers import eight_schools_params, load_cached_models, BaseArvizTest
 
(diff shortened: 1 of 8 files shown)


canyon289:
This is cool!

ColCarroll:
Oooh, there's also a surprise change in here for `plot_pair`, to allow datasets with no divergence data to use it.

arviz/tests/test_data.py
```diff
@@ -209,6 +210,30 @@ def get_inference_data(self):
             dims={'theta': ['school'], 'theta_tilde': ['school']},
         )
 
+class TestEmceeNetCDFUtils(BaseArvizTest):
```
canyon289: Not picky and not not a blocker but do want to ask. Do we want tests to be functional style or xunit style?

ColCarroll: Oof, I figured `BaseArvizTest` contained some extra tests, and I was pleased they all passed. 

What if we merge this, and I'll do a followup converting all the `NetCDFUtils` tests?

canyon289: Feel free to merge! It's great that there are tests. I just moreso wanted to understand where we *eventually* wanted the test style to be.

Issue #303, opened 2018-09-29
Tests running version python 3.7


canyon289:
Per @ahartikainen the tests are running in Python 3.7

I believe this is because we cache miniconda, which then skips the activation of the right python environment in our bash script.

https://github.com/arviz-devs/arviz/blob/master/scripts/install_miniconda.sh
https://travis-ci.org/arviz-devs/arviz/jobs/435001107#L1501

PR #304, opened 2018-09-29
Modify logic for CI env activation

canyon289:
Should fix https://github.com/arviz-devs/arviz/issues/303

I believe the issue was occurring due to the placement of `source activate` in the if else block.

https://travis-ci.org/arviz-devs/arviz/jobs/435001107#L607

### DIFF
--- a/scripts/create_testenv.sh
+++ b/scripts/create_testenv.sh
@@ -18,8 +18,9 @@ if [[ $* != *--global* ]]; then
         echo "Environment ${ENVNAME} already exists, keeping up to date"
     else
         conda create -n ${ENVNAME} --yes pip python=${PYTHON_VERSION}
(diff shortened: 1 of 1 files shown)


canyon289:
What does this line do `[[ $* != *--global* ]]`?

canyon289:
That makes sense but a follow up question is there something that's a "non root" conda env?

ColCarroll:
Er, this is the conversation where we originally added it (I cargo-culted that file from the pymc3 scripts): https://github.com/pymc-devs/pymc3/pull/1473#issuecomment-255619392

the `--global` flag will install stuff globally for conda, otherwise it creates a conda env.

canyon289:
Ah I see, its a command line switch for the bash script. Thanks for the explanation and link

PR #309, opened 2018-09-30
Add pyro converter

ColCarroll:
This adds pyro support. It is missing tests, but it is ready to be tested as soon as https://github.com/uber/pyro/issues/1426 is settled.

I have run this all manually, and it does ok. The below code gets you a centered eight schools model on this branch.

```python
import arviz as az
from arviz.tests import helpers

posterior = helpers.pyro_centered_schools(helpers.eight_schools_params(), 500, 500)

schools = np.array(['Choate', 'Deerfield', 'Phillips Andover', 'Phillips Exeter', 'Hotchkiss', 'Lawrenceville', "St. Paul's", 'Mt. Hermon'])
data = az.convert_to_inference_data(posterior,
                                    coords={'school': schools},
                                    dims={'theta': ['school'], 'obs': ['school']})

data
```
```
Inference data with groups:
	> posterior
	> observed_data
```

### DIFF
--- a/arviz/data/io_pyro.py
+++ b/arviz/data/io_pyro.py
@@ -1,0 +1,89 @@
+"""pyro-specific conversion code."""
+import numpy as np
+
+from .inference_data import InferenceData
+from .base import dict_to_dataset
+
+
+def _get_var_names(posterior):
+    """Extract latent and observed variable names from pyro.MCMC.
+
+    Parameters
+    ----------
+    posterior : pyro.MCMC
+        Fitted MCMC object from Pyro
+
+    Returns
+    -------
+    list[str], list[str]
+    observed and latent variable names from the MCMC trace.
+    """
+    sample_point = posterior.exec_traces[0]
+    nodes = [node for node in sample_point.nodes.values() if node['type'] == 'sample']
+    observed = [node['name'] for node in nodes if node['is_observed']]
+    latent = [node['name'] for node in nodes if not node['is_observed']]
+    return observed, latent
+
+
+class PyroConverter:
+    """Encapsulate Pyro specific logic."""
+
+    def __init__(self, posterior, *_, coords=None, dims=None):
+        """Convert emcee data into an InferenceData object.
+
+        Parameters
+        ----------
+        posterior : pyro.MCMC
+            Fitted MCMC object from Pyro
+        coords : dict[str] -> list[str]
+            Map of dimensions to coordinates
+        dims : dict[str] -> list[str]
+            Map variable names to their coordinates
+        """
+        self.posterior = posterior
+        self.observed_vars, self.latent_vars = _get_var_names(posterior)
+        self.coords = coords
+        self.dims = dims
+
+    def posterior_to_xarray(self):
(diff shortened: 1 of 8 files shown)


ColCarroll:
"but it is ready to be tested" means that it can be added to the existing feature and tested across all plots pretty quickly - the test model is already implemented.

ColCarroll:
It looks like the size of `pyro-ppl` is breaking the build. I'll have to think about this...

arviz/data/io_pyro.py
```diff
@@ -0,0 +1,89 @@
...
+    observed = [node['name'] for node in nodes if node['is_observed']]
+    latent = [node['name'] for node in nodes if not node['is_observed']]
+    return observed, latent
+
+
+class PyroConverter:
+    """Encapsulate Pyro specific logic."""
+
+    def __init__(self, posterior, *_, coords=None, dims=None):
+        """Convert emcee data into an InferenceData object.
```
canyon289: Docstring says emcee?

arviz/data/io_pyro.py
```diff
@@ -0,0 +1,89 @@
...
+            Map variable names to their coordinates
+        """
+        self.posterior = posterior
+        self.observed_vars, self.latent_vars = _get_var_names(posterior)
+        self.coords = coords
+        self.dims = dims
+
+    def posterior_to_xarray(self):
+        """Convert the posterior to an xarray dataset."""
+        from pyro.infer import EmpiricalMarginal
```
canyon289: Why does the import have to be here rather than at the top?

arviz/tests/test_plots.py
```diff
@@ -70,12 +71,11 @@ def fig_ax():
 @pytest.mark.parametrize("kwargs", [{"point_estimate": "mean"},
                                     {"point_estimate": "median"},
                                     {"outline": True},
-                                    {"colors": ["g", "b"]},
-                                    {"colors": "gb"},
+                                    {"colors": ["g", "b", "r"]},
```
canyon289: Thanks for catching this

scripts/create_testenv.sh
```diff
@@ -23,6 +23,12 @@ if [[ $* != *--global* ]]; then
     source activate ${ENVNAME}
 fi
 
+if [ "$PYTHON_VERSION" = "3.5" ]; then
```
canyon289: Why only test in one version of python? Does pyro not work in 3.6?

arviz/data/io_pyro.py
ColCarroll: Not making pyro a library requirement (the `io_pymc` does a similar trick)

scripts/create_testenv.sh
ColCarroll: they're good questions! leaving answers here in addition to code for posterity!  this manually grabs the python 3.5 and 3.6 wheels. it is super brittle, but the only way I can think of doing it while there's a massive download on pypi (torch 1.0 was just released, so that might make it easier?)

canyon289:
LGTM! Thanks for responses

PR #310, opened 2018-09-30
Increase tests for CmdStan

ahartikainen:
Also trims the source.


ahartikainen:
This is ready.

I edited `arviz/data/base.py` to only give coords that are found in dims.

arviz/data/io_cmdstan.py
```diff
@@ -372,15 +385,21 @@ def _read_output(path):
...
                 if line.startswith("#"):
                     timing_info.append(line)
                 else:
-                    msg = "Invalid input file. " \
-                          "Header information missing from combined csv. " \
-                          "Timing: {}".format(path)
-                    raise ValueError(msg)
+                    raise_timing_error = True
+                    break
+            no_elapsed_time = not any("elapsed time" in row.lower() for row in timing_info)
```
canyon289: This is interesting. What was happening here that required this conditional?

arviz/tests/saved_models/cmdstan/example_stan.data.R
```diff
@@ -0,0 +1,5 @@
+x <- 1
```
canyon289: Also just curious, is R required for cmdstan?

arviz/tests/test_data.py
```diff
@@ -359,33 +359,44 @@ def setup_class(cls):
...
+        ]
 
     def get_inference_data(self, output, **kwargs):
         return from_cmdstan(output=output, **kwargs)
 
     def test_sample_stats(self):
-        for _, path in self.paths.items():
+        for key, path in self.paths.items():
+            if 'missing' in key:
+                continue
```
canyon289: What does this do? Looks like it check's for missing then just keeps going?

canyon289: Nevermind, Shouldn't do code reviews before drinking coffee.

arviz/tests/test_data.py
```diff
@@ -431,15 +442,15 @@ def test_inference_data2(self):
             assert hasattr(inference_data, 'posterior_predictive')
             assert hasattr(inference_data, 'observed_data')
 
-    def test_inference_data3(self):
+    def test_inference_data_input_types2(self):
```
canyon289: Not required. For these tests a single line comment would be nice explaining the difference between this one and the other 3 similarly named tests. Makes it easier to figure out what the intent was for each one

arviz/data/io_cmdstan.py
ahartikainen: These are for combined csv `cat *.csv > comb.csv`

Assuming that time is missing it will either use break (end of the file) or after looping if it read header of the "second file"

arviz/tests/saved_models/cmdstan/example_stan.data.R
ahartikainen: No. `<-` is just part of Rdump format that cmdstan reads.

arviz/tests/test_data.py
ahartikainen: True. I can add them later.

ahartikainen:
I will merge this now.

There was a discussion should we do

    coords = {
        key : coord for key, coord in coords.items() \
        if any(key == dim for dim in dims)
    }

is arviz/base.py --> `generate_dims_coords`

But based on the docstring and in general what that function should produce

     Generate default dimensions and coordinates for a variable.

I think filtering non-valid coords is the right thing to do.

PR #311, opened 2018-10-01
Add from * methods to docs

canyon289:
To show of all the hard work you guys have been doing


ColCarroll:
Thanks! docs should update so we can actually see it, too :100:

PR #312, opened 2018-10-01
add clip_on for n_eff

ahartikainen:
Simple fix


ahartikainen:
Before 

image

After

image

After

image

After

image


ps. What is going on with the order here? After each plot I have restarted Notebook and imported eight_schools with arviz_.load_...? It randomly changes the order. Dataset order is static.

ColCarroll:
While you're here, do you want to add the fix in `plot_rhat`, too?

ahartikainen:
For plot_forest? It will be trickier.

ColCarroll:
Forgot the axes were set strangely - thanks for fixing!

Issue #246, opened 2018-09-17
Turn diagnostics into proper ufuncs


ColCarroll:
See https://docs.scipy.org/doc/numpy-1.15.0/reference/ufuncs.html

For a bunch of functions I "faked" a ufunc to allow xarray to broadcast calculations correctly, but careful rewriting should make cleaner, faster implementations of `effective_n`, `hpd`, `rhat`, to name a few.

canyon289:
After finishing the intro docs I'd like to give these a try

canyon289:
I don't quite understand this one. It looks like u_funcs are already used for effective_n. Is this implementation incorrect?

https://github.com/arviz-devs/arviz/blob/master/arviz/stats/diagnostics.py#L56

ColCarroll:
Despite opening the issue, I do not know much about this.

- I am not confident that the current implementation will work for higher dimensions, and automatically broadcast. 
- I am not even sure that a ufunc can require a 2d input and return 1d output. 
- The linked documentation is all Cython - I could not find anything on writing ufuncs in python. Do we gain a lot implementing it in Cython?

aloctavodia:
What about [Numba](https://numba.pydata.org/numba-doc/dev/user/vectorize.html)?

canyon289:
Hm for whoever takes a go at this, I would be really interested in understanding the details, and would be especially thrilled if someone would be willing to mentor me through implementing one of these myself!

Not required but figured I could ask

ColCarroll:
For what it is worth, I ported `bfmi` to a numba ufunc, and on a 4 x 500 energy array, I see about a 3x speedup.

Here is the code:

```python
@guvectorize([(float64[:], float64)], '(n)->()')
def bfmi(energy, res):
    n_rows = energy.shape[0]
    # moments
    first, second = 0, 0    
    squared_diff, trailing = 0, energy[0]
    
    for row in range(n_rows):
        first += energy[row]
        second += np.power(energy[row], 2)
        squared_diff += np.power(energy[row] - trailing, 2)
        trailing = energy[row]
    first = np.power(first, 2) / n_rows
    
    res = (n_rows * squared_diff) / ((n_rows - 1) * (second - first))
```

For reference, the current implementation is
```python
def bfmi(energy)
    energy_mat = np.atleast_2d(energy)
    return np.square(np.diff(energy_mat, axis=1)).mean(axis=1) / np.var(
        energy_mat, axis=1
    )
```

[...]

PR #314, opened 2018-10-03
fix hpd

aloctavodia:
HPD was computed over datapoints instead of over samples for 2D arrays


ColCarroll:
Thanks! we should confirm this did not break anything for #246, but tests still pass.

PR #315, opened 2018-10-03
Add attrs to datasets

### LINKED ISSUE #243: Actually use `attrs` to describe xarray datasets in convert functions
`attrs` are json metadata attached to xarray datasets - we could even put whole stan programs in there, and maybe pymc3 ones, using `inspect` if we're clever.

ColCarroll:
Fixes #243 

Heavily influenced by https://github.com/rasbt/watermark/blob/master/watermark/watermark.py

Essentially this gives some information about the machine and environment where a dataset was fit. This will add attributes by default, through the `dict_to_dataset` function. Additionally, you can pass in a *module* used for the inference (i.e., the `pymc3` from `import pymc3`) to get some version information there. 

I have regenerated the built-in data with this watermark. See the `Attributes` at the bottom of the repr here. I tried to not include information that is _too_ personally identifying (hostname, for example).

```python
centered.posterior


<xarray.Dataset>
Dimensions:  (chain: 4, draw: 500, school: 8)
Coordinates:
  * chain    (chain) int64 0 1 2 3
  * draw     (draw) int64 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 ...
  * school   (school) object 'Choate' 'Deerfield' 'Phillips Andover' ...
Data variables:
    mu       (chain, draw) float64 ...
    theta    (chain, draw, school) float64 ...
    tau      (chain, draw) float64 ...
Attributes:
    created_at:                 2018-10-03T13:43:45.415460
    inference_library:          pymc3
    inference_library_version:  3.5
    machine:                    x86_64
    platform:                   Linux-4.4.0-134-generic-x86_64-with-debian-st...
    python_implementation:      CPython
    python_version:             3.6.6
    python_compiler:            GCC 4.8.2 20140120 (Red Hat 4.8.2-15)
    architecture:               ['64bit', 'ELF']
    processor:                  x86_64
    libc_ver:                   ['glibc', '2.2.5']
    cpu_count:                  8
```


canyon289:
Is it possible to turn this off or will this happen by default each time a Pymc3 object is converted to Inference data?

I'm being paranoid about infosec here. There are situations where I would not want to post this much information about my system publicly

ColCarroll:
That's fair - Two "fixes" that I think would be a compromise - 

1. Default this to being on, but provide a boolean to turn off (`attrs=False` or similar)?  

2. On reflection, I only care about the timestamp and inference library. The inference library you already opt into (unless you use `az.from_*`). What if we only attached those ones? 

```
    created_at:                 2018-10-03T13:43:45.415460
    inference_library:          pymc3
    inference_library_version:  3.5
```
If there's interest in machine data in the future, we can just make it easier to add as an "opt-in", instead of a default.

ColCarroll:
This is a little less revealing now!
```
centered.posterior

<xarray.Dataset>
Dimensions:  (chain: 4, draw: 500, school: 8)
Coordinates:
  * chain    (chain) int64 0 1 2 3
  * draw     (draw) int64 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 ...
  * school   (school) object 'Choate' 'Deerfield' 'Phillips Andover' ...
Data variables:
    mu       (chain, draw) float64 ...
    theta    (chain, draw, school) float64 ...
    tau      (chain, draw) float64 ...
Attributes:
    created_at:                 2018-10-03T14:26:54.822913
    inference_library:          pymc3
    inference_library_version:  3.5
```

canyon289:
Very cool! I do think its a good feature to have in some cases! I'll do my best to remember you implemented this here so when we revisit this we can just copy and paste the code back in!

ahartikainen:
I think os could be good addition `unix`, `osx`, `win32`..

ColCarroll:
Merging this as at least a placeholder - system level info + timing could be cool eventually for benchmarking and setting expectations, but htis is a reasonable start.

PR #316, opened 2018-10-03
add code to generate logo

### LINKED ISSUE #298: Store ArviZ logo somewhere
This snippet will save three copies of the logo. We should probably put it somewhere.

```python
import numpy as np
import matplotlib
from matplotlib.transforms import Bbox
from matplotlib.path import Path
from matplotlib.patches import PathPatch
from scipy import stats

x = np.linspace(0, 1, 200)
pdfx = stats.beta(2, 5).pdf(x)

path = Path(np.array([x, pdfx]).transpose())
patch = PathPatch(path, facecolor='none', alpha=0)
plt.gca().add_patch(patch)

cmap = matplotlib.colors.LinearSegmentedColormap.from_list('', ["#00bfbf", "#00bfbf", "#126a8a"])

im = plt.imshow(np.array([[1, 0, 0], [1, 1, 0]]),
                cmap=cmap, interpolation='bicubic',
                origin='lower',extent=[0,1, 0.0, 5],
                aspect="auto", clip_path=patch, clip_on=True)

plt.axis('off')
plt.ylim(0, 5.5)
plt.xlim(0, 0.9)

bbox = Bbox([[0.75, 0.5], [5.4, 2.2]])

plt.savefig('logo_00.png', dpi=300,  bbox_inches=bbox, transparent=True)
plt.text(x=0.04, y=-0.01, s='ArviZ', fontdict={'name':'ubuntu mono', 'fontsize':62}, color='w')

plt.savefig('logo_01.png', dpi=300,  bbox_inches=bbox)
plt.savefig('logo_02.png', dpi=300,  bbox_inches=bbox, transparent=True)
```

aloctavodia:
This closes #298 I guess there is no need to add the actual image files, right?


ahartikainen:
Should this go under a new repo with some svg + different quality png / jpg?

aloctavodia:
Why a different repository? There is a line to generate a svg file. I think jpg is not recommended for this type of graphic. Do you mean to include the image files?

canyon289:
Same question. This matches the pattern in pymc3 which is nice.

I think there is value in having the files generated. This way if I want to link to an Arviz image on a blog, or facebook post or something, I can reference the image in the master branch. Makes it easier for others to use the logo elsewhere. Not needed for this PR, just wanted to mention the use case

https://github.com/pymc-devs/pymc3/tree/master/docs/logos

aloctavodia:
I can add the files, no problem!

ColCarroll:
Wow! This diff is now 71k lines. I assume that's just the format of the eps file, but can you confirm that the file sizes are still reasonable?

aloctavodia:
Wow! you are right the .eps file is 9.1 MB

ahartikainen:
It doesn't need to, just a multirepo - monorepo difference.

aloctavodia:
Now is smaller size, but still pretty large diff. Let me know if you want me to remove the .eps, probably having png and svg by default is enough. and if someone needs other format they can just run the script.

ColCarroll:
I think we can leave the .eps out, and maybe even leave out the .png, which is already in `/doc/_static/logo.png`, which is where the docs pick it up. I _love_ having the code that generates the logos checked in.

ColCarroll:
Yes. Let's do that in a separate PR, though (it involves editing `conf.py` in a few places)

ahartikainen:
LGTM

PR #317, opened 2018-10-03
Inference data cookbook

canyon289:
Added placeholders for the newer from_* methods to get the docs in faster


ColCarroll:
This is great! I'll link this from the other docs and readme.

PR #318, opened 2018-10-03
LRU Cache

MFreidank:
Hey everyone,

as hinted at before, this pull request has code that is *far from* being ready for merge. I am basically in the process of figuring out what the best way to add lru caching in arviz might look like.
As pointed out by @ahartikainen before -- it may be crucial to provide a convenient way for users to disable this feature. I am not yet sure if it is better to disable it on a *per-function* basis (e.g. by storing an attribute on the decorated function) or on a *global* basis across all functions. My current implementation serves the latter use-case. Also, I have a rather ugly global variable to control the cache -- not sure what a pretty solution for this would look like.
However, there is quite some flexibility to what I have here -- it is possible to disable caching and reenable it at a later point. 

Code I used to play with it:
```
from arviz.data.caching import f, g, CACHE_ENABLED
print(f(10, 20))
print(f.cache_info())
print(f(10, 20))
print(f.cache_info())
CACHE_ENABLED = False
print(f(10, 20))
CACHE_ENABLED = True
print(f.cache_info())
print(f(10, 20))
print(f(10, 20))
print(f(10, 20))
print(f.cache_info())
print(f.cache_info())
print(f.cache_info())
print(g(10, 20))
print(g(10, 20))
print(g(10, 20))

print(g.cache_info())
print(g.cache_info())
print(g.cache_info())
CACHE_ENABLED = False
print(f.cache_info())
```

... oh and I just randomly had the thought that this is overengineered, because users may just recover the wrapped function using `._wrapped` on a `lru_cache` decorated function. If you guys think that the use case where one wants to disable the cache is rare enough we may also simply decorate with `lru_cache`..

On a separate note: which functions do you think make the most sense? I think anything IO-bound (`data/io_*`) are probably good candidates.


ahartikainen:
io + stats + diagnostics and maybe plot_ppc?

ahartikainen:
What does `maxsize=128` mean? MB?

ColCarroll:
This is very cool - thanks! 
 A few random questions:
1. Would `plot_ppc` benefit from this? I thought most of the time was spent rendering plots.  
2. I wonder how equality of inputs is calculated. Is a pymc3 trace equal to itself? What if a random number generator advances somewhere? It sounds like you're already thinking about this, and I'm happy to wait for example code.

canyon289:
Pairplots take the longest to render by far, if we're trying to speed up plots. But it seems like this doesn't speed up rendering plots, just reading bits somewhere? When you say IO, io from what? RAM to cpu? Disk to ram?

To be clear as well the uncertainty is more me trying to learn than my doubts to the solution :)

ahartikainen:
io: from disk or from ram.

Because at least for huge models the bottleneck is to get data from fit --> inference_data due to `fit.extract` being slow (it's slow when model is wide --> one has 100k parameters, e.g. large log_likelihood). With lru_cache this input would not change so one could speed up plotting when user don't do inference_data object first

    az.plot_trace(fit)
    az.plot_forest(fit)

Let's assume that fit --> inference_data takes 5 minutes, then plotting multiple plots would take really long. (This is assuming that getting data from inference data is super fast)

I think what we could do is wrap `lru_cache` to az --> `az.lru_cache`  and have clearing and setting of the cache from there. We just need examples for the docstring how to do it.

MFreidank:
@ahartikainen if I understood you correctly, I might have implemented what you have in mind now, with explanations and examples in the docstring. Feel free to add suggestions or corrections if I misunderstood something entirely. A different question is whether some functions should use this decorator by default (I believe this might be useful for particularly IO-heavy stuff)

ahartikainen:
I think we close this PR for now and let's rethink this later when needed. We have new libraries and need to make sure there are no random components in the loading of the data. Also, the algorithms are revised at some points so lru_cache should be implemented after that.

PR #319, opened 2018-10-03
Update documentation to point to new notebooks

ColCarroll:
Small changes for consistency, but removed the roadmap and added the cookbook to the first page of the docs. Also linked the xarray stuff from the front page and from the cookbook.


aloctavodia:
LGTM

PR #324, opened 2018-10-04
Fix color bar problem and sort gallery alphabetically

canyon289:
Also remove extra cell in cookbook

Minimally reproducible example of what was causing problem. I generated docs locally and looked through the thumbnails. If you see something I missed let me know :)

```
import arviz as az

coords = {'school': ['Choate', 'Deerfield']}
az.plot_pair(centered, var_names=['theta', "mu", "tau"], kind='hexbin', coords=coords,
colorbar=True, divergences=True)

import matplotlib.pyplot as plt; plt.savefig("test.png", bbox_inches="tight")
```


ColCarroll:
Cool - thank you!

PR #328, opened 2018-10-06
var_name as a str

ahartikainen:
This is one proposal by fixing everything by hand.

Updated also tests for this behavior.


canyon289:
Ah I see what you're referring to in the issue.

PR #330, opened 2018-10-06
Save plots on testing

canyon289:
The code here lets you save plots on testing. If this functionality and interface is acceptable I'll write up documentation and submit in another PR. 

Can you guys try this out and let me know what you think? I have not tested on Windows yet but I'm pretty sure I handled all the problems that would arise from the file system being locked from people leave images open, or leaving Windows Explorer open in the images directory

I want to make it possible to develop plots for ArviZ in the following way.
1. Think of idea for a plot addition, feature, change, or new plot entirely
2. Write code in arviz
3. Write some side code to verify your plot is working
4. Write a test to generate image your code is working as expected
5. Skip step 3 because by writing the test you can generate the image and get the test done in one step!!

**Usage**
Standard behavior of running test and closing plots
```
pytest arviz/tests
```

Saves generated images in a default directory called test_images
```
pytest arviz/tests --save
```

Saves generated images in a user specified directory
```
pytest arviz/tests --save user_choice
```

Test a specific plot and save just that image
```
pytest arviz/tests --save user_choice - k "test_plot_forest_single_value"
```

### DIFF
--- a/arviz/tests/conftest.py
+++ b/arviz/tests/conftest.py
@@ -18,12 +18,9 @@ def save_figs(request):
     if fig_dir is not None:
         # Try creating directory if it doesn't exist
         print("Saving generated images in {}".format(fig_dir))
-        try:
-            os.mkdir(fig_dir)
-            print("Directory {} created".format(fig_dir))
 
-        except FileExistsError:
-            print("Directory {} exists".format(fig_dir))
+        os.makedirs(fig_dir, exist_ok=True)
+        print("Directory {} created".format(fig_dir))
 
         # Clear all files from the directory
         for file in os.listdir(fig_dir):
--- a/arviz/tests/conftest.py
+++ b/arviz/tests/conftest.py
@@ -18,22 +18,23 @@ def save_figs(request):
 
     if fig_dir is not None:
         # Try creating directory if it doesn't exist
+        print("Saving generated images in {}".format(fig_dir))
         try:
             os.mkdir(fig_dir)
             print("Directory {} created".format(fig_dir))
 
         except FileExistsError:
-            print("Directory {} already exists".format(fig_dir))
+            print("Directory {} exists".format(fig_dir))
 
         # Clear all files from the directory
         for file in os.listdir(fig_dir):
-            try:
-                os.remove(file)
-            except OSError:
-                print("Can't remove {}".format(file))
+            full_path = os.path.join(fig_dir, file)
 
-    else:
-        print("Not saving figures")
+            try:
+                os.remove(full_path)
+
+            except OSError:
+                print("Failed to remove {}".format(full_path))
 
     return fig_dir
 
--- a/arviz/tests/conftest.py
+++ b/arviz/tests/conftest.py
@@ -1,0 +1,40 @@
+"""
+Configuration for test suite
+"""
+import pytest
+import os
+
+
+def pytest_addoption(parser):
+    """Definition for command line option to save figures from tests"""
+    parser.addoption("--save", nargs="?", const="test_images",
+                     help="Save images rendered by plot")
+
+
+@pytest.fixture(scope="session")
+def save_figs(request):
+    """Enables command line switch for saving generation figures upon testing"""
+    fig_dir = request.config.getoption("--save")
+
+    if fig_dir is not None:
+        # Try creating directory if it doesn't exist
+        try:
+            os.mkdir(fig_dir)
+            print("Directory {} created".format(fig_dir))
+
+        except FileExistsError:
+            print("Directory {} already exists".format(fig_dir))
+
+        # Clear all files from the directory
+        for file in os.listdir(fig_dir):
+            try:
+                os.remove(file)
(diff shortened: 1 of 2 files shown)


canyon289:
Also I should change the print statements to logging. I'll create an issue for logging in general in ArviZ

arviz/tests/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+    if fig_dir is not None:
+        # Try creating directory if it doesn't exist
+        print("Saving generated images in {}".format(fig_dir))
+        try:
+            os.mkdir(fig_dir)
+            print("Directory {} created".format(fig_dir))
+
+        except FileExistsError:
+            print("Directory {} exists".format(fig_dir))
+
```
ahartikainen: Use 

`os.makedirs(path, exist_ok=True)`

No need to make a try-except here.

This will also enable multiple new subdirs.

arviz/tests/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+        # Clear all files from the directory
+        for file in os.listdir(fig_dir):
+            full_path = os.path.join(fig_dir, file)
+
+            try:
+                os.remove(full_path)
+
+            except OSError:
+                print("Failed to remove {}".format(full_path))
+
```
ahartikainen: Images are not locked, so overwriting is fine. Maybe catch exception when saving fails and then retry n-times (use maybe date-to-milliseconds as extra identifier)

canyon289: I think images lock on Windows if the user has it open in in Windows photo viewer but I could be wrong.

canyon289: Oh I also decided to remove the images up front rather than overwrite during testing.

The reason for this is that if I run a full test suite ~100 images will be generated, but I run one test afterward only one image will be generated and overwritten, and the other 99 would just remain in the directory.
This may lead to confusion because one directory could contain renderings from different version of ArviZ.

Figured it was cleaner to always clear the directory at the start of each test run.

Does this pattern make sense?

ahartikainen: Yes.

You can not remove file that is open, but you can overwrite it.

arviz/tests/test_plots.py
```diff
@@ -32,12 +33,27 @@ class Models:
 
 
 @pytest.fixture(scope='function', autouse=True)
-def clean_plots(request):
+def clean_plots(request, save_figs):
+    """Close plots after each test, optionally save if --save is specified during test invocation"""
     def fin():
+        if save_figs is not None:
+
+            # Retry save three times
```
canyon289: @ahartikainen Added plot save retry based on your previous comment. Does this implementation meet the ask?

canyon289:
No, I think that would be confusing because in that event you could end up with 4 plots at different time states and forget which plot came from which test run instance.

Trying to make sure that its unambiguous which plots were generated from a test run, by clearing all of them before testing, and only having one plot per test method after completion

ahartikainen:
Ok, in that case should it have some delay maybe from time.sleep??

canyon289:
Good idea!

arviz/tests/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+def pytest_addoption(parser):
+    """Definition for command line option to save figures from tests."""
+    parser.addoption("--save", nargs="?", const="test_images",
+                     help="Save images rendered by plot")
+
+
+@pytest.fixture(scope="session")
+def save_figs(request):
+    """Enable command line switch for saving generation figures upon testing."""
+    fig_dir = request.config.getoption("--save")
```
ColCarroll: it looks like if just `--save` is passed, `fig_dir` will be `None`, and nothing will get saved. Am I reading that right?

ColCarroll:
This looks awesome.

arviz/tests/conftest.py
canyon289: It'll default to a directory named test_images later. I did it this way to save users from typing in a a folder name if they didn't care, but if they wanted to do a side by side comparion they could check out one branch and name the directory `before_cool_feature` and then checkout their other branch and name it `after_cool_feature` to have both plots

Hope that makes sense

arviz/tests/conftest.py
```diff
@@ -0,0 +1,35 @@
+"""Configuration for test suite."""
+# pylint: disable=redefined-outer-name
+import os
+import pytest
+
+
+def pytest_addoption(parser):
+    """Definition for command line option to save figures from tests."""
+    parser.addoption("--save", nargs="?", const="test_images",
```
canyon289: @ColCarroll Defaulting to `test_images` is here

ColCarroll: Oop! Thanks. I'm good merging whenever you/Ari are, then.

canyon289: Wow this is a great idea. Jealous I didn't think of it :)

ahartikainen:
LGTM

PR #332, opened 2018-10-07
Fix potential bug in tests

canyon289:
I have a question on what the benefit of the conditional in the modified test is.

I commented it out and things seem to work alright. In it's original form I believe it was overriding the parameterization from the decorator, meaning that all the tests suite was really just repeating the same test over and over again. Is my understanding correct?


canyon289:
Ready for a review on this one, no rush of course

ColCarroll:
👍 thanks for keeping it tidy

canyon289:
Thanks for writing tests!

PR #333, opened 2018-10-07
figure autoscale

aloctavodia:
yet another try

### DIFF
--- a/arviz/plots/plot_utils.py
+++ b/arviz/plots/plot_utils.py
@@ -42,31 +42,42 @@ def _scale_fig_size(figsize, textsize, rows=1, cols=1):
     markersize : int
         markersize
     """
-    rc_width, rc_height = tuple(mpl.rc_params()['figure.figsize'])
-    rc_fontsize = mpl.rc_params()['font.size']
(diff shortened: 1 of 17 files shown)


arviz/plots/plot_utils.py
```diff
@@ -42,31 +42,42 @@ def _scale_fig_size(figsize, textsize, rows=1, cols=1):
...
+    if isinstance(rc_titlesize, str):
+        rc_titlesize = 16
+    if isinstance(rc_xt_labelsize, str):
+        rc_xt_labelsize = 14
+
     if figsize is None:
         width, height = rc_width, rc_height
-        width = width * cols
-        height = height * rows
+        sff = 1 if (rows == cols == 1) else 1.15
```
canyon289: How are these magic numbers decided?

aloctavodia: I was trying to make that for a figure with multiple subplots each subplot has the same size as a single figure.

canyon289: Got it!

ColCarroll:
Merging! It will be cool when #330 lands, and we can compare a bunch of plots for a change like this.

PR #334, opened 2018-10-08
[WIP] Add initial draft of autocorrelation documentation

canyon289:
I think one great opportunity with ArviZ is
1. write really great summary on each plot motivation and usage
2. Show how ArviZ implements each, 
3. Show easy to follow sample of the calculations
4.  Cite references to further reading

I took a run at this with Autocorrelation plot after lots of research, but I feel that it could be much better. I could definitely use heavy and super critical feedback, and once this is good enough to merge I can continue down the line until everything including the stats is well documented

In my opinion there's this great opportunity to put knowledge next to actual application and ArviZ has the opportunity to fill that gap!

### DIFF
--- a/doc/notebooks/AutoCorrelation.ipynb
+++ b/doc/notebooks/AutoCorrelation.ipynb
@@ -1,0 +1,212 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Autocorrelation Diagnostic"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Autocorrelation plots show the degree a time series signal is correlated with itself.  As a diagnostic for Markov Monte Carlo Chains they provides an estimate the indepedence of each draw from previous draws in the same chain. * This is important since draws from a mcmc chain that are highly correlated could indicate issues with chain convergence."
+   ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


ColCarroll:
This seems like a great idea, and a useful start! My only specific feedback is that I would like to see you critique a particular (or more than one) autocorrelation plot, and recommend what to do. In general, a Bayesian analysis might have a model that fails to fit the data, or an inference algorithm that fails to fit the model. Using Metropolis-Hastings on a high dimensional or hierarchical model should exhibit high autocorrelation, and indicate a problem of the second type. I am not sure how to come up with a problem of the first type. I would also make sure to show an autocorrelation plot that is fine, and say that it is!

My general thoughts on this are, given the nature of open source, I would not commit yourself (or ourselves :) ) to writing a whole book on model checking, but maybe we start by linking this in the docs and from the `az.autocorr` function documentation, and if/as it grows, add more scaffolding to supporting it?  

My other general thought is that prose notebooks are hard to go through code review, and I tend to support merging quickly and having followup PRs with you as the reviewer if there are nitpicks.

@aloctavodia may have thoughts on all this.

canyon289:
Per multiple discussions I'll come up with a proposal on 
* Expected end outcome
* Level of support needed to develop

and also propose a process for 
* creating documentation while isolating workflow from other ArviZ development
* Easy criticism and review of notebooks

PR #335, opened 2018-10-08
Update `az.summary`

ahartikainen:
Now it works for nD objects.

Default return type is "wide" format dataframe. Other options are "long" format dataframe and "xarray" xarray dataset.

~Also added `ddof=1` for std calculations for normal std and circ_std.~
Fixed also a bug where `circ_mean` was used also for `circ_std`.

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -588,7 +588,7 @@ def summary(data, var_names=None, fmt='wide', include_circ=None, stat_funcs=None
     include_circ : bool
         Whether to include circular statistics
     fmt : {'wide', 'long', 'xarray'}
-        Return format is either DataFrame {'wide', 'long'} or xarray.Dataset {'xarray'}.
+        Return format is either pandas.DataFrame {'wide', 'long'} or xarray.Dataset {'xarray'}.
     stat_funcs : None or list
         A list of functions used to calculate statistics. By default, the mean, standard deviation,
         simulation standard error, and highest posterior density intervals are included.
@@ -608,7 +608,7 @@ def summary(data, var_names=None, fmt='wide', include_circ=None, stat_funcs=None
 
     Returns
     -------
-    pd.DataFrame
+    pandas.DataFrame
         With summary statistics for each variable. Defaults statistics are: `mean`, `sd`,
         `hpd_3`, `hpd_97`, `mc_error`, `n_eff` and `Rhat`. `n_eff` and `Rhat` are only computed
         for traces with 2 or more chains.
@@ -754,7 +754,7 @@ def summary(data, var_names=None, fmt='wide', include_circ=None, stat_funcs=None
 
     joined = xr.concat(metrics, dim="metric").assign_coords(metric=metric_names)
 
-    if fmt.lower().startswith('wide'):
+    if fmt.lower() == 'wide':
         dfs = []
         for var_name, values in joined.data_vars.items():
             if len(values.shape[1:]):
@@ -771,7 +771,7 @@ def summary(data, var_names=None, fmt='wide', include_circ=None, stat_funcs=None
                 df = df.T
             dfs.append(df)
         return pd.concat(dfs)
-    elif fmt.lower().startswith('long'):
+    elif fmt.lower() == 'long':
         df = joined.to_dataframe().reset_index().set_index('metric')
         df.index = list(df.index)
         return df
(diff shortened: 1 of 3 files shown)


ahartikainen:
Ok, we are hitting a fixed bug in xarray. I will remove those `ddof` now and add a new issue so we can fix it later.

arviz/stats/stats.py
```diff
@@ -633,36 +635,25 @@ def summary(data, var_names=None, include_circ=None, stat_funcs=None,
...
-        def mc_error_ufunc(ary):
-            return _mc_error(ary.reshape(*ary.shape[:-2], -1).T, **kwargs)
-
-        return mc_error_ufunc
-
-    def make_hpd_ufunc(idx, **kwargs):
-        def hpd_ufunc(ary):
-            return hpd(ary.reshape(*ary.shape[:-2], -1).T, **kwargs)[..., idx]
+    if fmt is None:
+        fmt = 'wide'
```
ColCarroll: default is already `wide`

ahartikainen: Yes, but if users run this through script and insert None, then it will go to default.

ahartikainen: Okey, I think we can remove it. Let's assume our users know what they are doing.

arviz/tests/test_stats.py
```diff
@@ -60,6 +60,15 @@ def test_summary(include_circ):
     centered = load_arviz_data('centered_eight')
     summary(centered, include_circ=include_circ)
 
+@pytest.mark.parametrize('fmt', ['wide', 'long', 'xarray'])
```
canyon289: Excited to see parametrize!

arviz/stats/stats.py
```diff
@@ -587,6 +587,8 @@ def summary(data, var_names=None, include_circ=None, stat_funcs=None,
         Names of variables to include in summary
     include_circ : bool
         Whether to include circular statistics
+    fmt : {'wide', 'long', 'xarray'}
+        Return format is either DataFrame {'wide', 'long'} or xarray.Dataset {'xarray'}.
```
canyon289: I'm being so incredibly picky here
xarray.Dataset is prefixed with library, for consistency this could be pandas.DataFrame

Doesn't block a merge

arviz/stats/stats.py
```diff
@@ -633,36 +635,22 @@ def summary(data, var_names=None, include_circ=None, stat_funcs=None,
...
-        def mc_error_ufunc(ary):
-            return _mc_error(ary.reshape(*ary.shape[:-2], -1).T, **kwargs)
-
-        return mc_error_ufunc
-
-    def make_hpd_ufunc(idx, **kwargs):
-        def hpd_ufunc(ary):
-            return hpd(ary.reshape(*ary.shape[:-2], -1).T, **kwargs)[..., idx]
+    fmt_group = ("wide", "long", "xarray")
+    if not isinstance(fmt, str) or (fmt not in fmt_group):
```
canyon289: I think checking `str` object type here may be redundant here because of the check for `in fmt_group`

arviz/stats/stats.py
```diff
@@ -765,7 +753,40 @@ def hpd_ufunc(ary):
     metric_names.append("gelman-rubin statistic")
 
     joined = xr.concat(metrics, dim="metric").assign_coords(metric=metric_names)
-    return joined
+
+    if fmt.lower().startswith('wide'):
```
canyon289: Why starts with? If I read correctly fmt has to be exactly `"wide"`

arviz/stats/stats.py
ahartikainen: `.lower` for non str

arviz/stats/stats.py
ahartikainen: Yes, this is old feature.

arviz/stats/stats.py
canyon289: I don't quite understand. Is `.lower()` a method on non strings?

canyon289:
LGTM, feel free to merge after tests pass

arviz/stats/stats.py
canyon289: Oh I see, yes this makes sense, you're avoiding the exception that would be raised if a non string was passed

ahartikainen:
I will merge this now and we can later fix ddof if needed.

PR #337, opened 2018-10-09
Allow passing axes and linewidth to plot_posterior

ColCarroll:
This code used to throw an exception for `fig` being undefined, and another for passing `linewidth` twice. It now runs fine:

```python
fig, ax = plt.subplots(nrows=2, ncols=3, figsize=figsize, constrained_layout=True, sharex=True)
az.plot_posterior(rugby, var_names=['atts'], textsize=20, linewidth=6, ax=ax)
```
image


aloctavodia:
LGTM!

canyon289:
LGTM me as well. Merging

PR #338, opened 2018-10-09
fix summary for single chain and minor aesthetic tweaks

aloctavodia:
* Summary was failing for single chain posteriors
* Change the format for the hpd to only print significant digits
* Shorten some names (standard deviation --> sd, gelman-rubin statistics --> gelman-rubin). Why we change r-hat? It was not the proper name?
* add round_to argument (default to 2)
* I also change the indentation. Sorry about this, but I really dislike how the code was formatted with lines exclusively for parentheses and code indentation not matching previous level.

### DIFF
--- a/arviz/stats/diagnostics.py
+++ b/arviz/stats/diagnostics.py
@@ -108,7 +108,7 @@ def _get_neff(sample_array):
             rho_hat_t[t + 1] = (rho_hat_t[t - 1] + rho_hat_t[t]) / 2.
             rho_hat_t[t + 2] = rho_hat_t[t + 1]
         t += 2
-    ess = (n_chain * n_draws) / (-1. + 2. * np.sum(rho_hat_t))
+    ess = int((n_chain * n_draws) / (-1. + 2. * np.sum(rho_hat_t)))
     return ess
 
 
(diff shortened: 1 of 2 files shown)


aloctavodia:
BTW, we have "R-hat" and "Effective n" for foresplot and "effective samples"  and "gelman-rubin" for summary. We should unify the names and also the order of these statistics.

ahartikainen:
Is `gelman-rubin` the official name? At least prof Gelman dislikes that format (this is what I remember, I could be wrong about this)

Format changes looks good. This is more a personal thing than pep thing?

Effective n should probably be integer.

aloctavodia:
not sure about the official name, but not surprised Gelman dislikes it :-) 
right Effective n should be an integer

ahartikainen:
There should be a way to get raw values or rounded values. I'm fine that we have 2 or 3 decimals accuracy + eff_n int and then roundto=None gives raw values?

ColCarroll:
I say we follow https://avehtari.github.io/rhat_neff/rhat_neff.html . I think for functions that are meant for display (`summary`, `plot_forest`), using something readable is better ("R-hat" and "Effective samples"?), while internal functions might use stuff without spaces (`r_hat`, `effective_n`)

PR #340, opened 2018-10-09
plot_forest: Do not plot eff_n or r_hat by default

aloctavodia:
also change the order of plotting for eff_n and  r_hat and unify names with those proposed  for summary in #338


canyon289:
LGTM

PR #344, opened 2018-10-09
Allow more customization of forest and pairplot

ColCarroll:
This makes the forestplot layout a little nicer, and allows adjusting marker size in `plot_pair`.

### DIFF
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -107,7 +107,9 @@ def plot_forest(data, kind='forestplot', model_names=None, var_names=None, combi
                              ncols=ncols,
                              figsize=figsize,
                              gridspec_kw={'width_ratios': width_ratios},
-                             sharey=True)
+                             sharey=True,
+                             constrained_layout=True,
+                             )
 
     axes = np.atleast_1d(axes)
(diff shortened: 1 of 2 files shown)


arviz/plots/forestplot.py
```diff
@@ -107,7 +107,9 @@ def plot_forest(data, kind='forestplot', model_names=None, var_names=None, combi
                              ncols=ncols,
                              figsize=figsize,
                              gridspec_kw={'width_ratios': width_ratios},
-                             sharey=True)
+                             sharey=True,
+                             constrained_layout=True,
+                             )
```
canyon289: You and @aloctavodia have very different styles on end parentheses and braces :)

canyon289:
LGTM

arviz/plots/forestplot.py
ColCarroll: oy, I don't care one way or another, but know very nice people who do, and sometimes forget which convention I should be working with in which project!  if we can find the right pylint/black/pycodestyle check for that, happy to merge it.

aloctavodia: The dance of the lonely parenthesis.

PR #356, opened 2018-10-10
minor tweaks to examples and plot_posterior docstring

aloctavodia:
I made several few changes, like trying to make some of the plots look nicer, being consistent with quotes marks, use list to pass `var_names` instead of tuples (I think this is less prone to errors and also the docs says that a list is expected).


canyon289:
LGTM.

Issue #325, opened 2018-10-06
Improve `var_names` behavior


ColCarroll:
If var_names is a string: interpret it as a list of just that string

ahartikainen:
```
if isinstance(var_names, str):
    var_names = [var_names]
```

I have used this in io files.

canyon289:
We should make this a central function so we can alter input behavior more easily later

canyon289:
What are you referring to when you say manually fixing this?

I was referring to this
```
def get_var_names(var_names):
    if isinstance(var_names, str):
        var_names = [var_names]
    return var_names
```

ahartikainen:
That seems easier to handle and test. I will remove the current PR.

canyon289:
@ahartikainen Do you mind if I pick this one up?

ahartikainen:
Yes, you can take it :) notice that some plots need atleast two variables, but even then I think function should fail as normally with a list of one var.

I'm not sure if we should update our docstrings.

Issue #326, opened 2018-10-06
Misspecified var_names


ColCarroll:
In certain cases, if an element of `var_names` is not in the dataset, should we should throw an exception? Right not, it will plot any var_names it can find, and ignore the others. This behavior has actually been kind of nice when plotting a centered eight trace vs a non-centered eight trace, but confusing if you typo one of the var_names.

ahartikainen:
Could it throw warning? Sometimes there are "missing" parameters.

Also can do some word2vec similarity scoring what was probably meant? This could be helpful for complex names?

ColCarroll:
right now the behavior is equivalent to `data.posterior[var_names]` (and specifically, `data.posterior[var_names].sel(**coords)`, though, which I like: i.e., matching xarray's behavior.

canyon289:
Fair point, matching xarrays behavior is good idea! We can focus this discussion on how verbose we want to be to user.

michaelraczycki:
Is it still relevant? I'd like to add this check and warnings if it's still needed

OriolAbril:
Try someting like:

```
idata = az.load_arviz_data("rugby")
az.plot_posterior(idata, var_names=("atts", "unexisting_var"))
```

and see what happens

aloctavodia:
Just to update that the current behavior is to throw a KeyError

PR #363, opened 2018-10-12
Var_names handling

canyon289:
Fix for https://github.com/arviz-devs/arviz/issues/325

### DIFF
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -10,6 +10,7 @@ from .plot_utils import (
     xarray_var_iter,
     _create_axes_grid,
 )
+from ..utils import _var_names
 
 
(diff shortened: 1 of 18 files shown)


ColCarroll:
This is great as is - two other corner cases I've been thinking about:

- `var_names=['mu', 'tau', 'mu']`
- `var_names=['not', 'in', 'model']`

Current behavior is to include `mu` twice, and plot nothing, respectively. I think that's the best thing to do, but wanted to call attention to it.

canyon289:
That's this issue right? https://github.com/arviz-devs/arviz/issues/326

Just spitballing here but in this PR there's a function that has no xarray in scope, and then vars are actually selected from xarray in another function, where combine boolean can be passed as well

Maybe better to keep separate? ¯\_(ツ)_/¯

PR #366, opened 2018-10-13
Fix permuted loglik for PyStan tests

ahartikainen:
Fixes some edge cases with 2.17 in tests.


ahartikainen:
Ok, let's see if this works with 2.17.

PR #368, opened 2018-10-13
remove stats/summary in document

gentlerainsky:
- remove batches reference in document


ColCarroll:
Thanks a third time!

canyon289:
I should have caught this, Thanks for following up

PR #370, opened 2018-10-17
Start model zoo

ColCarroll:
This design borrows heavily from scikit-learn. 

I trained and uploaded a model [from the pymc3 docs to figshare](https://figshare.com/articles/radon_nc/7215344), and added the ability to download it using arviz. The first usage takes a little while, the second usage is cached. I think this feature could allow sharing models trained in many different libraries, which is super exciting.

Here is what the new "UI" looks like. Also pinging @fonnesbeck in case he has feedback on the description of the radon model (or other concerns):
```python
import arviz as az

print(az.list_datasets())
```
```pythontb
centered_eight
==============

A centered parameterization of the eight schools model. Provided as an example of a
model that NUTS has trouble fitting. Compare to `load_arviz_data("non_centered_eight")`.

The eight schools model is a hierarchical model used for an analysis of the effectiveness
of classes that were designed to improve students’ performance on the Scholastic Aptitude Test.

See Bayesian Data Analysis (Gelman et. al.) for more details.

local: /home/colin/projects/arviz/arviz/data/_datasets/centered_eight.nc

----------

non_centered_eight
==================

A non-centered parameterization of the eight schools model. This is a hierarchical model
where sampling problems may be fixed by a non-centered parametrization. Compare to
`load_arviz_data("centered_eight")`.

The eight schools model is a hierarchical model used for an analysis of the effectiveness
of classes that were designed to improve students’ performance on the Scholastic Aptitude Test.

See Bayesian Data Analysis (Gelman et. al.) for more details.

local: /home/colin/projects/arviz/arviz/data/_datasets/non_centered_eight.nc

----------

radon
=====

Radon is a radioactive gas that enters homes through contact points with the ground.
It is a carcinogen that is the primary cause of lung cancer in non-smokers. Radon
levels vary greatly from household to household.

This example uses an EPA study of radon levels in houses in Minnesota to construct a
model with a hierarchy over households within a county. The model includes estimates
(gamma) for contextual effects of the uranium per household.

See Gelman and Hill (2006) for details on the example, or
https://docs.pymc.io/notebooks/multilevel_modeling.html#Correlations-among-levels
by Chris Fonnesbeck for details on this implementation.

remote: https://ndownloader.figshare.com/files/13284311
```

### DIFF
--- a/arviz/data/datasets.py
+++ b/arviz/data/datasets.py
@@ -1,6 +1,4 @@
-"""
-Base IO code for all datasets. Heavily influenced by scikit-learn's implementation.
-"""
+"""Base IO code for all datasets. Heavily influenced by scikit-learn's implementation."""
 from collections import namedtuple
 import hashlib
 import itertools
@@ -121,6 +119,28 @@ def _sha256(path):
 
 
 def load_arviz_data(dataset=None, data_home=None):
+    """Load a local or remote pre-made dataset.
+
+    Run with no parameters to get a list of all available models.
+
+    The directory to save to can also be set with the environement
+    variable `ARVIZ_HOME`. The checksum of the dataset is checked against a
+    hardcoded value to watch for data corruption.
+
+    Run `az.clear_data_home` to clear the data directory.
+
+    Parameters
+    ----------
+    dataset : str
+        Name of dataset to load.
+
+    data_home : str, optional
+        Where to save remote datasets
+
+    Returns
+    -------
+    xarray.Dataset
+    """
     if dataset in LOCAL_DATASETS:
         resource = LOCAL_DATASETS[dataset]
         return load_data(resource.filename)
@@ -148,6 +168,7 @@ def load_arviz_data(dataset=None, data_home=None):
 
 
 def list_datasets():
+    """Get a string representation of all available datasets with descriptions."""
     lines = []
     for name, resource in itertools.chain(LOCAL_DATASETS.items(), REMOTE_DATASETS.items()):
 
--- a/arviz/data/datasets.py
+++ b/arviz/data/datasets.py
@@ -109,9 +109,9 @@ def _sha256(path):
     """Calculate the sha256 hash of the file at path."""
     sha256hash = hashlib.sha256()
     chunk_size = 8192
-    with open(path, "rb") as f:
+    with open(path, "rb") as buff:
         while True:
-            buffer = f.read(chunk_size)
+            buffer = buff.read(chunk_size)
             if not buffer:
                 break
             sha256hash.update(buffer)
--- a/arviz/data/datasets.py
+++ b/arviz/data/datasets.py
@@ -1,0 +1,162 @@
+"""
+Base IO code for all datasets. Heavily influenced by scikit-learn's implementation.
+"""
+from collections import namedtuple
+import hashlib
+import itertools
+import os
+import shutil
+from urllib.request import urlretrieve
+
+from .io_netcdf import load_data
+
+LocalFileMetadata = namedtuple("LocalFileMetadata", ["filename", "description"])
+
+RemoteFileMetadata = namedtuple(
+    "RemoteFileMetadata", ["filename", "url", "checksum", "description"]
+)
+_DATASET_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), "_datasets")
+
+LOCAL_DATASETS = {
+    "centered_eight": LocalFileMetadata(
+        filename=os.path.join(_DATASET_DIR, "centered_eight.nc"),
+        description="""
+A centered parameterization of the eight schools model. Provided as an example of a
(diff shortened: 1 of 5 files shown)


arviz/tests/test_data.py
```diff
@@ -30,6 +34,60 @@ def chains():
...
+
+
+def test_missing_dataset():
+    with pytest.raises(ValueError):
+        load_arviz_data("does not exist")
+
+
+def test_list_datasets():
+    datasets = list_datasets()
+    for key in ("centered_eight", "non_centered_eight", "test_remote", "bad_checksum"):
```
canyon289: It might be better to do a set compare. This way if the test fails it'll show all that are missing, not just report the first one it failed on
```suggestion
    for set("centered_eight", "non_centered_eight", "test_remote", "bad_checksum") == set(datasets)
```

arviz/tests/test_data.py
```diff
@@ -11,7 +11,11 @@
     from_pymc3,
     from_pystan,
     from_emcee,
+    load_arviz_data,
```
canyon289: These tests are great!

arviz/data/datasets.py
```diff
@@ -0,0 +1,183 @@
...
+from collections import namedtuple
+import hashlib
+import itertools
+import os
+import shutil
+from urllib.request import urlretrieve
+
+from .io_netcdf import load_data
+
+LocalFileMetadata = namedtuple("LocalFileMetadata", ["filename", "description"])
```
canyon289: Good use of named tuples

arviz/data/datasets.py
```diff
@@ -0,0 +1,183 @@
...
+
+The eight schools model is a hierarchical model used for an analysis of the effectiveness
+of classes that were designed to improve students’ performance on the Scholastic Aptitude Test.
+
+See Bayesian Data Analysis (Gelman et. al.) for more details.
+""",
+    ),
+}
+
+REMOTE_DATASETS = {
```
canyon289: This implementation of local and remote is very nicely done

arviz/data/datasets.py
```diff
@@ -0,0 +1,183 @@
...
+        return load_data(resource.filename)
+
+    elif dataset in REMOTE_DATASETS:
+        remote = REMOTE_DATASETS[dataset]
+        home_dir = get_data_home(data_home=data_home)
+        file_path = os.path.join(home_dir, remote.filename)
+        if not os.path.exists(file_path):
+            urlretrieve(remote.url, file_path)
+        checksum = _sha256(file_path)
+        if remote.checksum != checksum:
```
canyon289: Would it be beneficial to provide some instructions on how to delete the offending file to allow for a redownload? For newer users they might just keep retrying this command get and get frustrated that the checksum fails every time

arviz/tests/test_data.py
```diff
@@ -11,7 +11,11 @@
...
+    list_datasets,
+    clear_data_home,
 )
+from arviz.data.datasets import REMOTE_DATASETS, LOCAL_DATASETS, RemoteFileMetadata
```
canyon289: Consistency for absolute vs relative imports in ArviZ would be nice but not required.

arviz/tests/test_data.py
```diff
@@ -30,6 +34,60 @@ def chains():
     return 2
 
 
+@pytest.fixture(autouse=True)
+def no_remote_data(tmpdir):
+    """Delete all remote data and replace it with a local dataset."""
+    keys = list(REMOTE_DATASETS)
```
canyon289: Not required. This would be a good candidate for use of the `monkeypatch` fixture. This code effectively does the same thing so it's not a concern. Just wanted to mention it.

arviz/tests/test_data.py
```diff
@@ -30,6 +34,60 @@ def chains():
...
+    for key in keys:
+        del REMOTE_DATASETS[key]
+
+    centered = LOCAL_DATASETS["centered_eight"]
+    filename = os.path.join(str(tmpdir), os.path.basename(centered.filename))
+    url = "file://" + centered.filename
```
canyon289: No change required. `+` operator string formatting is an antipattern imo. Might be better to use `os.path.join` to avoid OS specific problems.

canyon289:
This is really cool!

arviz/tests/test_data.py
ColCarroll: Sorry, `list_datasets` is just a string representation of all the available datasets. This was making sure that at least the names of the datasets show up in that string. I have commented it a little more to make the intention clear.

arviz/data/datasets.py
ColCarroll: Good idea!

arviz/tests/test_data.py
ColCarroll: True! I ended up using `urlunsplit`, since the _scheme_ is `file`. I think this should make it OS portable.

canyon289:
LGTM! Thanks for the second round of comments

Issue #359, opened 2018-10-11
Markersize argument is unused in khat plot


canyon289:
Markersize input by user has no effect on plot as its overwritten in function

https://github.com/arviz-devs/arviz/blob/master/arviz/plots/khatplot.py#L40

canyon289:
Closed by https://github.com/arviz-devs/arviz/pull/371

PR #371, opened 2018-10-17
fix: markersize argument is unused for khat plots

MFreidank:
Addresses #359. My new approach only autoscales the markersize if it has not been explicitly specified by the user. Differently put: a user-specified markersize has preference over markersizes determined via autoscaling, otherwise autoscaled markersizes are used (as before).


canyon289:
Changes look good, thank you. You can use black to format everything to pass linting easily

MFreidank:
ready for merge?

canyon289:
LGTM

Issue #331, opened 2018-10-06
Add logging to ArviZ


canyon289:
Getting good logging in now will make our lives easier when debugging or trying to help users with ArviZ later

Area that might benefit from logging/capturing logs from other packages
1. Which matplotlib backend is being used
2. Logs during testing, which we can instruct users to run locally and then send to use for debugging
3. Capturing logs about which packages versions are being used

canyon289:
If taking on this pull request please don't use BasicConfig, and instead app to instantiate handlers manually. The reasoning for this is that while BasicConfig "does a lot for you" for this long term library the extra controls will come in handy down the line

riddell-stan:
You might look at what pymc3 and pystan do for this. They do the same 
thing; both avoid BasicConfig.

Logging is configured at the bottom of pymc3's __init__.py file: 
https://github.com/pymc-devs/pymc3/blob/35fa41d0fc48ce7fc919ad4e44504dd3d2d96365/pymc3/__init__.py

PR #372, opened 2018-10-17
init logging

gentlerainsky:
Try to implement #331
- log matplotlib backend

**PR In progress - need some idea on this**
e.g.
- What would you like to see in log?

### DIFF
--- a/arviz/__init__.py
+++ b/arviz/__init__.py
@@ -1,7 +1,6 @@
 # pylint: disable=wildcard-import,invalid-name,wrong-import-position
 """ArviZ is a library for exploratory analysis of Bayesian models."""
 __version__ = "0.2.1"
-from matplotlib import get_backend as mpl_get_backend
 from matplotlib.pyplot import style
 
 from .data import *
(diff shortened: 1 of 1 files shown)


arviz/__init__.py
```diff
@@ -1,8 +1,20 @@
 # pylint: disable=wildcard-import,invalid-name,wrong-import-position
 """ArviZ is a library for exploratory analysis of Bayesian models."""
 __version__ = "0.2.1"
+from matplotlib import get_backend as mpl_get_backend
```
canyon289: I dont think we should include this import for two reasons. One is that matplotlib should handle its own logging, the other is that this adds another object to the namespace

canyon289:
@gentlerainsky Thanks for submitting this. I agree with @ahartikainen's comments and left one inline.

canyon289:
@gentlerainsky are we holding you up on this pull request?

If you're looking for some ideas there's print statements in `io_cmdstan.py` and `conftest.py` that would be well suited to convert to logging

gentlerainsky:
@canyon289 @ahartikainen  Sorry that I'm quite for sometime. I've update the logging as you've suggested and I took the style from https://github.com/requests/requests/blob/master/requests/__init__.py#L128

For prints statement, I found only a couple of 'print' in the repo. so I'll try those first.

canyon289:
There's no rush, just wanted to be sure we weren't blocking you

canyon289:
@gentlerainsky Would you be ok if we merged this? Trying to unblock an issue that would use this logging code!

https://github.com/arviz-devs/arviz/issues/389

gentlerainsky:
@canyon289 Sorry. I've been occupied for the past few weeks. Yes, that would be better for now. Sorry again that I can't complete this.

canyon289:
@gentlerainsky I think you did a fine job! Thank you so much for your contribution. Just asking to be sure. 

Will double check and likely merge tomorrow

Thank you again

canyon289:
Merging, Thanks @gentlerainsky

PR #373, opened 2018-10-18
Change credible interval default to .42

canyon289:


aloctavodia:
We are using  .94 as the default for all plots and summaries.

ahartikainen:
I don't think we need to change this.

PR #374, opened 2018-10-22
DOC: add api examples for plot_ppc. #355

### LINKED ISSUE #355: Add API docs for plot_ppc
https://arviz-devs.github.io/arviz/generated/arviz.plot_ppc.html#arviz.plot_ppc

See Posterior Docs for example
https://arviz-devs.github.io/arviz/generated/arviz.plot_posterior.html#arviz.plot_posterior

mattboggess:
Adds api examples for the plot_ppc function. Closes #355 

Let me know if there are edits/other examples you want added. 

In my html builds, the legends were defaulting quite large and covering up a lot of the plots. I also used the provided radon example data rather than the eight schools since the distributions are less wonky and make for a cleaner example in my opinion. Let me know if you would prefer the eight schools to fit with the other examples on the site.

### DIFF
--- a/arviz/plots/ppcplot.py
+++ b/arviz/plots/ppcplot.py
@@ -37,6 +37,25 @@ def plot_ppc(
     Returns
     -------
     axes : matplotlib axes
+
+    Examples
+    --------
+    Plot the observed data KDE overlaid on posterior predictive KDEs.
+
+    .. plot::
+        :context: close-figs
+
+        >>> import arviz as az
+        >>> data = az.load_arviz_data('radon')
+        >>> az.plot_ppc(data)
+
+    Plot the overlay with empirical CDFs.
+
+    .. plot::
+        :context: close-figs
+
+        >>> az.plot_ppc(data, kind='cumulative')
+
     """
     for group in ("posterior_predictive", "observed_data"):
         if not hasattr(data, group):


mattboggess:
As I was looking through the plot, a lot of fixes and additional features came to mind that I thought would help improve the plot, I'd be happy to work on any/all of them that you think would be worthwhile and want to add. These are out of scope and off topic for this PR (with the exception of maybe the minor doc updates) so let me know if you'd rather I create a new issue to discuss.

DOC:
- data_pairs default behavior is not described (assumes observed and pp variables have the same name)
- could use a multi-observed variable model to showcase the auto grid feature of the function (I think it auto-grids right?)
- alpha and kind do not specify default behavior (these are obvious from the function declaration, but it might be good to just list in the description as well since some other obvious parameters have the default listed)
- data is not really 'array-like'? It should have a matching description as in plot_posterior referencing InferenceData convertability
- document passing extra arguments on to matplotlib

Functionality:
- add a var_names parameter to support subsets in multi-observed variable plots
- extend the kdes out to the edges of the x-axis
- Add an axes parameter
- Add parameters to customize color
- The current automatic x-axis labeling is potentially confusing with the legend labeling (it confused me at least). I would either recommend labeling the x-axis with the observed parameter name (as done currently), but make the legend labels not reference parameter names. Or I would not label the x-axis and keep the legend with parameter name references.
- Make the observed & pp mean lines solid, thicker, and bolder so they stand out from the pp KDEs more.
- Add a "sub-sample" parameter that would plot a random sub-sample of the posterior predictive distributions. This could really help with long plotting times when there are a lot of generated predictions.

ColCarroll:
This looks great to me, thank you!

I _think_ I had the same thoughts as you at some point - would you mind opening this comment as an issue for more discussion?

ColCarroll:
> I also used the provided radon example data rather than the eight schools since the distributions are less wonky and make for a cleaner example in my opinion.

I am happy someone is starting to use the extra provided datasets, and agree on the wonkiness! I think it would be nice for the 8 schools to produce 8 different plots with 1 point each, right?

ColCarroll:
just like that, you're in the docs: https://arviz-devs.github.io/arviz/generated/arviz.plot_ppc.html

mattboggess:
@ColCarroll You are totally right about the source of the wonkiness. I'll add it to the issue list. 

P.S. your latest pull request about sharing fitted models is a really cool idea!

PR #376, opened 2018-10-23
Add an implementation for tensorflow-probability

yaochitc:
An from_tfp method was added to the data module.
As the model structure is not included in the sample result of tfp, the InferenceData contains nothing but the posterior samples.
The convert method is not implemented because the list of samples are returned instead of library specific result.

### DIFF
--- a/arviz/data/io_tfp.py
+++ b/arviz/data/io_tfp.py
@@ -4,8 +4,10 @@ import numpy as np
 from .inference_data import InferenceData
 from .base import dict_to_dataset
 
+
 class TfpConverter:
     """Encapsulate tfp specific logic."""
+
     def __init__(self, posterior, *_, var_names=None, coords=None, dims=None):
         self.posterior = posterior
 
@@ -30,8 +32,6 @@ class TfpConverter:
             data[var_name] = np.expand_dims(self.posterior[i], axis=0)
         return dict_to_dataset(data, library=self.tfp, coords=self.coords, dims=self.dims)
 
-
-    """Encapsulate tfp specific logic."""
     def to_inference_data(self):
         """Convert all available data to an InferenceData object.
 
@@ -39,17 +39,11 @@ class TfpConverter:
         the `posterior` and `sample_stats` can not be extracted), then the InferenceData
         will not have those groups.
         """
-        return InferenceData(
-            **{
-                "posterior": self.posterior_to_xarray()
-            }
-        )
+        return InferenceData(**{"posterior": self.posterior_to_xarray()})
+
 
 def from_tfp(posterior, var_names=None, *, coords=None, dims=None):
     """Convert tfp data into an InferenceData object."""
     return TfpConverter(
-        posterior = posterior,
-        var_names = var_names,
-        coords=coords,
-        dims=dims
+        posterior=posterior, var_names=var_names, coords=coords, dims=dims
     ).to_inference_data()
--- a/arviz/data/io_tfp.py
+++ b/arviz/data/io_tfp.py
@@ -1,0 +1,55 @@
+"""Tfp-specific conversion code."""
+import numpy as np
+
+from .inference_data import InferenceData
+from .base import dict_to_dataset
+
+class TfpConverter:
+    """Encapsulate tfp specific logic."""
+    def __init__(self, posterior, *_, var_names=None, coords=None, dims=None):
+        self.posterior = posterior
+
+        if var_names is None:
+            self.var_names = []
+            for i in range(0, len(posterior)):
+                self.var_names.append("var_{0}".format(i))
+        else:
+            self.var_names = var_names
+
+        self.coords = coords
+        self.dims = dims
+
+        import tensorflow_probability as tfp
+
+        self.tfp = tfp
+
+    def posterior_to_xarray(self):
+        """Convert the posterior to an xarray dataset."""
+        data = {}
+        for i, var_name in enumerate(self.var_names):
+            data[var_name] = np.expand_dims(self.posterior[i], axis=0)
+        return dict_to_dataset(data, library=self.tfp, coords=self.coords, dims=self.dims)
+
+
+    """Encapsulate tfp specific logic."""
+    def to_inference_data(self):
+        """Convert all available data to an InferenceData object.
+
+        Note that if groups can not be created (i.e., there is no `trace`, so
+        the `posterior` and `sample_stats` can not be extracted), then the InferenceData
+        will not have those groups.
+        """
+        return InferenceData(
+            **{
+                "posterior": self.posterior_to_xarray()
+            }
+        )
+
+def from_tfp(posterior, var_names=None, *, coords=None, dims=None):
+    """Convert tfp data into an InferenceData object."""
+    return TfpConverter(
+        posterior = posterior,
+        var_names = var_names,
+        coords=coords,
+        dims=dims
+    ).to_inference_data()
--- a/arviz/tests/helpers.py
+++ b/arviz/tests/helpers.py
@@ -19,7 +19,6 @@ import tensorflow as tf
 from ..data import from_tfp
 
 
-
 def _emcee_neg_lnlike(theta, x, y, yerr):
     """Proper function to allow pickling."""
     slope, intercept, lnf = theta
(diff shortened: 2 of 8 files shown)


canyon289:
Thank you for this PR! Especially thrilled that you took the time to add tests.

I'm not super familiar with TensorFlow probability so I have to ask, do you know what the possible causes of this exception are?
https://travis-ci.org/arviz-devs/arviz/jobs/445022984#L1466

yaochitc:
> Thank you for this PR! Especially thrilled that you took the time to add tests.
> 
> I'm not super familiar with TensorFlow probability so I have to ask, do you know what the possible causes of this exception are?
> https://travis-ci.org/arviz-devs/arviz/jobs/445022984#L1466

Maybe due to the version, I'll try to fix it.

yaochitc:
Seems I found the reason. Both torch and tensorflow use a lot of TLS, so this error occured when importing them together. But I did't not find this error when I run pytest arvis/tests on my own computer.

yaochitc:
After searching through google, I found some description for this error. It is caused by the TLS limit of glibc, which has larger number in later version such as that installed on Ubuntu 16(my computer) and has a lower number in ealier version such as that installed on Ubuntu 14(used by travis ci).
Maybe there are two options to solve the problem:
1. Try to set the environment of travic ci, use higher version of glibc. I'm not sure if it is possible to do that.
2. Separate the tests of different implementation so that they can be executed in different process.

arviz/data/io_tfp.py
```diff
@@ -0,0 +1,49 @@
...
+import numpy as np
+
+from .inference_data import InferenceData
+from .base import dict_to_dataset
+
+
+class TfpConverter:
+    """Encapsulate tfp specific logic."""
+
+    def __init__(self, posterior, *_, var_names=None, coords=None, dims=None):
```
canyon289: I think this just needs to be * to prevent positional arguments. instead of accepting an arbitrary amount

arviz/tests/helpers.py
```diff
@@ -128,6 +132,55 @@ def pyro_centered_schools(data, draws, chains):
...
+def tfp_noncentered_schools(data, draws, chains):
+    """Non-centered eight schools implementation for tfp."""
+    del chains
+
+    def schools_model(num_schools, treatment_stddevs):
+        avg_effect = ed.Normal(loc=0.0, scale=10.0, name="avg_effect")  # `mu`
+        avg_stddev = ed.Normal(loc=5.0, scale=1.0, name="avg_stddev")  # `log(tau)`
+        school_effects_standard = ed.Normal(
+            loc=tf.zeros(num_schools), scale=tf.ones(num_schools), name="school_effects_standard"
+        )  # `theta_tilde`
```
canyon289: These comments are great!

arviz/tests/helpers.py
```diff
@@ -128,6 +132,55 @@ def pyro_centered_schools(data, draws, chains):
     return posterior
 
 
+def tfp_noncentered_schools(data, draws, chains):
+    """Non-centered eight schools implementation for tfp."""
+    del chains
```
canyon289: This is just my curiosity, why do chain needs to explicitly be deleted?

arviz/tests/helpers.py
```diff
@@ -11,8 +11,12 @@
 import pyro.distributions as dist
 from pyro.infer.mcmc import MCMC, NUTS
 import pystan
+import tensorflow_probability as tfp
```
canyon289: If you have some time could you try changing the import order to see if that fixes things, per your research and @ColCarroll's? 

If you don't that have time thats ok! We can do what @ColCarroll suggested which is merge and fix later

yaochitc: In fact I tried and got a segment fault error.

arviz/tests/helpers.py
yaochitc: Just to follow the other tests, no need in fact.

arviz/data/io_tfp.py
yaochitc: Well, I just tried to keep uniform with other converters here.

yaochitc:
Thanks for responding. As mentioned in the comment, I have tried solutions such as changing the import order and got a segment fault exception, and here is the output:
https://travis-ci.org/yaochitc/arviz/jobs/445472193

Maybe this should be solved by the community through some reconstruction.

ahartikainen:
This is a long shot.. could relative import help?

    from ... import ...

arviz/data/io_tfp.py
canyon289: Consistency is good!

The issue here is that this is not consistent. Check out the Pyro implementation
https://github.com/arviz-devs/arviz/blob/master/arviz/data/io_pyro.py#L82

In the current implementation the user can input an arbitrary amount of non keyword arguments, where instead we want to the user to input only one positional argument, and use keywords for the rest. The underscore makes all the difference!

This code snippet might make things more clear.
``` def test_one(posterior, *_, b=2):
    print(posterior)
    print(b)

test_one("posterior", 2, 3, 4, 5, 6, b=10)


def test_two(posterior, *, b=2):
    print(posterior)
    print(b)

test_two("posterior", 2, 3, 4, 5, 6, b=10) 
```

arviz/tests/helpers.py
canyon289: If it has no effect I would delete it. The other tests may need it because of the way the framework works. If tensorflow does not then no need to include it :)

arviz/tests/helpers.py
canyon289: That's frustrating, thank you for trying!

arviz/tests/helpers.py
ColCarroll: this is sort of a funny pattern to allow the function to have the same signature as the others, but make the intention of not using `chains` clear (see bottom of https://github.com/google/styleguide/blob/gh-pages/pyguide.md#214-decision). I think it is also used in the pyro example, where there are no chains. 

I like it in that it limits the number of ignored pylint warnings (my feelings about it are not strong).

canyon289: ah @yaochitc ignore my comment. Thanks all

arviz/data/io_tfp.py
canyon289: oh geez, yes you're right. I got my signals crossed. Ignore this entire chain.

yaochitc:
@ahartikainen Seems not, maybe a hard limit.

canyon289:
@yaochitc We're recognizing that inconsistent testing is a pain and are currently trying to see if docker will provide more consistency. I apologize for the delay

https://github.com/arviz-devs/arviz/pull/380

ahartikainen:
Before we get docker working, could you split tfp tests on its own file?

Let's fix test files after docker is running?

yaochitc:
@ahartikainen Sure, should I create separated helper and test_plots for tfp?
@canyon289 Ok, I'll check it.

ahartikainen:
@canyon289 had some idea for testing.

canyon289:
@yaochitc First off thanks again for your patience while we get everything in order, we really appreciate it.

Can two last things be done we merge this PR?
1. Add a test in test_data.py for tfp? This will ensure that the dataloading is working as expected, independent of any plots. The two links below are good reference implementations
https://github.com/arviz-devs/arviz/blob/master/arviz/tests/test_data.py#L306
https://github.com/arviz-devs/arviz/blob/master/arviz/tests/test_data.py#L340

2. Mark the failing tests from TravisCI as `xfail` in pytest. This way we get visibility both in the code, and in the test suite, that the tests are present but expected to fail, or conversely if they start passing that they're unexpected to pass.

The pytest docs have good examples
https://docs.pytest.org/en/latest/skipping.html#xfail-mark-test-functions-as-expected-to-fail
https://docs.pytest.org/en/latest/skipping.html#skip-xfail-with-parametrize

If the second one is too much let us know and we can take care of it on our end! Thanks again for contributing to ArviZ!

yaochitc:
Ok, I'll work on them.

yaochitc:
@canyon289 Hello, I have some questions when implementing.
1. Currently the InferenceData for tfp contains nothing but the posterior, seems nothing more can be added except something like this:
```
class TestTfpNetCDFUtils:
    @pytest.fixture(scope="class")
    def data(self, draws, chains):
        class Data:
            model, obj = load_cached_models(draws, chains)["tfp"]

        return Data

    def get_inference_data(self, data, eight_school_params):
        return data.obj

    def test_inference_data(self, data, eight_schools_params):
        inference_data1 = self.get_inference_data(data, eight_schools_params)
```

2.As the tls exception was cause be the import in helpers.py, I'm not sure if it will be solved by marking the  unit tests in test_plots.py, I'll have a try.

canyon289:
@yaochitc 
1. That's fine, having the test is preferable still to ensure the posterior functionality doesn't break, and also serve as a reference for anyone trying to understand the codebase

2. Based on your explanation for the second one I think you may be right. If you can try that'd be awesome. After that though if things still fail I think we'll just have to merge and make this a community problem, you've helped out a lot and we really appreciate it. @ahartikainen and @ColCarroll for a second opinion

ahartikainen:
Yes

if it won't work, let's fix tests later (high hopes for docker).

Have you (anyone?) tested this in a docker manually?

If all tests pass on a docker, I believe that we can merge this so user's could use it. << comment out imports and skip tests, right?

yaochitc:
canyon289 
The unit in test_data was added.
For the previous exception, I just checked the code, and found it difficult to just mark the tests because some logic are constructed through code instead of parameters, like this:
`    data = eight_schools_params()
    supported = (
        (tfp, tfp_noncentered_schools),
        (pystan, pystan_noncentered_schools),
        (pm, pymc3_noncentered_schools),
        (emcee, emcee_linear_model),
        (pyro, pyro_centered_schools),
    )`
I think some reconstruction should be done before I can help.

I am working at adding the posterior-predictive and observed to the module and will finish it soon.

canyon289:
Sounds good. I just merged docker testing into master. Once you're ready we can see if it works in there!

canyon289:
@yaochitc Good news! I rebased your branch onto one with docker testing and things seem to work! I did have to install tensorflow manually, perhaps it needs to be added to the requirements-dev.txt in addition to tensorflow probability.

yaochitc:
That's good. Please let me know if you need my help.

canyon289:
@yaochitc I wanted to again apologize for the wait. I promise we're working to get this merged in.

yaochitc:
@canyon289 That's fine. I've just finished adding posterior predictive,observed and sample stats to the module. I'll open another pr after this merged.

canyon289:
@yaochitc We added you as a coauthor here.
 https://github.com/arviz-devs/arviz/commit/22ce2494d51883c49af2518eec491794f0604566

I copied and pasted the username and email from the commits in this repo because github seems to get the attribution correct there
https://github.com/yaochitc/yaochitc.github.io/graphs/contributors

Thank you again for your contribution to the project, ArviZ is better because of you!

PR #378, opened 2018-10-25
smooth hpd

aloctavodia:
This add the option to smooth the output of a hpd, I hope this helps to get nicer plots and less confused students/audience. I did not do extensive testing but seems that the parameters I am using for `Savitzky-Golay` generally work, but I am open to suggestions. Should we let the user select those parameters? 

smooth


ahartikainen:
Before merge can you test that this "centering" the smoothing? E.g. try with spike func

    ary = [0, ..., 0, 1, 0, ..., 0]

canyon289:
Could I also ask for a test, at least one?

aloctavodia:
Thanks for the feedback. @ahartikainen I can confirm that this is centering the smoothing. @canyon289 I will add a test.

canyon289:
LGTM

Issue #375, opened 2018-10-23
plot_ppc fixes & improvements


mattboggess:
As I was looking through the plot_ppc function, a lot of fixes and improvements came to mind that I thought would help improve the plot. I'd be happy to work on any/all of them that you think would be worthwhile and want to add. There's a lot here and will likely want to be partitioned into multiple PRs/issues, but I wanted to get everything listed.

DOC:

- data_pairs default behavior is not described (assumes observed and pp variables have the same name)
- use a multi-observed variable model to showcase the auto grid feature of the function in an api example (I think it auto-grids right?). this would actually be the eight schools dataset with the fix in the first bullet point under functionality.
- alpha and kind do not specify default behavior (these are obvious from the function declaration, but it might be good to just list in the description as well since some other obvious parameters have the default listed)
- data is not really 'array-like'? It should have a matching description as in plot_posterior referencing InferenceData convertability
- document passing extra arguments on to matplotlib

Style:
- extend the kdes out to the edges of the x-axis
- add an axes parameter
- add parameters to customize color
- the current automatic x-axis labeling is potentially confusing with the legend labeling (it confused me at least). I would either recommend labeling the x-axis with the observed parameter name (as done currently), but make the legend labels not reference parameter names. Or I would not label the x-axis and keep the legend with parameter name references.
- make the observed & pp mean lines solid, thicker, and bolder so they stand out from the pp KDEs more.

Fixes & New Features:
- Fix grouped/vectorized parameter flattening:
  - observed variables that are vectorized should have the option to not be flattened across dimensions and instead plotted in a grid? This is why I think the eight school examples look kind of "wonky" since it's a mixture of multiple distributions with one observed data point each?
  - add a var_names parameter to support subsets in multi-observed variable plots.
- Provide sub-sample and uncertainty plotting options to reduce long plotting time with lots of generated predictions:
  - add a "sub-sample" parameter that would plot a random sub-sample of the posterior predictive distributions. This would be nice when there are a lot of generated predictions, but a subset would suffice for visualization.
  - add a parameter to allow plotting pp uncertainty shading with the pp mean instead of all of the KDEs.

ColCarroll:
The docs would be welcome additions. Here is what I was thinking about for multi-dimensional observations. In case of the radon dataset, note this would be 89 plots, each with a few observations!

![image](https://user-images.githubusercontent.com/2295568/47328310-a804bf80-d63e-11e8-9967-326ee0b73ae0.png)
[...]

PR #382, opened 2018-10-29
Fix parameter descriptions for plot_ppc #375

mattboggess:
Minor fixes to the plot_ppc parameter descriptions as noted in #375

### DIFF
--- a/arviz/plots/ppcplot.py
+++ b/arviz/plots/ppcplot.py
@@ -14,25 +14,30 @@ def plot_ppc(
 
     Parameters
     ----------
-    data : Array-like
-        Observed values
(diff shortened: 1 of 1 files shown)


mattboggess:
I'm not sure what went wrong with the CI? It terminated because it stalled for some reason on one of the builds. All tests and linters passed locally so I don't think it's my changes.

ColCarroll:
Oops - yes, I think canyon289 and ahartikainen are doing some work fixing up the CI (it is getting bogged down with all the inference libraries we need to install!)

This looks much nicer, thank you! Feel free to open PRs with prototypes if you are looking for feedback on further updates to the ppc plots.

PR #387, opened 2018-11-02
Trace divergence

### LINKED ISSUE #385: Add divergences to traceplot
A flag to add vertical and/or horizontal lines to the traceplot whenever there are divergences would be neat. That is `az.plot_trace(data, divergences=True)`.

I made a quick mockup here:
image

Inspired by [this slide](https://youtu.be/E8vdXoJId8M?t=685) from Jonah Gabry + Dan Simpson's recent talk, where Dan is saying that a traceplot "tells you *nothing*."

ColCarroll:
Fixes #385 

This adds vertical dashed lines wherever there are divergences. I have turned it on by default mostly to be provocative - I'm ok to have it off by default.  I sort of like that the markers are visually distracting whenever you *should* be distracted by divergences.

A lot of the shuffling of the code was done so that these lines (and the `lines` argument) would not change the x or y limits of the plots. Note that this happens on master right now!

Here is an example of the output for the centered and non-centered eight schools:

image

### DIFF
--- a/arviz/plots/traceplot.py
+++ b/arviz/plots/traceplot.py
@@ -89,7 +89,6 @@ def plot_trace(
         >>> coords = {'theta_t_dim_0': [0, 1], 'school':['Lawrenceville']}
         >>> az.plot_trace(data, var_names=('theta_t', 'theta'), coords=coords, lines=lines)
     """
-
     if divergences:
         try:
             divergence_data = convert_to_dataset(data, group="sample_stats").diverging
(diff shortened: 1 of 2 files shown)


junpenglao:
I think the Bayesflow style is a bit better as it is a bit cleaner: http://mc-stan.org/bayesplot/articles/visual-mcmc-diagnostics.html#mcmc_trace

aloctavodia:
I agree with @junpenglao the rug-plot style is cleaner, also no need to differentiate divergences from each chain, we can use the same color for all of them (I suggest using black by default).

ColCarroll:
Updated per comments. Here are the same two plots:
image

ahartikainen:
Could it take "bottom" / "top" / None or False

ColCarroll:
I had a plan to animate spinning divergences, but went with `{bottom | top | False}` instead :)

The default is "bottom", but in actuality, any truth-y argument will have divergences on the bottom, False-y will have none, and "top" will put them on top.

image

PR #392, opened 2018-11-05
WIP: Add parameter indexing & plot style support to plot_ppc #375

mattboggess:
This pull request adds two new sets of features to the plot_ppc function:

1. Variable subsetting: Added var_names, coords, and flatten parameters to allow for sub-selection and gridding of variables across dimensions.
2. Plot style control: Added obs_plot_style, pp_plot_style, and pp_flatten parameters to allow for better control of plotting styles. This was particularly in response to the "wonkiness" of density plots for datasets with few observations (like the current eight schools example).

Needs more work and clean up to be merged, but first wanted to get feedback on the new API design approach. Do these changes seem reasonable? Is there a better way to do it?

See below for examples and see the updated docstring for more in depth parameter descriptions.

### DIFF
--- a/arviz/plots/ppcplot.py
+++ b/arviz/plots/ppcplot.py
@@ -1,11 +1,20 @@
 """Posterior predictive plot."""
 import numpy as np
 from .kdeplot import plot_kde, _fast_kde
-from .plot_utils import _scale_fig_size, _create_axes_grid, default_grid
+from .plot_utils import (
+    xarray_var_iter,
+    _scale_fig_size,
+    default_grid,
+    make_label,
+    _create_axes_grid,
+    get_coords,
+)
+from ..utils import _var_names
 
 
 def plot_ppc(
-    data, kind="density", alpha=0.2, mean=True, figsize=None, textsize=None, data_pairs=None
+    data, kind="density", alpha=0.2, mean=True, figsize=None, textsize=None, data_pairs=None,
+    var_names=None, coords=None, flatten=None
 ):
     """
     Plot for Posterior Predictive checks.
@@ -38,6 +47,17 @@ def plot_ppc(
 
         If None, it will assume that the observed data and the posterior
         predictive data have the same variable name.
+    var_names : list
+        List of variables to be plotted. Defaults to all observed variables in the
+        model if None.
+    coords : dict
+        Dictionary mapping dimensions to selected coordinates to be plotted.
+        Dimensions without a mapping specified will include all coordinates for
+        that dimension. Defaults to including all coordinates for all
+        dimensions if None.
+    flatten : list
+        List of dimensions to flatten. Only flattens across the coordinates
+        specified in the coords argument. Defaults to flattening all of the dimensions.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


aloctavodia:
Thanks a lot for the PR and taking the time to contribute to ArviZ, I like were this PR is heading. Specially I like the idea of the 1D scatter (or rug-plot) for observed data, as using a KDE is certainly over-smoothing the data. And while this is also true for the posterior predictive distribution samples, if we just plot the posterior mean (an average over many samples) we hide the uncertainty of the predictions, something that is of central importance for posterior predictive checks. The  "wonkiness" of the density plots, while aesthetically not very pleasant, is a direct reflection of the uncertainty of making predictions with very few observations.

ColCarroll:
The pattern from the `radon` example is a pretty common one (at least in PyMC3) of passing a "ragged" array of observations for a number of categorical distributions (ragged meaning each category may have a different number of observations).

Consider running

```python
def gen_data(n):
    means = st.norm(2, 3).rvs(2)
    rvs = [st.norm(mean, 1) for mean in means]

    cats = st.randint(0, 2).rvs(n)
    values = np.array([rvs[cat].rvs() for cat in cats])
    return means, cats, values

means, cats, values = gen_data(1_000)
```

and then 

```python
    with pm.Model():
        means = pm.Normal('means', 2, 3, shape=2)
        obs = pm.Normal('obs', means[cats], 1, shape=sub_cats.shape[0], observed=values)
        trace = pm.sample()
        posterior = pm.sample_posterior_predictive(trace)
```

If I care about how well my model fit category 0 vs. category 1, I might have a hard time, even though that dimensional data can be encoded into the inference data object. I guess that is different from a rug plot vs. a kde though?

Just out of interest, here's the output of the above for a few numbers of observations:

image


image

image

mattboggess:
> And while this is also true for the posterior predictive distribution samples, if we just plot the posterior mean (an average over many samples) we hide the uncertainty of the predictions, something that is of central importance for posterior predictive checks. The "wonkiness" of the density plots, while aesthetically not very pleasant, is a direct reflection of the uncertainty of making predictions with very few observations.

@aloctavodia, you are absolutely right. I'm embarrassed at having gotten lost in the mechanics/aesthetics and mis-characterizing the "wonkiness"! I think the pp_flatten parameter should clearly return to just being an option to overlay the mean then. 

The kde for the posterior predictive samples does still present an issue as the number of observations approaches 1 though. For example, in the schools case, the current kde implementation errors when the number of observation is 3 or less. The question then would be whether to:
1. Fall back on a posterior mean only plot (potentially with a warning that this is misleading about the uncertainty) 
2. Catch and present the error to the user with a more informative error message about sample size
3. Another better alternative

Thoughts on which option would be preferred?

---

> If I care about how well my model fit category 0 vs. category 1, I might have a hard time, even though that dimensional data can be encoded into the inference data object. I guess that is different from a rug plot vs. a kde though?

@ColCarroll, regarding your comment if I understand correctly: I think that is what allowing the coord + flatten pairing can help address. A similar situation to your example using the radon dataset would be comparing the flattened variable to the unflattened for two of the counties (see below). By unflattening, you can see the separate variability for each county side by side. 

`az.plot_ppc(radon_data, coords={'observed_county': ['AITKIN', 'WRIGHT']}, flatten=[])`
split
`az.plot_ppc(radon_data, coords={'observed_county': ['AITKIN', 'WRIGHT']}, flatten=None)`
combined

In my opinion, this is separate from the kde vs. rug plot consideration for the most part other than allowing plotting of individual coordinates makes one more likely to produce plots with few observations where the rug plot vs. kde decision discussed above would come up. The eight schools is an extreme example where unflattening creates sample sizes of 1 which clearly causes a problems for kde plots. This example got me a bit confused which is why I think these multiple issues surrounding the plot got blended into this single PR in my head. Sorry for the confusion. I've tried to more clearly re-summarize below.

---
Thanks to both of you for your help clarifying everything! To sum up, these are the main considerations as I see them after incorporating your input:

1. Adding the var_names, coords, and flatten parameters to give users the ability to control indexing and flattening of variables and dimensions.
2. Determining how to handle the failure of the kde plots as the number of observations approaches 1.
3. Replacing the kind parameter with separate observed data and posterior predictive plot type parameters. This allows for adding more plot types for the observed data, notably a 1d scatter/rug option that can be more appropriate when there are not many observations. Having separate parameters gives more flexibility (pair an observed 'rug' plot with either a pp 'pdf' or 'cdf'), but it does replace the kind parameter and thus break the previous api. The other option would be to restrict to a single kind parameter where the posterior predictive plot type is determined automatically based on the observed plot type.

aloctavodia:
@mattboggess don't worry is easy to get loss. 

Regarding how to generate a `ppc plot` adapted to small datasets. One alternative is to use a _stacked rugplot_, something like this:

example

What do you think?

mattboggess:
Oh cool, this is a neat idea. I was hoping someone would have a good #3 alternative!

Let me play around with some examples.

aloctavodia:
I am glad you like it and thanks for taking the time to work on this. Probably with this type of plot the number of samples should be restricted, even if the user pass, let say 300 sample, we should probably only plot a stacked-rug for just a few, like 30-50 or so to avoid cluttering. Let us know what get from playing with some examples :-)

mattboggess:
Yeah that is exactly what I was worried about too and wanted to play around with. I'll see if I can think of some clever ways to scale, but I think you're right that there's probably going to always have to be a hard limit at some point for larger samples.

ColCarroll:
I really like how this plot looks. @aloctavodia might have thoughts about whether this should still be called `plot_ppc`, but I would definitely use something like this.  

- I agree with you that 30 seems like a reasonable default
- I love the different color for the actual observations vs the ppc samples
- It gives a good intuition for "what do samples from this distribution really look like?"



As a somewhat related aside, I've read both Allen Downey and Nate Silver talking about doing something like this with election results: each day for a month before the election, print results from a single "posterior predictive draw", and imagine what the news would look like if those were true election results.

aloctavodia:
I really like this!

This should definitively be called `plot_ppc`. Maybe in the future `plot_ppc` could be splinted in sub-functions but even then it would be nice to have a unified interface for ppc plots.

30 is the 42 of statistics, totally in favor! 

Fake news and post-truth are so 2018! We should have Posterior predictive news!!! and Posterior predictive truths!!!

arviz/plots/ppcplot.py
```diff
@@ -1,35 +1,70 @@
...
+)
+from ..utils import _var_names
 
 
-def plot_ppc(
-    data, kind="density", alpha=0.2, mean=True, figsize=None, textsize=None, data_pairs=None
-):
+def plot_ppc(data, var_names=None, coords=None, flatten=None,
+             obs_plot_style="auto", pp_plot_style="auto", pp_flatten="auto",
+             alpha=0.2, mean=True, figsize=None, textsize=None, data_pairs=None):
```
aloctavodia: `mean` is no longer and argument.

arviz/plots/ppcplot.py
```diff
@@ -68,38 +103,92 @@ def plot_ppc(
...
+
+    #if density_type.lower() not in ("pdf", "cdf"):
+    #    raise TypeError("`kind` argument must be either `pdf` or `cdf`")
 
     if data_pairs is None:
         data_pairs = {}
 
-    observed = data.observed_data
-    posterior_predictive = data.posterior_predictive
+    data_threshold = 30
```
aloctavodia: probably this should be an argument that defaults to 30

arviz/plots/ppcplot.py
mattboggess: See the new commits and discussion below. I reverted and fixed the API changes based on our discussion. The mean parameter is no longer gone.

arviz/plots/ppcplot.py
mattboggess: I agree. See my comment below. This is now num_pp_samples.

mattboggess:
Oops sorry. Github noob alert. I was getting divergence with my remote fork branch when I tried to rebase and I wasn't really sure why or what the best way to fix it was. I think I found the right fix now, but let me know if things still look wrong to you. 

I added a new commit with the random sampling we mentioned. It is really nice to repeatedly plot with different random samples, especially with small datasets. :) 

This should now be all of the API additions discussed so let me know if there is anything else that should still be modified. I'm going to get started on adding tests and examples.

To summarize, the three most recent commits add the following API additions:

- b8e961c: Adds num_pp_samples and random_seed parameters to allow for reproducible random sampling of posterior predictive samples to plot. Defaults to 30 for scatter plots and all of the samples otherwise.
- a27a4d1: Adds a 'scatter' option to the kind parameter to allow for a stacked scatter plot style. Default is still 'density'.
- 09c6744: Adds var_names, coords, and flatten parameters to give control of plotting across variables and dimensions subsets with or without flattening. Defaults to plotting all variables flattened across all dimensions to still match the previous default.

aloctavodia:
This is great. I think is is basically ready to merge once test pass and examples are added.

mattboggess:
I added inline examples to the API page. Should additional examples be added to the gallery as well?

Build failed on one of the jobs, but I don't think it is related to any of my changes? All tests and style checks are passing for me locally and on the first three jobs. I added an additional plot_ppc test that checks for correct # of axes when splitting across coordinates.

ColCarroll:
Thanks again, @mattboggess!

mattboggess:
Thank you both for all of the help and sorry it took a while!

This can close #375

PR #397, opened 2018-11-09
More docker test

canyon289:
No changes made to CI, but added more functionality to make local dev easier.

Roadmap after this is merged
* Make python and pystan version variables much more obvious
* Change TravisCI to actually run docker containers


canyon289:
If no comments will assume this one is good to merge in ~36 hours

PR #398, opened 2018-11-09
KDE: rugplot position

aloctavodia:
Changes:
* Rugplot are plotted below zero (avoid overlapping with the KDE).
* add a new keyword to `rug_kwargs` `space`, this keyword control the location of the rugplot (as a proportion of `max(density)`), this also allows to have more than one rugplot with-out overlapping.  See this example:

2rug

### DIFF
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -63,6 +63,8 @@ def plot_kde(
         Ignored for 2D KDE
     rug_kwargs : dict
         Keywords passed to the rug plot. Ignored if rug=False or for 2D KDE
+        Use `space` keyword (float) to control the position of the rugplot. The larger this number
+        the lower the rugplot.
     contour_kwargs : dict
         Keywords passed to the contourplot. Ignored for 1D KDE
     ax : matplotlib axes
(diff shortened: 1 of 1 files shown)


arviz/plots/kdeplot.py
```diff
@@ -162,12 +168,13 @@ def plot_kde(
...
             ax.set_xlim(0, auto=True)
-            rug_x, rug_y = np.zeros_like(values), values
+            rug_x, rug_y = np.zeros_like(values) - rug_space, values
         else:
             ax.set_ylim(0, auto=True)
-            rug_x, rug_y = values, np.zeros_like(values)
+            rug_x, rug_y = values, np.zeros_like(values) - rug_space
 
         if rug:
+            del rug_kwargs["space"]
```
aloctavodia: There is a better way to do this? I am deleting `space`, because `ax.plot`, does not have a space keyword

arviz/plots/kdeplot.py
```diff
@@ -147,11 +149,15 @@ def plot_kde(
...
         rug_kwargs.setdefault("linestyle", "None")
         rug_kwargs.setdefault("color", default_color)
+        rug_kwargs.setdefault("space", 0.2)
 
         plot_kwargs.setdefault("linewidth", linewidth)
         rug_kwargs.setdefault("markersize", 2 * markersize)
 
         density, lower, upper = _fast_kde(values, cumulative, bw)
+
+        rug_space = max(density) * rug_kwargs.get("space")
```
canyon289: @aloctavodia You could use a dictionary pop which will remove the keyword and value from the dictionary, saving you from having to manually delete later

`rug_space = max(density) * rug_kwargs.pop("space")`

Admittedly I haven't run the code yet, just what I surmise from static reading.

ahartikainen: pop will KeyError if not found. pop witj default will not

    dict.pop('a', '1')

aloctavodia: Thanks for the suggestions! Using `rug_kwargs.pop("space")` will always work in this scenario because space is set a few line above.

PR #399, opened 2018-11-12
Add logging to test helpers model generation

canyon289:
Generating models takes a long time in tests, providing messages should help with understanding status

### DIFF
--- a/arviz/tests/helpers.py
+++ b/arviz/tests/helpers.py
@@ -14,6 +14,10 @@ import pystan
 import scipy.optimize as op
 import torch
 
+import logging
+
+_log = logging.getLogger(__name__)
+
 
 def _emcee_neg_lnlike(theta, x, y, yerr):
     """Proper function to allow pickling."""
@@ -183,7 +187,7 @@ def pymc3_noncentered_schools(data, draws, chains):
 
 
 def load_cached_models(draws, chains):
-    """Load pymc3, pystan, and emcee models from pickle."""
+    """Load pymc3, pystan, emcee, and pyro models from pickle."""
     here = os.path.dirname(os.path.abspath(__file__))
     data = eight_schools_params()
     supported = (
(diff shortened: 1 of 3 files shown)


arviz/tests/helpers.py
```diff
@@ -14,6 +14,10 @@
 import scipy.optimize as op
 import torch
 
+import logging
+
+_log = logging.getLogger(__name__)
```
ColCarroll: should we use `logging.getLogger('arviz')` instead of `__name__`?

arviz/tests/helpers.py
```diff
@@ -201,9 +205,13 @@ def load_cached_models(draws, chains):
         )
         path = os.path.join(data_directory, fname)
         if not os.path.exists(path):
+
             with open(path, "wb") as buff:
+                _log.info("Generating and caching %s" % fname)
```
ColCarroll: should use `_log.info("Generating and caching %s", fname)`.  formatting will not occur until the logging happens, and services like sentry that aggregate logs can use the format string to group them together, even if the `fname` is different.

canyon289: Ah shoot, missed the comma. Thanks for the catch

arviz/tests/helpers.py
canyon289: I think it's functionally equivalent, but this comment swayed me enough to the "explicit over implicit".

Will make the change

canyon289: Ohh, I confused this with the `__init__.py` at the root directory of arviz.

I think this is better because itll log under `arviz.tests.helpers` which will capture where the logs are coming from. If we use just `arviz` it becomes harder to pinpint. 

The logging functionality is exactly the same because python logger inherit from the root logger, which is configured in '__init__.py` in arviz's root directory

My recommendation is to keep it as such, absent of some other knowledge that Im missing.

canyon289:
Is this one good for a merge?

ahartikainen:
I think this can go in.

PR #400, opened 2018-11-12
faster io_pystan

ahartikainen:
Extra tweaking making code faster and easier to read.


canyon289:
I took a look at this and the code is indeed easier to read

PR #403, opened 2018-11-13
Add optional 'group' param to plot_density.

rpgoldman:
This permits plot_density to be used to, e.g., plot density
information for prior samples.

### DIFF
--- a/arviz/plots/densityplot.py
+++ b/arviz/plots/densityplot.py
@@ -17,6 +17,7 @@ from ..utils import _var_names
 # pylint:disable-msg=too-many-function-args
 def plot_density(
     data,
+    group="posterior",
     data_labels=None,
     var_names=None,
     credible_interval=0.94,
@@ -28,7 +29,6 @@ def plot_density(
     bw=4.5,
     figsize=None,
     textsize=None,
-    group="posterior",
 ):
     """Generate KDE plots for continuous variables and histograms for discrete ones.
 
@@ -41,13 +41,13 @@ def plot_density(
         Any object that can be converted to an az.InferenceData object, or an Iterator returning
         a sequence of such objects.
         Refer to documentation of az.convert_to_dataset for details about such objects.
-    data_labels : Optional[List[str]]
-        List with names for the datasets passed as "data." Useful when plotting more than one
-        dataset.  Must be the same shape as the data parameter.  Defaults to None.
     group: Optional[str]
         Specifies which InferenceData group should be plotted.  Defaults to 'posterior'.
         Alternative values include 'prior' and any other strings used as dataset keys in the
         InferenceData.
+    data_labels : Optional[List[str]]
+        List with names for the datasets passed as "data." Useful when plotting more than one
+        dataset.  Must be the same shape as the data parameter.  Defaults to None.
     var_names: Optional[List[str]]
         List of variables to plot.  If multiple datasets are supplied and var_names is not None,
         will print the same set of variables for each dataset.  Defaults to None, which results in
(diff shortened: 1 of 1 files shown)


canyon289:
Thanks for this PR.

I like the simplicity of this change and its easily understandable. I'll let another person share their opinion to make sure there's a consensus and that I'm not missing anything

arviz/plots/densityplot.py
```diff
@@ -42,6 +43,10 @@ def plot_density(
     data_labels : list[str]
         List with names for the samples in the list of datasets. Useful when
         plotting more than one trace.
+    group: str, optional
+        Which sort of data should be plotted?  Defaults to 'posterior'.  Alternative
```
canyon289: I would change this docstring to a statement rather than a question. I would also move `Default to "posterior" to the end to match other doc strings.

`hpd_markers` and `point_estimate` docstrings in the same function are good examples.

ahartikainen:
Is there any easy way to include group specific info for the plot (e.g. `theta (prior)`; `theta (posterior)`)?

arviz/plots/densityplot.py
rpgoldman: I will make the suggested changes and push a revised commit.
But one question: when you say move the default to the end, I think the resulting docstring will be incoherent, because the sentence which will be left in the middle describes alternatives to the default.  It reads a lot better _after_ the specification of the default.

From my perspective as a new user, the notion of "group" is a little mysterious.  I *think* that the group could be any string, but in current practice is limited to 'posterior', 'prior', and 'posterior_predictive'.  Is that correct? Is this documented somewhere?

rpgoldman:
@ahartikainen Maybe this should be a new issue?  I don't claim to understand how the plotting works well enough to know how one would go about adding group-specific information.  For that matter, I don't really understand the `data_labels` parameter.
Am I right in thinking that this has to do with the fact that `data` can be a _list_ (or iterator?) of data objects, rather than a singleton, as described in discussion here: #404

ahartikainen:
Sounds good (another issue/PR)

rpgoldman:
There may be more to this than I had expected.

Here is a plot I made with pyplot, from inverse gamma distributions generated by `pymc3.sampling.sample_prior_predictive`:
```
plt.hist(samples['Inverse Gamma(1,2)'][...,1],bins=200,density=True,range=(0,20))
```
raw pyplot fig

Now here's the same data from arviz:
```
az.plot_density(samples['Inverse Gamma(1,2)'][...,1], colors='lightBlue', shade=0.75, outline=False)```
![arviz plot](https://user-images.githubusercontent.com/3274/48513735-2c072d00-e822-11e8-9d5c-676c317bf89d.png)



This really doesn't look like an inverse gamma to me.

**Comment by ahartikainen** [comment]
Interesting. Have you tested with different bandwidths?

@aloctavodia should we test bandwidth against empirical CDF with some heuristics?

**Comment by aloctavodia** [comment]
I don't think we need to adapt the bandwith. I may have a solution, I will try it tomorrow.

**Comment by rpgoldman** [comment]
Pushed an update to my PR to bring it up to sync with recent commits to master.

File: arviz/plots/densityplot.py [hunk 2]
```diff
@@ -28,6 +28,7 @@ def plot_density(
     bw=4.5,
     figsize=None,
     textsize=None,
+    group="posterior",
```
ColCarroll: if you've got a moment, could you move this up to be the 2rd argument (and make sure it matches that order in the documentation)? It feels like this will be a helpful general change everywhere, and feels like it should be the second argument.

rpgoldman:
I'll make this change now, and push.  I'll make it as a separate commit; you might want to squash-merge my PR.

aloctavodia:
@ rpgoldman thanks for this PR and also for finding the bug in `plot_density`

rpgoldman:
No thanks are necessary!  I appreciate all of the group's work making this library and helping me understand my data!  Thanks to you all!

PR #408, opened 2018-11-15
Density plot

aloctavodia:
This change the way the density is estimated following discussion in #403 and [discourse](https://discourse.pymc.io/t/bug-in-sample-prior-predictive/2210/6).

lala

### DIFF
--- a/arviz/plots/densityplot.py
+++ b/arviz/plots/densityplot.py
@@ -46,8 +46,7 @@ def plot_density(
         List of variables to plot (defaults to None, which results in all
         variables plotted).
     credible_interval : float
-        Credible intervals. Should be in the interval (0, 1]. For values larger than 0.999 it will
-        be approximated as 1. Defaults to 0.94.
+        Credible intervals. Should be in the interval (0, 1]. Defaults to 0.94.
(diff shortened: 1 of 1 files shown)


arviz/plots/densityplot.py
```diff
@@ -206,23 +210,25 @@ def _d_helper(
     ax : matplotlib axes
     """
     if vec.dtype.kind == "f":
-        density, lower, upper = _fast_kde(vec, bw=bw)
-        x = np.linspace(lower, upper, len(density))
-        hpd_ = hpd(vec, credible_interval)
-        cut = (x >= hpd_[0]) & (x <= hpd_[1])
+        if credible_interval < 0.999:
```
canyon289: Could the check be `credible_interval != 1` or is it important that it truly is less than `.999`?

aloctavodia: This seem a better choice.

canyon289:
Just a question, I see the changes but I'm too dense to get what the change was before and after. If you get a second could you describe what was causing the issue and what the fix was? This is request is optional :)

ahartikainen:
LGTM

PR #413, opened 2018-11-17
Add arviz-white and arviz-colors styles

aloctavodia:
Also adds one example and increase the default dpi from 72 to 100.

The `arviz-white` style is for those that dislike having a grid and a background (for _strict Tufte followers_), the `arviz-colors` style is for those that want to any style they might like but with our color discrete pallet.


ColCarroll:
Cool! 

Merging, but I think the example gallery will only pick up 1 figure, so we may need to mess with that.

aloctavodia:
I can make a single figure with subplots

ColCarroll:
also needs to include `import numpy as np` -- docs are currently broken :)

PR #414, opened 2018-11-17
make arviz styles available without copy them

### LINKED ISSUE #377: arviz-darkgrid not working in example from quickstart
This command is not being recognized.
I installed the program from github on Mac OS 10.13 using macports Python3.5 

aloctavodia:
Following discussion with @ahartikainen this PR proposes an alternative way to make ArviZ styles available without the need to copy files to matplotlib config folder. This should fix #377

### DIFF
--- a/arviz/__init__.py
+++ b/arviz/__init__.py
@@ -2,8 +2,15 @@
 """ArviZ is a library for exploratory analysis of Bayesian models."""
 __version__ = "0.2.1"
 
+import os
 import logging
 from matplotlib.pyplot import style
+
+# add ArviZ's styles to matplotlib's styles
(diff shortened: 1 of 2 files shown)


ColCarroll:
I think we ran into this problem before that `arviz` tries to install in a fresh environment before `matplotlib`

ColCarroll:
I misread - this is a tricky, and good solution (it just adds arviz's style sheet path to matplotlib's search path). Looks good to me!

canyon289:
I haven't run this but it looks like a very elegant solution

ahartikainen:
LGTM. Have anyone tested this with fresh env? Old env has still copies of the old ones (which is fine if we don't touch the old ones)

ColCarroll:
I'll test locally and then merge

PR #417, opened 2018-11-17
styles example: add imports

aloctavodia:
add missing imports!


canyon289:
LGTM. Merging to get the docs back up

PR #423, opened 2018-11-22
Fix test errors

canyon289:
Hopefully fixes (oddly) failing master branch

https://travis-ci.org/arviz-devs/arviz/builds/456930803


ahartikainen:
Is this ready?

canyon289:
Yes, if there's no feedback we can merge

ColCarroll:
Looks good to me! `module` fixtures run once per test suite?

canyon289:
Should be once per module, so if a fixture is used in two python files it runs one per. Session is once per test run.

PR #426, opened 2018-11-25
 emcee version 3 compatibility 

### LINKED ISSUE #425: Incompatibility with emcee3 pre-release
Trying to use the `from_emcee()` function raises the following error for `emcee3`:

> AttributeError: 'EnsembleSampler' object has no attribute 'args'

This is due to line 29 in `io_emcee.py`:

    num_args = len(sampler.args)

In version 2, the EnsembleSampler class stored this information, but only to pass this to `_function_wrapper` class, which also stored the same information. This wrapper was for the posterior probability, so, `sampler.args` is equivalent to `sampler.lnprobfn.args`.

In version 3, this has been slightly modified, and now only the `_FunctionWrapper` class stores this information, however, its name has also been modified, so in this case it should be retrieved with `sampler.log_prob_fn.args`.

OriolAbril:
Fixes  #425

### DIFF
--- a/arviz/data/io_emcee.py
+++ b/arviz/data/io_emcee.py
@@ -27,7 +27,8 @@ def _verify_names(sampler, var_names, arg_names):
         Defaults for var_names and arg_names
     """
     num_vars = sampler.chain.shape[-1]
-    num_args = len(sampler.args) if hasattr(sampler,'args') else len(sampler.log_prob_fn.args)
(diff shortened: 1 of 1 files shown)


ahartikainen:
LGTM

canyon289:
LGTM as well! Thank you for making ArviZ more useful for everyone!

PR #432, opened 2018-11-30
plot_pair: Refactor code and minor tweaks

aloctavodia:
This removes the need to use gridspec and use `plot` instead of `scatter`. I did not run a formal benchmark, but it seems to run faster now. I also changed how divergences are rendered and now the auto-scaling works better.

some pics :-)

lala_00

lala_01

lala_02


ColCarroll:
looks good to me!

PR #433, opened 2018-12-03
Add root directory to conf.py

ColCarroll:
Docs are currently broken. This fixes the docs on my local machine. I don't know why this only popped up now, but probably good to add the root directory to the sphinx build.

This looks like a big PR because `black` ran on `conf.py`, but the only real change is 

```python
sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```

which is a portable way to add `../` to the path.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_script:
 script:
   - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then . ./scripts/lint.sh ; fi
   - pytest -v arviz/tests/ --cov=arviz/
-  - travis-sphinx build --nowarn --source=doc -vvv
+  - travis-sphinx build --nowarn --source=doc
 
 after_success:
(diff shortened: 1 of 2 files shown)


canyon289:
Hm every TravisCI job is failing with a segmentation fault. This is really odd

ColCarroll:
will try adding `- travis-sphinx build --nowarn --source=doc -vvv` to the travis.yml later

ColCarroll:
Turns out there's no `-vvv` command! I am going to merge this so I can (more easily) build the docs locally, and maybe just push them manually for a little while.

ahartikainen:
Interesting, have there ever been that or is the support dropped?

ColCarroll:
This is the `travis-sphinx` command - not the sphinx command! So this is just for pushing the documentation over to github. 

Took me a while to figure that out...

ColCarroll:
Merging despite the segfault, since it puts us in no worse position with respect to docs building.

PR #441, opened 2018-12-06
wrap titles

aloctavodia:
I add `wrap=True` to several plots, as I realize tittle get cropped for the saved figures (but not to the figures rendered inside notebooks)


ColCarroll:
Thanks for tidying this all up!

PR #443, opened 2018-12-07
Tidy up before release 0.3

aloctavodia:

PR #445, opened 2018-12-08
Rerun notebooks

ColCarroll:
The pystan compiler got less chatty, and I left all the output in the notebooks this time. Let me know if you want to suppress any more of these warnings. This should replace most of #443


canyon289:
These look better. 

Not blocking merge but
InferenceDataCookbook and Introduction have empty cells at the bottom which can be deleted.
We should add from TFP in cookbook, if we don't do it on this PR  I can open an issue

aloctavodia:
I think we should remove the empty cells for this PR, and open a new one about adding TFP in cookbook.

Issue #451, opened 2018-12-14
Introduce burn-in parameter in `from_emcee`


OriolAbril:
Most of the mcmc sampling functions have some kind of _tune_ or _warmup_ phase before the actual sampling, whereas in emcee, the burn-in is done after the sample discarding some of them, cutting the resulting array in the draws dimension. 

It may also be interesting to allow a thinning coefficient, because the samples are not independent, in order to be able to mimic completely the emcee3 `get_chain()` method.

I don't think it should be much of a problem because the array returned by emcee always has 3 dimensions: `(nwalkers, draws, ndim)`. I can give it a go.

ahartikainen:
Burn-in sounds good. 

Thinning also, but are the cases where it is a good idea? Thinning just hides the problems with sampling.

kyleabeauchamp:
IIRC, thinning is the least bad idea when you are RAM-limited and want to fill system memory with `n_effective` decorrelated samples rather than `n` samples.

PR #452, opened 2018-12-14
[WIP] Add burn-in and thinning parameters in from_emcee()

OriolAbril:
Discussed in #451


ColCarroll:
This looks good to me (and thanks for keeping the emcee code up to date) -- I would usually suggest keeping all the data and using xarray's slicing to get the `burnin` and `thin`, but I can see how this might be easier and more natural for users.

The test failures are hitting that segmentation fault that is blocking builds :/ Manually inspecting it looks good. I'll merge in a few hours if no one else does (or if no one has objections).

canyon289:
Random question: If thin is meant to reduce memory consumption of a trace by "eliminating" draws, in the way this is implemented wouldn't all the draws already have to be in memory?

OriolAbril:
It would have sense when working with the `HDFBackend` (or another custom defined backend), when you store the mcmc run in an h5 file to analyze the results afterwards. Then, it is not implemented as "eliminating" some draws but instead as reading only some of them. 

However I have just realized that this is only emcee3 compatible, and `chain` does not work anymore when backends are used, `get_chain` must be used. The `posterior_to_xarray` would have to be modified in a similar way to the `observed_data_to_xarray`. 

Overall, the current implementation of thinning crashes with backends which is its only advantage over thinning afterwards with xarray.

I can implement it with a hasattr if you think it is worth it.

OriolAbril:
Starting with emcee3, you have the option to store the results to an hdf5 file, creating the `EnsembleSampler` as:

    filename = "tutorial.h5"
    backend = emcee.backends.HDFBackend(filename)
    sampler = emcee.EnsembleSampler(nwalkers, ndim, log_prob, backend=backend)

This allows to restart from the last checkpoint in case of crashes, analyzing the results somewhere else and so on. And it is actually this case the only scenario where it makes sense to implement the thinning parameter, because otherwise, the data is already in memory and it only eliminates some of it. To retrieve the saved data you would create a reader, and use it as your sampler:

    reader = emcee.backends.HDFBackend(filename)

However, the `sampler` and `reader` are not exactly the same. The current implementation of `from_emcee` is not compatible with reader-like objects, because they have nor `args` nor `log_prob_fun`, so they have no observed data. 

In addition, slicing with h5py is extremely slow, meaning that for 3 variables, 32 walkers and 1e6 draws:
* `reader.get_chain(discard=2000, thin=40)` (which uses h5py under the hood) spends more than 100 s
* `reader.get_chain()[:,2000 + 40 - 1::40,:]` spends 1 s, and around 6 times more memory than when using h5py

Overall I don't really think it's worth it to implement it, and if it is done it should warn that unless the memory consumption is really critical, the thinning should be done with xarray.

I can modify the code to solve the problem with the missing observed data and forget about burn-in and thinning.

canyon289:
Thank you for this super detailed answer. I'm personally traveling the next 4 days but will reread when I'm back home and not jetlagged.

canyon289:
@OriolAbril I read over your answers again (with much more sleep) and they're very helpful.

It sounds like from your last comment you'd like to remove the burnin/thinning logic and solve the problem with missing data. Is my understanding correct?

OriolAbril:
Yes, personally I don't think that the memory saved is worth all the extra time used reading the hdf5 file.

canyon289:
Just to ensure we're all on same page. We are waiting for burnin/thinning logic to be removed since the memory saved is not worth the disk i/o? If so I'll change this PR title to WIP

OriolAbril:
It will probably be better to close this pull request, and when I have time I'll fix the incompatibility with `from_emcee` and reading the results in an h5 file. I found it here analyzing the performance of thinning but it is actually completely unrelated.

PR #454, opened 2018-12-15
Travis tmp fix

ahartikainen:
Quick fix. Something fills up something.

Commenting both of these out, travis works.


ahartikainen:
Weird, first time it segfaulted everything, now after restart things work again.

Is this some OS problem with travis?

I guess we need to checkout docker.

PR #459, opened 2018-12-19
Add initial documents for paper

canyon289:
Submitting a paper to JOSS seems deceptively easy. I think this is all we need to do in the repo, besides one note below.

@aloctavodia
Is there a magic command you prefer to render .md files into PDFs?

I found this site but it seems there are many options so not sure which one is "best"
https://gist.github.com/maxogden/97190db73ac19fc6c1d9beee1a6e4fc8

For reference implementation I'm using gala's submission.
https://github.com/adrn/gala


paper/paper.md
```diff
@@ -0,0 +1,71 @@
+# ArviZ a unified library for exploratory analysis of Bayesian models in Python
```
aloctavodia: I think we should remove this line

aloctavodia:
I generally use something like `pandoc  --pdf-engine=xelatex --filter pandoc-citeproc  input.md -o output.pdf` 

But AFAIK we do not need to create the pdf. The JOSS system will do that for us.

canyon289:
Good to know. I'll make change and continue on submission process then

canyon289:
Using these other papers as reference for format as well
https://raw.githubusercontent.com/telegraphic/hickle/master/paper.md
https://raw.githubusercontent.com/boyanangelov/sdmbench/master/paper/paper.md

canyon289:
Merging for submission

PR #461, opened 2018-12-26
Add docs and exception to plot compare

canyon289:
Raise ValueError if WAIC or LOO is not present in compare_df rather than doing nothing
Add inline docs


ahartikainen:
LGTM

Issue #462, opened 2018-12-26
from_pyro with multiple chains


jeffpollock9:
I guess this is a similar problem to #456, that is, when I run the pyro sampler with 1 chain this works, but more than 1 chain does not.

Here is my example:

```python
import torch
import pyro
import pyro.distributions as dist
import pyro.infer.mcmc as mcmc
import arviz as az

torch.multiprocessing.set_sharing_strategy("file_system")
pyro.set_rng_seed(42)

NUM_WARMUP = 100
NUM_SAMPLES = 100
NUM_CHAINS = 1  # only works for NUM_CHAINS < 2
N = 100
P = 2

alpha_true = dist.Normal(42.0, 10.0).sample()
beta_true = dist.Normal(torch.zeros(P), 10.0).sample()
sigma_true = dist.Exponential(1.0).sample()

eps = dist.Normal(0.0, sigma_true).sample([N])
x = torch.randn(N, P)
y = alpha_true + x @ beta_true + eps


def model(x, y):
    alpha = pyro.sample("alpha", dist.Normal(0.0, 100.0))
    beta = pyro.sample("beta", dist.Normal(torch.zeros(P), 10.0))
    sigma = pyro.sample("sigma", dist.HalfNormal(10.0))
    mu = alpha + x @ beta
    return pyro.sample("y", dist.Normal(mu, sigma), obs=y)


kernel = mcmc.NUTS(model, jit_compile=True)

posterior = mcmc.MCMC(kernel,
                      num_samples=NUM_SAMPLES,
                      warmup_steps=NUM_WARMUP,
                      num_chains=NUM_CHAINS,
                      disable_progbar=True).run(x, y)

data = az.from_pyro(posterior)

az.summary(data)
```

Which when `NUM_CHAINS = 1` outputs:

```python
>>> az.summary(data)
          mean    sd  mc error  hpd 3%  hpd 97%
alpha    45.44  0.08      0.01   45.27    45.61
beta[0]   1.25  0.07      0.01    1.14     1.41
beta[1]   2.23  0.10      0.01    2.01     2.40
sigma     0.80  0.05      0.01    0.71     0.89
```

and for e.g. `NUM_CHAINS = 2` outputs:

```python
>>> az.summary(data)
            mean    sd  mc error  hpd 3%  hpd 97%
alpha[0]   45.43  0.02       NaN   45.41    45.45
alpha[10]  45.41  0.08       NaN   45.33    45.49
[...]

PR #463, opened 2018-12-26
fix for from_pyro with multiple chains

jeffpollock9:
As per #462 - we should only expand the dimension of the samples if the number of chains is 1. Please note I have not tested this for pyro versions less than 0.3.

For the example in the #462, this now correctly prints:

```python
>>> az.summary(data)
          mean    sd  mc error  hpd 3%  hpd 97%  eff_n  r_hat
alpha    45.46  0.08       0.0   45.30    45.59  142.0   1.00
beta[0]   1.25  0.08       0.0    1.11     1.41  164.0   1.01
beta[1]   2.21  0.09       0.0    2.05     2.40  141.0   1.00
sigma     0.79  0.06       0.0    0.69     0.89   77.0   1.00
```

when running with 2 chains.

### DIFF
--- a/arviz/data/io_pyro.py
+++ b/arviz/data/io_pyro.py
@@ -53,12 +53,10 @@ class PyroConverter:
         # Do not make pyro a requirement
         from pyro.infer import EmpiricalMarginal
 
-        def expand_if_single_chain(x):
-            return np.expand_dims(x, 0) if self.posterior.num_chains == 1 else x
-
         try:  # Try pyro>=0.3 release syntax
             data = {
-                name: expand_if_single_chain(samples.enumerate_support())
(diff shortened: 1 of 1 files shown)


ahartikainen:
I think you can skip the function and just duplicate your if-else structure, this way everything should work.

jeffpollock9:
@ahartikainen thanks. I __think__ `num_chains` is missing in pyro < 0.3 since they didn't support running multiple chains in parallel at that point - so the untested changes I made were dumb (serves me right for not properly testing). That is at least as far as I can tell looking at e.g. https://github.com/uber/pyro/blob/0.3.0-release/pyro/infer/mcmc/mcmc.py#L103

Hoping it is ok now.

ahartikainen:
LGTM

PR #465, opened 2018-12-27
Fix docs

ahartikainen:
Close old unused figures in examples.


ahartikainen:
I think we are hitting 1GB limit with sphinx.

There is a way to create figures outside the sphinx command. Can we do the same with notebooks?

canyon289:
We run sphinx on every TraviCI job? It seems like it.

When you say create figures outside of Sphinx what do you mean? For examples?

PR #467, opened 2018-12-27
Update .travis.yml

ahartikainen:
Update travis to use matrix structure.

Upgrade distribution to xenial.

Fix pyro code.

~Fix error in plotting code which caused tests to fail.~ Pyro fix was the answer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+dist: xenial
+
 language: python
 
 python:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,20 @@ dist: xenial
 
 language: python
 
-python:
-  - "3.6"
-  - "3.5"
-
-env:
-  - PYSTAN_VERSION=2.17
-  - PYSTAN_VERSION=latest
+matrix:
+  include:
+  - name: "Python 3.6 Unit Test"
+    python: 3.6
+    env: PYSTAN_VERSION=latest NAME="UNIT"
+  - name: "Python 3.5 Unit Test"
+    python: 3.5
+    env: PYSTAN_VERSION=latest NAME="UNIT"
+  - name: "Python 3.6 Lint"
+    python: 3.6
+    env: PYSTAN_VERSION=latest NAME="LINT"
+  - name: "Python 3.6 Sphinx"
+    python: 3.6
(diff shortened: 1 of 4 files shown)


ahartikainen:
This is ready now. Only think I don't like is the squeezing pyro output.

Maybe that could be done bu squeezing out only the last dimension if the shape is (N,1)?

arviz/plots/densityplot.py
```diff
@@ -171,10 +171,9 @@ def plot_density(
     for plotters in to_plot:
         length_plotters.append(len(plotters))
         for var_name, selection, _ in plotters:
-            label = make_label(var_name, selection)
+            label = make_label(var_name, selection).replace("\n", "_")
```
canyon289: I think it's better if `.replace` be done in `make_label`. perhaps by passing in a tuple. That way all the label making logic is in one place and `make_label` provides the finished label. 

Thoughts?

.travis.yml
```diff
@@ -1,12 +1,21 @@
...
 
-python:
-  - "3.6"
-  - "3.5"
+language: python
 
-env:
-  - PYSTAN_VERSION=2.17
-  - PYSTAN_VERSION=latest
+matrix:
```
canyon289: This is matrix approach is way better. Thanks for doing this

.travis.yml
```diff
@@ -1,12 +1,21 @@
...
+  include:
+  - name: "Python 3.6 Unit Test"
+    python: 3.6
+    env: PYSTAN_VERSION=latest NAME="UNIT"
+  - name: "Python 3.5 Unit Test"
+    python: 3.5
+    env: PYSTAN_VERSION=latest NAME="UNIT"
+  - name: "Python 3.6 Lint"
+    python: 3.6
+    env: PYSTAN_VERSION=latest NAME="LINT"
```
canyon289: No change required but why have a separate lint environment?

arviz/plots/densityplot.py
canyon289: Will fix on another PR per discussion. Thanks ari :)

ahartikainen:
If tests pass, then this is good to go.

We will need to think about pyro shape issue later and also update densityplot if multiple models contain different parameters.

    {'a', 'b'} + {'b','c'} --> should make 3 subplots, now it makes 2.

PR #474, opened 2018-12-30
[docs] Run all examples

ahartikainen:
WIP

Enable both ppc plots


canyon289:
LGTM I don't see any segmentation faults

PR #479, opened 2018-12-31
KDE: Display quantiles

aloctavodia:
This add the option to get a KDE segmented into quartiles (or more general quantiles). There are still a few things to fix and improve

gauss_q

expo_q

### DIFF
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -157,6 +157,7 @@ def plot_kde(
         rug_space = max(density) * rug_kwargs.pop("space")
 
         x = np.linspace(lower, upper, len(density))
+        x_q = x
         fill_func = ax.fill_between
         fill_x, fill_y = x, density
         if rotated:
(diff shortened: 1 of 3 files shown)


ahartikainen:
List of numbers between 0 and 100. Or quantiles => list of numbers between 0 and 1.

Also what algorithm should we use for quantiles etc? Data or kde results?

ahartikainen:
Also, would there be any potential to define ... "symmetric" quantile, e.g. distribution which is constrained and MAP is on one of the constrain?

E.g. 1-tail / 2-tail / negative 1-tail

aloctavodia:
Not sure I truly understand what these "symmetric" quantiles are

ahartikainen:
True, I probably don't understand them either.

I think it similar if we split data based on the density (starting from top, moving horizontal line down until q-quantile of the density is inside the density limits...

ahartikainen:
Good choice

aloctavodia:
Sure! I can add a test, but not until next Tuesday. So if you want to add it sooner than that, please go ahead.

canyon289:
LGTM

aloctavodia:
@ahartikainen are you also OK with this?

ahartikainen:
Lgtm

Issue #481, opened 2018-12-31
README doesn't mention non-master installation instructions.


malmaud:
In the README installation section, you don't mention `pip install arviz` as the way of installing the release version. I know it seems trivial, but some people might not realize if the instructions aren't explicit.

ahartikainen:
Oh, we have forgot to update readme.

Do you want to do a PR?

canyon289:
Definitely not trivial. This is a good suggestion.

canyon289:
Closed by https://github.com/arviz-devs/arviz/pull/482

PR #482, opened 2019-01-01
Add more installation instructions to README

canyon289:
Add installation instructions
https://github.com/arviz-devs/arviz/issues/481


canyon289:
Feel like this won't be controversial. As is such self merging

PR #485, opened 2019-01-02
Paper fix 2

canyon289:
Reduced paper down to two figures. 

Example attached.
[paper.pdf](https://github.com/arviz-devs/arviz/files/2722350/paper.pdf)


canyon289:
Self merging

PR #490, opened 2019-01-03
Normalize from_xyz functions

ahartikainen:
Use `kw1=None, *, kw2=None, ...` in all of the `from_xyz` functions.

Each library can still use their nomenclature if needed.

created aliases for `load_data` (`from_netcdf`) and `save_data` (`to_netcdf`)

### DIFF
--- a/arviz/data/io_netcdf.py
+++ b/arviz/data/io_netcdf.py
@@ -39,3 +39,7 @@ def save_data(data, filename, *, group="posterior", coords=None, dims=None):
     """
     inference_data = convert_to_inference_data(data, group=group, coords=coords, dims=dims)
     return inference_data.to_netcdf(filename)
+
+
+from_netcdf = load_data
+to_netcdf = save_data
--- a/arviz/data/io_pymc3.py
+++ b/arviz/data/io_pymc3.py
@@ -10,7 +10,7 @@ class PyMC3Converter:
     """Encapsulate PyMC3 specific logic."""
 
     def __init__(
-        self, *_, trace=None, prior=None, posterior_predictive=None, coords=None, dims=None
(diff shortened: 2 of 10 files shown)


arviz/data/io_netcdf.py
```diff
@@ -39,3 +39,7 @@ def save_data(data, filename, *, group="posterior", coords=None, dims=None):
     """
     inference_data = convert_to_inference_data(data, group=group, coords=coords, dims=dims)
     return inference_data.to_netcdf(filename)
+
+
+from_netcdf = load_data  # pylint: disable=invalid-name
```
canyon289: Can we just rename the functions rather than alias them and have two of the same function in the API?

ahartikainen: We could. Then we only need to mention this change in "what's changed". Or deprecate old functionality.

canyon289: I vote for deprecate old functionality. If we want to be safe we can wrap load_data and add a warning

arviz/data/io_pymc3.py
```diff
@@ -10,7 +10,7 @@ class PyMC3Converter:
     """Encapsulate PyMC3 specific logic."""
 
     def __init__(
-        self, *_, trace=None, prior=None, posterior_predictive=None, coords=None, dims=None
+        self, *, trace=None, prior=None, posterior_predictive=None, coords=None, dims=None
```
ColCarroll: thank you for catching this :/

arviz/data/io_netcdf.py
ColCarroll: i'm fine with deprecation, or using the fact that we're not 1.0 yet to just change it (starting up a RELEASE_NOTES.md is a good idea!)

ahartikainen:
I can do it later today.

PR #502, opened 2019-01-05
Add InferenceData tutorial in header

canyon289:
Will add InferenceData tutorial in header to meet paper review feedback.

In making this documentation I tried first creating a Tutorial page like Seaborn
https://seaborn.pydata.org/tutorial.html

However it seems getting to this will require more work. Seaborn doesn't use the nbconvert plugin, instead providing its own Makefiles and scripts that convert the notebooks to .rst and .html at the appropriate times
https://github.com/mwaskom/seaborn/blob/master/doc/Makefile
https://github.com/mwaskom/seaborn/tree/master/doc/tools
https://github.com/mwaskom/seaborn/blob/master/doc/tutorial/Makefile

I think it would be great to implement this eventually, but for now just want to get the existing tutorial visible for the paper review

### DIFF
--- a/doc/notebooks/XarrayforArviZ.ipynb
+++ b/doc/notebooks/XarrayforArviZ.ipynb
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Introduction to xarray for ArviZ"
+    "# Introduction to xarray, InferenceData, and netCDF for ArviZ"
    ]
   },
   {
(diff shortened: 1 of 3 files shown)


ahartikainen:
LGTM

doc/notebooks/XarrayforArviZ.ipynb
```diff
@@ -13,21 +13,33 @@
...
-    "* arviz.InferenceData\n",
-    "* NetCDF\n",
+    "They are  \n",
+    "\n",
+    "<ul class=nbtable>\n",
+    "  <!-- HTML Table to allow proper rendering in sphinx -->\n",
+    "  <li>xarray.Dataset</li>\n",
+    "  <li>arviz.InferenceData</li>\n",
+    "  <li>netCDF</li>\n",
+    "</ul> \n",
```
ahartikainen: Should these be hidden? If I check these with view, I can see them

    <ul ...
    </ul>

Maybe if they are translated to HTML they are hidden

canyon289: Aw shoot that's annoying. They seem to get hidden when Sphinx renders them, but not github apparently. I'll make another fix

PR #505, opened 2019-01-06
Rename N_effective to S_effective

canyon289:
Trying to make it more clear to users and newcomers to Bayesian methodology what calculation is being done and where to look for further reference
 
Review requested more for the idea of the change, and less so the code.

I can add deprecation or rename warning to `n_effective` to make this a non breaking change

### DIFF
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -7,7 +7,7 @@ import matplotlib.pyplot as plt
 
 from ..data import convert_to_dataset
 from ..stats import hpd
-from ..stats.diagnostics import _get_neff, _get_split_rhat
+from ..stats.diagnostics import _get_seff, _get_split_rhat
 from .plot_utils import _scale_fig_size, xarray_var_iter, make_label
 from .kdeplot import _fast_kde
 from ..utils import _var_names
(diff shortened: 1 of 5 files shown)


ahartikainen:
hmm, Should we then also move to split calculations

canyon289:
@ahartikainen On your question you and @aloctavodia are much better experts on this than I am.

The perspective I have is that I was very very confused when trying to learn how this method really worked because the naming, and the reference, did not match the code.

Whether we should move to split calculations though I'm not sure! Happy to take suggestion

**Code Note**
Rebased this PR on top of the pytest_pin branch

junpenglao:
I think we should move to split calculations.
cc @avehtari

ahartikainen:
I updated Rhat to use split function. Not sure about the Seff.

junpenglao:
@canyon289 maybe we can do the same - renaming it to ESS, and wait for Stan to update to split calculation.

avehtari:
The current best reference for what is currently used in Stan, is Stan manual. See sections 15.3 and 15.4
https://mc-stan.org/docs/2_18/reference-manual/notation-for-samples-chains-and-draws.html
https://mc-stan.org/docs/2_18/reference-manual/effective-sample-size-section.html

Rhat and ESS case study is the best reference what is planned for the future. Section 2 is review. Section 3 is new material. Sections 3.10-3.12 provide explicit proposal for changes. This case study will change to a paper and vignette during the spring.

PR #506, opened 2019-01-06
Pin pytest version to avoid coverage exceptions

canyon289:
Pytest updated to version 4.1.0 which is causing failures with coverage

Pinning version for now and can open issue ticket so we remember to check on this every now and then

Additional details for future reference
https://docs.pytest.org/en/latest/changelog.html

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,7 +9,7 @@ pylint
 pyro-ppl
 tensorflow
 tensorflow-probability
-pytest
+pytest=4.0.2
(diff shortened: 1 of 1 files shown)


ahartikainen:
Good, I need to rebase after merge.

canyon289:
Look like they've fixed this in the development version, just might not be on pypi yet.

Link for reference
https://github.com/pytest-dev/pytest-cov/issues/254

ahartikainen:
LGTM

PR #515, opened 2019-01-09
add conditional_jit decorator and speed up _fast_kde 2x

aloctavodia:
adds the conditional_jit decorator that works only if numba is installed. I applied the decorator to speed-up _fast_kde. I guess other functions (stats, diagnostics) could benefit more from using numba.

### DIFF
--- a/arviz/utils.py
+++ b/arviz/utils.py
@@ -24,7 +24,23 @@ def _var_names(var_names):
 
 
 def conditional_jit(function=None, **kwargs):  # noqa: D202
-    """Use numba's jit decorator if numba is installed."""
+    """Use numba's jit decorator if numba is installed.
+
+    Notes
+    -----
+        If called without arguments  then return wrapped function.
+
+        @conditional_jit
+        def my_func():
+            return
+
+        else called with arguments
+
+        @@conditional_jit(nopython=True)
+        def my_func():
+            return
+
+    """
 
     def wrapper(function):
         try:
@@ -34,6 +50,7 @@ def conditional_jit(function=None, **kwargs):  # noqa: D202
         except ImportError:
             return function
 
+    #
     if function:
         return wrapper(function)
     else:
--- a/arviz/utils.py
+++ b/arviz/utils.py
@@ -23,8 +23,9 @@ def _var_names(var_names):
         return var_names
 
 
-def conditional_jit(function=None, **kwargs):
+def conditional_jit(function=None, **kwargs):  # noqa: D202
     """Use numba's jit decorator if numba is installed."""
+
     def wrapper(function):
         try:
             numba = importlib.import_module("numba")
--- a/arviz/utils.py
+++ b/arviz/utils.py
@@ -22,11 +22,18 @@ def _var_names(var_names):
         return var_names
 
 
-def conditional_jit(function):
-    "use numba's jit decorator if numba is installed"
-    try:
-        from numba import jit
+def conditional_jit(function=None, **kwargs):
+    """Use numba's jit decorator if numba is installed."""
 
-        return jit(function)
-    except ImportError:
-        return function
+    def wrapper(function):
+        try:
+            from numba import jit
+
+            return jit(**kwargs)(function)
+        except ImportError:
+            return function
+
+    if function:
+        return wrapper(function)
+    else:
+        return wrapper
--- a/arviz/utils.py
+++ b/arviz/utils.py
@@ -36,7 +36,7 @@ def conditional_jit(function=None, **kwargs):  # noqa: D202
 
         else called with arguments
 
-        @@conditional_jit(nopython=True)
+        @conditional_jit(nopython=True)
         def my_func():
             return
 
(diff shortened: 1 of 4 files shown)


canyon289:
This is really cool. Surprised its so easy

Tests should be added and we should add numba to dev dependencies.

ahartikainen:
Should we enable `**kwargs` for specific functions? E.g. `nopython`, `nogil` etc

aloctavodia:
I will add numba to dev dependencies and fix the pylint issues. And try to think about the test. Right, we should have `**kwargs`!

canyon289:
I can help with tests as well. We might need to mock the module somehow to force the method to run in numba,  and not with numba. I _think_ we can do it without having to create a totally different env

aloctavodia:
BTW we can use the same pattern for other numba decorator, but I think we should follow a _just in time_ approach ;-) and add them as needed. 

I am not sure how to test this, @canyon289 let me know if you have any suggestion.

ColCarroll:
Have you verified that this works? I _think_ you need to use parens on the decorator the way it is currently defined.

If you want arguments for a decorator, I usually see a class used:

```python
class conditional_jit:
    def __init__(self, **numba_kwargs):
        self.numba_kwargs = numba_kwargs

    def __call__(self, func):
        try:
            from numba import jit

             return jit(**kwargs)(function)
        except ImportError:
            return function
```

ColCarroll:
For testing, you could also monkeypatch this decorator in `conftest.py` to just return `function`, right?

aloctavodia:
Yes, I tested locally. To avoid using a non-numba environment I manually change `from numba import jit` to  `from numbad import jit`

@ColCarroll  You can call it like `@conditional_jit` or `@conditional_jit()`or with arguments like `@conditional_jit(nopython=True)`

canyon289:
> For testing, you could also monkeypatch this decorator in `conftest.py` to just return `function`, right?

I believe you can monkeypatch in many places but I would prefer it be done closer to the test so its clear why things are being monkeypatched, and the scope of the change is only for these tests. In conftest.py it would apply to everything, even when its not needed

@aloctavodia Whenever you are done with code changes I can write the test

aloctavodia:
@canyon289 I have not further changes. It is all in your hands now :-) 
Thanks for helping with this.

canyon289:
I will be able to start writing tests in ~3 days (Don't have access to my desktop computer for a bit)

arviz/utils.py
```diff
@@ -20,3 +21,20 @@ def _var_names(var_names):
...
+        try:
+            numba = importlib.import_module("numba")
+            return numba.jit(**kwargs)(function)
+
+        except ImportError:
+            return function
+
+    if function:
+        return wrapper(function)
+    else:
```
canyon289: @aloctavodia This piece of code I'm confused on. In which situation would would  you need to use `conditional_jit` without a function?

ahartikainen: Not sure where this would be suitable and if it even works (probably), but isn't it kind of create a function that has all the settings done and then just apply it to some function (or maybe for many functions?)

aloctavodia: This allow to call it like `@conditional_jit` otherwise you will have to call it like `@conditional_jit()` or `@conditional_jit(some_keyword)`

aloctavodia:
LGTM!

arviz/utils.py
canyon289: What's confusing to me is that in the tests It is being used as `@conditional_jit` but the else block is not hit according to coveralls

https://coveralls.io/builds/21049537/source?filename=arviz/utils.py#L40

ahartikainen: Can you print what is going on?

What about `function=None`?

canyon289: Sure, I will add more tests and some prints in the next ~1 hour. My gut feel though is that the else block is not needed if this will be used as a decorator because there will always be an object passed to the decorator.

I need to test with kwargs as well to ensure those are present.

aloctavodia: You will hit the "if-branch" when you use the decorator as `@conditional_jit` and the "else-branch when you use it as `@conditional_jit()` or  `@conditional_jit()`.

canyon289: Will update in ~5 hours and then we should be good

arviz/utils.py
```diff
@@ -24,7 +24,23 @@ def _var_names(var_names):
...
+    -----
+        If called without arguments  then return wrapped function.
+
+        @conditional_jit
+        def my_func():
+            return
+
+        else called with arguments
+
+        @@conditional_jit(nopython=True)
```
aloctavodia: Typo "@@"

canyon289: Thanks

canyon289:
Rebased on master for TravisCI fix

canyon289:
I think this PR is fine? Not sure why the push build is failing when the PR one is passing

canyon289:
This is good so I'm just going to merge it.

PR #520, opened 2019-01-11
update version number to 0.3.2

ahartikainen:
Small update


aloctavodia:
shouldn't be just 0.3.2?

ahartikainen:
No, when we update to 0.3.2 it won't install the latest with pip if dev is already installed (is this correct?).

canyon289:
LGTM

ahartikainen:
https://www.python.org/dev/peps/pep-0440/

If we fould make releases then we could add number at the end, but I think we are not releasing dev versions so just `.dev` should be enough.

PR #522, opened 2019-01-11
add stable logsumexp

ahartikainen:
Implements stable logsumexp when `b>0` also can use b_inv which is better than 1/x when x is large.


aloctavodia:
LGTM!

PR #529, opened 2019-01-14
Fix pymc3 typo

ColCarroll:
I am not sure how this worked in the past, but it looks like we definitely want `pymc3.util` (https://github.com/pymc-devs/pymc3/blob/master/pymc3/util.py) instead of `pymc3.utils` (which resolves to `pymc3.glm.utils`, and does not have `get_default_varnames`).


ahartikainen:
So this has nothing to do with numpy=1.16 error?

ColCarroll:
Nope. Linking to pymc3 issue now, but can use arviz in PyMC3 with this change.

ahartikainen:
LGTM

PR #531, opened 2019-01-15
[WIP] Remove redundant docker line in travisyml

canyon289:
Seeing if this works


ahartikainen:
Why? Or why this is not needed?

canyon289:
Seeing if it was not needed, and since it's not I think we should remove it!

PR #532, opened 2019-01-15
test _histogram without numba

ahartikainen:
Numba wraps _histrogram function and this causes coverage not to update on that function.


ahartikainen:
First commit mess things up.

PR #534, opened 2019-01-15
Run tests without numba

ahartikainen:
By following recommendation in https://stackoverflow.com/questions/26875191/analyzing-coverage-of-numba-wrapped-functions

Basically it goes to numba, but returns the normal function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ script:
 
   # Tests command is default in dockerfile cmd
   - if [[ $NAME == UNIT ]]; then docker run  --mount type=bind,source="$(pwd)",target=/opt/arviz --name arviz_test arviz:latest /bin/bash -c \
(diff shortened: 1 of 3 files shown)


canyon289:
Ah, I know what's going on, this seems to break the test I made yesterday for the decorator test.

ahartikainen:
No, that last commit was my fault. Everything before last commit works. I'm just trying to see if there would be a way return results from wrapper and function :D

canyon289:
Ah ok, I'll leave you to it! Thanks for the fix

ahartikainen:
I think this will work.

canyon289:
That last one is a good idea!

ahartikainen:
Oh wait, I have even better

ahartikainen:
I was fooled by the lambda + comma

Issue #539, opened 2019-01-16
`SyntaxError: invalid syntax` upon import for arviz 0.3.2 Python==3.5.2


MFreidank:
**Describe the bug**
Simply importing `arviz` fails for the new release (< 1 day ago) `0.3.2`. Reverting to `0.3.1` makes the bug disappear. 

**To Reproduce**
`pip install arviz==0.3.2`
`python -c "import arviz"`

Fails with the following trace:
```
/home/moritz/projects/venv/lib/python3.5/importlib/_bootstrap.py:222: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/moritz/projects/venv/lib/python3.5/site-packages/arviz/__init__.py", line 22, in <module>
    from .data import *
  File "/home/moritz/projects/venv/lib/python3.5/site-packages/arviz/data/__init__.py", line 8, in <module>
    from .io_dict import from_dict
  File "/home/moritz/projects/venv/lib/python3.5/site-packages/arviz/data/io_dict.py", line 28
    ):
    ^
SyntaxError: invalid syntax
``` 

**Expected behavior**
The import should go through without error.

**Additional context**

`python --version`
Python 3.5.2
Arviz version: `0.3.2`
OS: Ubuntu 16.04 xenial

MFreidank:
UPDATE: I believe that this is somehow related to *keyword-only* function definitions (as in `arviz.data.io_dict.DictConverter.__init__`. However, the following minimal example works with the same interpreter: 

```
#!/usr/bin/python3
# -*- coding: iso-8859-15 -*-
def f(
    a=None,
    *,
    b=None,
    c=None
):
    return a + b +c

print(f(10, b=20, c=30))  # prints `60` without error
```
Quite confused as to what is going on..

ahartikainen:
Is there a trailing comma in arviz code?

MFreidank:
@ahartikainen Yes! And I think you are on to something, because my minimal example (function `f` above) fails if I add a trailing comma. Should I quickly do a PR for this?

ahartikainen:
Thanks, this is probably due to that we had this problem before, but travis told us to clean up this problem, then we changed to docker (which is great) but apparently we have failed with 3.5.

We have also a small problem with requirements.txt so it is good that we can fix this at the same time and make quickly a new bugfix release.

[...]

PR #541, opened 2019-01-16
Remove trailing commata in `arviz/data/io_dict.py`

MFreidank:
See also #539 .
Trailing commata in combination with keyword-only function definitions are only supported in Python >= 3.6 and thus break support for Python3.5.
As pointed out in #539, this is a problem introduced in `black` and subject to an issue there as well.

This PR removes trailing commas in `arviz/data/io_dict.py` manually. After these changes, `python -c "import arviz"` works without error.


ahartikainen:
LGTM

PR #550, opened 2019-01-18
Emcee reader support

OriolAbril:
To this date, there are 3 possible cases to be taken into account in `from_emcee`:

* emcee2
* emcee3 
* emcee3 "reader", which is using `emcee.backend` to store the posterior in file and reading it afterwards. It is different from the "regular" emcee3 sampler because the observed data is not stored.

The first two cases already work, this adds support to the 3rd case. I have run some tests manually in my computer to check they work, but I don't know how to update the tests to include different emcee versions.


canyon289:
@OriolAbril Can you write the test suite in pytest for the new emcee "reader" and check it works in your local environment?

After that I can make the changes in TravisCI to test the emcee reader as well

OriolAbril:
I added a test for the emcee reader in its own file inside the tests folder. However I am not sure about how to implement it directly in test_data.py, because it should only executed when emcee version is >=3 but without stopping to test the regular emcee3 sampler.

All tests will probably fail because I think actually none of them uses emcee3.

ColCarroll:
This would be a great use for a `pytest.skipif` fixture:

```python
needs_emcee3 = pytest.mark.skipif(emcee.__version__ < '3', reason="emcee3 required")

...

@needs_emcee3
def test_emcee3_stuff(...):
    ...
```

OriolAbril:
I used skipif to implement an extra test inside the class for emcee. Now the test should pass, but it may not mean I have implemented them properly because they will be skipped. What sould be changed to include emcee3 in the tests?

I did not run them locally because I do not have all libraries installed (i.e. stan). Splitting them as @ahartikainen proposed would ease running them locally, at least for me.

canyon289:
@OriolAbril 
Youre able to select tests to run using the `-k` parameter in pytest. I can provide an example if needed

arviz/tests/helpers.py
```diff
@@ -28,6 +28,7 @@
 
 _log = logging.getLogger(__name__)
 
+needs_emcee3 = pytest.mark.skipif(emcee.__version__ < '3', reason="emcee3 required")
```
canyon289: Will this string comparison work? If so why not do a string on line 99 as well?

OriolAbril: I checked in ipython interpreter and it worked. I can change it to use the same method in both cases.

canyon289:
@OriolAbril You're definitely not taking up too much of my time! I'll speak for everyone and say we really appreciate YOUR time :) 

If you'd like we can chat on gitter as well for faster collaboration

https://gitter.im/arviz-devs/community

OriolAbril:
For now I'd rather push the changes here, but I will start looking into docker. Can you recommend me some tutorial?

Now only the lint check is failing, I disabled all `no-member` and `invalid-arguments` in the emcee3 code. Also an `invalid-name` for `needs_emcee3`, but I wasn't really sure about disabling it or using uppercase

ahartikainen:
Is it for tests? Disable is fine.

ahartikainen:
> For now I'd rather push the changes here

TBH this is mostly how I do stuff. I just run `black arviz` and `pylint arviz` (or usually specific file I have changed) (I ignore all import errors) and then push to travis.

This is because I do dev work on different machines / phone, so I only need git tools and vim.

ahartikainen:
Sounds fine.

LGTM

canyon289:
Thanks @OriolAbril

PR #557, opened 2019-01-22
Maintain order of variables in xarray_var_iter

### LINKED ISSUE #554: Ordering of variables after coords set
## Short Description

I'm trying to use the coords/dims to set labels on posterior variables from a stan fit and the order of the dimensions gets shuffled when I set the labels (not sure what determines the order). Is there a way to (1) retain the existing order (A,B,C...) and (2) sort the order according to a custom index that I provide (e.g. pass a index array like [2,1,0,3, ...] to have the plot appear in the order [C, B, A, D ...])? 

Sorry if this is covered in the documentation somewhere, I didn't see it.

Without setting any coords/dims, the variables appear in the same order as one would expect from the stan code.

test1

Setting the coords/dims with the code below changes the order to this.

test2

## Code Example or link

```
azdata = az.from_pystan(posterior=fit, observed_data=['Y'], coords={'betas': ['A', 'B', 'C', 'D', 'E', 'F', 'G']}, dims={'b': ['betas']})
	
axes = az.plot_forest(azdata, var_names=('b'))
```
Using arviz 0.3.2

ColCarroll:
Fixes #554. My local environment is a little broken, so I want to make sure unittests pass while I write a test for this case. This works on the examples from the issue, though.

```
data = az.from_pymc3(trace=trace, coords={'schools': list('abcdefgh')}, dims={'theta': ['schools']})
az.plot_forest(data, var_names='theta')
```

image

### DIFF
--- a/arviz/plots/plot_utils.py
+++ b/arviz/plots/plot_utils.py
@@ -260,7 +260,7 @@ def xarray_var_iter(data, var_names=None, combined=False, skip_dims=None, revers
     for var_name in var_names:
         if var_name in data:
             new_dims = [dim for dim in data[var_name].dims if dim not in skip_dims]
-            vals = [set(data[var_name][dim].values) for dim in new_dims]
+            vals = [np.unique(data[var_name][dim].values) for dim in new_dims]
             dims = [{k: v for k, v in zip(new_dims, prod)} for prod in product(*vals)]
             if reverse_selections:
                 dims = reversed(dims)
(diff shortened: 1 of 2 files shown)


ColCarroll:
Ok, has a test now that fails on master and passes locally.

canyon289:
It looks like a linting error not a test. A couple of pylint ignores or fixes and you should be good to go!

ahartikainen:
LGTM

PR #563, opened 2019-01-23
Update conftest testing

ahartikainen:
Make dir in test_images (which raises OSError. Btw this is a good design, the user can make "old" folder for old images while testing)

Use khat to test saving. It makes only one plot.

### DIFF
--- a/arviz/tests/conftest.py
+++ b/arviz/tests/conftest.py
@@ -22,7 +22,7 @@ def save_figs(request):
         print("Directory {} created".format(fig_dir))
 
         # Clear all files from the directory
-        # Does not touch directories
+        # Does not alter or delete directories
         for file in os.listdir(fig_dir):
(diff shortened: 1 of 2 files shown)


arviz/tests/conftest.py
```diff
@@ -22,6 +22,7 @@ def save_figs(request):
         print("Directory {} created".format(fig_dir))
 
         # Clear all files from the directory
+        # Does not touch directories
```
canyon289: I'm being exceptionally nitpicky here but "touch" seems ambigious. I think `does not alter or delete directories` would be more clear.

Not required to change, just my backseat coding :)

ahartikainen: Good point. (It was just a quick fix to fix coveralls report). I will change it to

    Does not alter or delete directories

PR #567, opened 2019-01-24
Increase testing

ahartikainen:
- Add tests for animation.
- removed tfp model from plotting tests. The old code was already giving inference data, but now it gives a model and list of arrays, which needs some more wrapping. And this wrapping was a bit hard to implement as a class variables. (Other libs go to convert_infererence_data, but we don't have currently no way to support it, unless tfp outputs some simple class that should contain needed information (lib + data + variables)
- tfp edgecases (one left)


ahartikainen:
Going to merge this, I need to handle that one weird case later.

PR #570, opened 2019-01-26
Fix to plot ordering

daeh:
Previously the order returned was always alphabetical due to use of np.unique(). Now the order of the provided coords are respected.

### DIFF
--- a/arviz/plots/plot_utils.py
+++ b/arviz/plots/plot_utils.py
@@ -242,6 +242,25 @@ def xarray_var_iter(data, var_names=None, combined=False, skip_dims=None, revers
         The string is the variable name, the dictionary are coordinate names to values,
         and the array are the values of the variable at those coordinates.
     """
+
+    def ordered_unique(l):
+        """Remove duplicates from list while conserving order
+        Required arguments:
+            l: Iterable
+        Returns:
+            First occurences in order
+        """
+    
+        def purge_duplicates(l):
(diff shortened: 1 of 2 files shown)


arviz/plots/plot_utils.py
```diff
@@ -242,6 +242,25 @@ def xarray_var_iter(data, var_names=None, combined=False, skip_dims=None, revers
...
+    
+        def purge_duplicates(l):
+            seen = set() # for faster lookups than lists
+            for x in l:
+                if x in seen:
+                    continue
+                seen.add(x)
+                yield x
+    
+        return np.array(list(purge_duplicates(l)))
```
canyon289: Could the same functionality be achieved with
```
_list = []
for item in l:
    if item not in _list:
        _list.append(item)
return _list
```

canyon289:
Thanks for this contribution! This is a good idea

arviz/plots/plot_utils.py
canyon289: The biggest thing is it cuts down 26 lines of code to 5 :)

daeh: updated!

canyon289:
If I can ask for one more change we make the method part of the module scope, rather declared inside of another function scope? This way it can be reused and can be further tested.

canyon289:
LGTM

PR #573, opened 2019-01-28
Fix nan handling and fix n_eff to ess

ahartikainen:
This PR makes ess to return float if NaN is seen.

Until there exists a way to return integer dtype with NaN, this is the best option.

Let's follow how xarray evolves with this issue. https://github.com/pydata/xarray/issues/1194

Also I noticed that there was many `n_eff` left so I changed them to ess.

### DIFF
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -29,7 +29,7 @@ def plot_forest(
     credible_interval=0.94,
     rope=None,
     quartiles=True,
-    eff_n=False,
+    ess=False,
     r_hat=False,
     colors="cycle",
     textsize=None,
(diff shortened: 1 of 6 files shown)


ahartikainen:
Actually we return dataframe by default, but still we have an option to return xarray, so I think we should follow what xarray does.

aloctavodia:
LGTM!

arviz/plots/forestplot.py
```diff
@@ -29,7 +29,7 @@ def plot_forest(
     credible_interval=0.94,
     rope=None,
     quartiles=True,
-    eff_n=False,
```
canyon289: Thanks for catching this

canyon289:
LGTM, I don't know what is going on with coveralls

ahartikainen:
Yeah, it fails sometimes. I will restart one of the builds (e.g. PR) and let's see how that goes. (Inside coveralls is shows 100%)

Linter  one tests `--save`.

PR #574, opened 2019-01-28
negate variable names

aloctavodia:
This allow to pass a variable to `var_names` and get all the variables except the one that was passed.

Thus if `data` is a model with variables names ['mu', 'theta', 'tau']

```python
az.plot_joint(data, var_names=['~theta']);
```

returns

lala

One caveat is that variables names are then no allowed to starts with `~`.

### DIFF
--- a/arviz/utils.py
+++ b/arviz/utils.py
@@ -2,25 +2,37 @@
 import importlib
 
 
-def _var_names(var_names):
+def _var_names(var_names, data):
     """Handle var_names input across arviz.
-
     Parameters
     ----------
     var_names: str, list, or None
-
+    data : xarray.Dataset
+        Posterior data in an xarray
     Returns
     -------
     var_name: list or None
     """
-    if var_names is None:
-        return None
+    if var_names is not None:
 
-    elif isinstance(var_names, str):
-        return [var_names]
+        if isinstance(var_names, str):
+            var_names = [var_names]
 
-    else:
-        return var_names
+        if isinstance(data, (list, tuple)):
+            all_vars = []
+            for dataset in data:
+                dataset_vars = list(dataset.data_vars)
+                for var in dataset_vars:
+                    if var not in all_vars:
+                        all_vars.append(var)
+        else:
+            all_vars = list(data.data_vars)
+
+        exclude_vars = [i[1:] for i in var_names if i.startswith("~")]
+        if exclude_vars:
+            var_names = [i for i in all_vars if i not in exclude_vars]
+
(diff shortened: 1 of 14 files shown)


canyon289:
Can I like this feature more than once? This is a great idea

ahartikainen:
Good idea. I'm assuming `"~a"` is not valid name?

aloctavodia:
Thanks @canyon289 and @ahartikainen. Right names like "~a" are not valid.

ahartikainen:
How can we handle this situation?

    az.from_dict({"mu" : np.random.randn(2,30,3), "~mu" : np.random.randn(2,30)})

arviz/utils.py
```diff
@@ -2,25 +2,37 @@
...
+            all_vars = []
+            for dataset in data:
+                dataset_vars = list(dataset.data_vars)
+                for var in dataset_vars:
+                    if var not in all_vars:
+                        all_vars.append(var)
+        else:
+            all_vars = list(data.data_vars)
+
+        exclude_vars = [i[1:] for i in var_names if i.startswith("~")]
```
ahartikainen: exclude_vars = [i[1:] for i in var_names if i.startswith("~") and i not in all_vars]

aloctavodia: This will give the wrong result

aloctavodia: OK, with this change negation will work as expected except that if a variable name starts with '~' like `~mu` when passing `~mu` to `var_names` it will understand that you want the variable `~mu` and not to negate a variable named `mu`. Passing `~~mu` will be understood as negating the variable `~mu`. Is this OK?

canyon289: I think thats fine. We're only promising a negation if the leading character is a tilda

ahartikainen: But... how do you negate `mu` in that situation?

canyon289: you're saying if a model contains both `mu` and `~mu` how would you negative one but not the other?

ahartikainen: How do you negate mu. I'm not saying we need to handle it, but warn users

canyon289: For now I think that's reasonable. Scan the var_names in InferenceData and see if any are leading with a tilda. If so provide warning negation behavior might be weird.

If it comes up enough we can solve further?

aloctavodia: oops, I missed this conversation, I will add the warning.

ahartikainen:
LGTM

PR #579, opened 2019-02-01
Feature/summary unpack order

ahartikainen:
Have the opportunity to unpack nD objects (n > 1) against either C-order (iterate last dimension first) or F-order (iterate first dimension first).

Implemented only for "wide" format.

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -611,6 +611,7 @@ def summary(
     stat_funcs=None,
     extend=True,
     credible_interval=0.94,
+    order="C",
 ):
     """Create a data frame with summary statistics.
 
@@ -640,6 +641,8 @@ def summary(
     credible_interval : float, optional
         Credible interval to plot. Defaults to 0.94. This is only meaningful when `stat_funcs` is
         None.
+    order : {"C", "F"}
+        If fmt is "wide", use either C or F unpacking order. Defaults to C.
 
     Returns
     -------
@@ -680,6 +683,12 @@ def summary(
     fmt_group = ("wide", "long", "xarray")
     if not isinstance(fmt, str) or (fmt.lower() not in fmt_group):
         raise TypeError("Invalid format: '{}'! Formatting options are: {}".format(fmt, fmt_group))
+
+    unpack_order_group = ("C", "F")
+    if not isinstance(order, str) or (order.upper() not in unpack_order_group):
+        raise TypeError(
+            "Invalid order: '{}'! Unpacking options are: {}".format(order, unpack_order_group)
+        )
 
     alpha = 1 - credible_interval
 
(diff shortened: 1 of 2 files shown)


ahartikainen:
I still need to add test for the order.

arviz/tests/test_stats.py
```diff
@@ -134,10 +139,19 @@ def test_summary_nan(centered_eight):
 
 
 def test_summary_bad_fmt(centered_eight):
+    with pytest.raises(TypeError):
+        summary(centered_eight, fmt=1)
     with pytest.raises(TypeError):
         summary(centered_eight, fmt="bad_fmt")
 
 
+def test_summary_bad_unpack_order(centered_eight):
```
canyon289: Can these be parameterized? That way they'll fail independently and we'll get an accurate test count

arviz/tests/test_stats.py
```diff
@@ -134,10 +139,19 @@ def test_summary_nan(centered_eight):
 
 
 def test_summary_bad_fmt(centered_eight):
+    with pytest.raises(TypeError):
```
canyon289: Can these be parameterized? That way they'll fail independently and we'll get an accurate test count

arviz/stats/stats.py
```diff
@@ -786,7 +795,9 @@ def summary(
             if len(values.shape[1:]):
                 metric = list(values.metric.values)
                 data_dict = {}
-                for idx in np.ndindex(values.shape[1:]):
+                for idx in np.ndindex(values.shape[1:] if order == "C" else values.shape[1:][::-1]):
```
canyon289: What does the if statement do here?

ahartikainen: Small numpy indexing trick to use 'F' order.

- C = iterate last dim first
- F = iterate first dim first

arviz/tests/test_stats.py
ahartikainen: Yes

arviz/tests/test_stats.py
ahartikainen: Yes

ahartikainen:
I think this is good to go.

It is a small change, but really annoying to fix by hand.

canyon289:
LGTM

arviz/stats/stats.py
canyon289: But in this statement if `order =="C"` goes nowhere? Or in other words if `order == "C"` it doesn't look like there's any code that gets executed?

ahartikainen: Ah, conditional expression https://docs.python.org/3/reference/expressions.html#conditional-expressions

`values.shape[1:] if order == "C" else values.shape[1:][::-1]`

`value_bool_true if bool else value_bool_false`

you could also chain them

`value1 if bool else value2 if bool2 else value3 if ....`

PR #582, opened 2019-02-06
fix kde scaling

### LINKED ISSUE #581: KDE plot produces strange results
**Description**
arviz's `plot_kde` produces plot with y-axis that doesn't make sense.

**Reproduce**
https://gist.github.com/absolutelyNoWarranty/f417a052c4b8d05fe3da3b95542f8a0a

**Expected behavior**
The peak of the kde should be around 0.4 for the standard gaussian just like Pandas's kde plot.

**Additional context**
Version of `arviz`: 0.3.2

aloctavodia:
This fixes #581. Also add tests and add `xmin` and `xmax` arguments to `_fast_kde`. After internal discussion we decided not to add xmin-xmax arguments to `plot_kde`.

### DIFF
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -182,7 +182,7 @@ def plot_kde(
         if cumulative:
             density_q = density
         else:
-            density_q = np.cumsum(density)
+            density_q = density.cumsum() / density.sum()
         fill_func = ax.fill_between
         fill_x, fill_y = x, density
         if rotated:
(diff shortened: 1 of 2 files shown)


ahartikainen:
LGTM

ahartikainen:
I tested this with locally, and everything looks ok. The integral is maybe 1% off from the 1, but given that there are much bigger uncertainties in KDE estimation than that error and we are not even suggesting that there is no probability mass outside the estimated area,

I will merge this in.

pepperpo:
kde_plot does not normalize the distributions.
Version '0.4.1'
Tried with both stable and development

PR #583, opened 2019-02-09
Add dictionary option for stat_funcs in summary

ahartikainen:
This PR enables one to use a dictionary in stat_funcs. This way the user can name the functions and use lambda with the correct name.

    stat_funcs=[np.std]
    stat_funcs={"std" : np.std, "std_ddof1" : lambda x: np.std(x, ddof=1)}

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -629,13 +629,14 @@ def summary(
         Return format is either pandas.DataFrame {'wide', 'long'} or xarray.Dataset {'xarray'}.
     round_to : int
         Number of decimals used to round results. Defaults to 2.
-    stat_funcs : None or list
-        A list of functions used to calculate statistics. By default, the mean, standard deviation,
-        simulation standard error, and highest posterior density intervals are included.
+    stat_funcs : dict
+        A list of functions or a dict of functions with function names as keys used to calculate
+        statistics. By default, the mean, standard deviation, simulation standard error, and
+        highest posterior density intervals are included.
 
         The functions will be given one argument, the samples for a variable as an nD array,
         The functions should be in the style of a ufunc and return a single number. For example,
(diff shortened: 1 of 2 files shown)


arviz/stats/stats.py
```diff
@@ -629,13 +629,14 @@ def summary(
         Return format is either pandas.DataFrame {'wide', 'long'} or xarray.Dataset {'xarray'}.
     round_to : int
         Number of decimals used to round results. Defaults to 2.
-    stat_funcs : None or list
-        A list of functions used to calculate statistics. By default, the mean, standard deviation,
-        simulation standard error, and highest posterior density intervals are included.
+    stat_funcs : dict
```
canyon289: This can be a dict, or an iter right? It looks like if you pass a list or tuple you'll still hit this line

https://github.com/arviz-devs/arviz/pull/583/files#diff-f44bb9f0195d3e0d831b3e10ee0ae059R710

arviz/tests/test_stats.py
```diff
@@ -140,8 +140,13 @@ def test_summary_unpack_order(order):
         assert col1 == col2
 
 
-def test_summary_stat_func(centered_eight):
-    assert summary(centered_eight, stat_funcs=[np.var]) is not None
+@pytest.mark.parametrize(
+    "stat_funcs", [[np.var], {"var": np.var, "var2": lambda x: np.var(x) ** 2}]
```
canyon289: In regards to the comment above this test seems to pass a list and still functions.

arviz/stats/stats.py
ahartikainen: Oh yes, dict or sequence

arviz/tests/test_stats.py
ahartikainen: Yes, I wanted to keep original structure

PR #592, opened 2019-02-20
add distplot

### LINKED ISSUE #146: Implement `distplot` equivalent
There are a few places where we would like to use `arviz.kdeplot` if a variable is continuous, and a histogram if not (traceplots, for example).  

This might involve implementing a `histplot` first and then having `distplot` just inspect the `dtype`.

(happy if someone wants to change the names of any of these...)

aloctavodia:
Fix #146

This should not pass the tests as I change `plot_utils.get_bins`. A 2D histogram (I guess an hexbin) is missing also tests and docstring. So very early draft.

index

### DIFF
--- a/arviz/plots/distplot.py
+++ b/arviz/plots/distplot.py
@@ -1,0 +1,94 @@
+def plot_dist(
+    values,
+    values2=None,
+    color="C0",
+    kind="auto",
+    cumulative=False,
+    label=None,
+    rotated=False,
+    rug=False,
+    bw=4.5,
+    quantiles=None,
+    contour=True,
+    fill_last=True,
+    textsize=None,
+    hist_kwargs=None,
+    plot_kwargs=None,
+    fill_kwargs=None,
+    rug_kwargs=None,
+    contour_kwargs=None,
+    ax=None,
+):
+
+    if ax is None:
+        ax = plt.gca()
+
+    if hist_kwargs is None:
+        hist_kwargs = {}
+    hist_kwargs.setdefault("cumulative", cumulative)
+
+    if plot_kwargs is None:
+        plot_kwargs = {}
+
+    if rotated:
+        hist_kwargs.setdefault("orientation", "horizontal")
+    else:
+        hist_kwargs.setdefault("orientation", "vertical")
+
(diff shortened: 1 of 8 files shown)


arviz/plots/traceplot.py
```diff
@@ -114,9 +120,14 @@ def plot_trace(
 
     trace_kwargs.setdefault("alpha", 0.35)
 
-    if kde_kwargs is None:
-        kde_kwargs = {}
-
+    if hist_kwargs is None:
```
ColCarroll: `if hist_kwargs is None` is repeated on line 131

arviz/plots/distplot.py
```diff
@@ -0,0 +1,162 @@
...
+        bins = get_bins(values)
+        ax.hist(
+            values,
+            bins=bins,
+            color=color,
+            label=label,
+            rwidth=0.9,
+            align="left",
+            density=True,
+            **hist_kwargs
```
ColCarroll: elsewhere we would do something like this, to allow for defaults:

```python
kwargs = {
    'bins': bins,
    'color': color,
    'label': label,
    'rwidth': 0.9,
    'align': 'left'
    'density': True
}
kwargs.update(hist_kwargs)
ax.hist(values, **kwargs)
```

arviz/plots/distplot.py
```diff
@@ -0,0 +1,155 @@
...
+    Returns
+    -------
+    ax : matplotlib axes
+    """
+    if ax is None:
+        ax = plt.gca()
+
+    if hist_kwargs is None:
+        hist_kwargs = {}
+    hist_kwargs.setdefault("bins", None)
```
canyon289: Being really picky here but this could be put into a loop to reduce lines of code

aloctavodia: This will actually add lines of code as black will put the items in the list one per line, plus the "lonely bracket" and then the loop.

aloctavodia:
I guess this is ready (except the unrelated Pylint issues)

canyon289:
Can you merge in master for CI pass?

canyon289:
I meant the other way but looks like thing still works :D

PR #590, opened 2019-02-17
Added documentation example for combining InferenceData (#552) and also solving the issue of creaton of empty file for better behaviour with inference data(#486)

Shashankjain12:
Added example in the docstring as per the issue number #552

PR #586, opened 2019-02-09
fix xticks for discrete variables

aloctavodia:
I was not able to reproduce the error reported at #585, but this fix may fix #585 or at least a related problem.

PR #577, opened 2019-01-31
Improving the behaviour of an empty InferenceData object. (Issue no 486).

Ban-zee:
Addressing issue #486

PR #548, opened 2019-01-17
add gitter badge

aloctavodia:

PR #596, opened 2019-02-22
Bump minor version

### LINKED ISSUE #594: Installing arviz breaks pymc3 installation
**Describe the bug**
Installing Arviz breaks a pymc3 installation, which is unfortunate because they're built to be compatible. After installation, importing pymc3 throws the following error. 

> WARNING (theano.tensor.blas): Using NumPy C-API based implementation for BLAS functions.

The reason is because arviz installation requires numpy==1.15 rather than numpy>=1.15. If you have 1.16, it uninstalls it and re-installs 1.15. It's annoying to fix. I ended up having to scrap the whole virtual environment and start over.

**To Reproduce**
Install arviz if you have any version of numpy other than 1.15, then import pymc3. 

**Expected behavior**
Do not force downgrade of numpy.

ColCarroll:
Fixes #594, and brings some nice new features onto pypi.

## New features

- Plot ppc supports multiple chains (#526)
- Plot titles now wrap (#441)
- plot_density uses a grid (#379)
- emcee reader support (#550)
- Animations in plot_ppc (#546)
- Optional dictionary for stat_funcs in summary (#583)
- Can exclude variables in selections with negated variable names (#574)

## Maintenance/fixes

- Order maintained with xarray_var_iter (#557)
- Testing very improved (multiple)
- Fix nan handling in effective sample size (#573)
- Fix kde scaling (#582)
- xticks for discrete variables (#586)
- Empty InferenceData saves consistent with netcdf (#577)

## Documentation

- JOSS and Zenodo badges (#537)
- Gitter badge (#548)
- Docs for combining InferenceData (#590)

### DIFF
--- a/arviz/__init__.py
+++ b/arviz/__init__.py
@@ -1,6 +1,6 @@
 # pylint: disable=wildcard-import,invalid-name,wrong-import-position
 """ArviZ is a library for exploratory analysis of Bayesian models."""
-__version__ = "0.3.2"
+__version__ = "0.3.3"
 
 import os
 import logging


ahartikainen:
Save & read funcs are deprecated?

ColCarroll:
Good catch -- they list v0.3.2, but weren't in those release notes. I guess we should remove them before a 0.4 release.

Issue #470, opened 2018-12-28
Improve API for plot_kde when used with inference data


canyon289:
## Tell us about it
Two issues
1. `plot_kde` is awkward to use with `az.InferenceData` objects as it requires a coercion to a numpy array to work
2. Rug plots cause an exception in the following

Code to reproduce
```
import arviz as az
eight = az.load_arviz_data("centered_eight")
az.plot_kde(eight.posterior.to_array(), rug=True)

```
## Thoughts on implementation
Would be nice to implement a group and fix rugplot exception

Ban-zee:
A few questions:
1)Will the group be specified by the user ? (I guess it will be) 
2)For a given group with n data variables, there should be n rugplots included in the same plot, right?

aloctavodia:
Not sure changing `plot_kde` is the best approach, why not adapting `plot_posterior`?

Ban-zee:
Maybe raise an exception when the user tries to use `plot_kde` on an inference data object, directing them to use the `plot_posterior` instead?

PR #600, opened 2019-02-26
Addressing issue number #470 (using plot_kde with inference data object) 

Ban-zee:
Made the following changes:
 * Using inference data or xr.dataset with plot_kde now raises an error prompting the user to use plot_posterior instead.
 * Rug plots displayed by default for plot_posterior plots.
 * Not included the tests yet.

### DIFF
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -148,7 +148,11 @@ def plot_kde(
     figsize, *_, xt_labelsize, linewidth, markersize = _scale_fig_size(figsize, textsize, 1, 1)
 
     if isinstance(values, xr.Dataset):
-        raise ValueError("Xarray dataset object detected. Use plot_posterior instead of plot_kde")
+        raise ValueError(
+            "Xarray dataset object detected. Use plot_posterior, plot_denstiy, plot_joint or plot_pair  "
+            "instead of "
+            "plot_kde"
+        )
     if isinstance(values, InferenceData):
         raise ValueError(" Inference Data object detected. Use plot_posterior instead of plot_kde")
     if values2 is None:
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -4,7 +4,8 @@ import matplotlib.pyplot as plt
 from scipy.signal import gaussian, convolve, convolve2d  # pylint: disable=no-name-in-module
 from scipy.sparse import coo_matrix
 from scipy.stats import entropy
-
+import xarray as xr
+from ..data.inference_data import InferenceData
 from ..utils import conditional_jit
 from .plot_utils import _scale_fig_size
 
@@ -146,6 +147,10 @@ def plot_kde(
 
     figsize, *_, xt_labelsize, linewidth, markersize = _scale_fig_size(figsize, textsize, 1, 1)
 
+    if isinstance(values, xr.Dataset):
+        raise ValueError("Xarray dataset object detected. Use plot_posterior instead of plot_kde")
+    if isinstance(values, InferenceData):
+        raise ValueError(" Inference Data object detected. Use plot_posterior instead of plot_kde")
     if values2 is None:
         if plot_kwargs is None:
             plot_kwargs = {}
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -149,9 +149,8 @@ def plot_kde(
 
     if isinstance(values, xr.Dataset):
         raise ValueError(
-            "Xarray dataset object detected.Use plot_posterior,plot_denstiy,plot_joint or plot_pair"
-            "instead of "
-            "plot_kde"
+            "Xarray dataset object detected.Use plot_posterior, plot_density, plot_joint"
+            "or plot_pair instead of plot_kde"
         )
     if isinstance(values, InferenceData):
         raise ValueError(" Inference Data object detected. Use plot_posterior instead of plot_kde")
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -149,10 +149,11 @@ def plot_kde(
 
     if isinstance(values, xr.Dataset):
         raise ValueError(
-            "Xarray dataset object detected. Use plot_posterior, plot_denstiy, plot_joint or plot_pair  "
+            "Xarray dataset object detected.Use plot_posterior,plot_denstiy,plot_joint or plot_pair"
             "instead of "
             "plot_kde"
         )
+        raise ValueError("Xarray dataset object detected. Use plot_posterior instead of plot_kde")
     if isinstance(values, InferenceData):
         raise ValueError(" Inference Data object detected. Use plot_posterior instead of plot_kde")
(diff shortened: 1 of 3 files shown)


Ban-zee:
pylint failing because of the following :
`E:301,17: Instance of 'InferenceData' has no 'posterior' member (no-member)`
for the following code:
`with pytest.raises(ValueError, match="Xarray"):
        plot_kde(eight.posterior)`
Any idea on how to resolve this issue?

canyon289:
The first one you likely need to disable

The second one might be a real error. I would check the exception to ensure its the correct one thats being raised. The pytest documentation can help you understand how the exception is being handled
https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

arviz/tests/test_plots.py
```diff
@@ -6,7 +6,7 @@
 import numpy as np
 import pytest
 import pymc3 as pm
-
+import arviz as az
```
canyon289: This should be a relative import from arviz, not an absolute import

Ban-zee: pylint keeps throwing the following exception: Inference Data has no 'posterior' member. Any explanation as to why this happens?

canyon289: Its because `az.InferenceData` has no posterior member :) at least during static code introspection.

If you look at the code for InferenceData posterior isn't explicitly called out, it's generated during initialization of the class. But pylint only statically analyzes code so when it "looks" it can't find anything and throws the exception.

In our case that error should be surpressed. The reason you in particular are getting so many issues is because pylint released a new version in the last couple of days, and now this PR is getting hit with all the linting errors. 

My suggestion to you is to ignore the linting errors in CI for now. Just make sure it passes in your local environment. Someone will likely make a PR in the next couple of days that handles all the new linting rules, at which point it'll get merged, and this PR can merge or rebase in the new changes. Hope that helps

https://github.com/arviz-devs/arviz/blob/master/arviz/data/inference_data.py#L24

Ban-zee:
Thank you.

arviz/tests/test_plots.py
Ban-zee: The tests do pass on my local environment. Thank you for the explanation. In the meantime, I'll start working on a different issue.

arviz/plots/kdeplot.py
```diff
@@ -146,6 +147,10 @@ def plot_kde(
 
     figsize, *_, xt_labelsize, linewidth, markersize = _scale_fig_size(figsize, textsize, 1, 1)
 
+    if isinstance(values, xr.Dataset):
+        raise ValueError("Xarray dataset object detected. Use plot_posterior instead of plot_kde")
```
aloctavodia: I  would suggest the user to use plot_posterior, plot_density, plot_joint or plot_joint.

arviz/plots/posteriorplot.py
```diff
@@ -358,6 +358,7 @@ def format_axes():
             fill_kwargs={"alpha": kwargs.pop("fill_alpha", 0)},
             plot_kwargs={"linewidth": linewidth},
             ax=ax,
+            rug=True,
```
aloctavodia: The default value should be `False`, just to keep this plot close to the originals by Kruschke

ahartikainen:
This PR needs to be rebased against the current Master. It contains now many commits from previous PRs.

This probably should work with forks:

So in your ArviZ folder (forked repo) typed to terminal

First select `master` branch

    git checkout master
    git remote add upstream git://github.com/arviz-devs/arviz.git
    git fetch upstream
    git pull upstream master

Then select your branch

    git checkout i-470
    git rebase master

You can check your log that everything looks good.

    git log --oneline

Push with force to update repo.

    git push -f

Ban-zee:
Hmm, the commits were copied, should I revert back to a previous commit and push again?

ahartikainen:
That looks correct? Those are commits belonging to this PR right?

Ban-zee:
Yes these commits belong to the same PR. It was an error on my part as I just read the commit messages and assumed that the commits were somehow copied :(.

aloctavodia:
LGTM!

arviz/plots/kdeplot.py
```diff
@@ -146,6 +147,14 @@ def plot_kde(
 
     figsize, *_, xt_labelsize, linewidth, markersize = _scale_fig_size(figsize, textsize, 1, 1)
 
+    if isinstance(values, xr.Dataset):
+        raise ValueError(
+            "Xarray dataset object detected.Use plot_posterior,plot_denstiy,plot_joint or plot_pair"
+            "instead of "
+            "plot_kde"
```
canyon289: Seems weird there's an extra line break here. Doesn't hurt anything but just curious why its there?

arviz/plots/kdeplot.py
```diff
@@ -146,6 +147,14 @@ def plot_kde(
 
     figsize, *_, xt_labelsize, linewidth, markersize = _scale_fig_size(figsize, textsize, 1, 1)
 
+    if isinstance(values, xr.Dataset):
+        raise ValueError(
+            "Xarray dataset object detected.Use plot_posterior,plot_denstiy,plot_joint or plot_pair"
```
canyon289: Density is spelled incorrectly, spaces between the different plots would be helpful when reading

Ban-zee: I'll make the changes.

canyon289: No worries, this PR is basically at the finish line. Very minor things.

Thanks so much for the contribution

arviz/plots/kdeplot.py
Ban-zee: The initial line was too long and pylint was failing. Added an extra line for the aesthetics :).

canyon289: Just to make sure I understand "instead of plot_kde" was added for aesthetics?

Ban-zee:
Made the changes.

arviz/plots/kdeplot.py
Ban-zee: Yes.

arviz/tests/test_plots.py
```diff
@@ -7,7 +7,7 @@
 import pytest
 import pymc3 as pm
 
-
+from ..data import load_arviz_data
```
canyon289: Being nitpicky here, can this import be combined with the import from `..data` below?

arviz/tests/test_plots.py
```diff
@@ -313,6 +313,14 @@ def test_plot_kde_quantiles(continuous_model, kwargs):
     assert axes
 
 
+def test_plot_kde_inference_data():
+    eight = load_arviz_data("centered_eight")
```
canyon289: Docstring for this test would be great even if its oneline. I know we don't have docstrings for all tests but I'm trying to reverse that trend.

canyon289:
@Ban-zee Apologies for not getting all the comments in on the last round of review. I think you're just about there after these last changes!

canyon289:
Thanks for help!

Issue #353, opened 2018-10-10
Add API docs for plot_joint


canyon289:
https://arviz-devs.github.io/arviz/generated/arviz.plot_joint.html#arviz.plot_joint

See Posterior Docs for example
https://arviz-devs.github.io/arviz/generated/arviz.plot_posterior.html#arviz.plot_posterior

Issue #352, opened 2018-10-10
Add API docs for plot_pair


canyon289:
https://arviz-devs.github.io/arviz/generated/arviz.plot_pair.html#arviz.plot_pair

See Posterior Docs for example
https://arviz-devs.github.io/arviz/generated/arviz.plot_posterior.html#arviz.plot_posterior

Issue #350, opened 2018-10-10
Add api docs for plot_forest


canyon289:
https://arviz-devs.github.io/arviz/generated/arviz.plot_forest.html#arviz.plot_forest

See Posterior Docs for example
https://arviz-devs.github.io/arviz/generated/arviz.plot_posterior.html#arviz.plot_posterior

kishan3:
@canyon289 Sorry, I want to pick this but not sure what to do. The link you shared has already explained the function. Can you explain what is to be done?

canyon289:
Same instructions as here https://github.com/arviz-devs/arviz/issues/351#issuecomment-428910514

PR #601, opened 2019-02-27
Add inline documentation for forest, joint and pairplot

agustinaarroyuelo:
Related to issues  #353,  #352 and #350

### DIFF
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -99,13 +99,26 @@ def plot_forest(
 
     Examples
     --------
-    Ridgeplot
+    Forestpĺot
 
     .. plot::
         :context: close-figs
 
         >>> import arviz as az
         >>> non_centered_data = az.load_arviz_data('non_centered_eight')
+        >>> fig, axes = az.plot_forest(non_centered_data,
+        >>>                            kind='forestplot',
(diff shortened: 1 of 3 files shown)


arviz/plots/forestplot.py
```diff
@@ -96,6 +96,24 @@ def plot_forest(
     Returns
     -------
     gridspec : matplotlib GridSpec
+
+    Examples
```
canyon289: These are great! Are you planning on adding examples for some of the other options/arguments as well?

arviz/plots/jointplot.py
```diff
@@ -56,6 +56,32 @@ def plot_joint(
     axjoin : matplotlib axes, join (central) distribution
     ax_hist_x : matplotlib axes, x (top) distribution
     ax_hist_y : matplotlib axes, y (right) distribution
+
+    Examples
```
canyon289: Same question for jointplot

arviz/plots/forestplot.py
agustinaarroyuelo: Hi! I just added a forestplot example (before it was only ridgeplot) and one more jointplot.

canyon289: Sounds good. Feel free to disregard my comment :)

PR #610, opened 2019-03-09
kde warning and empty plot

aloctavodia:
Currently `plot_kde` fails if all the elements of the input array are `nan` or `inf` or exactly the same number, or for very short array (less than 3 or so). Functions like `plot_trace` will fail for those cases too, and only variables before the failing point will be plotted.

With this PR if `plot_kde` fails (at least for the cases mentioned before.) we get an empty plot and a warning message.

### DIFF
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -1,4 +1,5 @@
 """One-dimensional kernel density estimate plots."""
+import warnings
 import numpy as np
 import matplotlib.pyplot as plt
 from scipy.signal import gaussian, convolve, convolve2d  # pylint: disable=no-name-in-module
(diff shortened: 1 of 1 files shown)


ColCarroll:
This looks good to me. There's an issue in pymc3 right now about not wanting warning boxes on plots with failed kdes, an this is a good compromise.

canyon289:
Can we keep this PR open until tonight? I'm going to use it as a demo for how PRs work during a talk

arviz/plots/kdeplot.py
```diff
@@ -289,6 +295,10 @@ def _fast_kde(x, cumulative=False, bw=4.5, xmin=None, xmax=None):
     std_x = entropy(x - xmin) * bw
 
     n_bins = min(int(len_x ** (1 / 3) * std_x * 2), n_points)
+    if n_bins < 2:
+        warnings.warn("kde plot failed, you may want to check your data")
```
canyon289: Duplication is not great, but if we make this a function then it doesn't save any lines of code

PR #612, opened 2019-03-10
fix Compareplot column indexing

junpenglao:
This is to fix the compatible issue: https://github.com/pymc-devs/pymc3/issues/3401

### DIFF
--- a/arviz/plots/compareplot.py
+++ b/arviz/plots/compareplot.py
@@ -91,9 +91,11 @@ def plot_compare(
 
     yticks_labels = [""] * len(yticks_pos)
 
-    _information_criterion = ["waic", "loo", "WAIC", "LOO"]
-    for information_criterion in _information_criterion:
-        if information_criterion in comp_df.columns:
+    _information_criterion = ["waic", "loo"]
+    for ic in _information_criterion:
+        if ic in comp_df.columns.str.lower():
+            information_criterion = comp_df.columns[
+                comp_df.columns.str.lower() == ic].values[0]
             break
     else:
         raise ValueError(
--- a/arviz/plots/compareplot.py
+++ b/arviz/plots/compareplot.py
@@ -91,7 +91,7 @@ def plot_compare(
 
     yticks_labels = [""] * len(yticks_pos)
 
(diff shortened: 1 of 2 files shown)


arviz/plots/compareplot.py
```diff
@@ -91,7 +91,7 @@ def plot_compare(
 
     yticks_labels = [""] * len(yticks_pos)
 
-    _information_criterion = ["waic", "loo"]
+    _information_criterion = ["waic", "loo", "WAIC", "LOO"]
```
canyon289: We probably should just the comparison case insensitive, especially because I think this might break arviz now

arviz/plots/compareplot.py
```diff
@@ -91,7 +91,7 @@ def plot_compare(
 
     yticks_labels = [""] * len(yticks_pos)
 
-    _information_criterion = ["waic", "loo"]
+    _information_criterion = ["waic", "loo", "WAIC", "LOO"]
     for information_criterion in _information_criterion:
         if information_criterion in comp_df.columns:
```
canyon289: I think we should change this this to be `comp_df.colums.str.lower()`

arviz/plots/compareplot.py
canyon289: Realized this wont break arviz but still feel like making the whole comparison case insensitive is the way to go

arviz/plots/compareplot.py
junpenglao: Yes, but then the indexing below needs to change as well - or you generate a copy of comp_df and overwrite the columns.

canyon289: What do you think about overwriting comp_df columns in place?

junpenglao: That would be a pretty big side effect...

canyon289:
Looks good. Once CI passes I can merge

junpenglao:
Hmm actually there is also dSE --> dse, SE --> se, pWAIC --> p_waic ...
I think maybe a better way to do is a small wrapper in `pymc3.compareplot` to map the columns. Closing this...

PR #615, opened 2019-03-15
Fix  #453

### LINKED ISSUE #453: Return value of plot_pair is numvars the same
The return value of `plot_pair` is a list that contains `numvars` times the array of subplots. I guess it comes from the original version with gridspec. 

Line 168 defines `axs` empty list, which is no longer used, line 219 appends `ax` to `axs`, but `ax` already contains all the subplots. Eventually, `axs` is returned instead of `ax`.

In addition, maybe the docstring should be updated to specify that `ax` (both as input and as return value) is actually an array of axes, not an axes object. It cannot be a single axes because there are many, but I am not sure if everyone will see it right away.

OriolAbril:


OriolAbril:
I don't understand what is `assert ax` supposed to do, nor if this failure should be solved using `np.any()`, `np.all()` or none of them.

ColCarroll:
I believe the goal of the test was to make sure a plot was produced, and returned 1 or more axis object. If you know the _shape_ of the axes that are returned, I would change the test to `assert axes.shape == (8,)` (or whatever). Otherwise `ax.all()` suffices, since each axis item is `True`-thy.

aloctavodia:
thanks @OriolAbril

Issue #421, opened 2018-11-20
Add more instructions on how to use docker for local dev


canyon289:
## Tell us about it
Docker helps install ArviZ consistently across environments and can be used for local dev.

Adding instructions on how to use the container would be great.
Adding to the bash script to automatically start a notebook on a port would be better

OriolAbril:
I will restart work on this issue soon with the goal of definitely closing it. Here is the list of changes I have in mind, feel free to add some more to the list:

* Change the option of `container.sh` from `--clear_cache` to `--clear-cache`: I think it is much more intuitively like this, as it follows the usual style for this kind of commands. I still have to decide between leaving the current option like this anyways, or make `--clear_cache` print something like "Please run --clear-cache".
* Remove the section about docker in README and leave a single line referring to CONTRIBUTING
* Add explanation about copying files from the container to the host machine. Thus, it will be possible to use docker for development.

Also, I have found that in some cases I cannot execute `pip install .` even though I have created the container with a clear cache before, but running `container.sh --clear_cache` again solves the problem. Has anybody experienced this problem, I am not sure if it is worth mentioning (as there is always the possibility of my memory betraying me).

PR #616, opened 2019-03-15
Update Docker instructions

OriolAbril:
I am no docker expert, but as I am trying to develop and test arviz using it, I think it will be useful to update the instructions with the things I have had to do different from what is written. 

I updated some filenames which were incorrect, and some basic instructions on how to run a notebook. `jupyter` is not installed by default in the docker image, therefore one of the steps is installing it, however, the docker image could also be modified.

Related to #421.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -182,22 +182,24 @@ tools:
 ## Developing in Docker
 
 We have provided a Dockerfile which helps for isolating build problems, and local development.
-Install [Docker](https://www.docker.com/) for your operating system, clone this repo, then
-run `./scripts/container.sh --build` (you may need root privileges). This should start a local docker image called `arviz`. 
-The repo will be running the code from your local copy of `arviz`, so it is good for development.
+Install [Docker](https://www.docker.com/) for your operating system, clone this repo. Docker will generate an environment with your local copy of `arviz` with all the packages in Dockerfile.
 
-Afterwards, you can start a container form the image with `docker run -it -p 8888:8888 arviz`, which will open an interactive bash shell. This can be used to run code inside the docker, with the packages listed in the Dockerfile. For instance, to open a terminal, type `jupyter notebook --ip 0.0.0.0 --no-browser --allow-root` inside the docker container. This will output something similar to `http://(<docker container id> or <ip>):8888/?token=<token id>`, and can be accessed at `http://localhost:8888/?token=<token id>`.
+### Testing in Docker
+Testing the code using docker consists of executing the same file 3 times (you may need root privileges to run it). First run `./scripts/container.sh --build`. This starts a local docker image called `arviz`. Then run `./scripts/container.sh --clear-cache` and eventually, run the tests in a container with the built image with `./scripts/container.sh --test`. This should be quite close to how the tests run on TravisCI.
 
-Or to run the tests instead, type `pytest -v arviz/tests/ --cov=arviz/` also inside the docker container.
+### Using the Docker image interactively
+Once the Docker image is built with `./scripts/container.sh --build`, interactive containers can also be run. 
 
-This should be quite close to how the tests run on TravisCI.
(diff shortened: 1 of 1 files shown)


canyon289:
Reviewing on my phone and this looks great. I'll merge when I get back to
my computer.


Thanks for fixing this.

CONTRIBUTING.md
```diff
@@ -182,29 +182,24 @@ tools:
...
-need the notebook instances token to work with the notebook. This token can be
-accessed with
-
-```
-docker exec -it arviz jupyter notebook list
-```
+Install [Docker](https://www.docker.com/) for your operating system, clone this repo. Docker will generate an environment with your local copy of `arviz` with all the packages in Dockerfile.
+
+### Testing in Docker
+Testing the code using docker consists of executing the same file 3 times (you may need root privileges to run it). First run `./scripts/container.sh --build`. This starts a local docker image called `arviz`. Then run `./scripts/container.sh --clear-cache` and eventually, run the tests in a container with the built image with `./scripts/container.sh --test`. This should be quite close to how the tests run on TravisCI.
```
canyon289: I feel like this sentence is cut off. 

`and eventually`

Should there be more to it?

CONTRIBUTING.md
```diff
@@ -182,29 +182,24 @@ tools:
...
+
+### Using the Docker image interactively
+Once the Docker image is built with `./scripts/container.sh --build`, interactive containers can also be run. 
+
+To start a bash shell inside Docker, run:
+
+    $ docker run -it arviz bash
+
+To start a jupyter notebook, there are two steps, first run:
```
canyon289: Thanks much for providing instructions on how to do this

CONTRIBUTING.md
OriolAbril: I will probably be better to remove the comma and `in a container with the built image`. It is not necessary to know the differences between a docker image and a container. It messes up the explanation instead of helping.

CONTRIBUTING.md
OriolAbril: You proposed a script, but I think this ended up being simple enough, so for me a script would be an overkill.

canyon289: Agreed, these instructions are good. Keep them as is!

CONTRIBUTING.md
canyon289: Sounds good :)

CONTRIBUTING.md
```diff
@@ -182,29 +182,24 @@ tools:
...
-need the notebook instances token to work with the notebook. This token can be
-accessed with
-
-```
-docker exec -it arviz jupyter notebook list
-```
+Install [Docker](https://www.docker.com/) for your operating system, clone this repo. Docker will generate an environment with your local copy of `arviz` with all the packages in Dockerfile.
+
+### Testing in Docker
+Testing the code using docker consists of executing the same file 3 times (you may need root privileges to run it). First run `./scripts/container.sh --build`. This starts a local docker image called `arviz`. Then run `./scripts/container.sh --clear-cache` and eventually run the tests with `./scripts/container.sh --test`. This should be quite close to how the tests run on TravisCI.
```
canyon289: Ah ok, now that I see this I have one comment.

The cache should be cleared before the container is built, otherwise they get copied into the image and cause failures with pytest because the models from the host file system don't unserialize correctly in the docker container.

Let me know if you like the working below


Testing the code using docker consists of executing the same file 3 times (you may need root privileges to run it). 
First run `./scripts/container.sh --clear-cache`. Then run `./scripts/container.sh --build`.  This starts a local docker image called `arviz`. Finally run the tests with `./scripts/container.sh --test`. This should be quite close to how the tests run on TravisCI.

OriolAbril: Fixed, thanks.

canyon289:
Looks great thanks!

PR #618, opened 2019-03-18
Added orderedDict to forestplot

### LINKED ISSUE #613: plot_forest() should maintain the order of azdata
**Describe the bug**
When using plot_forest the ordering of variable is not the same as the one in azdata. In my opinion order should be maintained across methods. I can take a look when I have a few spare minutes but currently I don't even understand how the order is determined.

**To Reproduce**
```
data = az.load_arviz_data('centered_eight')

data.posterior

<xarray.Dataset>
Dimensions:  (chain: 4, draw: 500, school: 8)
Coordinates:
  * chain    (chain) int64 0 1 2 3
  * draw     (draw) int64 0 1 2 3 4 5 6 7 8 ... 492 493 494 495 496 497 498 499
  * school   (school) object 'Choate' 'Deerfield' ... "St. Paul's" 'Mt. Hermon'
Data variables:
    mu       (chain, draw) float64 ...
    theta    (chain, draw, school) float64 ...
    tau      (chain, draw) float64 ...
Attributes:
    created_at:                 2018-10-03T14:26:54.822913
    inference_library:          pymc3
    inference_library_version:  3.5

_ = az.plot_forest(data=data, rope=(0,0), combined=True)
```
demo

**Expected behavior**
The plot should begin with mu, then theta (starting with Choate school and ending with Mt. Hermon) and tau

**Additional context**
arviz 0.3.3

GWeindel:
I added OrderedDict module from collections to keep the structure of the InferenceData object. This applies to both the variable names and the dimensions within the variable names. Fixes #613

### DIFF
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -39,7 +39,7 @@ def plot_forest(
     ridgeplot_overlap=2,
     figsize=None,
 ):
-    """Forest plot to compare credible intervals from a number of distributions, generates a forest plot of 100*(credible_interval)% credible intervals from a trace or list of traces.
+    """Forest plot to compare credible intervals from a number of distributions. Generates a forest plot of 100*(credible_interval)% credible intervals from a trace or list of traces.
 
     Parameters
     ----------
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -571,4 +571,4 @@ class VarHandler:
         if self.combined:
             end_y += self.group_offset
 
-return end_y + 2 * self.group_offset
+    return end_y + 2 * self.group_offset
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -494,7 +494,7 @@ class VarHandler:
                 for _, selection, values in datum_iter:
                     label = make_label(self.var_name, selection, position="beside")
                     if label not in label_dict:
-                        label_dict[label] = {}
+                        label_dict[label] = OrderedDict()
                     if name not in label_dict[label]:
                         label_dict[label][name] = []
                     label_dict[label][name].append(values)
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -245,7 +245,10 @@ class PlotHandler:
         self.model_names = list(reversed(model_names))  # y-values are upside down
 
         if var_names is None:
-            self.var_names = list(reversed(*[OrderedDict(datum.data_vars) for datum in self.data]))
+            if len(self.data) > 1:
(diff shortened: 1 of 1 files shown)


GWeindel:
Just saw that I have to add other changes as orderedDict doesn't work when comparing models

GWeindel:
I had to reintroduce set().union when var_names are not specified by the user and when there is more than one model.

At the end we have the following behavior : 
- If one model is provided : 
    - var_names is specified by the user, then the plot is ordered according to the users input
    - var_names is not specified, the plot is ordered according to the InferenceData Object
- If model > 1 : 
    -  var_names is specified by the user, then the plot is ordered according to the users input
    - var_names is not specified, the plot is ordered randomly as before

The ordering of the dimensions is maintained across all scenarios.

The behavior when model > 1 and var_name is not specified is sub-optimal but in my opinion it is an acceptable shortcut.

ColCarroll:
Thanks @GWeindel! 

Looks like you need to run `black` on `foresplot.py`. 

This looks good to me when tests pass, but I would expect the `OrderedDict` only fixes some issues on Python 3.5 (last version without `dict` being ordered), which [has 1.5 years to end of life](https://devguide.python.org/#status-of-python-branches)! I would be interested in whether this also fixes anything on 3.6+ (though definitely that's not necessary to merge, just interested).

It looks like you also got some corner cases for all versions.

GWeindel:
Sorry I'm still not very familiar with github but I promise that for the next PR I will make it flawless on the first try ! Concerning Python 3.6 the dimensions are ordered but the var_names aren't because I guess that the set() is breaking order in python 3.6 also.

canyon289:
Don't worry about github or being perfect. We mess stuff up all the time :)


Thanks for the contributions

ahartikainen:
My PR are usually 50+ commits that I merge to 5 which is then merged to 1.

    git log --oneline
    (Select hash)
    git rebase -i hashhh

GWeindel:
Let's rephrase it otherwise. Next time I will have read the Contributing readme before making a PR :-) 

I will take care of the tests for this PR soon (as part of my learning tasks)

GWeindel:
The last pylint error is linked to the fact that the summary line of forestplot() is too long (or on two lines). I don't know how you handle these so I let it up to you.

ahartikainen:
Is it complaining with this

    """Forest plot to compare credible intervals from a number of distributions.

    Generates a forest plot of 100*(credible_interval)% credible intervals from 
    a trace or list of traces.

    Parameters
    ...

GWeindel:
Great, thanks ! LGTM

ColCarroll:
Thanks for sticking with it @GWeindel! :grin: (and ari for pushing over the line)

PR #619, opened 2019-03-18
Remove tight_layout

### LINKED ISSUE #608: densityplot warning tight_layout
plot density returns the following warning:

_UserWarning: This figure was using constrained_layout==True, but that is incompatible with subplots_adjust and or tight_layout: setting constrained_layout==False. 
  warnings.warn("This figure was using constrained_layout==True, ")_

The solution is to do not use tight_layout()

OriolAbril:
Fixes #608


canyon289:
LGTM. Thanks again

PR #620, opened 2019-03-18
Add logging arviz wide

OriolAbril:
Start implementing logging. Part of #331.
  * [x] Replace print statements
  * [ ] Replace warnings (and modify tests accordingly)
  * [ ] Log error traces in exceptions

### DIFF
--- a/arviz/__init__.py
+++ b/arviz/__init__.py
@@ -16,7 +16,7 @@ _log = logging.getLogger("arviz")
 
 if not logging.root.handlers:
     handler = logging.StreamHandler()
-    formatter = logging.Formatter('%(name)s - %(levelname)s - %(message)s')
+    formatter = logging.Formatter("%(name)s - %(levelname)s - %(message)s")
     handler.setFormatter(formatter)
     _log.setLevel(logging.INFO)
     _log.addHandler(handler)
--- a/arviz/__init__.py
+++ b/arviz/__init__.py
@@ -16,6 +16,8 @@ _log = logging.getLogger("arviz")
 
 if not logging.root.handlers:
     handler = logging.StreamHandler()
+    formatter = logging.Formatter('%(name)s - %(levelname)s - %(message)s')
+    handler.setFormatter(formatter)
     _log.setLevel(logging.INFO)
     _log.addHandler(handler)
 
--- a/arviz/data/io_cmdstan.py
+++ b/arviz/data/io_cmdstan.py
@@ -4,6 +4,7 @@ from copy import deepcopy
 from glob import glob
 import linecache
 import os
+import logging
 import re
 
 
@@ -13,6 +14,9 @@ import xarray as xr
 
 from .inference_data import InferenceData
 from .base import requires, dict_to_dataset, generate_dims_coords
+
+
+_log = logging.getLogger(__name__)
 
 
 class CmdStanConverter:
@@ -42,7 +46,7 @@ class CmdStanConverter:
                     for i, path in enumerate(posterior, 1)
                 )
                 len_p = len(posterior)
-                print("glob found {} files for 'posterior':\n{}".format(len_p, msg))
+                _log.info("glob found %d files for 'posterior':\n%s", len_p, msg)
         self.posterior_ = posterior
         if isinstance(posterior_predictive, str):
             posterior_predictive_glob = glob(posterior_predictive)
@@ -53,7 +57,7 @@ class CmdStanConverter:
                     for i, path in enumerate(posterior_predictive, 1)
                 )
                 len_pp = len(posterior_predictive)
-                print("glob found {} files for 'posterior_predictive':\n{}".format(len_pp, msg))
+                _log.info("glob found %d files for 'posterior_predictive':\n%s", len_pp, msg)
         if isinstance(prior, str):
             prior_glob = glob(prior)
             if len(prior_glob) > 1:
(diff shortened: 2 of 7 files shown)


ahartikainen:
I think `.format` was the structure we are using for string replacement.

I think there are only few cases where `warning.warns` needs `%`?

DidI remember this correctly?

OriolAbril:
Yes, I changed it because otherwise the code would not pass the pylint test.

canyon289:
> I think `.format` was the structure we are using for string replacement.
> 
> I think there are only few cases where `warning.warns` needs `%`?
> 
> DidI remember this correctly?
For logging % is the method you're supposed to use because it defers string interpolation

ahartikainen:
Sounds good.

OriolAbril:
Now I think what is missing is updating the contributing with some instructions on how and when to use logging, and replacing some of the `warnings.warn()` by `logger.warning()`. 

Here are the ones I would replace:
  * [`utils` L38-L44](https://github.com/arviz-devs/arviz/blob/master/arviz/utils.py#L38-L44)
  * [`base` L58-L64](https://github.com/arviz-devs/arviz/blob/master/arviz/data/base.py#L58-L64) and [L118-L124](https://github.com/arviz-devs/arviz/blob/master/arviz/data/base.py#L118-L124)
  * [`ppcplot` L168-L171](https://github.com/arviz-devs/arviz/blob/master/arviz/plots/ppcplot.py#L168-L171)
  * [`kdeplot` L281](https://github.com/arviz-devs/arviz/blob/master/arviz/plots/kdeplot.py#L281) and [L299](https://github.com/arviz-devs/arviz/blob/master/arviz/plots/kdeplot.py#L299)
  * [`stats` L452-L458](https://github.com/arviz-devs/arviz/blob/master/arviz/stats/stats.py#L452-L458), [L470-L474](https://github.com/arviz-devs/arviz/blob/master/arviz/stats/stats.py#L470-L474), [L1013-L1018](https://github.com/arviz-devs/arviz/blob/master/arviz/stats/stats.py#L1013-L1018) and [L1028-L1032](https://github.com/arviz-devs/arviz/blob/master/arviz/stats/stats.py#L1028-L1032)

And this two I would probably replace them with an ordinary print or a logger.info instance:
  * [`diagnostics` L351](https://github.com/arviz-devs/arviz/blob/master/arviz/stats/diagnostics.py#L351) and [L353](https://github.com/arviz-devs/arviz/blob/master/arviz/stats/diagnostics.py#L353)
  
And the ones I would leave as `warnings.warn()`:
  * [`io_netcdf` L59-L62](https://github.com/arviz-devs/arviz/blob/master/arviz/data/io_netcdf.py#L59-L62) and [L94-L97](https://github.com/arviz-devs/arviz/blob/master/arviz/data/io_netcdf.py#L94-L97)

As a side note, currently there is only one `logger.warning()` instance in the whole package at [`io_dict` L47-L50](https://github.com/arviz-devs/arviz/blob/master/arviz/data/io_dict.py#L47-L50).

Once I have updated these according to your criterion, I will try to explain when is each case used in the Contributing.

OriolAbril:
I have pushed an example of how to reproduce the actual behavior, using the warnings for tests with `logging.warning` instead of `warnings.warn`. It uses pytest's [caplog fixture](https://docs.pytest.org/en/latest/logging.html#caplog-fixture) to check that a message is added to the logger and to check its level is warning.

If it is ok with you, I will wait for reviews on the code or the TODO list and afterwards go ahead implementing logging, otherwise it will progressively become more work to use logging in all of ArviZ.

ColCarroll:
Thanks, @OriolAbril! I like the idea of using `logging.warning` instead of `warnings.warn` -- easier to filter, and also more visible by default. I would encourage us to use `logging.info` instead of `print` as well, for the same reason.

This code looks great to me so far -- please flag when you would like this reviewed + merged.

OriolAbril:
I tried to follow the [when to use logging](https://docs.python.org/3.6/howto/logging.html#when-to-use-logging) section in the `logging` tutorial. Thus, I decided to leave as `warnings.warn` the messages related to wrong (or discouraged) use of ArviZ, and to convert to `logging.warning` all the other messages, like for instance warnings about the model. 

Therefore, using draws as the first dimension generates a warning, the kde failing with less than 2 bins too; whereas, the pareto_k having values larger than 0.7 or the matplotlib animation warning are issued via logging.

I find that even following these recommendations from the `logging` docs it is still a subjective choice though. Both of them allow to ignore the messages really easily, but there are some differences. The `warnings` module allows to treat its warnings as exceptions if desired, but its formatting is less powerful and con only be modified via monkeypatching. On the other hand, `logging` does not allow to treat warnings as exceptions but its formatting is extremely powerful and customizable. 

I will wait until new stats and diagnostics are merged before modifying these files. All the other files should already follow this pattern. Please comment if this is what you had in mind to do with logging.

canyon289:
@OriolAbril You're right that the logging docs are more subjective than the other python docs. We trust your judgement.

I think this PR can be merged just fine if you'd like it to. It does all the right things and LGTM but happy to keep it open if you want to keep plugging away

OriolAbril:
It is fine by me to merge it already. My main concern here was choosing the use cases of each of the libraries due to the subjectivity. That is why I did a proposal, but the last word is yours.

canyon289:
@OriolAbril Honestly you're the logging expert now :)

It lgtm

Issue #621, opened 2019-03-20
New R-hat, ESS and quantile-MCSE functions + related plots + rank plots


avehtari:
Add new R-hat, ESS, quantile-MCSE and new plots described in

- Aki Vehtari, Andrew Gelman, Daniel Simpson, Bob Carpenter, Paul-Christian Bürkner (2019): Rank-normalization, folding, and localization: An improved R-hat for assessing convergence of MCMC. [arXiv preprint arXiv:1903.08008](http://arxiv.org/abs/1903.08008).

The reference code in R is available in [monitornew.R](https://github.com/avehtari/rhat_ess/blob/master/code/monitornew.R) and in [monitorplot.R](https://github.com/avehtari/rhat_ess/blob/master/code/monitorplot.R). I think porting required computation from monitornew.R is easy and porting the plots may require a bit more work, but I'm not expert on Python graphics.

There might be also something useful in the corresponding issues for rstan stan-dev/rstan#617 and bayesplot stan-dev/bayesplot#178

ColCarroll:
@avehtari thanks to you and collaborators for the nice paper, releasing code at the same time, and for reaching out to help get this all implemented!

Two questions on rank plots:
1. Does it make sense to think about putting error bars on rank plots, like in the simulation based calibration paper? Part of me thinks that if you want to automatically "score" your rank plot you should just use R-hat.
2.  Any thoughts on how to set a y-axis for "mean centered" rank plots? I like that they show positives and negatives very clearly, but it is hard to tell scale. The error bars might help with this, I suppose.

Centered Eight Schools:
![image](https://user-images.githubusercontent.com/2295568/54924936-b6078680-4ee3-11e9-9eef-20de2c6add9d.png)

Non-centered Eight Schools:
![image](https://user-images.githubusercontent.com/2295568/54924966-c15ab200-4ee3-11e9-8874-7e2f41f4caaf.png)

avehtari:
> Does it make sense to think about putting error bars on rank plots, like in the simulation based calibration paper?

Yes, but that they are not sufficient as mentioned in SBC paper. I think the benefit of rank plots is the simple interpretation. Additional figure type would be ECDF difference as in Fig 14c in SBC paper. It has a better envelope, but it requires more explanation and thus both are useful to have.

>  Part of me thinks that if you want to automatically "score" your rank plot you should just use R-hat.

[...]

PR #623, opened 2019-03-21
New R-hat and ESS

ahartikainen:
Fixes issue #621 (Only data part)

Still needs some working. 
Most of the needed functions are translated. 

Now there is still need to push these changes to summary.


arviz/stats/stats.py
```diff
@@ -854,9 +872,36 @@ def summary(
...
+            mcse = list(mcse.items())
+        for q in mcse:
+            if isinstance(q, (list, tuple, np.ndarray)):
+                key, q = q
+            else:
+                key = "mcse_q{}".format(q)
+            if not (0 < q < 1):
+                _log.warning("mcse quantile values need to be in range of 0 to 1.")
+                continue
+            print(key)
```
canyon289: Whats this print statement for?

ahartikainen: Haha, my print debug (mistake)

ahartikainen:
Ok, I think we need to make some decisions.

Do we want to keep old functionality 

(split-rhat? --> I can reimplement this)

    split_rhat = lambda x: _get_rhat(_split_chain(x))

There is now possibility to use mcse so should we go with this (mcse mean + sd), also there is possibility to use quantiles.

Then we have an option to have even more diagnostics variables, I have not yet implemented these.

Also we start to have alot of functionality with `main_func + make_ufunc + real_func`, so do we want to restructure these?

I think we also need to update docstrings.

ps. mcse quantile need better name (also for summary column name).

This branch is btw on the main lib, so you can try if you want. `pip install -e .`

ColCarroll:
This is nice. 

Do you have any thoughts about implementing these (or at least lower level parts of this) with `numba`? [There's a decorator for that](https://numba.pydata.org/numba-doc/dev/user/vectorize.html). That is one solution to the repetitive code you mention. That would require adding it to the requirements, which we've talked about, but never decided on in the past.

ahartikainen:
Maybe. Faster option would be making ufuncs that returns multiple objects and then making these special function that return multiple items.

```
def _make_ufunc(func, n_output=1, index=Ellipsis, **kwargs):  # noqa: D202
    """Make ufunc from function."""

    def _ufunc(ary):
        target = np.empty(ary.shape[:-2])
        for idx in np.ndindex(target.shape):
            target[idx] = np.asarray(func(ary[idx].ravel(), **kwargs))[index]
        return target

    def _multi_ufunc(ary):
        targets = [np.empty(ary.shape[:-2]) for _ in range(n_output)]
        for idx in np.ndindex(target.shape):
            results = func(ary[idx].ravel(), **kwargs)
            for i, res in enumerate(results):
                target[i][idx] = np.asarray(res)[index]
        return *targets

    n_output > 1:
        return _multi_ufunc
    else:
        return _ufunc
```

arviz/stats/diagnostics.py
```diff
@@ -83,7 +215,6 @@ def _get_ess(sample_array):
         raise TypeError("Effective sample size calculation requires multiple chains.")
 
     acov = np.asarray([_autocov(sample_array[chain]) for chain in range(n_chain)])
-
     chain_mean = sample_array.mean(axis=1)
     chain_var = acov[:, 0] * n_draws / (n_draws - 1.0)
     acov_t = acov[:, 1] * n_draws / (n_draws - 1.0)
```
ahartikainen: Is this correct?

arviz/stats/diagnostics.py
```diff
@@ -83,7 +215,6 @@ def _get_ess(sample_array):
         raise TypeError("Effective sample size calculation requires multiple chains.")
 
     acov = np.asarray([_autocov(sample_array[chain]) for chain in range(n_chain)])
-
     chain_mean = sample_array.mean(axis=1)
     chain_var = acov[:, 0] * n_draws / (n_draws - 1.0)
```
ahartikainen: Should we just calculate the mean first

`mean_var = np.mean(chain_var)`

arviz/stats/diagnostics.py
ahartikainen: It is used in here

rho_hat_odd = 1.0 - (mean_var - np.mean(acov_t)) / var_plus

ahartikainen: It is not scaled here

https://github.com/avehtari/rhat_ess/blob/master/code/monitornew.R

Lines: 106+
```
acov <- lapply(seq_len(chains), function(i) autocovariance(sims[, i])) 
acov <- do.call(cbind, acov)
rho_hat_odd <- 1 - (mean_var - mean(acov[t + 2, ])) / var_plus
```
but is done for the first part

`mean_var <- mean(acov[1, ]) * n_samples / (n_samples - 1) `

ahartikainen: @avehtari do we have a bug in here?

avehtari: @ahartikainen can you clarify? (you can also email or slack me, although I probably don't answer until tomorrow)

ahartikainen: in monitor.R

```
ess_rfun <- function(sims) {
...
  #### acov create ###
  acov <- lapply(seq_len(chains), function(i) autocovariance(sims[, i])) 
  acov <- do.call(cbind, acov)
  chain_mean <- apply(sims, 2, mean)
  #### first column in acov scaled with n_samples / (n_samples  - 1) (so basically 1/N-1)
  mean_var <- mean(acov[1, ]) * n_samples / (n_samples - 1) 
  #### rescaled back (so basically 1/N)
  var_plus <- mean_var * (n_samples - 1) / n_samples
  if (chains > 1) 
    var_plus <- var_plus + var(chain_mean)
  
  # Geyer's initial positive sequence
  rho_hat_t <- rep.int(0, n_samples)
  t <- 0
  rho_hat_even <- 1;
  rho_hat_t[t + 1] <- rho_hat_even
  #### here there is no scaling for the second column
  rho_hat_odd <- 1 - (mean_var - mean(acov[t + 2, ])) / var_plus
  rho_hat_t[t + 2] <- rho_hat_odd
  t <- 2  
  while (t < nrow(acov) - 1 && !is.nan(rho_hat_even + rho_hat_odd) && 
         (rho_hat_even + rho_hat_odd > 0)) {
    rho_hat_even = 1 - (mean_var - mean(acov[t + 1, ])) / var_plus
    rho_hat_odd = 1 - (mean_var - mean(acov[t + 2, ])) / var_plus
    if ((rho_hat_even + rho_hat_odd) >= 0) {
      rho_hat_t[t + 1] <- rho_hat_even
      rho_hat_t[t + 2] <- rho_hat_odd
    }
    t <- t + 2
  }
```

I know 1/N and 1/(N-1) don't really matter, just want to be certain should we scale the second column too?

btw, order could be simplified here (maybe?)

```
mean_var <- mean(acov[1, ]) * n_samples / (n_samples - 1) 
var_plus <- mean_var * (n_samples - 1) / n_samples
```
-->
```
var_plus <- mean(acov[1, ]) 
mean_var <- mean_var * n_samples / (n_samples - 1)
```

arviz/stats/diagnostics.py
```diff
@@ -4,13 +4,104 @@
...
+    ----------
+    x : Numpy array
+        An array containing MCMC samples
+
+    Returns
+    -------
+    acov: Numpy array same size as the input array
+    """
+    acorr = autocorr(x)
+    varx = np.var(x, ddof=1) * (len(x) - 1) / len(x)
```
ahartikainen: What is the point with this?

    np.var(x, ddof=1) * (len(x) - 1) / len(x)

is same as

    np.var(x, ddof=0)

arviz/stats/diagnostics.py
```diff
@@ -353,3 +458,339 @@ def ks_summary(pareto_tail_indices):
...
+    rhat = max(z_split_rhat, z_fsplit_rhat)
+    if round_to is None:
+        return rhat
+    else:
+        return np.round(rhat, round_to)
+
+
+def _ess(sample_array):
+    """Compute the effective sample size for a 2D array."""
+    shape = np.asarray(sample_array).shape
```
ahartikainen: need to transform first

    sample_array = np.asarray(sample_array)

arviz/stats/diagnostics.py
```diff
@@ -353,3 +458,339 @@ def ks_summary(pareto_tail_indices):
...
+    if len(shape) != 2:
+        raise TypeError("Effective sample size calculation requires 2 dimensional arrays.")
+    n_chain, n_draws = shape
+    if n_chain <= 1:
+        raise TypeError("Effective sample size calculation requires multiple chains.")
+
+    acov = np.asarray([_autocov(sample_array[chain]) for chain in range(n_chain)])
+    chain_mean = sample_array.mean(axis=1)
+    chain_var = acov[:, 0] * n_draws / (n_draws - 1.0)
+    acov_t = acov[:, 1] * n_draws / (n_draws - 1.0)
```
ahartikainen: this line is not needed and the scaling is not right

canyon289:
Lots of fixes huh. How you feeling about this one? Ready to merge?

ahartikainen:
Haha, a lot of fixes.

I think we are missing these
- I'm not satisfied with the functionality; There is a lot of code duplication. Wrapping is not nice.
- Are functions returning the correct value
- Split chain before vs after
- Docstrings need update

ahartikainen:
I rewrote those ufunc + wrapper

~I'm not satisfied with the functionality; There is a lot of code duplication. Wrapping is not nice.~
### Are functions returning the correct value
Can we test this against something?

### Split chain before vs after (ess and rhat)
I think this is minor point, currently ess_mean should return the old ess value. 

### Docstrings need update
I think we can start to work these at some point later (currently the main points should be addressed)

So this is ready for inspection and if you think some of these issues are "dealbreaker" let's fix them.

At least let's have some idea how to test the algos are working as intended. Was there some example dataset? (e.g. see discussion https://github.com/stan-dev/rstan/pull/618 )

arviz/stats/stats_utils.py
```diff
@@ -0,0 +1,322 @@
...
+        kwargs=func_kwargs,
+        input_core_dims=input_core_dims,
+        output_core_dims=output_core_dims,
+        **kwargs
+    )
+
+
+def update_docstring(ufunc, func, n_output=1):
+    """Update ArviZ generated ufunc docstring."""
+    module = ""
```
canyon289: Do you think this would be easier to write with fstrings?

ahartikainen: Can we use fstrings with py3.5?

aloctavodia: no, we need Python > 3.6

canyon289: I'm still curious though, would it be easier with fstrings? If is it I'd like to make an issue that if we drop 3.5 support we can refactor this

ahartikainen: I don't think so. This is quite explicit now. Even with fstrings it woukd need to be explicit.

And this is quite minor thing, only good after people start to use this more (ufunc wrapper).

canyon289:
I feel bad that we're letting this one linger. Is there anything we can do to help @ahartikainen

ahartikainen:
No, not really, unless you want to double check the code.

There is still algorithm validation that I need to implement against RStan (I have the numbers)

arviz/stats/stats_utils.py
```diff
@@ -0,0 +1,334 @@
...
+
+def check_valid_size(ary, msg, n=2):
+    """Validate 2D array shape."""
+    ary = np.asarray(ary)
+    shape = ary.shape
+    if len(shape) != 2:
+        raise TypeError("{} calculation requires 2 dimensional array.".format(msg))
+    n_chain, n_draw = shape
+    if n_chain < n:
+        raise TypeError("{} calculation requires multiple chains.".format(msg))
```
OriolAbril: I would include the n in this message too, like in the draws error message. This is the error risen in the tests currently failing for forest plots with `ess=True`. The data used for testing has only 2 chains, and ess checks for <3, thus, this error is risen even though there are already multiple chains present.

arviz/stats/stats.py
```diff
@@ -885,67 +834,20 @@ def summary(
         summary_df = df
     else:
         summary_df = joined
```
OriolAbril: Here joined is still an xarray object. It should probably be `joined.to_dataframe()`.

ahartikainen: I think this is on purpose

ahartikainen:
Currently the code is still broken for some edge cases, so I will keep fixing it. After that is done, I can address the comments given.

I have also the "testing" setup "done" so, I don't think this will take too long anymore.


There are some (edge) cases we need to handle.

Do we allow 1 chain?
Do we raise exception or return nan if given 1 chain?
Do we return nan if draws < 2 or 3?

arviz/stats/stats.py
OriolAbril: But then it will raise an error afterwards when calling `summary_df.columns`. pandas Dataframes have this attribute, but xarray objects do not.

avehtari:
> There are some (edge) cases we need to handle.

Good questions.

> Do we allow 1 chain?
> Do we raise exception or return nan if given 1 chain?

Stan/RStan/PyStan C++ code allows 1 chain. monitor.R returns NaN for rhat as between variance is not defined, but returns 1 chain ess without using rhat. Stan dev team discussion result was to allow 1 chain, and possibly add a warning.

> Do we return nan if draws < 2 or 3?

monitor.R returns NaN if draws per chain < 3

in addition rstan will give warning if ess is small or rhat big, so there will be warnings if the total number of draws is small

ahartikainen:
Now everything should work (as in algo) and now I just need to do final polishing.

@aloctavodia @ColCarroll @canyon289 
What is your opinion for the restrictions.

@avehtari is 3 draws enough for the mcse_quantile? It does the split_chain twice.

ahartikainen:
Ok, I think this is now ready for review

Basic diagnostics workflow is --> main function for idata / ndarrays
- idata -> ufunc wrapper -> call arr func ->  dataset
- ndarray -> call arr func -> value

New methods for

### `az.rhat(inference_data, method="rank")`
- rank
- split (this is not the same split as we do now, but close enough)
- folded
- z_scale
- identity (call rhat function without transformations)

### `az.ess(inference_data, method="bulk", relative=False)`
### `az.ess(inference_data, method="quantile", prob=0.2, relative=False)`
- bulk (recommended by Vehtari et all)
- tail   (recommended by Vehtari et all)  #kw prob, optional, defaults to 0.05, 0.95
- quantile #kw prob
- mean (old ess)
- sd
- median (quantile, prob=0.5)
- mad (mean absolute deviance)
- z_scale
- folded
- split
- identity (call ess function without transformations)

`az.effective_sample_size` deprecated

### `az.mcse(inference_data, method="mean")`
### `az.mcse(inference_data, method="quantile", prob=0.2)`
- mean
- sd
- quantile

New ufunc wrapper and ufunc maker implemented. Can work with multidimensional data and can output multiple values or multiple datasets.

Autocovariance and autocorrelation updated (follows RStan monitor)

BFMI interface updated

Some (important) rhat, ess and mcse values are validated against RStan monitor values (test docstring contains specific hash and R code to create reference table). Reference data is also found from the saved_models and the same data is used for the reference in Stan.

Edge cases are currently variable, depending how many splits are done.

Basically,
ess --> min_chain = 1, min_draws = 3
rhat --> min_chain = 2, min_draws = 3
mcse --> min_chain = 1, min_draws = 6 (due to call to ess where another split is done)
Returns `nan`, if values are not satisfied.

Added some helper functions for stats.

### Summary table updated to include bulk, tail, mean and sd ess, also mean and sd mcse.
I'm planning to add quantile summary table and Stan summary tables in another PR

Question: Do we add bulk and tail mcse??

canyon289:
I reviewed code at high level and looks good. I can deep dive into questions tonight

avehtari:
> mcse --> min_chain = 1, min_draws = 6 (due to call to ess where another split is done)

I don't think there should be another split. Quickly looking, I couldn't find this in monitor.R reference code

avehtari:
> Question: Do we add bulk and tail mcse??

No. Bulk ESS and Tail ESS are diagnostic values, and what you have for mcse are fine.

ahartikainen:
> > mcse --> min_chain = 1, min_draws = 6 (due to call to ess where another split is done)
> 
> I don't think there should be another split. Quickly looking, I couldn't find this in monitor.R reference code

It's for `mcse_quantile`

```
mcse_quantile <- function(sims, prob) {
  conv_quantile(sims, prob)$mcse
}
```

1st split
```
conv_quantile <- function(sims, prob) {
  if (is.vector(sims)) {
    dim(sims) <- c(length(sims), 1)
  }
  ess <- ess_quantile(split_chains(sims), prob)
  p <- c(0.1586553, 0.8413447, 0.05, 0.95)
  a <- qbeta(p, ess * prob + 1, ess * (1 - prob) + 1)
  ssims <- sort(sims)
  S <- length(ssims)
  th1 <- ssims[max(round(a[1] * S), 1)]
  th2 <- ssims[min(round(a[2] * S), S)]
  mcse <- (th2 - th1) / 2
  th1 <- ssims[max(round(a[3] * S), 1)]
  th2 <- ssims[min(round(a[4] * S), S)]
  data.frame(mcse = mcse, Q05 = th1, Q95 = th2, ess = ess)
}
```
2nd split
```
ess_quantile <- function(sims, prob) {
  I <- sims <= quantile(sims, prob)
  ess_rfun(split_chains(I))
}
```

avehtari:
> 1st split

Thanks. That's an error.

ahartikainen:
Ok, I updated code too (also the reference file was updated).

Now 
ess and mcse work when
    draws > 3, chains > 0
rhat
    draws > 3, chains > 1

With 3 or less draws + split returns a matrix with only 1 pseudo-draw, which results to `nan`.

arviz/stats/stats_utils.py
ahartikainen: I think this should have been fixed. If not, clean these in other PR.

arviz/stats/stats.py
ahartikainen: Yes, this is true, I fixed this problem by returning not-rounded version for xarray.

ahartikainen:
> in addition rstan will give warning if ess is small or rhat big, so there will be warnings if the total number of draws is small

I think ArviZ could also have some similar diagnostics as PyStan currently has (also PyMC3).

Also, I think after we have this PR merged we can start to work on our diagnostic output. E.g. Stan-like text based output and some "customized" object containing pandas dataframe + text box e.g.

These text-based outputs are also needed for loo etc cc. @OriolAbril

arviz/stats/diagnostics.py
```diff
@@ -47,130 +106,135 @@ def effective_sample_size(data, *, var_names=None):
...
+    }
+
+    if method not in methods:
+        raise TypeError(
+            "ESS method {} not found. Valid methods are:\n{}".format(method, "\n    ".join(methods))
+        )
+    if relative:
+        ess_func = methods_relative[method]
+    else:
+        ess_func = methods[method]
```
OriolAbril: I feel like I am missing something really obvious. Why are different methods defined depending on ess or ress? Would it be possible to first calculate the ess and then divide by n_draws*n_chains if `relative=True` or do nothing otherwise?

arviz/stats/stats_utils.py
```diff
@@ -0,0 +1,378 @@
...
+    """
+    if ufunc_kwargs is None:
+        ufunc_kwargs = {}
+    if func_args is None:
+        func_args = tuple()
+    if func_kwargs is None:
+        func_kwargs = {}
+
+    n_output = ufunc_kwargs.pop("n_output", 1)
+    callable_ufunc = make_ufunc(ufunc, n_output=n_output, **ufunc_kwargs)
```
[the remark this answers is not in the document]
ahartikainen: `ufunc_kwargs.pop("n_output", 1)` is same as: find "n_output" and remove from dict if found, else return 1

This can be other than 1.

`make_ufunc` can be used without `ufunc_wrapper` (not remember what was the name).

arviz/stats/diagnostics.py
ahartikainen: Yes, that could be done. (etc different `ess` functions could take `relative` keyword).

In this case I wanted to keep these separate, so each function is explicit and also the `ess` functions would be as clean as possible.

ahartikainen: And `ess` function can also eat InferenceData or other objects, so handling relativity is easier with a func.

arviz/stats/stats_utils.py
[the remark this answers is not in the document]
OriolAbril: Yes, what I mean is: if `ufunc_kwargs` has "n_output" as key, `n_output` will be different than one. If "n_output" is not a key in `ufunc_kwargs`, `n_output` will be 1. 
This is the same behavior as passing only `**ufunc_kwargs` to `make_ufunc`: if "n_output" is a key, it will override the default in `make_ufunc`, whereas if it is not a key, the default will be used, which is 1.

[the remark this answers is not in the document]
ahartikainen: Oh yes, that is true. I will fix that tomorrow.

OriolAbril:
Everything looks good to me. I will start working on the text outputs once it is merged. Thanks for all the work!

OriolAbril:
I have just seen that ess_mean and ess_split perfom exactly the same calculation, and the same happens in monitornew.R that Aki linked. Why is that?

Also, if this is intended, we could remove one of them and make both methods map to the same function.

ahartikainen:
Good catch, you can make new PR, just remove split function and let `split` method refer to `mean`.

avehtari:
> I have just seen that ess_mean and ess_split perfom exactly the same calculation, and the same happens in monitornew.R that Aki linked. 

There is no ess_split in monitornew.R. The behavior of some functions including ess_mean were fixed after making the issue, which may have caused confusion.

OriolAbril:
Sorry, I was referring to the variable in [line 655](https://github.com/avehtari/rhat_ess/blob/1c0ddafdb146f2b5824a14fffcf45e9e0fb14db2/code/monitornew.R#L655) of monitornew. Now I have checked the current version of monitor in RStan, and monitor_extra is not present, which is probably where the misunderstanding comes. 

What would you recommend about ess_split, to map it to ess_mean (as we were thinking about) or directly removing it?

avehtari:
I would remove ess_split. monitor_extra had more variants to verify that the choices we made were the best. These different variants can be now composed from the current functions in monitor (or monitornew). I recommend having functions which compute recommended things by default, and if you want to allow experimenting then make it require additional option or use of several functions.

Issue #390, opened 2018-11-03
Write helper method for tests in ArvizTestData


canyon289:
There are numerous test functions `test_data.py` that have multiple asserts to check everything was loaded up properly.

It would be better to turn this into an iterator rather than line the asserts up linearly

Example
https://github.com/arviz-devs/arviz/blob/master/arviz/tests/test_data.py#L426

Possible Solution
https://github.com/arviz-devs/arviz/pull/388#discussion_r230570675

ahartikainen:
Fixed by #624

PR #624, opened 2019-03-21
Helper method for tests

OriolAbril:
In order to adress Issue #390 I started creating a function that performs multiple `hasattr` checks and returns a list of tuples containing `(object, attribute)` for the failed tests, as proposed. I have included an extra argument `parent` to account for the fact that most of the checks are on the same object and/or on childs of the same object. 

I have replaced one of the multiple checks instances for a call to this helper function, to show its possible usage. I have used `assert len(failed_checks) == 0` instead of `assert failed_checks == []` because its message when failing is much more clear to me:

**`len == 0` message**

    E  AssertionError: assert 2 == 0
    E    +  where 2 = len([(<xarray.Dataset>\nDimensions:  (A_dim_0: 2, B_dim_0: 5, B_dim_1: 2, chain: 2, draw: 10)\nCoordinates:\n  * chain    (ch... draw, D_dim_0, D_dim_1) float64 0.7048 0.64 ... 0.1206\nAttributes:\n    created_at:  2019-03-21T00:14:24.493919, 'Ci')])

**`list == []` message**

    E  AssertionError: assert [(<xarray.Dat...585079, 'Ci')] == []
    E    Left contains more items, first extra item: (<xarray.Dataset>\nDimensions:  (A_dim_0: 2, B_dim_0: 5, B_dim_1: 2, chain: 2, draw: 10)\nCoordinates:\n  * chain    (cha...draw, B_dim_0, B_dim_1) float64 -0.2893 -0.418 ... 1.061\nAttributes:\n    created_at:  2019-03-21T00:14:39.583937, 'Ai')

At this level of verbosity none of the options shows clearly which checks failed, but the first option shows clearly how many of them failed.

### DIFF
--- a/arviz/tests/test_data.py
+++ b/arviz/tests/test_data.py
@@ -525,7 +525,7 @@ class TestDictIONetCDFUtils:
             "posterior_predictive": [],
             "prior_predictive": [],
             "sample_stats_prior": [],
-            "observed_data": [],
+            "observed_data": ["J", "y", "sigma"],
         }
         fails = check_multiple_attrs(test_dict, inference_data)
         assert not fails
@@ -760,7 +760,7 @@ class TestPyMC3NetCDFUtils:
             pm.Normal("y2", x, 1, observed=y2_data)
             trace = pm.sample(100, chains=2)
         inference_data = from_pymc3(trace=trace)
-        test_dict = {"posterior": ["x"], "observed_data": ["y1", "y2"]}
+        test_dict = {"posterior": ["x"], "observed_data": ["y1", "y2"], "sample_stats": ["lp"]}
         fails = check_multiple_attrs(test_dict, inference_data)
         assert not fails
         assert not hasattr(inference_data.sample_stats, "log_likelihood")
@@ -942,7 +942,7 @@ class TestPyStanNetCDFUtils:
             model = StanModel(model_code=model_code)
             fit = model.sampling(iter=10, chains=2, check_hmc_diagnostics=False)
             posterior = from_pystan(posterior=fit)
-            test_dict = {"posterior": ["y", "z"]}
+            test_dict = {"posterior": ["y"], "sample_stats": ["lp"]}
             fails = check_multiple_attrs(test_dict, posterior)
             assert not fails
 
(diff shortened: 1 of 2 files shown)


canyon289:
This is shaping up to be great! Thank you!

OriolAbril:
I have rewritten the helper function from scratch. The previous version would end up in `AttributeError` in many cases, because after checking that a dataset is not present in the InferenceData object it would try to access it anyway to check the attributes of the dataset. 

Now the function takes as input a dictionary summarizing the tests (i.e. `{dataset1_name: [var1, var2], dataset2_name: [var]}`) and the InferenceData object. Therefore it is straightforward to perform attribute checks only once the dataset existence has been checked.

arviz/tests/test_data.py
```diff
@@ -176,12 +177,9 @@ def test_addition():
...
     assert new_idata is not None
-    assert hasattr(new_idata, "posterior")
-    assert hasattr(new_idata, "prior")
-    assert hasattr(new_idata.posterior, "A")
-    assert hasattr(new_idata.posterior, "B")
-    assert hasattr(new_idata.prior, "C")
-    assert hasattr(new_idata.prior, "D")
+    test_dict = {"posterior": ["A", "B"], "prior": ["C", "D"]}
+    fails = check_multiple_attrs(test_dict, new_idata)
+    assert len(fails) == 0
```
canyon289: I would change this line to  `assert not fails` for two reasons. One is that the code matches the english expectation in that we don't want any fails. The second is that when pytest produces an assertion failure we can actually see the failed attrs. Right now if it fails there we'd only see something like 
> assertationerror: 7 != 2

OriolAbril: Okey, no problem. As I explained above, I prioritized seeing how many checks failed because none of the approaches showed which ones failed with the old `check_multiple_attrs` because the list was too long and therefore not fully printed, now all the approaches will show all the failed checks tough, either on the first or the second line of the message.
If you agree I will start to replace all multiple checks with this function and `assert not fails`.

canyon289: Feel free to start replacing all the asserts.

The length of the diff in pytest is configurable, but either way since all the comparison logic is contained in one function, if a logic change is needed it only needs to be changed in one place! Great design :)


https://docs.pytest.org/en/latest/assert.html#making-use-of-context-sensitive-comparisons

canyon289:
@OriolAbril Ready for a merge?

OriolAbril:
Unless there are any more reviews yes!

canyon289:
LGTM. Thank you!

PR #626, opened 2019-03-25
Add rankplot

ColCarroll:
This implements `plot_rank`, as described in #621. This still lacks tests, but wanted to get comments on some styling choices I was already making, specifically:

1. Dotted line in the background to show what a uniform histogram would look like.
2. All chains plotted in one axis
3. Using `bins='auto'` by default. In the paper, it looks like a lot of bins were chosen, but I do not understand why. You can replicate that behavior by manually passing a number of bins (see examples below).

It might make sense to also make this an optional panel in a traceplot, in which case the "artist" should get factored out.

```python
data = az.load_arviz_data('centered_eight')
az.plot_rank(data, ['mu', 'tau', 'theta'], coords={'school': ['Choate', 'Deerfield']})
```
image

```python
az.plot_rank(data, ['mu', 'tau', 'theta'], coords={'school': ['Choate', 'Deerfield']})
```
image

```python
data = az.load_arviz_data('centered_eight')
az.plot_rank(data, var_names='tau', bins=50, figsize=(10, 7));
```
image

### DIFF
--- a/arviz/plots/rankplot.py
+++ b/arviz/plots/rankplot.py
@@ -34,7 +34,16 @@ def _sturges_formula(dataset, mult=1):
     return int(np.ceil(mult * np.log2(dataset.draw.size)) + 1)
 
 
-def plot_rank(data, var_names=None, coords=None, bins=None, ref_line=True, figsize=None, axes=None):
+def plot_rank(
+    data,
+    var_names=None,
+    coords=None,
+    bins=None,
+    ref_line=True,
+    mean_centered=False,
+    figsize=None,
+    axes=None,
+):
 
     posterior_data = convert_to_dataset(data, group="posterior")
     if coords is not None:
@@ -64,25 +73,31 @@ def plot_rank(data, var_names=None, coords=None, bins=None, ref_line=True, figsi
         gap = all_counts.max() * 1.05
         width = bin_ary[1] - bin_ary[0]
 
-        y_ticks = []
-
         # Center the bins
         bin_ary = (bin_ary[1:] + bin_ary[:-1]) / 2
+
+        y_ticks = []
         for idx, counts in enumerate(all_counts):
             y_ticks.append(idx * gap)
-            if ref_line:
+            if ref_line and not mean_centered:
                 # Line where data is uniform
                 ax.axhline(y=y_ticks[-1] + counts.mean(), linestyle="--", color="C1")
             # fake an x-axis
-            ax.axhline(y=y_ticks[-1], color="k")
+            ax.axhline(y=y_ticks[-1], color="k", lw=1)
             ax_color = ax.get_facecolor()
+            if mean_centered:
+                heights = counts - counts.mean()
+                color = [f"C{abs(j - 1)}" for j in (heights > 0).astype(int)]
+            else:
+                heights = counts
+                color = "C0"
             ax.bar(
                 bin_ary,
-                counts,
+                heights,
                 bottom=y_ticks[-1],
                 width=width,
                 align="center",
-                color="C0",
+                color=color,
                 edgecolor=ax_color,
             )
         ax.set_xlabel("Rank (all chains)", fontsize=ax_labelsize)
--- a/arviz/plots/rankplot.py
+++ b/arviz/plots/rankplot.py
@@ -1,4 +1,5 @@
 import numpy as np
+import scipy.stats
 
 from ..data import convert_to_dataset
 from .plot_utils import (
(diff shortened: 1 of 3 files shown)


ahartikainen:
Is `bins='auto'` what we want?

arviz/plots/rankplot.py
```diff
@@ -0,0 +1,58 @@
...
+    if axes is None:
+        rows, cols = default_grid(len(plotters))
+
+        figsize, ax_labelsize, titlesize, _, _, _ = _scale_fig_size(
+            figsize, None, rows=rows, cols=cols
+        )
+        _, axes = _create_axes_grid(len(plotters), rows, cols, figsize=figsize, squeeze=False)
+
+    for ax, (var_name, selection, var_data) in zip(axes.ravel(), plotters):
+        ranks = var_data.argsort(axis=None).reshape(*var_data.shape)
```
ahartikainen: This will not give ranks.

Adhoc way would do argsort twice which give ranks, but we actually want to use `scipy.stats.rankdata` that averages the rank for equal values (this is needed for int parameters)

arviz/plots/rankplot.py
```diff
@@ -0,0 +1,58 @@
...
+        all_counts = []
+        for row in ranks:
+            # If `bins` was auto, it gets overwritten here, and carried forward
+            # using the `range` argument makes sure the bins cover all the data
+            counts, bins = np.histogram(row, bins=bins, range=(0, ranks.size))
```
ahartikainen: This is not what we want. It will not return the consistent amount of bins (tested with out example data)

We want to make bins only based on N, not based on data (what "auto" will give us).

Either we need to iterate all chains and take the maximum bin_count and use that or we can use our own "sturges" here (what R uses)

    def sturges(ary):
        return int(np.ceil(np.log2(ary.size)) + 1)

arviz/plots/rankplot.py
```diff
@@ -0,0 +1,58 @@
...
+        y_ticks = []
+        for idx, counts in enumerate(all_counts):
+            y_ticks.append(idx * gap)
+            # Line where data is uniform
+            ax.axhline(
+                y=y_ticks[-1] + counts.sum() / bins.shape[0], linestyle="--", color="C1", zorder=-10
+            )
+            # fake an x-axis
+            ax.axhline(y=y_ticks[-1], color="k")
+            ax.bar(bins[:-1], counts, bottom=y_ticks[-1], width=width, align="edge", color="C0")
```
ahartikainen: edges are not correct here

We should use `align="center"` and `bins=(bins[1:]+bins[:-1])/2`

image

arviz/plots/rankplot.py
```diff
@@ -0,0 +1,58 @@
...
+        all_counts = np.array(all_counts)
+        gap = all_counts.max() * 1.05
+        width = (bins[1] - bins[0]) * 0.95
+
+        y_ticks = []
+        for idx, counts in enumerate(all_counts):
+            y_ticks.append(idx * gap)
+            # Line where data is uniform
+            ax.axhline(
+                y=y_ticks[-1] + counts.sum() / bins.shape[0], linestyle="--", color="C1", zorder=-10
```
ahartikainen: Should this be `np.mean(counts)`, because counts.size = bins.size - 1

arviz/plots/rankplot.py
ahartikainen: I think it would be even better if we would go with 

    def double_sturges(ary):
        return int(np.ceil(2 * np.log2(ary.size)) + 1)

arviz/plots/rankplot.py
ahartikainen: Also, I think this line look better on top.

This should probably be optional ?

ahartikainen:
We could also have option to center against the mean rank

```
count_pos = np.clip(counts-mean_count, 0, None)
    count_neg = np.clip(counts-mean_count, None, 0)
    bars = ax.bar(bins, count_pos, bottom=y_ticks[-1]+mean_count, width=width, align="center", color="C0")
    bars = ax.bar(bins, count_neg, bottom=y_ticks[-1]+mean_count, width=width, align="center", color="C1")
```
this code assumes that bins is the recalculated value mentioned before and uses the `double_sturges`

centered_eight:

image

non_centered_eight:

image


These can also be done with width=1

centered_eight

image


non_centered_eight

image

arviz/plots/rankplot.py
ColCarroll: i guess this is easy for us, since `ary.size` will be just `data.posterior.draw.size`.

arviz/plots/rankplot.py
ColCarroll: optional is a good idea, as is `np.mean`. I like it better in back, but you've made a lot of good points here, so i'm trusting you!

arviz/plots/rankplot.py
ColCarroll: good catch, thank you

ColCarroll:
Responded to all but the last comment, and also replaced the gap between bars with a line around the edge with width 1 and color taken from the background. This looks _much_ better with lots of bars.

Here are the plots from above with the new code. Notice the `tau` plot now matches the plot from the paper:

image

image

image

ahartikainen:
This look good. Could the fake-zero line have `lw=1`?

ColCarroll:
Added `mean_centering` as a keyword argument. I like this approach, but we will need to think about how to scale the y-axis. Perhaps a confidence interval to give absolute y-scales? 

The three plots now look like

image
image
image

junpenglao:
cant wait to try! Personally, I prefer the bar plots style with 20 - 25 bins (10 is a bit too few IMO)

ColCarroll:
Hope to get tests out on the train home in a few hours, but added documentation and in-line examples.

ColCarroll:
(also I agree that ari's suggestion of double Sturges' formula is a good one for this plot)

ColCarroll:
I think this is ready now -- I am punting the ecdf stuff for a follow-up, but this is an ok start.

ravinkumar:
LGTM

canyon289:
LGTM

Issue #497, opened 2019-01-05
Examples and tests for stats.psislw and plot_khat


ahartikainen:
khat functionality needs better documentation and examples.

Ban-zee:
which dataset would be appropriate for documentation?

ahartikainen:
I think we can start with 8-schools? Maybe wait until loo PR is done.

Also, we probably need to update `plot_khat` before we add these examples.

Ban-zee:
Any good reading material about khat and Pareto's indices?

ahartikainen:
See http://mc-stan.org/loo/articles/loo2-example.html and references mentioned there

Ban-zee:
Thank you.

PR #627, opened 2019-03-25
Added example for plot_khat(#497)

Ban-zee:
Added an example for plot_khat and added a test for stats.psislw.


Ban-zee:
Ready for review.

canyon289:
@Ban-zee This work looks great, the next comment is more for @ColCarroll @aloctavodia @ahartikainen, is this really what the API looks like? It takes all these steps to get to a khat plot?

ahartikainen:
I can update psislw function to eat InferenceData now when I'm refactoring `stats`.

Still, nice example

OriolAbril:
Wouldn't it be easier to get the pareto k data from the `loo` function with `pointwise=True`? It is shorter and accounts for the proper relative MCMC efficiency. This should do the plot:

```python
>>> import arviz as az
>>> data = az.load_arviz_data('centered_eight')
>>> loo_data = az.loo(data, pointwise=True)
>>> az.plot_khat(loo_data.pareto_k)```

**Comment by Ban-zee** [comment]
Thank you @OriolAbril for pointing this out.

**Comment by Ban-zee** [comment]
Any more changes to be made?

File: arviz/tests/test_stats.py [hunk 1]
```diff
@@ -262,9 +262,26 @@ def test_loo_warning(centered_eight):
 
 
 def test_psislw():
+    np.random.seed(8)
     linewidth = np.random.randn(20000, 10)
     _, khats = psislw(linewidth)
     assert_array_less(khats, 0.5)
+    assert khats.shape == (10,)
```
ColCarroll:
```suggestion
    assert khats.shape == (linewidth.shape[0],)
```

arviz/tests/test_stats.py
```diff
@@ -262,9 +262,26 @@ def test_loo_warning(centered_eight):
...
+    assert_almost_equal(
```
ColCarroll: this seems pretty brittle/hard to maintain - I might leave the random seed, but get rid of the hardcoded k-hats. @canyon289 any thoughts?

ColCarroll: alternatively, adding a comment explaining _what_ this testing and why it is here. i just imagine being confused seeing this list of numbers in a year or two!

Ban-zee: Instead of using random data, we could use eight schools for the test. Assert that the values of loo.pareto_k and psislw and equal. We would get rid of these harcoded indices as well.

Ban-zee:
Ready for review.

canyon289:
LGTM

aloctavodia:
Thanks @Ban-zee!

PR #628, opened 2019-03-30
Fix doc build, drop travis-sphinx

ColCarroll:
`travis-sphinx` is a nice project with not much code. The logic here is _almost_ the same as the logic there (they also do a check for cname, that we don't use, and tags on non-master branches).

I'm marking this as a draft: once tests pass, I would like to merge on _my_ fork to make sure nothing gets nuked.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,21 +4,21 @@ language: python
 
 matrix:
   include:
-  - name: "Python 3.6 Unit Test"
-    python: 3.6
-    env: PYTHON_VERSION=3.6 PYSTAN_VERSION=latest PYRO_VERSION=latest EMCEE_VERSION=2 NAME="UNIT"
-  - name: "Python 3.6 Unit Test - PyStan=3 Pyro=0.2.1 Emcee=3"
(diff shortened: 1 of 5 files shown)


ColCarroll:
This is ready to merge, but there's some new stylesheet being applied too, so maybe don't merge yet. It:

1. built successfully on my fork: https://colcarroll.github.io/arviz/
2. did not destroy any branches while doing it: https://github.com/ColCarroll/arviz

anyways, check it out: it is thinner than the current version!

ColCarroll:
I found it: there was a `class='body'` being applied on the main content div by `bootstrap-sphinx`, and then `basic.css`, which was being loaded by `bootstrap-sphinx.css` (TIL css can have imports), which set a `max-width` on `body.div`. Which is to say 
1. this is ready to go 
2. i am a frontend ninja now

canyon289:
LGTM, I'll wait 24 hours in case someone else wants to review

ColCarroll:
Restarted current PRs, too.

Issue #629, opened 2019-03-30
az.plot_forest inline docs are not displaying


canyon289:
**Describe the bug**
The inline docs for az.plot_forest do not seem to be rendering correctly
https://arviz-devs.github.io/arviz/generated/arviz.plot_forest.html#arviz.plot_forest

**To Reproduce**
None

**Expected behavior**
Plots should appear similar to https://arviz-devs.github.io/arviz/generated/arviz.plot_posterior.html#arviz.plot_posterior

**Additional context**
None

Issue #607, opened 2019-03-09
documentation, add link to github repo


aloctavodia:
The ArviZ page should have a link to this github repository,  It would be nice to have the octocat logo on the upper-right corner

Issue #409, opened 2018-11-15
Arviz Docs Navbar has site dropdown with duplicate links


canyon289:
#### Summary:
Arviz Docs Navbar has site dropdown with duplicate links. Would be nice to delete the link

#### Description:
https://github.com/ryan-roemer/sphinx-bootstrap-theme/blob/master/sphinx_bootstrap_theme/bootstrap/navbar.html#L21

<img width="939" alt="screen shot 2018-11-15 at 12 16 28 pm" src="https://user-images.githubusercontent.com/7213793/48579124-76df7e00-e8d0-11e8-980f-fbf6a2cb1c65.png">

#### Expected Output:
No site drop down

PR #630, opened 2019-04-01
Documentation fixes

OriolAbril:
I am working in various documentation related issues:

* #629: the vertical spacing was wrong, and the plots were not added as images in the docs.
* #409: I followed the instructions in [sphinx-bootstrap-theme #185](https://github.com/ryan-roemer/sphinx-bootstrap-theme/issues/185) to remove the "Site" dropdown with the repeated links
* I checked the pymc3 docs source to try to do the same and add the github logo (following #607) but I could not understand the source code. I only know how to use the "Fork me on GitHub" ribbon.
* I added a powered by numfocus badge. However, I am not sure on if it should be there; or if it has to say something else; or nothing at all as arviz is affiliated instead of sponsored (compared to pymc).


ColCarroll:
Thanks! I am building this locally to check it out.

ColCarroll:
- plot forest is fixed
- site dropdown is gone
- The sticky navbar will make adding a github banner like that a little funny looking. let's punt that on this already excellent PR.
- Here's the front page now. It feels a little cluttered with badges? I would get rid of the JOSS badge (since it is below, and the DOI badge serves a similar purpose) and the black badge (which is hip, but doesn't effect users).

image

ColCarroll:
This is great! I want to give others a bit of time to chime in about which badges to keep and get rid of. We will need to manually update the README to have the same set of badges, too.

OriolAbril:
I agree with the ribbon being weird in ArviZ's case. Do you know if it is possible to mimic pymc's icon on the upper right corner in a simple way? Or some documentation that may explain it? I searched in the `sphinx-bootstrap` docs with no luck. There seems to be no possibility to do it from `conf.py` and trying to use a custom `layout.html` inside `_templates` I only managed to edit the body, not the navbar.

ColCarroll:
Yes, I think you would have to do a custom template -- I _think_ it can inherit from a base template, but I do not recall where I saw that (and it is also up to the `sphinx-bootstrap` devs to have given you hooks into inheriting).

ahartikainen:
Can we remove all badges and then have badge page?

aloctavodia:
What about having only the NumFOCUS badge at the top, and the JOSS one below. BTW, what is the purpose of the Zenodo badge given that we have the JOSS one?

OriolAbril:
I have managed to add the GitHub icon already in the navbar. Now the front page looks like this:

arviz_homepage_with_github_logo

ColCarroll:
I personally care most about the travis and coveralls (or some sort of CI/testing badges) when I visit a project. I think either of the JOSS badge/Zenodo badges lends us some academic credence, too. It is nice to have numfocus -- we can do that via a badge or with a logo somewhere on the page?

For the sake of being able to merge this, what if we go

[travis][coveralls][JOSS][NumFOCUS]

for now, and can do a followup once we decide?

ColCarroll:
(GH badge looks fantastic, btw)

OriolAbril:
I have modified the badges to leave: [travis][coveralls][Zenodo][NumFOCUS] on top and [JOSS] below. Now it looks like this:

image

I have not modified the readme though (they are already different, the readme has the gitter badge and it makes sense to be there). They should probably be different.

aloctavodia:
I like it!

canyon289:
I like it too. Merging

PR #634, opened 2019-04-03
Colin's name was backwards

ColCarroll:
I *thought* it looked wrong, but kept looking for a misspelling (lot's of "L"s to double or not double), and didn't notice it was backwards.


aloctavodia:
Thanks Collinn! :-)

Issue #633, opened 2019-04-03
Change text on front page of docs


ColCarroll:
## Tell us about it

I want to change **ArviZ is under heavy development** to **ArviZ is ready for general use**

![image](https://user-images.githubusercontent.com/2295568/55444818-3a15da00-5586-11e9-8ae4-7efdf439d6a7.png)

PR #635, opened 2019-04-03
Changed the index page text #633

Ban-zee:


aloctavodia:
We may just remove that line.

ColCarroll:
LGTM once tests pass (I can't imagine how this would break anything, but good practice). 

@Ban-zee - github has some magic built in so if your PR message includes the text "Fixes #633", it will close the issue once the PR merges. It isn't a big deal (we will just go close the issue manually), but wanted to let you know it exists!

Issue #351, opened 2018-10-10
Add API docs for plot_parallel


canyon289:
https://arviz-devs.github.io/arviz/generated/arviz.plot_parallel.html#arviz.plot_parallel

See Posterior Docs for example
https://arviz-devs.github.io/arviz/generated/arviz.plot_posterior.html#arviz.plot_posterior

kishan3:
@canyon289 Sorry, I want to pick this but not sure what to do. The link you shared has already explained the function. Can you explain what is to be done?

canyon289:
Hey Kishan! Thanks for wanting to take this on

In the source code for `posteriorplot` you'll see code blocks like the one below. These code blocks are Sphinx directives that generate documentation. We would like all major functionality to be documented 

The steps will be
1. Add sphinx directive in documentation
2. Verify documentation by running `make html` in the `docs` directory to generate the documentation. It will appear under `_build`.

Let me know if this helps!

https://github.com/arviz-devs/arviz/blob/master/arviz/plots/posteriorplot.py#L103-L106
```
 .. plot::
        :context: close-figs

        >>> import arviz as az
        >>> data = az.load_arviz_data('centered_eight')
        >>> az.plot_posterior(data)
```

kishan3:
Hey @canyon289! thanks for taking the time to explain this. I will be glad to help.

PR #636, opened 2019-04-04
Added normalization in plot_parallel. Fixes #419 and #351.

### LINKED ISSUE #419: Allow normalization in plot_parallel
It would be great if plot_parallel had a keyword arg `normalize` (or standardize), that centers and rescales the variables before plotting. That would make it easier to see things if some posteriors are much more tight than others:
grafik

Ban-zee:
Made the following changes:
* Added two normalization methods, normal and minmax
* Not added the tests yet.
 
Please review my normalization method to ensure that the method has been implemented properly.

### DIFF
--- a/arviz/plots/parallelplot.py
+++ b/arviz/plots/parallelplot.py
@@ -2,10 +2,10 @@
 import matplotlib.pyplot as plt
 import numpy as np
 
+from scipy.stats.mstats import rankdata
 from ..data import convert_to_dataset
 from .plot_utils import _scale_fig_size, xarray_to_ndarray, get_coords
 from ..utils import _var_names
-from scipy.stats import rankdata
 
 
 def plot_parallel(
@@ -77,17 +77,17 @@ def plot_parallel(
     var_names, _posterior = xarray_to_ndarray(
         get_coords(posterior_data, coords), var_names=var_names, combined=True
     )
+    if len(var_names) < 2:
+        raise ValueError("This plot needs at least two variables")
     if normalize:
         if norm_method == "normal":
             _posterior = (_posterior - np.mean(_posterior)) / np.std(_posterior)
         elif norm_method == "minmax":
             _posterior = _posterior - np.min(_posterior) / (np.max(_posterior) - np.min(_posterior))
         elif norm_method == "rank":
-            _posterior = rankdata(_posterior)
+            _posterior = rankdata(_posterior, axis=1)
         else:
             raise ValueError("{} is not supported. Use normal, minmax or rank.".format(norm_method))
-    if len(var_names) < 2:
-        raise ValueError("This plot needs at least two variables")
 
     figsize, _, _, xt_labelsize, _, _ = _scale_fig_size(figsize, textsize, 1, 1)
 
--- a/arviz/plots/parallelplot.py
+++ b/arviz/plots/parallelplot.py
@@ -99,9 +99,9 @@ def plot_parallel(
     if norm_method != "none":
         if norm_method == "normal":
             mean = np.mean(_posterior, axis=1)
-            sd = np.std(_posterior, axis=1)
+            stan_dev = np.std(_posterior, axis=1)
             for i in range(0, np.shape(mean)[0]):
-                _posterior[i, :] = (_posterior[i, :] - mean[i]) / sd[i]
+                _posterior[i, :] = (_posterior[i, :] - mean[i]) / stan_dev[i]
         elif norm_method == "minmax":
             min_elem = np.min(_posterior, axis=1)
             max_elem = np.max(_posterior, axis=1)
--- a/arviz/plots/parallelplot.py
+++ b/arviz/plots/parallelplot.py
@@ -5,6 +5,7 @@ import numpy as np
 from ..data import convert_to_dataset
 from .plot_utils import _scale_fig_size, xarray_to_ndarray, get_coords
 from ..utils import _var_names
+from scipy.stats import rankdata
 
 
 def plot_parallel(
@@ -28,6 +29,7 @@ def plot_parallel(
 
     Parameters
     ----------
+
     data : obj
         Any object that can be converted to an az.InferenceData object
         Refer to documentation of az.convert_to_dataset for details
@@ -55,8 +57,9 @@ def plot_parallel(
     ax : axes
         Matplotlib axes.
     norm_method : str
-        Method for normalizing the data. Methods include normal, minmax as of now.
+        Method for normalizing the data. Methods include normal, minmax and rank.
         Defaults to normal.
+
     Returns
     -------
     ax : matplotlib axes
@@ -80,10 +83,10 @@ def plot_parallel(
             _posterior = (_posterior - np.mean(_posterior)) / np.std(_posterior)
         elif norm_method == "minmax":
             _posterior = _posterior - np.min(_posterior) / (np.max(_posterior) - np.min(_posterior))
+        elif norm_method == "rank":
+            _posterior = rankdata(_posterior)
         else:
-            raise ValueError(
-                "{} is not supported. Use either normal or minmax.".format(norm_method)
-            )
+            raise ValueError("{} is not supported. Use normal, minmax or rank.".format(norm_method))
     if len(var_names) < 2:
         raise ValueError("This plot needs at least two variables")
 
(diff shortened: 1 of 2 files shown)


ahartikainen:
Could you add also `rank`?

`scipy.stats.rankdata`

Ban-zee:
As another normalization method?

ahartikainen:
Yes

Ban-zee:
Given the different methods which rank has, should I use the "average" method or let the user decide by adding another argument?

ahartikainen:
I think we should use average.

ahartikainen:
Oh, btw great PR. Normal and min-max are correct.

Ban-zee:
Sorry for updating the PR this late, had some pending college work. Added rankdata and will add tests by tomorrow.

Ban-zee:
Ready for review.

arviz/plots/parallelplot.py
```diff
@@ -18,6 +20,7 @@ def plot_parallel(
     colord="C1",
     shadend=0.025,
     ax=None,
+    norm_method="normal",
```
aloctavodia: What about removing `normalize` argument and setting `norm_method = None` by default.

arviz/plots/parallelplot.py
```diff
@@ -69,9 +77,17 @@ def plot_parallel(
     var_names, _posterior = xarray_to_ndarray(
         get_coords(posterior_data, coords), var_names=var_names, combined=True
     )
-
     if len(var_names) < 2:
         raise ValueError("This plot needs at least two variables")
+    if normalize:
+        if norm_method == "normal":
+            _posterior = (_posterior - np.mean(_posterior)) / np.std(_posterior)
```
aloctavodia: Is this OK, or the axis need to be specified?

Ban-zee: Mean and S.D of an individual row should be used, however I'll make the changes after further discussion.

Ban-zee: Specified the axis.

Ban-zee:
Ready for review once again after the implementing the suggested changes.

arviz/tests/test_plots.py
```diff
@@ -236,12 +236,17 @@ def test_plot_parallel_raises_valueerror(df_trace):  # pylint: disable=invalid-n
...
+    assert plot_parallel(obj, var_names=["mu", "tau"], norm_method="normal")
+    assert plot_parallel(obj, var_names=["mu", "tau"], norm_method="minmax")
+    assert plot_parallel(obj, var_names=["mu", "tau"], norm_method="rank")
 
 
 def test_plot_parallel_exception(models):
     """Ensure that correct exception is raised when one variable is passed."""
     with pytest.raises(ValueError):
         assert plot_parallel(models.pymc3_fit, var_names="mu")
+    with pytest.raises(ValueError):
```
canyon289: I would parametrize this test rather than have two asserts be checked in one

arviz/tests/test_plots.py
```diff
@@ -236,12 +236,17 @@ def test_plot_parallel_raises_valueerror(df_trace):  # pylint: disable=invalid-n
 def test_plot_parallel(models, model_fit):
     obj = getattr(models, model_fit)
     assert plot_parallel(obj, var_names=["mu", "tau"])
+    assert plot_parallel(obj, var_names=["mu", "tau"], norm_method="normal")
```
canyon289: I would parametrize this test as well

arviz/tests/test_plots.py
```diff
@@ -233,12 +233,11 @@ def test_plot_parallel_raises_valueerror(df_trace):  # pylint: disable=invalid-n
...
 @pytest.mark.parametrize("model_fit", ["pymc3_fit", "stan_fit"])
-def test_plot_parallel(models, model_fit):
+@pytest.mark.parametrize("norm_method", ["normal", "minmax", "rank"])
+def test_plot_parallel(models, model_fit, norm_method):
     obj = getattr(models, model_fit)
     assert plot_parallel(obj, var_names=["mu", "tau"])
-    assert plot_parallel(obj, var_names=["mu", "tau"], norm_method="normal")
-    assert plot_parallel(obj, var_names=["mu", "tau"], norm_method="minmax")
-    assert plot_parallel(obj, var_names=["mu", "tau"], norm_method="rank")
+    assert plot_parallel(obj, var_names=["mu", "tau"], norm_method=norm_method)
```
canyon289: This one could be parametrized too. The trick here is to only have one assert in the test, and only use `pytest.mark.parametrize` once

Ban-zee: Done.I'll do so in the next test as well once this one passes.

arviz/plots/parallelplot.py
```diff
@@ -99,9 +99,9 @@ def plot_parallel(
     if norm_method != "none":
         if norm_method == "normal":
             mean = np.mean(_posterior, axis=1)
-            sd = np.std(_posterior, axis=1)
+            stan_dev = np.std(_posterior, axis=1)
```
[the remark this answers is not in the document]
Ban-zee: Saw the build and all the tests failed :). I guess it would be better if we add another parametrize line.

arviz/tests/test_plots.py
```diff
@@ -232,16 +232,20 @@ def test_plot_parallel_raises_valueerror(df_trace):  # pylint: disable=invalid-n
         plot_parallel(df_trace)
 
 
-@pytest.mark.parametrize("model_fit", ["pymc3_fit", "stan_fit"])
-def test_plot_parallel(models, model_fit):
+@pytest.mark.parametrize(
+    "model_fit", ["pymc3_fit", "stan_fit"],["normal", "minmax", "rank"]
```
canyon289: The error is coming from here. Because `norm_method` is an argument the tester is looking for it in either a fixture or in parameterization.  

To debug faster try running the tests locally to see what fails. I would look at the plotting tests to see a pattern where we pass multiple kwargs to one function

Ban-zee: I wrote the initial test following the pattern of `test_plot_joint` which has a similar situation with their parameter `kind`.

Ban-zee:
Look's like we'll have to add an extra parametrize statement. Combined the multiple assert statements into one as suggested. Any other changes required?

arviz/tests/test_plots.py
```diff
@@ -233,15 +233,18 @@ def test_plot_parallel_raises_valueerror(df_trace):  # pylint: disable=invalid-n
...
+@pytest.mark.parametrize("norm_method", [None, "normal", "minmax", "rank"])
+def test_plot_parallel(models, model_fit, norm_method):
     obj = getattr(models, model_fit)
-    assert plot_parallel(obj, var_names=["mu", "tau"])
+    assert plot_parallel(obj, var_names=["mu", "tau"], norm_method=norm_method)
 
 
-def test_plot_parallel_exception(models):
+@pytest.mark.parametrize("var_names", [None, "mu", ["mu", "tau"]])
+@pytest.mark.parametrize("norm_method", ["foo"])
```
canyon289: This one is confusing to me. If there is only one thing you're testing for why not put this directly line 247?

Sort of like this `assert plot_parallel(models.pymc3_fit, var_names=var_names, norm_method='foo)`

Ban-zee: Don't know why this did not cross my mind. Anyway, I have changed it now.

Ban-zee:
Any other change to be made?

arviz/plots/parallelplot.py
```diff
@@ -69,9 +94,23 @@ def plot_parallel(
...
         get_coords(posterior_data, coords), var_names=var_names, combined=True
     )
-
     if len(var_names) < 2:
         raise ValueError("This plot needs at least two variables")
+    if norm_method is not None:
+        if norm_method == "normal":
+            mean = np.mean(_posterior, axis=1)
+            standard_deviation = np.std(_posterior, axis=1)
+            for i in range(0, np.shape(mean)[0]):
```
ColCarroll: Maybe I am not thinking of the right example, but is this block equivalent to 
```python
_posterior = (_posterior - _posterior.mean(axis=1)) / np.std(_posterior, axis=1)
```

Ban-zee: Had to perform each operation using loops as the above method threw broadcasting issues.

ahartikainen: _posterior = (_posterior - _posterior.mean(axis=1, keepdims=True)) / np.std(_posterior, axis=1, keepdims=True)

What is the shape of the `_posterior`?

Ban-zee: iirc (2000,2).

PR #637, opened 2019-04-04
Pair plot fixes and kde extra options. Fixes #602

### LINKED ISSUE #602: Change pair_plot behavior for MCMC methods that do not produce divergences
**Describe the bug**
When using metropolis hasting sampler in pymc3, then converting to inference data, then using `az.pair_plot(divergences=True)` raises an index error

**To Reproduce**
```
import arviz as az
import pymc3 as pm

x = [1,2]
y = [1,2]
with pm.Model() as model_g:
    α = pm.Normal('α', mu=0, sd=10)
    β = pm.Normal('β', mu=0, sd=1)
    ϵ = pm.HalfCauchy('ϵ', 5)

    μ = pm.Deterministic('μ', α + β * x)
    y_pred = pm.Normal('y_pred', mu=μ, sd=ϵ, observed=y)

# NUTS sampler works when plotting divergences
"""
with model_g:
    trace_nuts_non_centered = pm.sample()

nuts_non_centered_dataset = az.from_pymc3(trace=trace_nuts_non_centered)
az.plot_pair(nuts_non_centered_dataset, var_names=['α', 'β', 'ϵ'], divergences=True)

plt.show()
"""

# Metropolis Hastings raises indexerror
with model_g:
    step = pm.Metropolis()
    trace_mh_non_centered = pm.sample(step=step)

az.plot_pair(trace_mh_non_centered, var_names=['α', 'β', 'ϵ'], divergences=True)
plt.show()
```

**Expected behavior**
~~Divergences are plotted as expected~~
Show a warning and continue plotting

**Additional context**
Versions of `arviz` and other libraries used, operating system used, and anything else that may be useful.

Arviz Version  '0.3.3'
PyMC3 Version : '3.6'

OriolAbril:
It should fix the issue with missing divergence data (I choose to use a `warnings.warn` instead of logging following #620) and added kwarg arguments for contourf and pcolormesh in `plot_kde`.

### DIFF
--- a/arviz/plots/pairplot.py
+++ b/arviz/plots/pairplot.py
@@ -5,7 +5,7 @@ import matplotlib.pyplot as plt
 from matplotlib.ticker import NullFormatter
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 
-from ..data import convert_to_dataset
+from ..data import convert_to_dataset, convert_to_inference_data
 from .kdeplot import plot_kde
 from .plot_utils import _scale_fig_size, xarray_to_ndarray, get_coords
 from ..utils import _var_names
@@ -134,6 +134,7 @@ def plot_pair(
     divergences_kwargs.setdefault("lw", 0)
 
     # Get posterior draws and combine chains
+    data = convert_to_inference_data(data)
     posterior_data = convert_to_dataset(data, group="posterior")
     var_names = _var_names(var_names, posterior_data)
     flat_var_names, _posterior = xarray_to_ndarray(
@@ -142,8 +143,8 @@ def plot_pair(
 
     # Get diverging draws and combine chains
     if divergences:
-        divergent_data = convert_to_dataset(data, group="sample_stats")
-        if hasattr(divergent_data, "diverging"):
+        if hasattr(data, "sample_stats") and hasattr(data.sample_stats, "diverging"):
+            divergent_data = convert_to_dataset(data, group="sample_stats")
             _, diverging_mask = xarray_to_ndarray(
                 divergent_data, var_names=("diverging",), combined=True
             )
--- a/arviz/plots/pairplot.py
+++ b/arviz/plots/pairplot.py
@@ -1,4 +1,5 @@
 """Plot a scatter or hexbin of sampled parameters."""
+import warnings
 import numpy as np
 import matplotlib.pyplot as plt
 from matplotlib.ticker import NullFormatter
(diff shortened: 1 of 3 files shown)


canyon289:
Can tests be added for these?

arviz/plots/pairplot.py
```diff
@@ -141,11 +143,20 @@ def plot_pair(
...
+        if hasattr(data, "sample_stats") and hasattr(data.sample_stats, "diverging"):
+            divergent_data = convert_to_dataset(data, group="sample_stats")
+            _, diverging_mask = xarray_to_ndarray(
+                divergent_data, var_names=("diverging",), combined=True
+            )
+            diverging_mask = np.squeeze(diverging_mask)
+        else:
+            divergences = False
+            warnings.warn(
+                "diverging field not found in sample stats. "
```
canyon289: Great message

arviz/tests/test_plots.py
```diff
@@ -387,6 +401,18 @@ def test_plot_pair_bad(models, model_fit):
         plot_pair(obj, var_names=["mu"])
 
 
+@pytest.mark.parametrize("has_stats", [True, False])
+def test_plot_pair_divergences_warning(has_stats):
+    data = load_arviz_data("centered_eight")
+    if has_stats:
+        data.sample_stats = data.sample_stats.rename({"diverging": "diverging_missing"})
```
canyon289: So if `has_stats` is true that means we set `diverging` to `diverging_mission`?

Should this be the opposite case semantically?

OriolAbril: The warning can be due to 2 reasons, either the InferenceData object has no `sample_stats` group, or the `sample_stats` dataset has no `diverging` variable. In the test, `has_stats=True` indicates the second case, whereas `has_stats=False` is the first case, the input only has a posterior group. 
Maybe I should modify the message of the warning as it is basically thought for the second case.

arviz/plots/pairplot.py
aloctavodia: I like the message, but maybe someone may think that the diverging field _should_ be there. And most likely what is happening is that the sampling method does not provide that information.

arviz/tests/test_plots.py
```diff
@@ -272,15 +272,29 @@ def test_plot_joint_bad(models, model_fit):
     [
         {"plot_kwargs": {"linestyle": "-"}},
         {"contour": True, "fill_last": False},
+        {
+            "contour": True,
+            "contourf_kwargs": {"cmap": "plasma"},
```
aloctavodia: Why `plasma` and not `viridis`, that is the matplotlib default?

OriolAbril: To test different values, no particular reason.

aloctavodia: I have the impression more people like plasma than viridis...

arviz/plots/pairplot.py
OriolAbril: What about ```"Divergences data not found, plotting without divergences. Make sure the sample method provides divergences data and that it is present in the `diverging` field of `sample_stats` or set divergences=False"```

canyon289:
This LGTM. Will leave for a day for others to review

PR #645, opened 2019-04-16
replace entropy by log

aloctavodia:
This will give exactly the same result as before but with a simpler computation (and slightly faster).


aloctavodia:
self-merging as this is a minor change

Issue #354, opened 2018-10-10
Add API docs for plot_khat


canyon289:
https://arviz-devs.github.io/arviz/generated/arviz.plot_khat.html#arviz.plot_khat

See Posterior Docs for example
https://arviz-devs.github.io/arviz/generated/arviz.plot_posterior.html#arviz.plot_posterior

PR #647, opened 2019-04-19
Fix Sphinx warnings

ColCarroll:
There was a warning about a section being `Example` instead of `Examples`, and the restructured text in `ppcplot` was getting interpreted in a funny way.

I also removed the roadmap, which was not linked from anywhere, and was quite out of date.


OriolAbril:
I built the docs locally, and it fixes #354. Probably due to the correction of Examples' misspelling.

PR #648, opened 2019-04-20
Update docker instructions

### LINKED ISSUE #421: Add more instructions on how to use docker for local dev

OriolAbril:
Update CONTRIBUTING.md to guide the full development process in docker.  Closes #421

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -189,18 +189,24 @@ Testing the code using docker consists of executing the same file 3 times (you m
 First run `./scripts/container.sh --clear-cache`. Then run `./scripts/container.sh --build`. This starts a local docker image called `arviz`. Finally run the tests with `./scripts/container.sh --test`. This should be quite close to how the tests run on TravisCI.
 
 ### Using the Docker image interactively
-Once the Docker image is built with `./scripts/container.sh --build`, interactive containers can also be run. 
+Once the Docker image is built with `./scripts/container.sh --build`, interactive containers can also be run. Therefore, code can be edited and executed inside the docker container. Once finished, the results can be copied back to the host machine.
 
 To start a bash shell inside Docker, run:
 
     $ docker run -it arviz bash
 
-To start a jupyter notebook, there are two steps, first run:
+Alternatively, to start a jupyter notebook, there are two steps, first run:
 
     $ docker run --name jupyter-dock -it -d -p 8888:8888 arviz
     $ docker exec -it jupyter-dock pip install jupyter
     $ docker exec -it jupyter-dock jupyter notebook --ip 0.0.0.0 --no-browser --allow-root
 
 This will output something similar to `http://(<docker container id> or <ip>):8888/?token=<token id>`, and can be accessed at `http://localhost:8888/?token=<token id>`.
+
+Once the code has been edited in the docker container, it can be copied back to the host machine with the following command (on the host machine):
+
+    $ docker cp <containerId>:/file/path/within/container /host/path/target
+    
+Where the `containerId` can be found using `docker ps` if the flag `--name` was not used when creating the container. It should not be necessary to keep the container running until everything has been copied, but it is highly recommended to do so. If the container crashes however, many of the files could still be recovered.
 
 #### This guide was derived from the [scikit-learn guide to contributing](https://github.com/scikit-learn/scikit-learn/blob/master/CONTRIBUTING.md)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -193,11 +193,11 @@ Once the Docker image is built with `./scripts/container.sh --build`, interactiv
 
 To start a bash shell inside Docker, run:
 
-    $ docker run -it arviz bash
+    $ docker run --mount type=bind,source="$(pwd)",target=/opt/arviz/ -it arviz bash
 
 Alternatively, to start a jupyter notebook, there are two steps, first run:
 
-    $ docker run --name jupyter-dock -it -d -p 8888:8888 arviz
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -189,18 +189,24 @@ Testing the code using docker consists of executing the same file 3 times (you m
...
-To start a jupyter notebook, there are two steps, first run:
+Alternatively, to start a jupyter notebook, there are two steps, first run:
 
     $ docker run --name jupyter-dock -it -d -p 8888:8888 arviz
     $ docker exec -it jupyter-dock pip install jupyter
     $ docker exec -it jupyter-dock jupyter notebook --ip 0.0.0.0 --no-browser --allow-root
 
 This will output something similar to `http://(<docker container id> or <ip>):8888/?token=<token id>`, and can be accessed at `http://localhost:8888/?token=<token id>`.
 
+Once the code has been edited in the docker container, it can be copied back to the host machine with the following command (on the host machine):
```
canyon289: This part I'm curious about. The container should share a filesystem with the host in the default configuration which means things shouldn't need to be copied. Is this not happening?

README.md
```diff
@@ -175,30 +175,3 @@ Additional information can be found in the [Contributing Readme](https://github.
 ## Code of Conduct
 ArviZ wishes to maintain a positive community. Additional details
 can be found in the [Code of Conduct](https://github.com/arviz-devs/arviz/blob/master/CODE_OF_CONDUCT.MD)
-
-
-### Developing
-
-A typical development workflow is:
```
canyon289: Why remove this section of the doc?

scripts/container.sh
```diff
@@ -15,7 +15,7 @@ if [[ $* == *--build* ]]; then
         --rm
 fi
 
-if [[ $* == *--clear_cache* ]]; then
+if [[ $* == *--clear_cache* ]] || [[ $* == *--clear-cache* ]]; then
```
canyon289: What does this do? Why the duplication?

canyon289: Nevermind I see it. Lets remove the `clear_cache argument` and stick with one command rather than two similar arguments

CONTRIBUTING.md
canyon289: Specifically the bind mount here should allow development in the container to be reflected on the host file system
https://github.com/arviz-devs/arviz/blob/master/scripts/container.sh#L26

I might have messed up though so if that's the case we should fix that rather than copy I think?

kyleabeauchamp: Are you sure that the built containers will automatically mount the dir?  E.g., I would imagine you need to pass the `-v` in the above commands in this README?

OriolAbril: The bind in L26 is in the `--test` option. It does indeed bind the container with the working directory, but the bind has to be specified for every `docker run` instance. Between `--mount type=bind,...` and `-v`, the former is recommended even tough there are no plans to deprecate the latter.

I have updated the guide to use `--mount`.

README.md
OriolAbril: Ii is only a really short summary of CONTRIBUTING, therefore I thought it was better to remove it and leave only the link to CONTRIBUTING, otherwise, whenever we update the contributing guide this section has to be updated too. Moreover, I don't think this short summary was complete enough for new contributors (as opposed to the contributing guide).

CONTRIBUTING.md
OriolAbril: Now that I was thinking about it though, precisely in the `--test` flag case, wouldn't it be better to not bind the container?
 I am not totally sure about what exactly does the clear cache flag, but it looks like it basically removes caches from running the tests. If this is the case, removing the bind would solve these problems. In addition, as only the build flag is used in Travis, this won't have any effect on tests there nor on coveralls.

README.md
canyon289: You've convinced me. All sound like great reasons. Thanks for putting the thought into this

CONTRIBUTING.md
canyon289: Clearing the cache deletes pycache files and cached models. What was happening was that if a user ran a test on their local computer, python would create .pyc files and arviz would cache models. Subsequently if a user tried testing in docker, both python and/or pystan would throw an exception because the C code was now being run in an environment it wasn't compiled in.

By clearing the cache files, ArviZ and its dependencies in docker would have anything to load and this error would be passed.


As for whether to bind or not during tests, the trouble with not binding is that updates were made on the local filesystem, and somebody wanted to test those changes, they would have to rebuild the whole container just to reflect code changes, or do something like copy them in. I figured with ArviZ our recommendation should to bind containers to the local filesystem as our containers are meant for reproducible and rapid development, more so than anything else like app deployment or scaling.

This is all up for debate though! Let me know what you think

OriolAbril: It definitely is much easier with binding. I think that we should indicate that clear cache may have to be executed after testing though.

canyon289: Just to be explicit it might need to be run after the user tested on their local computer, but before they try testing in docker

canyon289:
@OriolAbril Let me know when you want me to review again and I'd be happy to do so. I don't want to keep you waiting

canyon289:
I also did not mean to close this, that was an accident :(

OriolAbril:
I have already included everything I had in mind. Unless there are any more requests I believe it is ready to merge now.

canyon289:
Thank you!

PR #652, opened 2019-04-25
Improving the legend display in plot_kde and plot_energy. Fixes #643

### LINKED ISSUE #643: Color scheme not displayed in plot_energy.
The color scheme is not displayed in the legend for a default energy plot.
arviz-plot_energy-1

This behavior is not observed when we set `kind='hist'`.
arviz-plot_energy-2

Ban-zee:
This commit fixes the legend display in `plot_kde` as discussed in #643. The issue of plot_energy legend has not been resolved yet.

### DIFF
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -2,6 +2,7 @@
 import warnings
 import numpy as np
 import matplotlib.pyplot as plt
+from matplotlib.patches import Patch
 from scipy.signal import gaussian, convolve, convolve2d  # pylint: disable=no-name-in-module
 from scipy.sparse import coo_matrix
 import xarray as xr
@@ -167,6 +168,7 @@ def plot_kde(
         plot_kwargs.setdefault("color", "C0")
 
         default_color = plot_kwargs.get("color")
+
         if fill_kwargs is None:
             fill_kwargs = {}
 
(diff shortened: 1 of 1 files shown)


aloctavodia:
LGTM!

Ban-zee:
`plot_energy` legend still need to be resolved. Ideally, this commit should have resolved that as well but for some reason, it does not.

arviz/plots/kdeplot.py
```diff
@@ -227,7 +229,8 @@ def plot_kde(
             fill_func(fill_x, fill_y, **fill_kwargs)
 
         if legend and label:
-            ax.legend()
+            legend_element = [Patch(edgecolor=default_color, label=label)]
+            ax.legend(handles=legend_element)
```
OriolAbril: This only overrides the previous behavior to always show a square of color `default_color` in the legend. In my opinion, 2 different cases must be detected to generate the legend in different ways.

1. `alpha=0` in `fill_kwargs`: The legend should show a line, not a rectangle. This is what the code used to do in all the cases. It can be left like that or it can be done using a `Line2D` object.
1. `alpha!=0` in `fill_kwargs`: the legend should show a rectangle, whose `facecolor` should be the color in `fill_kwargs`, `edgecolor` is the color in `plot_kwargs`, and both `fill_kwargs` and `plot_kwargs` should be passed (at least partially) to the `Patch` object. This is the reason why I would recommend using `linewidth=0` instead of `alpha=0` to hide line plots (and alpha for area/fill plots).

OriolAbril:
> `plot_energy` legend still need to be resolved. Ideally, this commit should have resolved that as well but for some reason, it does not.

I would tackle this after completely solving the behavior of `plot_kde`. For now the legend in `plot_energy` is based in the fact that the plots are generated with labels _always_, but the legend is only shown if `legend==True`. Thus, plotting several plots and generating the legend at the end will collect all the labels from all plots.

PR #653, opened 2019-04-25
plot_autocorr: add ax argument

aloctavodia:


canyon289:
LGTM.
I think we need to add axes tests to all plotting functions. I can make issue for that

Issue #656, opened 2019-05-01
`from_pymc3` fails when dims and coords are given


arabidopsis:
**Describe the bug**

`az.from_pymc3` fails with exception on multidimensional data when coordinates
are given.
Exception is `ValueError: conflicting sizes for dimension 'exp': length 30 on the data but length 15 on coordinate 'exp'`

**To Reproduce**

```
import pymc3 as pm
import numpy as np
import arviz as az
import xarray as xr

shape = (2,)
data = np.random.randn(15, *shape)
print(data.shape)
with pm.Model() as m:
    mu = pm.Normal("mu", 0, 1, shape=shape)
    dt = pm.Normal("data", mu=mu, sd=1, observed=data, shape=shape)

    trace = pm.sample(23)

d = az.from_pymc3(
    trace, dims=dict(mu=["x"], data=["exp", "x"]),
 coords=dict(x=["a", "b"], exp=list(range(data.shape[0])))
)
```

**Expected behavior**

A valid inference_data object.

**Additional context**

`(az.__version__, pm.__version__,np.__version__,xr.__version__)
== ('0.3.3', '3.6', '1.15.4', '0.12.1')`

also `xr.show_versions()` gives

INSTALLED VERSIONS
------------------
commit: None
python: 3.7.3 (default, Mar 27 2019, 22:11:17) 
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 4.15.0-48-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_AU.UTF-8
LOCALE: en_AU.UTF-8
libhdf5: 1.10.4
libnetcdf: 4.6.1

xarray: 0.12.1
pandas: 0.24.2
numpy: 1.15.4
scipy: 1.2.1
netCDF4: 1.4.2
pydap: None
h5netcdf: None
h5py: 2.9.0
Nio: None
zarr: None
cftime: 1.0.3.4
nc_time_axis: None
PseudonetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: 1.2.1
dask: 1.2.0
distributed: 1.27.0
matplotlib: 3.0.3
cartopy: None
seaborn: 0.9.0
setuptools: 40.6.3
pip: 18.1
conda: 4.6.14
pytest: 4.4.1
IPython: 7.4.0
sphinx: 2.0.1

canyon289:
Is this issue fixed? Looks like a PR was merged that addressed it

PR #657, opened 2019-05-01
keep shape of observations fixes issue #656

arabidopsis:
Fixes issue #656

Log likelihood should keep the dimensions of the original observations. And is essential
if you dimension your data.


ahartikainen:
Will this affect `loo` / `waic` / `compare`

ColCarroll:
This looks good to me -- @ahartikainen it looks like the test was updated to address your question. I'll merge tomorrow if there are no other objections.

PR #658, opened 2019-05-02
Vectorize effective sample size

fehiepsi:
This PR vectorizes effective sample size in all dimensions: variable dimensions, chain dimension, draw dimension. In addition, I think that we don't need multi-chains to be able to compute this value, so I also remove that restriction in this PR (the same thing happens for `split-R-hat`: we can vectorize and don't need multi-chains to be able to compute its value (but I would like to defer the discussion about it to another PR).

Test on `centered_eight` posterior, this PR gives nearly identical result to the current one:
```
    mu       float64 192.0
    theta    (school) float64 377.0 372.0 418.0 301.0 290.0 397.0 347.0 359.0
    tau      float64 130.0
```
```
    mu       float64 192.0
    theta    (school) float64 377.0 372.0 418.0 300.0 290.0 397.0 347.0 359.0
    tau      float64 130.0
```
The difference 301 vs 300 comes from the inconsistent behaviour of `acov_t` vs the remaining `acov[:, t]` values in the current implementation.

By the way, thanks for making this awesome library!

### DIFF
--- a/arviz/stats/diagnostics.py
+++ b/arviz/stats/diagnostics.py
@@ -97,9 +97,9 @@ def _get_ess(sample_array):
 
     # Geyer's initial monotone sequence
     # here we split out the initial value and take the accumulated min of the remaining sequence
-    P_t = np.concatenate([P_t[..., :1],
-                          np.minimum.accumulate(P_t[..., 1:].clip(min=0), axis=-1)],
-                         axis=-1)
+    P_t = np.concatenate(
+        [P_t[..., :1], np.minimum.accumulate(P_t[..., 1:].clip(min=0), axis=-1)], axis=-1
+    )
 
     ess = np.floor((n_chain * n_draws) / (-1.0 + 2.0 * np.sum(P_t, axis=-1)))
     return ess
@@ -122,7 +122,7 @@ def autocorr(x):
     y = x - x.mean(axis=-1, keepdims=True)
     len_y = y.shape[-1]
     result = fftconvolve(y, y[..., ::-1], axes=-1)
-    acorr = result[..., result.shape[-1] // 2:]
+    acorr = result[..., result.shape[-1] // 2 :]
     acorr /= np.arange(len_y, 0, -1)
     acorr /= acorr[..., :1]
     return acorr
--- a/arviz/stats/diagnostics.py
+++ b/arviz/stats/diagnostics.py
@@ -73,6 +73,7 @@ def _get_ess(sample_array):
 
     It is required that the last two dimensions are chain dimension and draw dimension.
     """
+    # pylint: disable=no-member
     if sample_array.ndim == 1:
         sample_array = sample_array[None, ...]
     n_chain, n_draws = sample_array.shape[-2], sample_array.shape[-1]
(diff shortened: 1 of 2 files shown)


arviz/stats/diagnostics.py
```diff
@@ -67,65 +65,44 @@ def _ess_ufunc(ary):
...
+    if sample_array.ndim == 1:
+        sample_array = sample_array[None, ...]
+    n_chain, n_draws = sample_array.shape[-2], sample_array.shape[-1]
 
-    acov = np.asarray([_autocov(sample_array[chain]) for chain in range(n_chain)])
+    acov = _autocov(sample_array)
 
-    chain_mean = sample_array.mean(axis=1)
-    chain_var = acov[:, 0] * n_draws / (n_draws - 1.0)
-    acov_t = acov[:, 1] * n_draws / (n_draws - 1.0)
```
fehiepsi: Here `acov_t = acov[:, 1] * n_draws / (n_draws - 1.0)`

arviz/stats/diagnostics.py
```diff
@@ -67,65 +65,44 @@ def _ess_ufunc(ary):
...
+    chain_mean = sample_array.mean(axis=-1, keepdims=True)
+    chain_var = acov[..., :1] * n_draws / (n_draws - 1.0)
+    mean_var = np.mean(chain_var, axis=-2)
     var_plus = mean_var * (n_draws - 1.0) / n_draws
-    var_plus += np.var(chain_mean, ddof=1)
-
-    rho_hat_t = np.zeros(n_draws)
-    rho_hat_even = 1.0
-    rho_hat_t[0] = rho_hat_even
-    rho_hat_odd = 1.0 - (mean_var - np.mean(acov_t)) / var_plus
```
fehiepsi: and `acov_t` is used to compute the first value of `rho_hat_odd`

arviz/stats/diagnostics.py
```diff
@@ -67,65 +65,44 @@ def _ess_ufunc(ary):
...
-    rho_hat_t[1] = rho_hat_odd
+    if n_chain > 1:
+        var_plus += np.var(chain_mean, axis=-2, ddof=1)
 
     # Geyer's initial positive sequence
-    max_t = 1
-    t = 1
-    while t < (n_draws - 2) and (rho_hat_even + rho_hat_odd) >= 0.0:
-        rho_hat_even = 1.0 - (mean_var - np.mean(acov[:, t + 1])) / var_plus
-        rho_hat_odd = 1.0 - (mean_var - np.mean(acov[:, t + 2])) / var_plus
```
fehiepsi: while we don't multiply with `n_draws / (n_draws - 1.0)` for the remaining value

ahartikainen:
Thanks, but all of this is updated in here https://github.com/arviz-devs/arviz/pull/623

So after it is merged, we can re-evaluate this.

These values could be calculated for one chain, but we shouldn't do it, there are statistical problems with this.

arviz/stats/diagnostics.py
ahartikainen: This behaviour is from the ess paper.

fehiepsi:
Oh, I didn't know that there are some overlapping with #623. The main purposes of this PR are to vectorize variable dimensions, chain dimension, draw dimension (instead of `for v in variable dimension: for d in draw dimension: we calculate rho_hat_t`). And as far as I know, effective sample size definition does not necessary involving chains. Stan devs do an additional job to make this statistics work for chain of samples. So what I take from your comment is: there are some problems with the traditional effective sample size definition? I am just curious because I have not followed recent research on this.

arviz/stats/diagnostics.py
fehiepsi: Yup, this is correct to me. I just pointed out the inconsistent of current code, to explain for why there are a tiny difference between the current result and this PR result.

fehiepsi:
Closed because there are conflicts with #623 where for loop is preferred than vectorized. A lot of work would be done to be able to understand and vectorize many changes in #623. Please feel free to adapt my ideas in this PR if some devs are interested in vectorize ess/rhat calculation to save computation (when many sites and draws are involved).

PR #665, opened 2019-05-11
improving the legend in plot_energy #643

Ban-zee:
Made the changes suggested by @OriolAbril  in #652. Still working on the plot_energy issue.

### DIFF
--- a/arviz/plots/energyplot.py
+++ b/arviz/plots/energyplot.py
@@ -142,9 +142,11 @@ def plot_energy(
             handles, labels = ax.get_legend_handles_labels()
             m_patch = Patch(facecolor=plot_kwargs["color"], label=labels[0])
             e_patch = Patch(facecolor=fill_kwargs["color"], label=labels[1])
+            patches = [m_patch, e_patch]
+            patches.extend(handles[2:len(handles)])
             ax.legend(
                 labels=labels,
-                handles=[m_patch, e_patch, handles[2], handles[3], handles[4], handles[5]],
+                handles=patches
             )
 
     ax.set_xticks([])
--- a/arviz/plots/energyplot.py
+++ b/arviz/plots/energyplot.py
@@ -1,7 +1,6 @@
 """Plot energy transition distribution in HMC inference."""
 import numpy as np
 import matplotlib.pyplot as plt
-from matplotlib.patches import Patch
 from ..data import convert_to_dataset
 from ..stats import bfmi as e_bfmi
 from .kdeplot import plot_kde
--- a/arviz/plots/energyplot.py
+++ b/arviz/plots/energyplot.py
@@ -115,6 +115,7 @@ def plot_energy(
                 fill_kwargs=fill_kwargs,
                 plot_kwargs=plot_kwargs,
                 ax=ax,
+                legend=False
             )
     elif kind == "hist":
         for alpha, color, label, value in series:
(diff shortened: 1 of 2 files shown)


arviz/plots/kdeplot.py
```diff
@@ -182,6 +182,7 @@ def plot_kde(
         rug_kwargs.setdefault("space", 0.2)
 
         plot_kwargs.setdefault("linewidth", linewidth)
+        plot_kwargs.setdefault("linestyle", "-")
```
OriolAbril: This is not necessary, the solid line is the default in matplotlib, thus if no linestyle is passed, the plot will have a solid line.

arviz/plots/kdeplot.py
```diff
@@ -229,8 +230,19 @@ def plot_kde(
...
-            legend_element = [Patch(edgecolor=default_color, label=label)]
-            ax.legend(handles=legend_element)
+            if fill_kwargs["alpha"] != 0:
+                legend_element = [
+                    Patch(
+                        facecolor=fill_kwargs["color"],
+                        edgecolor=plot_kwargs["color"],
+                        label=label,
+                        alpha=fill_kwargs["alpha"],
+                        linestyle=plot_kwargs["linestyle"],
```
OriolAbril: I would try to merge `fill_kwargs` and `plot_kwargs`, after removing some of the arguments like the color instead of passing them all "by hand". Other arguments like linewidth or hatch may also be relevant.

OriolAbril: Even though, it looks like from all the possible arguments of a Patch only these two I pointed out are relevant for a legend.

arviz/plots/kdeplot.py
Ban-zee: plot_energy throws an exception if this line is not included.

arviz/plots/kdeplot.py
OriolAbril: I would merge both kwargs dicts, something like this should work:

    plot_options = ("ls", "linestyle", "lw", "linewidth")
    fill_options = ("alpha", "hatch", "capstyle", "joinstyle")
    kwargs1 = {key: value for key, value in fill_kwargs.items() if key in fill_options} 
    kwargs2 = {key: value for key, value in plot_kwargs.items() if key in plot_options}
    patch_kwargs = {**kwargs1, **kwargs2}

And then, when calling patch, pass the 2 colors and the label manually, and then `**patch_kwargs`.

arviz/plots/kdeplot.py
OriolAbril: I thinl that the recommendation in the other comment will solve this

Ban-zee:
The code might look grotesque as of now :(. I will refine it in my upcoming commits :).

Ban-zee:
Ready for review!!

aloctavodia:
LGTM! What do you think @OriolAbril?

arviz/plots/energyplot.py
```diff
@@ -113,11 +112,10 @@ def plot_energy(
                 bw=bw,
                 label=label,
                 textsize=xt_labelsize,
-                plot_kwargs=plot_kwargs,
                 fill_kwargs=fill_kwargs,
+                plot_kwargs=plot_kwargs,
                 ax=ax,
```
OriolAbril: There should be a `legend=False`, otherwise, the legend of the kde will always be present.

arviz/plots/kdeplot.py
OriolAbril: Now this line can be safely removed. It is not needed afterwards anymore.

ColCarroll:
Thanks for the contribution @Ban-zee , and for the nice review @OriolAbril !

PR #667, opened 2019-05-12
Quick fix for diagnostics

ahartikainen:
After discussion let's trim ress functionality and explicit split_ess, which is now ess_mean.

Also there was errors in docs.

### DIFF
--- a/arviz/stats/diagnostics.py
+++ b/arviz/stats/diagnostics.py
@@ -195,7 +195,9 @@ def ess(data, *, var_names=None, method="bulk", relative=False, prob=None):
     if isinstance(data, np.ndarray):
         data = np.atleast_2d(data)
         if prob is not None:
-            return ess_func(data, prob=prob, relative=relative)
+            return ess_func(  # pylint: disable=unexpected-keyword-arg
+                data, prob=prob, relative=relative
+            )
         else:
             return ess_func(data, relative=relative)
 
--- a/arviz/stats/diagnostics.py
+++ b/arviz/stats/diagnostics.py
@@ -33,8 +33,8 @@ def bfmi(data):
     Parameters
     ----------
     data : obj
-        Any object that can be converted to an az.InferenceData object
-        Refer to documentation of az.convert_to_dataset for details
+        Any object that can be converted to an az.InferenceData object.
+        Refer to documentation of az.convert_to_dataset for details.
         If InferenceData, energy variable needs to be found.
 
     Returns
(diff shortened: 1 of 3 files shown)


doc/api.rst
```diff
@@ -55,15 +55,10 @@ Diagnostics
...
-    effective_sample_size_bulk
-    effective_sample_size_tail
-    effective_sample_size_quantile
+    ess
+    effective_sample_size
     rhat
-    msce_mean
-    msce_sd
-    msce_quantile
+    msce
```
OriolAbril: There is a typo, should be mcse.

arviz/stats/diagnostics.py
```diff
@@ -657,20 +637,22 @@ def _ess(ary):
     ess = ess / tau_hat
     if np.isnan(rho_hat_t).any():
         ess = np.nan
+    if relative:
+        ess /= ary.size
```
OriolAbril: I was thinking on `ess = 1 if relative else n_chain * n_draw` some lines before instead of `ess = n_chain * n_draw`. What do you think about it?

doc/api.rst
```diff
@@ -73,8 +68,10 @@ Stats utils
 .. autosummary::
     :toctree: generated/
 
+    autocov
     autocorr
     make_ufunc
+    wrap_xarray_ufunc
```
OriolAbril: Nor `make_ufunc` nor `wrap_xarray_ufunc` are in "stats/__init__.py", thus, the docs are not properly generated as they can not be imported.

doc/api.rst
ahartikainen: Good catch

doc/api.rst
ahartikainen: Thanks for catching this. I think they could be added to init, because they are quite flexible functions.

On the other hand, I think we can skip them now and let's re-evaluate this point after the summer.

arviz/stats/diagnostics.py
ahartikainen: Yes, this makes sense

doc/api.rst
OriolAbril: Same here, do whatever you feel like. I only commented it here and not in init because init was not modified.

arviz/stats/diagnostics.py
ahartikainen: I did something close to this. (1/log10(1)) is not good :)

ahartikainen:
I will merge this.

PR #669, opened 2019-05-15
Emcee improvements

OriolAbril:
I have started working on some emcee related improvements. It all started because of a warning about the "draws" and "chains" of the observed data, and little by little I started adding more functionalities:
* observed data is no longer created via a call to `dict_to_dataset`, thus, it won't have draw and chain dims anymore.
* `from_emcee` now has support for blobs, allowing to easily store the log_likelihood, posterior_predictive... I have tested it in both emcee2 and 3 locally, I will add tests and examples when I find some time.
* I added a `.sel()` method to InferenceData objects. It is basically thought for performing a burn-in on emcee objects (emcee has no warm-up nor burn-in by default, so at some point, the first iterations must be removed). It only takes kwargs, loops over all groups and performs dataset.sel(**kwargs) on all groups. I left it quite general so that it can be used for things different than burn-ins, but if it is really specific to emcee, I will search another way.

### DIFF
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -110,12 +110,11 @@ class InferenceData:
         **kwargs : mapping
             It must be accepted by Dataset.sel()
         """
-
         for group in self._groups:
             dataset = getattr(self, group)
             for key, item in kwargs.items():
                 if key in list(dataset.dims):
-                    dataset = dataset.sel(key=item)
+                    dataset = dataset.sel(**{key: item})
                     setattr(self, group, dataset)
 
 
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -98,7 +98,7 @@ class InferenceData:
         """Concatenate two InferenceData objects."""
         return concat(self, other, copy=True, inplace=False)
 
-    def sel(self, **kwargs):
+    def sel(self, inplace=True, **kwargs):
         """Perform an xarray selection on all groups.
 
         Loops over all groups to perform Dataset.sel(key=item)
@@ -110,12 +110,16 @@ class InferenceData:
         **kwargs : mapping
             It must be accepted by Dataset.sel()
         """
+        out = self if inplace else deepcopy(self)
         for group in self._groups:
             dataset = getattr(self, group)
-            for key, item in kwargs.items():
-                if key in list(dataset.dims):
-                    dataset = dataset.sel(**{key: item})
-                    setattr(self, group, dataset)
+            valid_keys = set(kwargs.keys()).intersection(dataset.dims)
+            dataset = dataset.sel(**{key: kwargs[key] for key in valid_keys})
+            setattr(out, group, dataset)
+        if inplace:
+            return None
+        else:
+            return out
 
 
 # pylint: disable=protected-access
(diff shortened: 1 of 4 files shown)


OriolAbril:
I have uploaded an example in an ipython notebook using the blobs functionality to directly get the log_likelihood and posterior_predictive in the InferenceData object directly with `az.from_emcee`. I also did the same exercise in PyMC3, to check everything made sense. They can be found [here](https://github.com/OriolAbril/Bayesian-Inference-examples/tree/dev/Multiple_Linear_Regression)

arviz/data/inference_data.py
```diff
@@ -98,6 +98,25 @@ def __add__(self, other):
         """Concatenate two InferenceData objects."""
         return concat(self, other, copy=True, inplace=False)
 
+    def sel(self, **kwargs):
```
ColCarroll: what happens if the group is not in one of the groups/datasets?

ColCarroll: er, the selection, rather (will having `sampler_statistics` break this?)

OriolAbril: There is a check for every pair _key_-_group_ to ensure that the key is one of the dimensions in the group. Thus, no error is risen when performing the burn in on the whole inference data even though the observed_data group has no draw dimension. Is this what you mean?

ColCarroll: actually, what do you think about

```python
for group in self._groups:
    dataset = getattr(self, group)
    valid_keys = set(kwargs.keys()).intersection(dataset.dims)
    dataset = dataset.sel(**{key: kwargs[key] for key in valid_keys})
    setattr(self, group, dataset)
```

I think this is significantly faster for certain use cases.

OriolAbril: Definitely much better. I wasn't really sure if this would be useful to someone else other than me. I can add copy argument too.

OriolAbril:
It should be ready for review and merge now. 

Eventually, the "changelog" of this PR has become:

* Add `.sel()` method to inference data objects. It loops over groups to apply `Dataset.sel()` in each dataset, but only for the dims present in each group to avoid errors. Tests have been implemented, and docs are present, even though I don't know where to find them in the documentation. Examples are missing.
* Fix issue with observed data in `from_emcee`. It used to contain chain and draw dims, now it doesn't.
* Add support for the blob functionality of emcee, along with tests and examples on how to use it in ArviZ.
* Add extra argument `slices`, to account for the fact that _theta_ may not be an array of scalar variables. Tests and examples present too.

arviz/data/inference_data.py
```diff
@@ -110,12 +110,16 @@ def sel(self, **kwargs):
...
             It must be accepted by Dataset.sel()
         """
+        out = self if inplace else deepcopy(self)
         for group in self._groups:
             dataset = getattr(self, group)
-            for key, item in kwargs.items():
-                if key in list(dataset.dims):
-                    dataset = dataset.sel(**{key: item})
-                    setattr(self, group, dataset)
+            valid_keys = set(kwargs.keys()).intersection(dataset.dims)
```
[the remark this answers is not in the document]
OriolAbril: Is order important? I thought that the order of kwarg arguments did not matter in `sel()`

OriolAbril:
I have just realized that the `sel()` method is actually like `get_coords()` in plot utils, we can mimic the behaviour and raise errors too in `sel()` instead of letting xarray handle them. 

Another option (which may not be incompatible) would be to "generalize" `get_coords` to work on both datasets and inferencedata objects.

PR #671, opened 2019-05-17
implement from_cmdstanpy

ahartikainen:
Implements `from_cmdstanpy`

Also fixes `convert` function to correctly handle different special groups and enables also `from_cmdstan` if input is a `list` or `str` that endswith `.csv`.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -324,7 +324,7 @@ ignore-docstrings=yes
 ignore-imports=no
 
 # Minimum lines number of a similarity.
-min-similarity-lines=15
+min-similarity-lines=50
 
 
 [TYPECHECK]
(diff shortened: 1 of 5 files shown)


ahartikainen:
I think I will add tests later. This currently at least working locally.

(I need to think how we should do the testing, because I don't want to install CmdStan to our test env, might need to.)

canyon289:
> I think I will add tests later. This is currently working locally.
> 
> (I need to think how we should do the testing, because I don't want to install CmdStan to our test env, might need to.)

Could the outputs of cmdstan be stored and we test against that?

ahartikainen:
Not really. (Or maybe). It is still in beta, so things might change in future.

Maybe we could read the csv files and transform them to PosteriorSample.

ahartikainen:
Will merge now, let's figure out testing later (cmdstanpy is still beta, so it might still change)

PR #676, opened 2019-05-22
Applying Numba on plot_kde[WIP]

Ban-zee:
This is the first pull request of my project(applying Numba across ArviZ). Open for review.

### DIFF
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -4,9 +4,9 @@ import numpy as np
 import matplotlib.pyplot as plt
 from scipy.signal import gaussian, convolve, convolve2d  # pylint: disable=no-name-in-module
 from scipy.sparse import coo_matrix
+from numba import njit
 import xarray as xr
 from ..data.inference_data import InferenceData
-from ..utils import conditional_jit
 from .plot_utils import _scale_fig_size
 
 
@@ -262,6 +262,7 @@ def plot_kde(
     return ax
 
 
+@njit(cache=True)
 def _fast_kde(x, cumulative=False, bw=4.5, xmin=None, xmax=None):
     """Fast Fourier transform-based Gaussian kernel density estimate (KDE).
 
@@ -331,12 +332,13 @@ def _fast_kde(x, cumulative=False, bw=4.5, xmin=None, xmax=None):
     return density, xmin, xmax
 
 
-@conditional_jit
+@njit(cache=True)
 def _histogram(x, n_bins, range_hist=None):
     grid, _ = np.histogram(x, bins=n_bins, range=range_hist)
     return grid
 
 
+@njit(cache=True)
 def _fast_kde_2d(x, y, gridsize=(128, 128), circular=False):
     """
     2D fft-based Gaussian kernel density estimate (KDE).
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -4,10 +4,10 @@ import numpy as np
 import matplotlib.pyplot as plt
 from scipy.signal import gaussian, convolve, convolve2d  # pylint: disable=no-name-in-module
 from scipy.sparse import coo_matrix
-from numba import njit
 import xarray as xr
 from ..data.inference_data import InferenceData
 from .plot_utils import _scale_fig_size
+from ..utils import conditional_jit
 
 
 def plot_kde(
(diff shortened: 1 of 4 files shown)


arviz/plots/kdeplot.py
```diff
@@ -4,9 +4,9 @@
 import matplotlib.pyplot as plt
 from scipy.signal import gaussian, convolve, convolve2d  # pylint: disable=no-name-in-module
 from scipy.sparse import coo_matrix
+from numba import njit
```
ahartikainen: do we still keep numba as an optional dependency?

canyon289:
@Ban-zee From my understanding you're still working on this?

Ban-zee:
@canyon289 Yes, waiting for feedback.

Ban-zee:
Not optional if everything works well

arviz/plots/kdeplot.py
aloctavodia: I think we want to keep this optional. not sure how easy is to install numba if not using conda

arviz/plots/kdeplot.py
```diff
@@ -262,6 +262,7 @@ def plot_kde(
     return ax
 
 
+@njit(cache=True)
```
aloctavodia: Is this really faster? Last time I tried this (and if my memory is not failing me) this was slower/or did not give any speed-up. That's why numba was restricted to  `_histogram`. It will be nice if every PR related to numba comes with  the speed-up. Maybe we should also have test that ensure the _numbified_ function is faster than the non-modified one.

arviz/plots/kdeplot.py
ahartikainen: I think it is not too bad, but arch and other smaller distros could have a problem

arviz/plots/kdeplot.py
ahartikainen: That sounds interesting.

Ban-zee: I agree, the improvement isn't much and sometimes the function is a bit slower. This is what I wanted to get checked out at first so as to ensure that the lack of speed-up was not because of some error in the code. Regarding tests, I agree that it is an interesting idea.

Ban-zee:
Removed jit from `fast_kde` and `fast_2d_kde` as the enhancement was not much on these functions. Instead of a test as suggested, I added a function which gives the enhancement provided after applying the numba decorator. Up for review.

arviz/utils.py
```diff
@@ -80,3 +81,19 @@ def wrapper(function):
...
+
+
+def wrapper(function, *args, **kwargs):
+    def wrapped():
+        return function(*args, **kwargs)
+
+    return wrapped
+
+
+def to_numba_or_not_to_numba(function, *args, **kwargs):
```
aloctavodia: I understand the idea of this function if for us to manually check if a adding numba is helping or not. Right? I still think we should also have tests in order to detect regressions (not only from our side, but just in case numba changes in some unexpected way). @ColCarroll what do you think about this?

ahartikainen: How accurate timing is on travis. 

Can we send those timings to some server so we can plot against each time they are tested?

Ban-zee: Yes, this gives us an estimate of the numba enhancement. I am thinking of adding a loop in this method so as to get a proper estimate of the time. If everyone is fine with this particular method of testing numba then I'll add the test.

ColCarroll: We use [airspeed velocity](https://github.com/airspeed-velocity/asv) on pymc3. The benchmarks are here: http://pandas.pydata.org/speed/pymc3/, and we check them manually whenever we want to. 

One of the anaconda devs (and all around good person) helped us set that up - I could check whether we can also run `arviz` benchmarks on that machine.

Ban-zee:
Added the test and improved the enhancement function. On plot_kde, only histogram showed a considerable enhancement after applying conditional_jit. Jit will have to be disabled for the test to pass or else it would throw an error. If everyone is fine with these changes, then I'll start to apply numba on stats tomorrow as mentioned in my timeline.

canyon289:
I vote optional on Numba, it's hard to install on some systems

Ban-zee:
The python tests pass every time on system. My guess is that jit is disabled which leads to this failure. I might be completely wrong as well. :(

aloctavodia:
@canyon289 maybe you can help here with the tests. I am not sure what to do.

Ban-zee:
NUMBA_DISABLE_JIT is set to 1 in travis.yml, changing it to 0 should resolve the issue.

canyon289:
> @canyon289 maybe you can help here with the tests. I am not sure what to do.

Just saw this sorry. We want to test both circumstances with JIT and without JIT so we'll need another test pipeline.

Ban-zee:
Closing this. Opened a new PR for the same.

Issue #501, opened 2019-01-05
Diagnostics + Stats to use xarray


ahartikainen:
Use InferenceData as a default data-structure for diagnostics and stats. Use dim names to do calculations correctly. (`chain`,`draw`).

ahartikainen:
I think most of the functions could eat InferenceData and then use some default (bfmi, etc). We do this already in `loo`.

What we could change, is that user can define those default values. --> InferenceData, value_name(, group)

OriolAbril:
I would like to start working on that @ahartikainen . But I have some questions.

For the `loo` function it is quite straightforward to use xarray in some more places as the input data is an InferenceData object, whereas many of the others are functions for numpy ndarrays which are applied to xarray objects via `apply_ufunc`. How should that scheme change? (in case it should change)

I will start working on the loo function and submit a PR to show it is a work in progress. For example:

    log_likelihood = inference_data.sample_stats.log_likelihood
    n_samples = log_likelihood.chain.size * log_likelihood.draw.size
    new_shape = (n_samples,) + log_likelihood.shape[2:]
    log_likelihood = log_likelihood.values.reshape(*new_shape)

can be modified to:

    log_likelihood = inference_data.sample_stats.log_likelihood.stack(n_samples=('chain','draw')) 

Which poses the question of whether or not to maintain the xarray datatype and modify the `psislw` to use xarray datatypes instead of numpy arrays. I would modify both functions, but I think this affects the library quite deeply, and probably some conventions will need to be chosen.

ahartikainen:
I have done most of the rewrite already for stats+diagnostics, so I think maybe we can come back to this after the PR is ready.

OriolAbril:
I checked this a little further, and even though the code may look cleaner, the xarray version runs around 10 times slower in my computer.So I don't think it is worth it. I leave the xarray version (the previous post is already outdated) here just in case:

    log_likelihood = inference_data.sample_stats.log_likelihood
    log_likelihood = log_likelihood.stack(n_samples=('chain','draw'), data_points=log_likelihood.dims[2:])
    n_samples, n_data_points = log_likelihood.shape

Also, fun fact, this will only work on python>=3.6, because from 3.6 on, kwargs maintain their order.

ahartikainen:
Interesting insight.

[...]

PR #678, opened 2019-05-23
pointwise elpd diagnostics (text formatting and plot)

### LINKED ISSUE #496: Improve loo functionality
`loo` functionality should be more verbose and return more informative data. (see Model Selection lecture in StanCon Helsinki by @avehtari)

Also docstrings needs extensive examples for loo-usage.

### LINKED ISSUE #660: plot pointwise WAIC/LOO
ArviZ is missing function to compare models in terms of pointwise ELPD. See [figure 10](https://arxiv.org/abs/1709.01449). 

Such a function could work for 2 models (like figure 10) or for more than 2 and return a grid of non-redundant subplots, like `plot_pair`.

OriolAbril:
PR to improve the output of waic and loo to make it more verbose and to create the pointwise elpd comparison plot. Will fix #496 and will fix #660.

This first version of the plot already works, but there is still much work needed.

### Plots
* [x] coloring?
image

Data from the toy model in [this notebook](https://github.com/OriolAbril/Bayesian-Inference-examples/blob/master/Binomial_on_2D_data/binomial_on_2D_data_intention_pystan.ipynb). This was generated with `color="river_distance"` and `legend=True`.

* [x] tick labels?
image

* [x] highlight worst points?
image
Added a `threshold` argument in order to show labels of point further away than threshold times elpd_i.std(). In the example, `threshold=1` to force showing the labels.

* [x] move common coloring/labeling data functions to `plot_utils.py` to have the same functionality in `plot_elpd` and `plot_khat`

### Stats functions
* [x] improve verbosity
image
image

* [x] Use xarray for all computations.
See #501 for some more detail.

### Both
* [x] check behaviour on more than 3D items (not only chain, draw and one extra dim).

Everything working after some corrections, still debating between reshaping pointwise loo/waic to original shape or work with multi-index objects from there on (eventually the scatter plot is flattened, thus at some point they are needed)

* [x] Optimize API: find a way to call waic and loo only once per InferenceData object ideally. 

Idea: Store pointwise loo, waic and pareto_k as dataarrays in the `ELPDData` object instead of as a flattened array and call `plot_elpd` with a dict of `ELPDData` and `plot_khat` with an `ELPDData` object instead of the array of pareto shape values. Using the dataarray will allow coloring, ticklabels, selection of a subset of observations and so on, and thanks to the overwritten `__str__` method, including this extra information it the `ELPDData` object won't clutter the relevant info when printed.

### DIFF
--- a/arviz/plots/__init__.py
+++ b/arviz/plots/__init__.py
@@ -6,7 +6,7 @@ from .energyplot import plot_energy
 from .forestplot import plot_forest
 from .kdeplot import plot_kde, _fast_kde, _fast_kde_2d
 from .parallelplot import plot_parallel
-from .pointwiseelpdplot import plot_pointwise_elpd
+from .elpdplot import plot_elpd
 from .posteriorplot import plot_posterior
 from .traceplot import plot_trace
 from .pairplot import plot_pair
@@ -29,7 +29,7 @@ __all__ = [
     "_fast_kde",
     "_fast_kde_2d",
     "plot_parallel",
-    "plot_pointwise_elpd",
+    "plot_elpd",
     "plot_posterior",
     "plot_trace",
     "plot_pair",
--- a/arviz/plots/__init__.py
+++ b/arviz/plots/__init__.py
@@ -6,6 +6,7 @@ from .energyplot import plot_energy
 from .forestplot import plot_forest
 from .kdeplot import plot_kde, _fast_kde, _fast_kde_2d
 from .parallelplot import plot_parallel
+from .pointwiseelpdplot import plot_pointwise_elpd
 from .posteriorplot import plot_posterior
 from .traceplot import plot_trace
 from .pairplot import plot_pair
@@ -28,6 +29,7 @@ __all__ = [
     "_fast_kde",
     "_fast_kde_2d",
     "plot_parallel",
+    "plot_pointwise_elpd",
     "plot_posterior",
     "plot_trace",
     "plot_pair",
--- a/arviz/plots/pointwiseelpdplot.py
+++ b/arviz/plots/pointwiseelpdplot.py
@@ -1,0 +1,147 @@
+"""Plot pointwise elpd estimations of inference data."""
+import numpy as np
+import xarray as xr
+import matplotlib.pyplot as plt
+from matplotlib.ticker import NullFormatter
+
+from ..data import convert_to_inference_data
+from .plot_utils import _scale_fig_size
+from ..stats import waic, loo
+
+
+def plot_pointwise_elpd(
+    idata_dict,
+    ic="waic",
+    color=None,
+    figsize=None,
+    textsize=None,
+    coords=None,
+    ax=None,
+    scale="deviance",
+    plot_kwargs=None,
+):
+    """
+    Plot a scatter or hexbin matrix of the sampled parameters.
+
+    Parameters
+    ----------
+    idata_dict : mapping, str -> InferenceData
+        A dictionary mapping the model name to the object containing its inference data.
+        Refer to az.convert_to_inference_data for details on possible dict items
+    ic : str, optional
+        Information Criterion (WAIC or LOO) used to compare models. Default WAIC.
+    color : str or array_like, optional
+        Colors of the scatter plot, if color is a str all dots will have the same color,
+        if it is the size of the observations, each dot will have the specified color,
+        otherwise, it will be interpreted as a list of the dims to be used for the color code
+    figsize : figure size tuple, optional
+        If None, size is (8 + numvars, 8 + numvars)
+    textsize: int, optional
+        Text size for labels. If None it will be autoscaled based on figsize.
+    coords : mapping, optional
+        Coordinates of points to plot. **All** values are used for computation, but only a
+        a subset can be plotted for convenience.
+    ax: axes, optional
+        Matplotlib axes
+    scale : str, optional
+        scale argument passed to az.waic or az.loo, see their docs for details
+    plot_kwargs : dicts, optional
+        Additional keywords passed to ax.plot
+    Returns
+    -------
+    ax : matplotlib axes
+
+    Examples
+    --------
+    """
+    valid_ics = ["waic", "loo"]
+    ic = ic.lower()
+    if ic not in valid_ics:
+        raise ValueError(
+            ("Information Criteria type {} not recognized."
+             "IC must be in {}").format(ic, valid_ics)
+        )
+
+    if coords is None:
+        coords = {}
+
+    if plot_kwargs is None:
+        plot_kwargs = {}
+        plot_kwargs.setdefault("marker", ".")
+        plot_kwargs.setdefault("lw", 0)
+
+
+    def get_pointwise_as_dataarray(idata, coords=coords, ic=ic, scale=scale):
+        if ic == "waic":
+            pointwise_elpd = waic(idata, pointwise=True, scale=scale).waic_i
+        else:
+            pointwise_elpd = loo(idata, pointwise=True, scale=scale).loo_i
+        like_dataarray = idata.sample_stats.log_likelihood
+        dims = [dim for dim in like_dataarray.dims if dim not in ["chain", "draw"]]
+        coords = {dim: like_dataarray.coords.indexes[dim] for dim in dims}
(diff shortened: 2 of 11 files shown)


OriolAbril:
Example of the current plot:

```python
import arviz as az
idata1 = az.load_arviz_data("centered_eight")
idata2 = az.load_arviz_data("non_centered_eight")

az.plot_pointwise_elpd({"centered model": idata1, "non centered model": idata2})
```

image

    az.plot_pointwise_elpd(
        {"centered model": idata1, "non centered model": idata2}, 
        coords={"school": ["Choate", "Deerfield", "Mt. Hermon"]}
    )

image

doc/api.rst
```diff
@@ -24,6 +24,7 @@ Plots
     plot_khat
     plot_pair
     plot_parallel
+    plot_pointwise_elpd
```
ahartikainen: Are we going to have elpd plot, or should we make this a bit simpler?

    plot_elpd

arviz/plots/__init__.py
```diff
@@ -6,6 +6,7 @@
 from .forestplot import plot_forest
 from .kdeplot import plot_kde, _fast_kde, _fast_kde_2d
 from .parallelplot import plot_parallel
+from .pointwiseelpdplot import plot_pointwise_elpd
```
ahartikainen: Could this be just `elpdplot`

arviz/plots/pointwiseelpdplot.py
```diff
@@ -0,0 +1,242 @@
...
+    """
+    Plot a scatter or hexbin matrix of the sampled parameters.
+
+    Parameters
+    ----------
+    idata_dict : mapping, str -> InferenceData
+        A dictionary mapping the model name to the object containing its inference data.
+        Refer to az.convert_to_inference_data for details on possible dict items
+    ic : str, optional
+        Information Criterion (WAIC or LOO) used to compare models. Default WAIC.
```
ahartikainen: Is there a reason why waic is the default?

arviz/stats/stats.py
```diff
@@ -12,7 +12,7 @@
 
 from ..data import convert_to_inference_data, convert_to_dataset
 from .diagnostics import _multichain_statistics, _mc_error, ess
-from .stats_utils import make_ufunc as _make_ufunc, logsumexp as _logsumexp
+from .stats_utils import make_ufunc as _make_ufunc, logsumexp as _logsumexp, ELPDData
```
ahartikainen: ~let's keep everything lowercase~ is this a class?

arviz/plots/__init__.py
OriolAbril: I will change it tomorrow

arviz/plots/pointwiseelpdplot.py
OriolAbril: I followed the default in `az.compare`, maybe it can be included as a package option too?

arviz/stats/stats.py
OriolAbril: Yes, it is a children class of `pd.Series` with the `__str__` and `__repr__` methods overwritten. I could create a custom class from zero too if there are no strong reasons in favor of using a Series.

arviz/plots/pointwiseelpdplot.py
ahartikainen: Sounds ok, let's follow `az.compare`

arviz/stats/stats.py
ahartikainen: Sounds fine with the `Series` backend.

arviz/stats/stats.py
```diff
@@ -403,12 +408,10 @@ def loo(data, pointwise=False, reff=None, scale="deviance"):
...
-    n_chains, n_draws, *shape = log_likelihood.shape
-    n_samples = n_chains * n_draws
-    n_data_points = np.product(shape)
-    dims = [dim for dim in log_likelihood.dims if dim not in ["chain", "draw"]]
-    present_coords = {dim: log_likelihood.coords.indexes[dim] for dim in dims}
-    log_likelihood = log_likelihood.values.reshape(n_samples, n_data_points)
+    log_likelihood = log_likelihood.stack(samples=("chain", "draw"))
+    shape = log_likelihood.shape
+    n_samples = shape[-1]
+    n_data_points = np.product(shape[:-1])
```
ahartikainen: What if loglikelihood is nD?

OriolAbril: It still works (i checked locally with the same model as the first plot and the test for this passed too). I decided to not stack the observations dimensions in order to keep this information also in the pointwise loo and waic.

OriolAbril: I think that the only braking change is the shape of psislw input. Which now, in the 2D (sample, and obs) is transposed from what it used to be. In the case of dataarray inputs, this should not affect (as long as the dimension is calles "samples") but for array inputs it does. It could be changed to the original shape and then call `np.rollaxis` to get the `n_samples` axis to the last position (as expected in xarray apply ufunc).

OriolAbril:
Ready for review! 

I can't decide on the yaxis ticks in the case of more than 2 models. Right now the axis are not shared, thus, the ticks on the left do not correspond to the values in the other plots. I came across with two options, but none of them really convince me. The first is to share the axis (or force the ylims to be common throughout each row) and the second is to put yticklabels in all plots.

OriolAbril:
# PR summary

This PR has adressed elpd statistics, loo and waic, and created a plot to compare various models according to pointwise elpd statistics. It adds a completely new plot to ArviZ, and modifies various stats functions. The main goal of all these changes in the stats functions is to use xarray for the computation. This allows to work easily with nd objects while maintaining the coords information. Eventually, plots like `plot_elpd` of `plot_khat` (on which I will work once this PR is merged) can use this coords information to automatically get relevant labels, use a color code generated from the coord values or plot only a subgroup of the observations.

It also adds the class `ELPDData` to ArviZ, but as it is a subclass of `pd.Series`, its only effect is when printing the result of loo or waic. Everything else stays the same.

### Modified functions
#### `loo` and `waic`
Their arguments stay the same. The computation method has been modified to use `wrap_xarray_ufunc`, so that the result of the pointwise calculations is a dataarray (instead of an array) and it maintains the original shape (instead of being flattened). Therefore, the output of `loo().loo_i` is now a dataarray instead of an array. In addition, the output class is now ELPDData class, so that it is more verbose, organized and relevant when printed (i.e. in the case of waic, waic_i is never printed, but it is stored if pointwise is true).

#### `psislw`
it has also been modified to work wrapping ufuncs on xarray or array objects. The output should be the same type and shape as the input. 

However, for convenience I have modified the dimension order of the input data. It currently supports only 2d inputs of shape (n_samples, n_observations), this PR proposes it to work on nd arguments of shape (..., n_samples) (which in the 2D case is the transposed). 
In the case of dataarray objects it does not have much relevance, because the "samples" dimension is automatically moved to the last position by apply_ufunc, but in the case of arrays it does. 

### Created
#### `ELPDData` 
Class to contain the result of loo and waic. It is a children of pd.Series, with the only difference of having the methods `__str__` and `__repr__` overwritten. 

#### `plot_elpd`
Plot to compare different models according to waic_i or loo_i differences. It accepts an dictionay of InferenceData objects (for which the ic is computed before the plot) or a dictionary of ELPDData objects (which avoids recalculating the ic every time plot_elpd is called). For more examples on various parameters, see the first comment.

OriolAbril:
On the testing side, tests have already been included, covering nearly 100% of the lines and taking into account multidimensional objects.

arviz/stats/stats.py
```diff
@@ -430,29 +447,60 @@ def loo(data, pointwise=False, reff=None, scale="deviance"):
...
+
+    lppd = np.sum(
+        _wrap_xarray_ufunc(
+            _logsumexp,
+            log_likelihood,
+            func_kwargs={"b_inv": n_samples},
+            ufunc_kwargs=ufunc_kwargs,
+            **kwargs
+        ).values
+    )
```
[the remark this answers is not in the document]
OriolAbril: None. The computations performed are the same, I have not modified how logsumexp works. I did this so that there is no need to convert to arrays and the named dims and coords info is not lost.

ahartikainen:
LGTM

I checked the code, and everything looks good.

aloctavodia:
Thanks @OriolAbril this looks great. Looking forward to try these changes!

Issue #609, opened 2019-03-09
Restrict the max number of plots by default


aloctavodia:
Following discussion in #94, naively or inadvertently calling plotting functions for models with a large number of variables can easily hang a computer. Thus we may want to restrict by default the max number of plots. 

If possible it would be nice to set this at a global level and not per plot function

ahartikainen:
So env variable

    ARVIZ_MAX_AXES

canyon289:
ENV works. I want to throw in pandas style options as a proposal design as well
https://pandas.pydata.org/pandas-docs/stable/user_guide/options.html

aloctavodia:
Notice that #679 addressed this for the particular case of traceplots

ahartikainen:
Is this fixed?

OriolAbril:
Not really, it is only fixed for `plot_trace` and should be done with all plots and with global options, not function specific.

OriolAbril:
For future reference, this issue was actually fixed in #787.

Issue #94, opened 2018-05-23
Traceplot parity


ColCarroll:
Is it intentional to have a different traceplot from pymc3?  Each element of an RV with shape > 1 gets a plot in `arviz`, while they are all the same for `pymc3`.  

Explicitly, with eight schools (I only capture the first few plots from arviz):

```
J = 8
y = np.array([28,  8, -3,  7, -1,  1, 18, 12])
sigma = np.array([15, 10, 16, 11,  9, 11, 10, 18])

with pm.Model() as hierarchical:
    eta = pm.Normal('eta', 0, 1, shape=J)
    mu = pm.Normal('mu', 0, sd=1e6)
    tau = pm.HalfCauchy('tau', 5)
    theta = pm.Deterministic('theta', mu + tau*eta)
    obs = pm.Normal('obs', theta, sd=sigma, observed=y)
    trace_h = pm.sample(1000)
```

![image](https://user-images.githubusercontent.com/2295568/40398462-3c442368-5e06-11e8-8020-337346b185e9.png)

aloctavodia:
Yes, this is intentional for traceplot and other plots.

ColCarroll:
Thanks!

junpenglao:
Is there a plan to add an option to have the old pymc3 style plot? 

Otherwise it might be useful to restrict the number of plot traceplot (or any other plots, for that matter) products. I am re-running some old codes and if there is a lot of dimension it will break arviz (since it is flattening the dimensions)

aloctavodia:
Do you mean to limit, by default, the output to a given number of plots? 

You can use the variable names and coordinates to, explicitly, reduce the number of plots, do you think this is not enough?

ColCarroll:
I could imagine doing this at the pymc3 level (you can do fancy things with the axes to "overplot"), so that `pm.traceplot` looks the same, or adding an option in `arviz` like (somewhat tongue-in-cheek) `style={'modern'|'classic'}`, or (more reasonable) `style={'facet'|'compact'}`.

ahartikainen:
This is just my opinion/question: How often all the components in one parameter describe the same thing?

What could be the future, is to create a book for all the plots or create interactive visualization where user can select the number of "rows" and "cols" for number of plots and also if parameters are combined or not (even on parameter level; e.g. select only theta).

But book would be cool.

junpenglao:
> Do you mean to limit, by default, the output to a given number of plots?

[...]

PR #679, opened 2019-05-24
Traceplot fixes

### LINKED ISSUE #654: Is there some way to plot individual chains (from PyMC3 data)?
## Short Description

I have a data set with four chains where one of them is going badly wrong:

image

It would be nice if I could plot the individual chains separately (i.e., in subplots), and if the plots would display the index of the chain in the plot legend (i.e., add a key to the plot which would show the mapping between chain index and color).

Thanks!

## Library versions:
Arviz 0.3.3 and PyMC3 3.6

OriolAbril:
For now still minor fixes. Still not gone into facet/compact styles. Should fix #654.

**EDIT**
Already gone into facet/compact styles. Reviews and advise welcome. The idea is to end up fixing both #94 and #609.

### DIFF
--- a/arviz/plots/traceplot.py
+++ b/arviz/plots/traceplot.py
@@ -1,4 +1,5 @@
 """Plot kde or histograms and values from MCMC samples."""
+from itertools import cycle
 import matplotlib.pyplot as plt
 from matplotlib.lines import Line2D
 import numpy as np
@@ -122,7 +123,12 @@ def plot_trace(
         lines = ()
 
     num_colors = len(data.chain) + 1 if combined else len(data.chain)
-    colors = [prop["color"] for _, prop in zip(range(num_colors), plt.rcParams["axes.prop_cycle"])]
+    colors = [
+        prop
+        for _, prop in zip(
+            range(num_colors), cycle(plt.rcParams["axes.prop_cycle"].by_key()["color"])
+        )
+    ]
 
     if compact:
         skip_dims = set(data.dims) - {"chain", "draw"}
@@ -253,15 +259,9 @@ def plot_trace(
                 line_values = [vlines]
             else:
                 line_values = np.atleast_1d(vlines).ravel()
-            axes[idx, 0].vlines(
-                line_values, *ylims[0], colors=colors[idx][0], linewidth=1.5, alpha=0.75
-            )
+            axes[idx, 0].vlines(line_values, *ylims[0], colors="black", linewidth=1.5, alpha=0.75)
             axes[idx, 1].hlines(
-                line_values,
-                *xlims[1],
-                colors=colors[idx][0],
-                linewidth=1.5,
-                alpha=trace_kwargs["alpha"]
+                line_values, *xlims[1], colors="black", linewidth=1.5, alpha=trace_kwargs["alpha"]
             )
         axes[idx, 0].set_ylim(bottom=0, top=ylims[0][1])
         axes[idx, 1].set_xlim(left=data.draw.min(), right=data.draw.max())
--- a/arviz/plots/traceplot.py
+++ b/arviz/plots/traceplot.py
@@ -245,7 +245,7 @@ def plot_trace(
                 alpha=trace_kwargs["alpha"]
             )
         axes[idx, 0].set_ylim(bottom=0, top=ylims[0][1])
-        axes[idx, 1].set_xlim(left=0, right=data.draw.max())
+        axes[idx, 1].set_xlim(left=data.draw.min(), right=data.draw.max())
         axes[idx, 1].set_ylim(*ylims[1])
     return axes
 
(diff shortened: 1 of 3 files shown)


arviz/plots/traceplot.py
```diff
@@ -97,11 +100,13 @@ def plot_trace(
     """
     if divergences:
         try:
-            divergence_data = convert_to_dataset(data, group="sample_stats").diverging
+            divergence_data = get_coords(
+                convert_to_dataset(data, group="sample_stats"), coords
+            ).diverging
```
OriolAbril: This is not good and there are many possible solutions, however, I feel like the cleanest one would be:

```python
""" #docstring end
data = convert_to_inference_data(data)
data = data.sel(inplace=False, **coords)
```
So that afterwards, the posterior and the divergence data are already retrieved with the correct coords.

OriolAbril: But it requires too much memory though, as the whole inference data object is copied. Open to suggestions.

OriolAbril:
Now `plot_trace` should be compatible with all kind of coords, even when plotting divergences, and in addition, the selected draws will be reflected in the xaxis. I modified the combined behaviour to combine the samples only on the distplot, but this can be modified. Below there are 2 examples of the centered schools data trace followed by the previous behaviour:

    az.plot_trace(idata, var_names=["tau", "mu"]);

Current behaviour:

new behaviour

which is the same as before this changes:

old behaviour

    az.plot_trace(idata, var_names=["tau", "mu"], combined=True, coords={"draw":slice(100,None,3), "chain":slice(1,None)}, legend=True);

Current behaviour:

new_plot_trace

in this case, the previus version does not allow all this options, thus, after modifying some of them to avoid an error:

    az.plot_trace(idata, var_names=["tau", "mu"], combined=True, coords={"draw":slice(100,None), "chain":slice(1,None)}, divergences=False);

old_plot_trace

Note that even though only the last 400 draws are used, the axis go from 0 to 500, and there are always values, because as the array is flattened, they reach up to 1200. What is seen above is the first chain and part of the second:

old_plot_trace2-axis

arviz/plots/traceplot.py
```diff
@@ -16,7 +16,9 @@ def plot_trace(
     figsize=None,
     textsize=None,
     lines=None,
+    compact=False,
```
aloctavodia: I agree this should be `False` by default, we should also have a max number of plots by default.

ahartikainen:
Do we need top, left, right, (bottom) spines?

If we drop spines and create pseudo 0-spine with `ax.axhline` we could move divergent values under the "spine". Would probably need some axis limit magic, or duplicate axis so div lines are then always the same

OriolAbril:
> Do we need top, left, right, (bottom) spines?
>
> If we drop spines and create pseudo 0-spine with `ax.axhline` we could move divergent values under the "spine". Would probably need some axis limit magic, or duplicate axis so div lines are then always the same

I can see that all the divergences in the right column are the same, but I can't follow you with the solution.

arviz/plots/traceplot.py
```diff
@@ -25,6 +25,7 @@ def plot_trace(
     rug_kwargs=None,
     hist_kwargs=None,
     trace_kwargs=None,
+    max_plots=40,
```
OriolAbril: I have set a preliminary maximum here, which will be removed when we decide on the implementation of global variables #609

aloctavodia: We should have a warning, telling users that want to plot more than `max_plot` plots that only `max_plots` are actually plotted.

aloctavodia: Maybe `max_plot` should also accepts `None`, in case users does not want to put a limit

OriolAbril: I agree on the warning, I will add it as soon as possible. About accepting `None`, I think it is not necessary, if no limit is wanted, it is easy to set it to 10000 and let matplotlib hanging be the limit.

OriolAbril: I think it may be better to merge this once it the warging is added, and afterwards do another PR to implement package parameters.

aloctavodia: Right about setting something to a large number but is more elegant to have a special keyword for that case. Anyway, I agree we should add the warning and then merge.

OriolAbril: I added both, `None` option and warning

arviz/plots/traceplot.py
```diff
@@ -240,3 +291,45 @@ def _histplot_op(ax, data, **kwargs):
     bins = get_bins(data)
     ax.hist(data, bins=bins, align="left", density=True, **kwargs)
     return ax
```
OriolAbril: This function should probably be deleted. It looks like kde/hist is handled by plot_dist

aloctavodia: good catch!

ColCarroll:
This is great, @OriolAbril, thank you!

PR #686, opened 2019-05-30
Fix warning from io_pymc3

ColCarroll:
This fixes the following warning whenever `from_pymc3` is called: 

```
/arviz/arviz/data/io_pymc3.py:56: FutureWarning: arrays to stack must be passed as a "sequence" type such as list or tuple. Support for non-sequence iterables such as generators is deprecated as of NumPy 1.16 and will raise an error in the future.
  chain_likelihoods.append(np.stack(log_like))
```


canyon289:
LGTM

PR #687, opened 2019-05-31
Numba notebook

Ban-zee:
Code Profiling and bottleneck identification. **NOT MEANT TO BE MERGED**


Ban-zee:
Notebook is the main one here. I made some random changes in the other files while experimenting.They can be ignored.

arviz/tests/test_utils.py
```diff
@@ -119,3 +119,16 @@ def placeholder_func():
...
     assert function_results == "output"
+
+
+def test_numba_check():
+    test_data = np.random.rand(200).reshape(10, 20)
+
+    def placeholder(x):
+        """A simple function to test numba enhancement"""
+        dummy = 0
+        for i in range(0, x.shape[0]):
```
canyon289: In ArviZ we would normally vectorize a function like this with numpy. Do you think a plain python for loop is representative of our typical use case?

Ban-zee: placeholder is just a dummy method to check the numba enhancement. The kind of code in the function is one which numba always speeds up. This is just to check that that my (numba_check) function works properly. It is not a representative of any arviz function.

Ban-zee:
Added loo and compare(1/3 complete). Will complete compare and add summary by tomorrow.

Ban-zee:
Numba-Notebook P.R details:


The pull request includes a jupyter notebook whose primary aim is to show the performance of different functions after applying numba. This pull request is not meant to be merged.

Link to the pull request: https://github.com/arviz-devs/arviz/pull/687

Methodology
* Profile the given function on centered_school and a custom random dataset.
* Identify bottlenecks.
* Try to implement a given numpy or scipy function from scratch in python and jit it.
* Compare their respective times.
*  Put the modified methods in the primary method and then compare it's speed to the one included in ArviZ.


Summary
Method 1: Hpd
Bottleneck: scipy circmean
Performance of hpd_new with jitted circmean:
Similar performance on both datasets, sometimes slower on the custom dataset.


Method 2: r2_score
Bottleneck: np.var
Performance of r2_score_new with jitted _var_1d and _var_2d_methods:
Reasonable speedup, almost 50% faster.


Method 3: waic
replaced np.var by _var_1d and _var_2d.
Reasonable speedup, a gain of 100ms on the custom dataset


Method 4: _gpdfit
Bottlenecks: np.log1p and np.exp
Experimented with jitted log_1p and exp functions. Their performance was similar to or worse than their numpy counterparts.  The _gpdfit_new(replace np.log_1p with jitted log_1p) method did not give any definite speedup.


Method 5: _gpinv
Bottlenecks: np.log1p and np.expm1
Experiment with jitted expm1. The performance was similar ~ worse than numpy.expm1. The _gpinv _new(replaced np.log1p by jitted log_1p) also did not give any definite speedup.


Method 6: psislw
Bottlenecks: logsumexp,_gpinv,_gpdfit
Replaced  _gpinv and _gpdfit with _gpinv_new and _gpdfit_new. Not much improvement compared to ArviZ psislw. Some scope of improvement still present as logsumexp, which is a part of diagnostics could be further enhanced.


Method 7: loo
Bottlenecks: psislw, logsumexp
Replaced psislw with psislw_new.  An average of 0.1 s gained on the large custom dataset. Almost similar performance on schools. Some scope of improvement still present as logsumexp, which is a part of diagnostics could be further enhanced.


Method 8: Compare:
Replaced loo and waic in compare with loo_new and waic_new.
waic_new gave a gain of 2-3 microseconds; loo_new gave similar performance as original loo.
Jitting the different functions under stacking, pseudo-bma and bb-pseudo bma slowed down the code substantially.


Method 9: _ic_matrix
Used numba to lift the loop in _ic_matrix. This change slowed down the overall process by approx 100ms.


Method 10: Summary
Majority of summary involves stats_utils. Therefore I'll see it's performance improvement after I start working on diagnostics and stats_utils.


Possible causes of slowdown:
The main reason behind the slowdown is the overall efficiency of numpy and in some instances, the failure of numba to compile things in nopython mode.


My take:
Add _var_1d and _var_2d in the main code to gain substantial speedup on r2_score, waic and compare.

Ban-zee:
I have finished the stats notebook and updated the pr. It is up for review. If anyone notices some new way of speedup that I may have missed, I'll be happy to test it out. If no one has any issues then I'll add the modified code into the main code and write the required tests.

canyon289:
> I have finished the stats notebook and updated the pr. It is up for review. If anyone notices some new way of speedup that I may have missed, I'll be happy to test it out. If no one has any issues then I'll add the modified code into the main code and write the required tests.

@Ban-zee This sounds good. I believe there is a couple of next steps, some which I'll mark optional, all which can be debated

1. (Optional) Write a blog post about what you found, how you tested it, and how you evaluated the results. This will help you with technical communication and also give you something to talk about later during job interviews etc. Plus I think 4 blog posts is required for GSOC so might as well get this out of the way
2. Propose how we should benchmark our tests to verify Numba speedups. Colin posted about airspeed velocity, and I found this plugin for pytest in some cursory googling https://github.com/ionelmc/pytest-benchmark/. See how we can benchmark in CI and tell us what you think is the best method. This also could be a blog post
3. Implement the Numba speeds up in the code, as well as the benchmark tests. If all go well then we'll be able to see the code and be able to reproduce the results!

canyon289:
I must add, I think the tests valuable for the following reason as well. Some users may ask if Numba speedup is worthwhile on their machine. If there are benchmark tests we can instruct users to run the tests with and without Numba to get an objective answer to their question

Ban-zee:
I'll start benchmarking once I am done with diagnostics(that should be in a couple of days).

ahartikainen:
```
@numba.njit
def _var_1d(data):
    a,b = 0,0
    for i in data:
        a = a+i
        b = b+i*i
    return b/len(data)-((a/len(data))**2)

@numba.njit
def _var_2d(data):
    a,b = data.shape
    var = np.zeros(a)
    for i in range(0,a):
        var[i] = _var_1d(data[i,:])
    return var
```

`_var_1d` is not numerically stable.

Could we use Welfords algo https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm

```
# add nopython here?
@numba.njit
def _var_1d(data, ddof=1):
    if len(data) < 2:
        return np.nan
    m2 = 0
    mean = 0
    for count, datum in enumerate(data, 1):
        delta = datum - mean
        mean += delta / count
        m2 += (datum - mean) * delta
    var = m2 / (count - ddof)
    return var
```

and  2d should have a possibility for axis argument

```
@numba.njit
def _var_2d(data, ddof=1, axis=0):
    a,b = data.shape
    # better handling for this (renormalize negative back to axis_len)
    # also None --> ravel
    if axis == 0:
        var = np.zeros(a)
        for i in range(a):
            var[i] = _var_1d(data[i], ddof=ddof)
    else:
        var = np.zeros(b)
        for i in range(b):
            var[i] = _var_1d(data[:, i], ddof=ddof)
    return var
```

ahartikainen:
Ok, I tested and it is slower. I guess that algo is fastest that can be done

Ban-zee:
@ahartikainen should I go on with the initial _var_1d?

ahartikainen:
Maybe yes, can you check are they basically using the same algorithm in numpy?

Also, if numba is not installed, numpy version should be used, right?

Ban-zee:
Here is numpy algorithm
`var = mean(abs(x - x.mean())**2)`

Yes,i f numba's not installed then the original numpy methods will be used.

canyon289:
@Ban-zee What do you want to do with this one?

Ban-zee:
I'll keep this open till I finish up with data and plots.

Ban-zee:
Done with data, pr will be up soon. I have delayed the numba-ahead of time compilation pr to after the data pr as I felt that it might be appropriate if I pick it up after numba has been applied across the codebase.

Ban-zee:
Closing this pr

PR #688, opened 2019-06-01
Azure pipelines

canyon289:
Testing out Azure Pipelines

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,19 @@
+# Starter pipeline
+# Start with a minimal pipeline that you can customize to build and deploy your code.
+# Add steps that build, run tests, deploy, and more:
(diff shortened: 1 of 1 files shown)


canyon289:
Is anyone else able to see the CI details of the azure pipeline? I want to be sure they're accessible to everyone

kyleabeauchamp:
yes

canyon289:
Ready for review. I'm thinking we side by side this with TravisCI for a while to see how it all works out.

After that if we're ready we can take down Travis Unit Test and Lint. I haven't figured out ghp-import for the Sphinx Docs yet, but talking to the Pandas team they still use TravisCI for their docs so I guess I can be lazy and leave it be for a while. Happy to hear thoughts

ahartikainen:
LGTM

PR #696, opened 2019-06-07
LOO-PIT

### LINKED ISSUE #80: [Feature request]Graphical check of leave-one-out cross-validated probability integral transform (LOO-PIT)
Example see figure 9 in https://arxiv.org/pdf/1709.01449.pdf

Reference usage in bayesplot: http://mc-stan.org/bayesplot/reference/PPC-loo.html

Related code in bayesplot: https://github.com/stan-dev/bayesplot/blob/master/R/ppc-loo.R

Computation of loo_pit: 
https://github.com/stan-dev/rstantools/blob/fd84944454f8f0fe1ad9d456211f73a24feaa0ef/R/loo-functions.R#L64-L74

OriolAbril:
Add Leave-One-Out (LOO) predictive checks using probability integral transformation (PIT). Fixes #80.

EDIT: 

#### `loo_pit`
Added a new function to stats module to calculate LOO-PIT values on idata/array-like objects. Its call signature (`loo_pit(idata, y, y_hat, log_weights)`) is very flexible, allowing to combine array inputs and data retrieved from the inference data object. It enforces hard boundaries on 0 and 1 so that numerical precision issues do not make `loo_pit` return values outside the range.
#### `plot_loo_pit`
Added a new function to plot the results of `loo_pit`. It has two main working modes:
1. The first one is to calculate the kde of the loo pit values, plot it and overlay the kde for several uniform variable realizations having each realization the same number of observations as the loo pit values.
image

1. The second one is to calculate the Empirical CDF of the loo pit values and compare them with their theoretical value (uniform distribution). Then, the difference between the loo pit ecdf and the uniform cdf is plotted along with its envelope.
image

#### `apply_test_function`
Added a new function to ease the calculation of bayesian test quantities by retrieving the relevant data from idata objects, wrapping `wrap_xarray_ufunc` using some sensible defaults and then storing the result in the idata object itself. For example, it can be used like the examples in BDA p. 146-150 or pointwise mode to use for LOO-PIT checks (which are generally done on _T(y\_i,theta)=y\_i_ but can be applied to any test function result).

#### `$ARVIZ_LOAD`
Added env variable `$ARVIZ_LOAD` to regulate the behaviour of `InferenceData.from_netcdf`. The default has not been modified, it is still lazy loading, but when working with many files simultaneously (i.e. tests or generating docs) it is necessary to load the data into memory. Setting `$ARVIZ_LOAD="EAGER"` (case insensitive) loads the data into memory. I cannot think of any case other than tests or docs where this can happen, so I have not included anything about it on the documentation.

#### `plot_posterior`
Added the `group` argument like in plot_density.
#### `make_ufunc` and `wrap_xarray_ufunc`
Modified their call signature slightly to make them compatible with multiple input datasets. In addition, they are now public and present in the api docs.

### DIFF
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -1,4 +1,5 @@
 """Data structure for using netcdf groups with xarray."""
+import os
 from collections import OrderedDict
 from collections.abc import Sequence
 from copy import copy as ccopy, deepcopy
@@ -60,9 +61,13 @@ class InferenceData:
         with nc.Dataset(filename, mode="r") as data:
             data_groups = list(data.groups)
 
+        arviz_load_mode = os.environ.get("ARVIZ_LOAD", "lazy").lower()
         for group in data_groups:
             with xr.open_dataset(filename, group=group) as data:
-                groups[group] = data
+                if arviz_load_mode == "eager":
+                    groups[group] = data.load()
+                else:
+                    groups[group] = data
         return InferenceData(**groups)
 
     def to_netcdf(self, filename, compress=True):
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -47,6 +47,9 @@ class InferenceData:
         """Initialize object from a netcdf file.
 
         Expects that the file will have groups, each of which can be loaded by xarray.
+        By default, the datasets of the InferenceData object will be lazily loaded. To
+        modify this behaviour, the environment variable ``ARVIZ_LOAD`` must be set to
+        ``EAGER`` (case insensitive) in order to load objects in memory instead.
 
         Parameters
         ----------
--- a/arviz/plots/loopitplot.py
+++ b/arviz/plots/loopitplot.py
@@ -67,10 +67,8 @@ def plot_loo_pit(
         raise ValueError("idata must be of type InferenceData")
 
     if ax is None:
-        (figsize, ax_labelsize, _, xt_labelsize, linewidth, markersize) = _scale_fig_size(
-            figsize, textsize, 1, 1
-        )
-        fig, ax = plt.subplots(1, 1, figsize=figsize)
+        (figsize, _, _, xt_labelsize, linewidth, _) = _scale_fig_size(figsize, textsize, 1, 1)
+        _, ax = plt.subplots(1, 1, figsize=figsize, constrained_layout=True)
 
     if plot_kwargs is None:
         plot_kwargs = {}
@@ -121,6 +119,7 @@ def plot_loo_pit(
             ax.plot(x_vals, unif_density, **plot_unif_kwargs)
     ax.plot(x_vals, loo_pit_kde, **plot_kwargs)
 
+    ax.tick_params(labelsize=xt_labelsize)
     if legend:
         if not use_hpd:
             ax.plot([], label="Uniform", **plot_unif_kwargs)
--- a/arviz/plots/loopitplot.py
+++ b/arviz/plots/loopitplot.py
@@ -49,8 +49,9 @@ def plot_loo_pit(
         Plot the difference between the LOO-PIT Empirical Cumulative Distribution Function
         (ECDF) and the uniform CDF instead of LOO-PIT kde.
         In this case, instead of overlaying uniform distributions, the beta ``credible_interval``
-        interval around the theoretical uniform CDF is shown. For more information, see
-        `Vehtari et al. (2019)`, `Appendix G <https://avehtari.github.io/rhat_ess/rhat_ess.html>`_
+        interval around the theoretical uniform CDF is shown. This approximation only holds
+        for large S and ECDF values not vary close to 0 nor 1. For more information, see
+        `Vehtari et al. (2019)`, `Appendix G <https://avehtari.github.io/rhat_ess/rhat_ess.html>`_.
     ecdf_fill : bool, optional
         Use fill_between to mark the area inside the credible interval. Otherwise, plot the
         border lines.
--- a/arviz/plots/loopitplot.py
+++ b/arviz/plots/loopitplot.py
@@ -1,9 +1,12 @@
 import numpy as np
 import matplotlib.pyplot as plt
+from matplotlib.colors import to_rgb, rgb_to_hsv, hsv_to_rgb
 
-from ..stats import loo_pit as _loo_pit
+from ..data import InferenceData
+from ..stats import loo_pit as _loo_pit, psislw as _psislw
 from .plot_utils import _scale_fig_size
 from .kdeplot import _fast_kde
+from .hpdplot import plot_hpd
 
 
 def plot_loo_pit(
@@ -11,17 +14,56 @@ def plot_loo_pit(
     y,
     y_hat=None,
     n_unif=100,
+    use_hpd=False,
     figsize=None,
     textsize=None,
+    color="C0",
+    legend=True,
     ax=None,
     plot_kwargs=None,
     plot_unif_kwargs=None,
+    hpd_kwargs=None,
 ):
-    """Plot LOO-PIT."""
-    if plot_kwargs is None:
-        plot_kwargs = {}
-    if plot_unif_kwargs is None:
-        plot_unif_kwargs = {}
+    """Plot Leave-One-Out (LOO) probability integral transformation (PIT) predictive checks.
+
+    Parameters
+    ----------
+    idata : InferenceData
+        InferenceData object with groups `observed_data`, `posterior_predictive` and
+        `sample_stats`. Objects that can be converted to InferenceData are not accepted.
+    y : str
+        Name of the observed_data variable to use.
+    y_hat : str, optional
+        Name of the posterior_predictive variable to use. If None, it will be taken as equal
+        to y.
+    n_unif : int, optional
+        Number of datasets to simulate and overlay from the uniform distribution.
+    use_hpd : bool, optional
+        Use plot_hpd to fill between hpd values instead of overlaying the uniform distributions.
+    figsize : figure size tuple, optional
+        If None, size is (8 + numvars, 8 + numvars)
+    textsize: int, optional
+        Text size for labels. If None it will be autoscaled based on figsize.
+    color : str or array_like, optional
+        Color of the LOO-PIT estimated pdf plot.
+    legend : bool, optional
+        Show the legend of the figure.
+    ax : axes, optional
+        Matplotlib axes
+    plot_kwargs : dict, optional
+        Additional keywords passed to ax.plot, for LOO-PIT line
+    plot_unif_kwargs : dict, optional
+        Additional keywords passed to ax.plot, for overlaid uniform distributions.
+    hpd_kwargs : dict, optional
+        Additional keywords passed to az.plot_hpd
+
+    Returns
+    -------
+    axes : axes
+        Matplotlib axes
+    """
+    if not isinstance(idata, InferenceData):
+        raise ValueError("idata must be of type InferenceData")
 
     if ax is None:
         (figsize, ax_labelsize, _, xt_labelsize, linewidth, markersize) = _scale_fig_size(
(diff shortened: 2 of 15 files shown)


ahartikainen:
also, is ppc related to posterior values?

OriolAbril:
> also, is ppc related to posterior values?

You mean in pymc?

ahartikainen:
Yes

aloctavodia:
@OriolAbril PyMC3 takes samples from all the chains and combine them. I guess it should not be too difficult to change that to get the results separated per chain (optionally).

OriolAbril:
@aloctavodia  I have given a quick look to the code for posterior predictive sampling in pymc, and arviz datasets contain only 500 samples is probably because it was created with the same code in the coockbook, restricting the number of posterior predictive samples to 500. The code already calculates one pp sample per posterior sample (there is no combination between chains).

There is an issue though, which is that the resulting array from `sample_posterior_predictive` is flattened. Therefore when imported to arviz without restricting the number of samples, it reads the posterior predictive array as one chain and 2000 draws, instead of 4x500. 

I am not sure whether this should be taken into account in arviz, or if it should be modified directly in pymc

aloctavodia:
That's what I meat by "combined", sorry about the confusion you are right _flattened_ is a better description. I guess the best route is to change the PyMC code to _optionally_ return the non-flattened array.

arviz/stats/stats_utils.py
```diff
@@ -123,7 +123,15 @@ def _ufunc(*args, out=None, **kwargs):
                 msg += " Correct shape is {}".format(arys[-1].shape[:-n_dims])
                 raise TypeError(msg)
         for idx in np.ndindex(out.shape):
-            arys_idx = [ary[idx].ravel() if ravel else ary[idx] for ary in arys]
+            try:
```
ahartikainen: do we need try except here? Could this all be done with

                arys_idx = []
                for ary in arys:
                    idx_broadcast = tuple(
                        0 if dim_len == 1 else idx_i for idx_i, dim_len in zip(idx, ary.shape)
                    )
                    arys_idx.append(ary[idx_broadcast].ravel() if ravel else ary[idx_broadcast])

OriolAbril:
## Test function application examples

Note: the data used is from [this toy model](https://github.com/OriolAbril/Bayesian-Inference-examples/tree/master/Binomial_on_2D_data)

Test quantities can be applied on all the observations at the same time (BDA p. 146-150), yielding a single scalar value for all the observations and one scalar value per posterior predictive sample, or they can be applied pointwise to each observation (BDA p. 152-153, i.e. LOO-PIT). Both options are available in `apply_test_function`.

Below there are 2 examples of using `apply_test_function`. They are really simple and can all be done without this function, but there may be cases where the function cannot be easily vectorized. I will add some recommendations in the docs. Their only goal is to show the function works and to show possible uses of test functions.

### Example

```python
pointwise = False # True
for idata in [idata1, idata2]:
    az.apply_test_function(
            idata,
            lambda y, theta: np.min(y),
            group="both",
            var_names={"observed_data": ["Right"], "posterior_predictive": ["Right_hat"]},
            out_name_data="T_min",
            pointwise=pointwise,
        )

    az.apply_test_function(
            idata,
            lambda y, theta: np.mean(y),
            group="both",
            var_names={"observed_data": ["Right"], "posterior_predictive": ["Right_hat"]},
            out_name_data="T_mean",
            pointwise=pointwise,
        )
    _, axes = plt.subplots(1,2)
    for ax, name in zip(axes, ["T_min", "T_mean"]):
        if pointwise:
            az.plot_loo_pit(idata, y=name, y_hat=name+"_hat", ax=ax)
        else:
            az.plot_posterior(
                idata.posterior_predictive[name+"_hat"].values,
                ref_val=float(idata.observed_data[name]),
                ax=ax,
            )
```

#### T(y, theta), `pointwise=False`

In this case, following BDA, the posterior_predictive of the test quantity is obtained, then, the single value from the observed data is compared with this distribution, to see if it could have been generated with the same process. Below there is an example of a good model, where the posterior predictive fits the data, and bad model where it does not.

good_test_hpd

bad_test_hpd

I will try to add the group parameter to `plot_posterior`.

#### T(y_i, theta), `pointwise=True`

In this case, both T_min and T_mean will yield the same result, as observations are scalars, and there is no theta contribution. In fact, here not only the two test functions will be equal, they will be equal to the observed variable and posterior predictive. Now, to see if the model agrees with the data, we will use LOO-PIT. The models are the same:

good_loo_pit

bad_loo_pit

ahartikainen:
do we possibility to select random sample vs envelope?

(Main line could be a bit thicker by default)

OriolAbril:
You mean envelope instead of overlaying 100 uniform distributions right? I am not really sure if I have calculated it properly. I used `plot_hpd` on the 100 uniform distribution kde values, it ends up being something like this:

image

ahartikainen:
Envelope = theoretical limits

OriolAbril:
I have already calculated the envelope for loo_pit values:

#### Centered 8 schools (only data from first chain)
image


#### Radon (only data from first chain)
image


I will commit this new option soon, using the color scheme and linestyles from the current `plot_loo_pit`.

arviz/plots/loopitplot.py
```diff
@@ -101,23 +111,40 @@ def plot_loo_pit(
...
-    x_vals = np.linspace(0, 1, len(loo_pit_kde))
-    if use_hpd:
-        unif_densities = np.empty((n_unif, len(loo_pit_kde)))
-        for idx in range(n_unif):
-            unif_densities[idx, :], _, _ = _fast_kde(unif[idx, :], xmin=0, xmax=1)
-        plot_hpd(x_vals, unif_densities, **hpd_kwargs)
+        loo_pit = _loo_pit(y.values, y_hat.values, log_weights.values).flatten()
+
+    if ecdf:
+        loo_pit = np.sort(loo_pit)
```
ahartikainen: loo_pit.sort() to do it inplace?

arviz/stats/stats.py
```diff
@@ -1126,3 +1136,232 @@ def waic(data, pointwise=False, scale="deviance"):
                 "waic_scale",
             ],
         )
+
+
+def loo_pit(y, y_hat, log_weights):
```
ahartikainen: Could we have a wrapper for inference data, where y and yhat could be strings? Or is there one already?

OriolAbril: There is no wrapper, but the conversion from idata+strings is already done in the plot, so making the wrapper is no problem at all. Once it is done though, should this function and the wrapper be public? Should they be the same function?

ahartikainen: You could follow ess / rhat format? So main function would be public.

OriolAbril: I am not sure on how to export their format here. ess and rhat have the basic ufunc, which is private and the public function. And the public function `data` argument accepts either inference data or array. The basic ufunc is already present, and to follow the ess/rhat behaviour, the public function should accept both `loo_pit(idata: InferenceData, y: str, y_hat: str)` and `loo_pit(y: array-like, y_hat: array-like, log_weights: array-like)`, which do not seem very compatible

ahartikainen: I think that format is ok

    idata, y: str, yhat: str
    y: ndarray, yhat: ndarray

Do you think there would be cases where 1) user calculates `yhat` after sampling -->

    idata
    y : str
    yhat : ndarray

or 2) where user has forgot to include data to idata -->

    idata
    y: ndarray
    yhat: str

And are we assuming that yhat is in posterior_predictive and y in observed_data. Maybe there could be kwarg that can change these defaults.

I know that is a bit complicated, but atleast I don't always remember to do posterior_predictive, and some interfaces don't have observed_data by default. So having flexible interface is great.

OriolAbril: I have modified the call signature to `loo_pit(idata, y, y_hat, log_weights)` and the same in `plot_loo_pit()` (arguments are passed directly to `loo_pit`). Now, the following possibilities are accepted:

1. `loo_pit(idata: InferenceData, y: str)`
1. `loo_pit(idata: InferenceData, y: str, y_hat: str)`
1. `loo_pit(idata: InferenceData, y: str, y_hat: array-like)`
1. `loo_pit(idata: InferenceData, y: array-like, y_hat: str)`
1. `loo_pit(idata: InferenceData, y: array-like, y_hat: array-like)`
1. `loo_pit(idata: InferenceData, y: str, y_hat: str, log_weights: array-like)`
1. `loo_pit(idata: InferenceData, y: str, y_hat: array-like, log_weights: array-like)`
1. `loo_pit(idata: InferenceData, y: array-like, y_hat: str, log_weights: array-like)`
1. `loo_pit(y: array-like, y_hat: array-like, log_weights: array-like)`

With the array inputs taking priority over strings or None, that is, if `log_weights` is array-like, these values are used instead of checking for a `log_likelihood` variable in the InferenceData. For now, they are all optional _positional_ arguments, which is not ideal. I would set _all_ 4 of them to be optional _named_ arguments, what do you think? Leaving `idata` as positional and set the other three to named arguments also makes sense.

ahartikainen:
Should you rebase, not merge?

arviz/plots/loopitplot.py
```diff
@@ -0,0 +1,208 @@
...
+        equal to y, thus, y must be str too.
+    log_weights : array or DataArray
+        Smoothed log_weights. It must have the same shape as ``y_hat``
+    ecdf : bool, optional
+        Plot the difference between the LOO-PIT Empirical Cumulative Distribution Function
+        (ECDF) and the uniform CDF instead of LOO-PIT kde.
+        In this case, instead of overlaying uniform distributions, the beta ``credible_interval``
+        interval around the theoretical uniform CDF is shown. This approximation only holds
+        for large S and ECDF values not vary close to 0 nor 1. For more information, see
+        `Vehtari et al. (2019)`, `Appendix G <https://avehtari.github.io/rhat_ess/rhat_ess.html>`_.
```
OriolAbril: Do you think this explanation will be enough or should I add a warning when `ecdf=True` and the number of observations is below some number?

OriolAbril:
Ready for review/merging. The changelog in the initial comment is quite thorough.

arviz/tests/test_stats.py
```diff
@@ -334,6 +349,133 @@ def test_multidimensional_log_likelihood(func):
...
+        loo_pit_data = loo_pit(idata=centered_eight, y=y, y_hat=y_hat, log_weights=log_weights)
+    else:
+        loo_pit_data = loo_pit(idata=None, y=y_arr, y_hat=y_hat_arr, log_weights=log_weights_arr)
+    assert np.all((loo_pit_data >= 0) & (loo_pit_data <= 1))
+
+
+def test_loo_pit_bad_input(centered_eight):
+    arr = np.random.random((8, 200))
+    with pytest.raises(ValueError):
+        loo_pit(idata=arr, y="obs")
```
canyon289: II have a slight preference that this be parameterized so if there's failure its clear which one is failing

arviz/tests/test_stats.py
```diff
@@ -334,6 +349,133 @@ def test_multidimensional_log_likelihood(func):
...
+    kwargs[arg] = 2
+    with pytest.raises(ValueError):
+        loo_pit(idata=centered_eight, **kwargs)
+
+
+def test_loo_pit_bad_input_shape():
+    y = np.random.random(8)
+    y_hat = np.random.random((8, 200))
+    log_weights = np.random.random((8, 200))
+    with pytest.raises(ValueError):
```
canyon289: Same comment here. Additionally if you're trying to capture a specific `ValueError` it would be good to check the error text to ensure the intended `ValueError` is being raised.

arviz/tests/test_stats.py
```diff
@@ -334,6 +349,133 @@ def test_multidimensional_log_likelihood(func):
...
+        loo_pit(y=y, y_hat=y_hat[None, :], log_weights=log_weights)
+    with pytest.raises(ValueError):
+        loo_pit(y=y, y_hat=y_hat[1:3, :], log_weights=log_weights)
+    with pytest.raises(ValueError):
+        loo_pit(y=y, y_hat=y_hat[:, :100], log_weights=log_weights)
+
+
+@pytest.mark.parametrize("group", ["both", "observed_data", "posterior_predictive"])
```
canyon289: The combinatorics on the parameterization here mean that a whole lot of tests are being run. Do we need to test for every combination of variation? If not would it be easier to enumerate the minimum number of tests needed by hand?

It might not be just asking

arviz/tests/test_stats.py
```diff
@@ -334,6 +349,133 @@ def test_multidimensional_log_likelihood(func):
...
+    if group == "both":
+        test_dict = {"observed_data": "T", "posterior_predictive": ["T"]}
+    else:
+        test_dict = {group: "T"}
+
+    fails = check_multiple_attrs(test_dict, idata_out)
+    assert not fails
+
+
+def test_apply_test_function_bad_group(centered_eight):
```
canyon289: Docstrings on all these tests would be nice, even if they;re really simple and are mostly in the method name. For example this one, at a glance I don't know what bad group is versus missing group. 

Does bad group mean that a group string is not matched?

arviz/tests/test_plots.py
```diff
@@ -981,3 +983,26 @@ def test_plot_ess_no_divergences(models):
...
+        {"ecdf": True, "credible_interval": 0.97, "fill_kwargs": {"hatch": "/"}},
+    ],
+)
+def test_plot_loo_pit(models, kwargs):
+    axes = plot_loo_pit(idata=models.model_1, y="y", **kwargs)
+    assert axes
+
+
+def test_plot_loo_pit_error(models):
+    with pytest.raises(ValueError):
```
canyon289: Why is an error expected with this one?

arviz/plots/loopitplot.py
```diff
@@ -0,0 +1,208 @@
...
+    if plot_kwargs is None:
+        plot_kwargs = {}
+    plot_kwargs["color"] = color
+    plot_kwargs.setdefault("linewidth", linewidth * 1.4)
+    plot_kwargs.setdefault("label", "LOO-PIT ECDF" if ecdf else "LOO-PIT")
+    plot_kwargs.setdefault("zorder", 5)
+
+    if plot_unif_kwargs is None:
+        plot_unif_kwargs = {}
+    light_color = rgb_to_hsv(to_rgb(plot_kwargs.get("color")))
```
ahartikainen: should we do this only if the user has not provided the color?

canyon289:
I left some comments, mostly so uninformed uneducated people (like me) will have an easier time understanding what tests do so we can dive in more quickly.

LGTM to me though overall! Thank you!

arviz/stats/stats.py
```diff
@@ -1132,3 +1143,311 @@ def waic(data, pointwise=False, scale="deviance"):
                 "waic_scale",
             ],
         )
+
+
+def loo_pit(idata=None, *, y=None, y_hat=None, log_weights=None):
+    """Compute LOO-PIT values.
```
ahartikainen: Where did LOO-PIT come from (acronym?)

arviz/stats/stats.py
```diff
@@ -1132,3 +1143,311 @@ def waic(data, pointwise=False, scale="deviance"):
...
+    results.
+
+    .. plot::
+        :context: close-figs
+
+        >>> import arviz as az
+        >>> idata = az.load_arviz_data("centered_eight")
+        >>> az.apply_test_function(idata, lambda y, theta: np.min(y))
+        >>> T = np.asscalar(idata.observed_data.T)
+        >>> az.plot_posterior(idata, var_names=["T"], group="posterior_predictive", ref_val=T)
```
ahartikainen: is this example correct? does is need inplace?

arviz/tests/test_stats.py
OriolAbril: Probably not all cases are extremely necessary, but whatever the subset, some errors may not trigger a test failure. It is also not the most important function though, so they can probably be reduced.

arviz/tests/test_plots.py
OriolAbril: ecdf and use_hpd are not compatible, I will add an explanation in the docstring

arviz/data/inference_data.py
```diff
@@ -60,9 +61,13 @@ def from_netcdf(filename):
         with nc.Dataset(filename, mode="r") as data:
             data_groups = list(data.groups)
 
+        arviz_load_mode = os.environ.get("ARVIZ_LOAD", "lazy").lower()
```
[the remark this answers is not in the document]
OriolAbril: It is not a bad idea. Should I update `InferenceData.from_netcdf` docstring only or also the function in `io_netcdf`?

arviz/plots/loopitplot.py
OriolAbril: Do you mean that the default when color is present should be that loo-pit line and unif lines have the same color? Or that when plot_unif_kwargs has the key color this calculation should be avoided?

OriolAbril: I decided on this behaviour so that loo-pit line and unif lines have similar but different color, unless the user sets color in plot_unif_kwargs. I think that they should be slightly different, which is why using exactly equal colors is not default nor _defaultish_

arviz/data/inference_data.py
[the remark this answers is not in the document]
ahartikainen: I think both

arviz/stats/stats.py
OriolAbril: The default with `apply_test_function` is `inplace=True`

arviz/plots/loopitplot.py
ahartikainen: Ok, now I understand, sounds good

arviz/stats/stats.py
OriolAbril: Now that I think about it though, there should be a warning or error before overwriting already present variables.

arviz/plots/loopitplot.py
OriolAbril: I will try to clarify this behaviour on the docs.

ahartikainen:
LGTM

PR #697, opened 2019-06-07
Lint fix

canyon289:


azure-pipelines.yml
```diff
@@ -59,14 +59,18 @@ jobs:
...
       # Tests saving plots and deleting files
       if [[ $NAME == LINT ]]; 
-      then docker run  --mount type=bind,source="$(pwd)",target=/opt/arviz --name arviz_test arviz:latest bash -c \
+      then docker run  -t --mount type=bind,source="$(pwd)",target=/opt/arviz --name arviz_test arviz:latest bash -c \
       "mkdir -p test_images/new_folder && pytest -v arviz/tests/ --save -k plot_khat --junitxml=junit/test-results.xml --cov=arviz/ --cov-report=xml --cov-report=html";
         fi
+
+      if [ $? -ne 0 ]; then
+        exit 1
+        if
```
OriolAbril: Shouldn't this be a `fi`? looks like a typo.

OriolAbril:
Also, I was wondering if having the test and the lint in two different scripts would allow to run the lint in the python3.6 unit test. It would reduce the number of jobs from 4 to 3. I was thinking about removing the `if name == unit` to execute the test always, and set the name lint to `Python_36_Unit_Test` which would become `Python_36_Unit_Test_Lint` or something similar. As they are in different scripts, if one works but the other does not, it should still be visible where is the problem. I am not sure if this approach would have any advantages to the current one, it is mainly to understand azure and experiment a little.

azure-pipelines.yml
canyon289: Yea definitely

canyon289:
> Also, I was wondering if having the test and the lint in two different scripts would allow to run the lint in the python3.6 unit test. It would reduce the number of jobs from 4 to 3. I was thinking about removing the `if name == unit` to execute the test always, and set the name lint to `Python_36_Unit_Test` which would become `Python_36_Unit_Test_Lint` or something similar. As they are in different scripts, if one works but the other does not, it should still be visible where is the problem. I am not sure if this approach would have any advantages to the current one, it is mainly to understand azure and experiment a little.

We used to have this but got different lint failings in 3.5 and 3.6 which was annoying. It also wasn't fun having in in the testing pipeline because then we couldn't tell if the lint was failing or tests were failing all that easily.

canyon289:
If there's no objections I'll merge this in three days. I think we should still continue evaluating Azure Pipelines for another month or so. I have a list of pros and cons and can write it up after scipy tutorial

ahartikainen:
I think this can go in, when you are ready.

I agree, that we should still run both.

Also we need to evaluate if we can deploy stuff from azure (numba compilation etc)

PR #702, opened 2019-06-11
Improve pymc posterior predictive handling

OriolAbril:
`form_pymc` currently stores all posterior predictive samples in 1 chain. Therefore, the default case, where an array of shape `(nchains*ndraws, *shape)` is returned, the posterior_predictive dataset has shape `(1,nchains*ndraws, *shape)`, instead of being `(nchains, ndraws, *shape)` (which would be the same shape as the log_likelihood for instance). 

I have modified for it to work in the following cases:

* `(nchains, ndraws, *shape)` -> `(nchains, ndraws, *shape)`
* `(nchains*ndraws, *shape)` -> `(nchains, ndraws, *shape)`

and for nsamples different than nchains*ndraws:
* `(nsamples, *shape)` -> `(1, nsamples, *shape)` plus a `logging.warning`

### DIFF
--- a/arviz/tests/test_data_pymc.py
+++ b/arviz/tests/test_data_pymc.py
@@ -35,21 +35,50 @@ class TestDataPyMC3:
             dims={"theta": ["school"], "eta": ["school"]},
         )
 
-    def test_posterior(self, data, eight_schools_params):
+    def test_from_pymc(self, data, eight_schools_params):
         inference_data = self.get_inference_data(data, eight_schools_params)
-        assert hasattr(inference_data, "posterior")
+        test_dict = {
+            "posterior": ["mu", "tau", "eta", "theta"],
+            "sample_stats": ["diverging", "log_likelihood"],
+            "posterior_predictive": ["obs"],
+            "prior": ["mu", "tau", "eta", "theta"],
+            "observed_data": ["obs"],
+        }
+        fails = check_multiple_attrs(test_dict, inference_data)
+        assert not fails
 
-    def test_sampler_stats(self, data, eight_schools_params):
-        inference_data = self.get_inference_data(data, eight_schools_params)
-        assert hasattr(inference_data, "sample_stats")
+    def test_posterior_predictive_reshaped(self, data, chains, draws, eight_schools_params):
+        with data.model:
+            posterior_predictive = pm.sample_posterior_predictive(data.obj)
 
-    def test_posterior_predictive(self, data, eight_schools_params):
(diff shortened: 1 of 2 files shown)


arviz/tests/test_data_pymc.py
```diff
@@ -35,21 +35,50 @@ def get_inference_data(self, data, eight_schools_params):
...
+        with data.model:
+            posterior_predictive = pm.sample_posterior_predictive(data.obj)
 
-    def test_posterior_predictive(self, data, eight_schools_params):
-        inference_data = self.get_inference_data(data, eight_schools_params)
-        assert hasattr(inference_data, "posterior_predictive")
+        inference_data = from_pymc3(
+            trace=data.obj,
+            posterior_predictive={
+                k: v.reshape((chains, draws, *v.shape[1:])) for k, v in posterior_predictive.items()
```
ahartikainen: I assume they also use `C` order?

arviz/tests/test_data_pymc.py
```diff
@@ -35,21 +35,50 @@ def get_inference_data(self, data, eight_schools_params):
...
+            trace=data.obj,
+            posterior_predictive=posterior_predictive,
+            coords={"school": np.arange(eight_schools_params["J"])},
+            dims={"theta": ["school"], "eta": ["school"]},
+        )
+        records = caplog.records
+        shape = inference_data.posterior_predictive.obs.shape
+        assert np.all([obs_s == s for obs_s, s in zip(shape, (1, 370, eight_schools_params["J"]))])
+        assert len(records) == 1
+        assert records[0].levelname == "WARNING"
```
ahartikainen: should you test against ravelled and reshaped values (slicing)

arviz/tests/test_data_pymc.py
OriolAbril: Yes

arviz/tests/test_data_pymc.py
ahartikainen:
```
for key, values in posterior_predictive.items():
    ivalues = inference_data.posterior_predictive[key]
    for chain in range(nchain):
        assert np.all(ivalues[chain] == values[chain * ndraw : (chain + 1) * ndraw]
```

OriolAbril: Okey, now I get it, thanks for the explanation. It is interesting to test this too. I will add it.

ahartikainen:
I think this needs black and is quite ready.

@ColCarroll can you review this?

ColCarroll:
This is nice, thank you!

PR #705, opened 2019-06-13
[WIP] Add significant digits option for plotting posterior

rpgoldman:


rpgoldman:
Here's a picture from a notebook showing the difference between significant digits (new) and rounding (old):
image
If people are happy with this, it's possible that it should spread to other plots.

ColCarroll:
I like how this looks -- I would tend to welcome a PR that did this.

My continuing caveat is that I'm always worried how much complexity we add:

- There are 3 number elements on each plot ("mean=11.3", "94% HPD", "7.5/14.9") looks like this applies to the first and the last?
- This is significant figures, and not decimals (thanks for choosing an example where we could see!)

I guess issues like these are where I keep coming back to wanting a grammar of graphics approach, so each text element could have this feature, and it is more "opt-in".

rpgoldman:
Pylint didn't like this.  New version should make it happier.

rpgoldman:
> I like how this looks -- I would tend to welcome a PR that did this.
> 
> My continuing caveat is that I'm always worried how much complexity we add:
> 
> * There are 3 number elements on each plot ("mean=11.3", "94% HPD", "7.5/14.9") looks like this applies to the first and the last?

Yes, that's right.  Significant digits really don't make a lot of sense for percentages, and that second number is a user-set parameter, not a function of the information in the distribution.

> * This is significant figures, and not decimals (thanks for choosing an example where we could see!)

Yes, I had some variables (regression coefficients applied to the square of a predictor) that had what looked like empty HPDs because they had small values (should I have been rescaling these?).
> 
> I guess issues like these are where I keep coming back to wanting a grammar of graphics approach, so each text element could have this feature, and it is more "opt-in".

I *think* I agree, if I understand you correctly, this might be something like dims and coords, where you supply a dictionary to get the full configurability.

I don't know my way around the arviz code base, so if someone could point me at what other plots might be improved by having a significant figures option, I would appreciate it.

aloctavodia:
Why not using the `g` string formater here, and make the `round_to`argument specify the significant figures? About the "3 numbers" I think is totally fine to leave the percentage out of the formatting

rpgoldman:
> Why not using the `g` string formater here, and make the `round_to`argument specify the significant figures? About the "3 numbers" I think is totally fine to leave the percentage out of the formatting

I did use the `%g` formatter, but I was reluctant to change the existing functionality to *replace* rounding by significant figures, so I just made them an alternative. If the community generally thought significant figures was a more appropriate alternative, I will be happy to revise the code in the PR.

What's the procedure for establishing consensus on something like this?

rpgoldman:
P.S. I think that `master` is broken now, right? I don't think the pystan error I see in Travis is related to my PR.

aloctavodia:
OK, but why we need the function `to_sig_figs` if we use the `g` formatter? What I am missing?

I will keep the `round_to` argument but change the interval behaviour (and docstring) to use significant figures.  

About the consensus process, we generally propose stuff and wait a few days to see if someone has strong feelings and arguments against the proposal, otherwise we merge.

aloctavodia:
Right that seems unrelated.

ahartikainen:
Yes, it's fixed in concat PR

rpgoldman:
> OK, but why we need the function `to_sig_figs` if we use the `g` formatter? What I am missing?

It's not needed in `display_point_estimate()` -- I just cargo-culted from the way `round_to` was handled. I have a revision to the PR that I will push in a moment.

But there's something going on in the calls to `Axes.text()` in `display_hpd()` that I don't understand.  The `s` argument to this method is specified to be a string, but we pass in something that is either a number or an `np.array`.  I imagine that this is going to be automagically stringified, but it's kind of messy.  I'll modify this for your review.

rpgoldman:
Second time today: I had better put black formatting into my pre-commit hook!

ColCarroll:
What do you think about changing `round_to` behavior to give significant digits? It keeps the API from getting more complicated, should reduce the implementation complexity, and seems like what a user would desire most of the time.

This PR looks in good shape, and I'd be happy to do the code shuffling to make that happen, but interested in your (or other's) thoughts.

PR #708, opened 2019-06-17
New ess plots

OriolAbril:
Add the plots for new ess (#621).

Example plots on the centered eight dataset. The default appearance is quite simple:

image

image

image

But it can be customized to look like the plots in the reference paper for example:

image

image

I have set the same default appearance for both quantile and local, differing from the paper, because I think that using a hist like plot can be misleading. Now the only difference is the ylabel, which can be confusing to someone not familiar with this kind of plots. Would you rather have different appearance between quantile and local ess plots?

### DIFF
--- a/arviz/plots/essplot.py
+++ b/arviz/plots/essplot.py
@@ -149,6 +149,8 @@ def plot_ess(
 
     data = convert_to_dataset(idata, group="posterior")
     var_names = _var_names(var_names, data)
+    n_draws = len(data.draw)
+    n_samples = n_draws * len(data.chain)
 
     if kind == "quantile":
         probs = np.arange(1 / n_points, 1, 1 / n_points)
@@ -179,9 +181,7 @@ def plot_ess(
             dim="ess_dim",
         )
     else:
-        n_draws = len(data.draw)
         first_draw = data.draw.values[0]
-        n_samples = n_draws * len(data.chain)
         ylabel = "{}"
         xdata = np.linspace(n_samples / n_points, n_samples, n_points)
         draw_divisions = np.linspace(n_draws // n_points, n_draws, n_points, dtype=int)
@@ -267,11 +267,11 @@ def plot_ess(
             mask = idata.sample_stats[rug_kind].values.flatten()
             values = np.argsort(values)[mask]
             rug_space = np.max(x) * rug_kwargs.pop("space")
-            rug_x, rug_y = values/(len(mask)-1), np.zeros_like(values) - rug_space
+            rug_x, rug_y = values / (len(mask) - 1), np.zeros_like(values) - rug_space
             ax_.plot(rug_x, rug_y, **rug_kwargs)
             ax_.axhline(0, color="k", linewidth=_linewidth, alpha=0.7)
 
-        ax_.axhline(min_ess, **hline_kwargs)
+        ax_.axhline(400 / n_samples if relative else min_ess, **hline_kwargs)
 
         ax_.set_title(make_label(var_name, selection), fontsize=titlesize, wrap=True)
         ax_.tick_params(labelsize=xt_labelsize)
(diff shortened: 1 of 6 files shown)


arviz/plots/essplot.py
```diff
@@ -0,0 +1,244 @@
...
+        Show relative ess in plot.
+    coords : dict, optional
+        Coordinates of var_names to be plotted. Passed to `Dataset.sel`
+    figsize : tuple, optional
+        Figure size. If None it will be defined automatically.
+    textsize: float, optional
+        Text size scaling factor for labels, titles and lines. If None it will be autoscaled based
+        on figsize.
+    n_points : int
+        Number of quantiles for which to plot their quantile/local ess or number of subsets
```
aloctavodia: Number of points for...

arviz/plots/essplot.py
```diff
@@ -0,0 +1,244 @@
...
+    ----------
+    data : obj
+        Any object that can be converted to an az.InferenceData object
+        Refer to documentation of az.convert_to_dataset for details
+    var_names : list of variable names, optional
+        Variables to be plotted, two variables are required.
+    kind : str, optional
+        Options: ``local``, ``quantile`` or ``change``, specify the kind of plot.
+    relative : bool
+        Show relative ess in plot.
```
aloctavodia: Show relative ess in plot `ress = ess / N`.

arviz/plots/essplot.py
```diff
@@ -0,0 +1,244 @@
...
+):
+    """Plot quantile, local or change of effective sample sizes (ESS).
+
+    Parameters
+    ----------
+    data : obj
+        Any object that can be converted to an az.InferenceData object
+        Refer to documentation of az.convert_to_dataset for details
+    var_names : list of variable names, optional
+        Variables to be plotted, two variables are required.
```
aloctavodia: Variables to be plotted.

OriolAbril: True, I copy-pasted too much

OriolAbril:
I think I have not forgotten any feature. If you have time to review and it turns out I did miss something, please say so. I will start writing tests so it can be merged soon.

arviz/plots/essplot.py
```diff
@@ -149,6 +149,8 @@ def plot_ess(
 
     data = convert_to_dataset(idata, group="posterior")
     var_names = _var_names(var_names, data)
+    n_draws = len(data.draw)
```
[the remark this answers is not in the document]
OriolAbril: As the last value of the dimension? I think I am not following

OriolAbril:
> One question (is that evolution calculated as `ess(ary[:i])`?

Yes, the first `i` draws (but all chains) are selected to calculate the ess evolution in ess "change"

arviz/plots/essplot.py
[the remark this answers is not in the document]
ahartikainen: data.draw is coordinate, right? The draw dim should contain value n?

ahartikainen:
Is change desciptive name? Could it be `evolution` or something similar?

arviz/plots/essplot.py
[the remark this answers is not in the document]
OriolAbril: Ohh! okey, yes, indeed. Thanks! :)

arviz/plots/essplot.py
```diff
@@ -164,7 +164,7 @@ def plot_ess(
             dim="ess_dim",
         )
     elif kind == "local":
-        probs = np.arange(0, 1, 1 / n_points)
+        probs = np.linspace(0, 1 - 1 / n_points, n_points)
```
ahartikainen: Is this same as 

    np.linspace(0,1,N, endpoint=False)

(could be different)

aloctavodia:
LGTM!

ahartikainen:
Similar plot for mcse quantile would be cool.

PR #709, opened 2019-06-17
Advanced concat for InferenceData

ahartikainen:
If dim=None --> concat unique groups
Valid dim values are {"chain", "draw"}.

~Needs tests.~
~Needs examples.~ (let's update tutorial in another PR)

### DIFF
--- a/arviz/tests/test_data.py
+++ b/arviz/tests/test_data.py
@@ -158,7 +158,7 @@ def test_addition():
 @pytest.mark.parametrize("copy", [True, False])
 @pytest.mark.parametrize("inplace", [True, False])
 @pytest.mark.parametrize("sequence", [True, False])
-def test_concat(copy, inplace, sequence):
+def test_concat_group(copy, inplace, sequence):
     idata1 = from_dict(
         posterior={"A": np.random.randn(2, 10, 2), "B": np.random.randn(2, 10, 5, 2)}
     )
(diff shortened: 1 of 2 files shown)


OriolAbril:
LGTM.

I also think that in addition to the tutorial, we could add the InferenceData class to the docs. Otherwise, the detailed description of the arguments in `concat` or `sel` cannot be found anywhere. I can create an issue to keep this in mind if you agree.

ahartikainen:
Sounds good.

arviz/tests/test_data.py
```diff
@@ -192,6 +192,58 @@ def test_concat(copy, inplace, sequence):
...
+        new_idata = idata1
+    assert new_idata is not None
+    test_dict = {"posterior": ["A", "B"], "observed_data": ["C", "D"]}
+    fails = check_multiple_attrs(test_dict, new_idata)
+    assert not fails
+    if inplace:
+        assert id(new_idata) == original_idata1_id
+    else:
+        assert id(new_idata) != id(idata1)
+    print(getattr(new_idata.posterior, dim).size)
```
canyon289: Not the biggest of deals. Can this print be changed to a log statement?

scripts/create_testenv.sh
```diff
@@ -56,7 +56,7 @@ if [ "$PYSTAN_VERSION" = "latest" ]; then
 else
   if [ "$PYSTAN_VERSION" = "preview" ]; then
     # try to skip other pre-releases than pystan
-    pip --no-cache-dir install numpy uvloop marshmallow PyYAML
+    pip --no-cache-dir install numpy uvloop marshmallow==3.0.0rc6 PyYAML
```
canyon289: Not asking for a change but curious. What UVloop, and pyYAML used for in pystan? I remember pieces of this conversation in slack but seeing this here am curious.

arviz/tests/test_data.py
ahartikainen: There should be no print, it's my debug line

scripts/create_testenv.sh
ahartikainen: 5s explanation: They are used by `httpstan` / `pystan3` to do communication and parsing io

https://github.com/stan-dev/httpstan

ColCarroll:
:+1:

PR #710, opened 2019-06-18
Numba

Ban-zee:
Numba has been kept optional. Added benchmarks tests for the modified methods(mainly variance, histogram and stats.circstd). These modified methods are twice as fast as the original numpy and scipy methods. Along with this, I have added a class in utils to allow the user to enable or disable numba at will. The required tests have been added as well.


arviz/stats/diagnostics.py
```diff
@@ -510,13 +511,22 @@ def geweke(ary, first=0.1, last=0.5, intervals=20):
         last_slice = ary[int(end - last * (end - start)) :]
 
         z_score = first_slice.mean() - last_slice.mean()
-        z_score /= np.sqrt(first_slice.var() + last_slice.var())
+        if numba_check():
```
canyon289: This comment is not required. While this if else pattern works I challenge us both to come up with a "nicer" pattern. I have some ideas and we can talk together about this.

Again don't worry about it right now but did want to make the comment

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,84 @@
+# Write the benchmarking functions here.
+# See "Writing benchmarks" in the asv docs for more information.
+import numpy as np
+import scipy.stats as st
+
+
+class Hist:
+	def time_histogram(self):
```
canyon289: These look like tabs not spaces :) 

Can you run black over this file for consistent formatting of all .py files?

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,84 @@
...
+						var[i] = stats_variance_1d(data[:, i], ddof=ddof)
+				return var
+			return stats_variance_2d(data)
+		except ImportError:
+			data = np.random.randn(10000, 10000)
+			return np.var(data, axis=1)
+
+
+class CircStd:
+	def time_circ_std(self):
```
canyon289: Looks like its still tabs in here. I ask that this be converted to spaces

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,84 @@
...
+			@numba.njit(cache=True)
+			def _hist(data):
+				return np.histogram(data, bins=100)
+			return _hist(data)
+		except ImportError:
+			data = np.random.rand(10000, 1000)
+			return np.histogram(data, bins=100)
+
+
+class Variance:
```
canyon289: Testing independent logic is a good start and makes it easy to see how numba should be implemented. Is it possible to also benchmark the arviz functions themselves with and without numba?

benchmarks/benchmarks.py
Ban-zee: Sure, look's like something went wrong with black.

benchmarks/benchmarks.py
Ban-zee: This is an issue. This was my initial plan but importing arviz breaks the build and the benchmarks do not execute. In my notebook, the reduction in the timings of the arviz methods was somewhat proportionate to the reduction in the time of the modified function; so I thought that if importing arviz does not work then I can benchmark the modified methods to ensure that the things are going in the right direction.

canyon289: Strategy makes sense. Let us know how we can help you figure out how to keep the build from breaking. Doesn't need to be done on this pr

arviz/stats/diagnostics.py
canyon289: @Ban-zee In the last 20 minutes I thought about this and think we should change on this PR before we move.

We shouldn't do a numba check everytime a function is called. I think a preferred pattern would be to check if Numba is installed once when ArviZ is loaded and then use that constant throughout the entire program.

I also think this will fit in well because users can optionally turn numba off, even if they have it installed.

If you want I can set some time to talk about this synchronously if you want :)

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,86 @@
+# Write the benchmarking functions here.
+# See "Writing benchmarks" in the asv docs for more information.
+import numpy as np
+import scipy.stats as st
+
+
+class Hist:
+    def time_histogram(self):
+        try:
+            data = np.random.rand(10000, 1000)
```
canyon289: Small thing. In these checks the data line is repeated but it seems it doesn't have to be

Ban-zee: It can be changed. It's just that asv takes into account the data initialization time as well so I had to include it in both blocks so as to nullify this effect.

arviz/stats/diagnostics.py
Ban-zee: This can be done. I have a few questions in mind and I am up for discussion anytime. :)

OriolAbril:
Could it be that `asv.conf.json` should be ignored and not added to ArviZ? It is full of local paths and `.asv/` was added to `.gitignore`

OriolAbril:
Also, out of curiosity, have you done any experiments or benchmarks with numba on `make_ufunc`?

Ban-zee:
I have removed the relative path; I had added it for my ease and forgot to replace it :| 
Regarding its inclusion, the conf file is needed for running the benchmarks locally. .asv contains environments and the results of the tests so I decided to gitignore it.
 
Unfortunately, I haven't tested `make_ufunc` yet :(.

benchmarks/benchmarks.py
canyon289: Got it. Today I learned something

asv.conf.json
```diff
@@ -0,0 +1,129 @@
...
+    // determined by looking for tools on the PATH environment
+    // variable.
+    "environment_type": "conda",
+
+    // timeout in seconds for installing any dependencies in environment
+    // defaults to 10 min
+    //"install_timeout": 600,
+
+    // the base URL to show a commit for the project.
+    "show_commit_url": "https://github.com/Ban-zee/arviz/commit/",
```
OriolAbril: Should this be modified to ArviZ instead of fork?

Ban-zee: Don't know how I missed that :( , I'll change it by my next commit

canyon289:
Just curious who is @ABM8? @Ban-zee is this another account that you have? 
image

Ban-zee:
That's me.

canyon289:
> That's me. Looks like I did not rebase :(. I'll resolve this after I wake up.

Out of curiosity why not use one account?

Ban-zee:
It has stuck with me ever since I made the account and I did not bother changing it. :|

Ban-zee:
Just need to add one or two more tests for increasing the code coverage. Apart from that, the pr is up for review.

arviz/stats/diagnostics.py
```diff
@@ -13,10 +13,11 @@
     autocov as _autocov,
     not_valid as _not_valid,
     wrap_xarray_ufunc as _wrap_xarray_ufunc,
+    stats_variance_2d as svar,
 )
 from ..data import convert_to_dataset
-from ..utils import _var_names
-
+from ..utils import _var_names, conditional_jit, conditional_vect
+from .. import Numba
```
OriolAbril: I _think_ that the cyclic import problem comes from this line. This line is in both diagnostics.py and stats.py, thus, it would explain why the cyclic import warnings finish always in one of these two files, and it looks like it is importing from _all_ ArviZ, hence the cycle. I don't really know how to import things from the __init__ file from within the same library though.

Ban-zee: init too imports stats, plots and data. Hence the cyclic import I guess.

Ban-zee:
Up for review!!

arviz/plots/elpdplot.py
```diff
@@ -12,7 +12,8 @@
     format_coords_as_labels,
     set_xticklabels,
 )
-from ..stats import waic, loo, ELPDData
+from ..stats import waic, loo
+from ..stats import ELPDData
```
OriolAbril: `from ..stats import waic, loo, ELPDData`

arviz/tests/test_diagnostics.py
```diff
@@ -4,6 +4,7 @@
 import numpy as np
 from numpy.testing import assert_almost_equal, assert_array_almost_equal
 import pandas as pd
+import scipy.stats as st
```
OriolAbril: `from scipy.stats import circstd`? I may have missed some call to st, but it looks like only circstd is used

arviz/utils.py
```diff
@@ -80,3 +80,58 @@ def wrapper(function):
...
+        """To disable numba."""
+        cls.numba_flag = False
+
+    @classmethod
+    def enable_numba(cls):
+        """To enable numba."""
+        if not cls.numba_flag and numba_check():
+            cls.numba_flag = True
+        else:
+            raise ValueError("Numba is already enabled or not installed")
```
OriolAbril: Should an error be raised when calling `Numba.enable_numba()` if numba is already enables? It feels more natural to just leave numba activated and do nothing (like calling `%matplotlib inline` when inline backend is already set.

arviz/stats/diagnostics.py
```diff
@@ -445,6 +441,12 @@ def mcse(data, *, var_names=None, method="mean", prob=None):
     )
 
 
+@conditional_vect
+def _sqr(a_a, b_b):
+    return math.sqrt(a_a + b_b)
```
ahartikainen: Is math really needed? also name should be `_sqrt`?

    (a_a + b_b)**0.5

arviz/stats/diagnostics.py
```diff
@@ -859,22 +891,32 @@ def _conv_quantile(ary, prob):
...
     """Compute the Markov Chain mean error."""
+    _numba_flag = Numba.numba_flag
     ary = np.asarray(ary)
     if _not_valid(ary, shape_kwargs=dict(min_draws=4, min_chains=1)):
         return np.nan
     ess = _ess_mean(ary)
-    sd = np.std(ary, ddof=1)
+    if _numba_flag:
+        ary = np.ravel(ary)
+        sd = np.sqrt(svar(ary, ddof=1))
```
ahartikainen: _sqrt?

ary.ravel() to same line?

arviz/stats/diagnostics.py
```diff
@@ -859,22 +891,32 @@ def _conv_quantile(ary, prob):
...
     """Compute the Markov Chain sd error."""
+    _numba_flag = Numba.numba_flag
     ary = np.asarray(ary)
     if _not_valid(ary, shape_kwargs=dict(min_draws=4, min_chains=1)):
         return np.nan
     ess = _ess_sd(ary)
-    sd = np.std(ary, ddof=1)
+    if _numba_flag:
+        ary = np.ravel(ary)
+        sd = np.sqrt(svar(ary, ddof=1))
```
ahartikainen: Should there be ssd/sstd function?

arviz/stats/stats_utils.py
```diff
@@ -425,3 +426,30 @@ def __str__(self):
...
+
+@conditional_jit
+def stats_variance_1d(data, ddof=0):
+    a_a, b_b = 0, 0
+    for i in data:
+        a_a = a_a + i
+        b_b = b_b + i * i
+    var = b_b / (len(data)) - ((a_a / (len(data))) ** 2)
+    var = var * (len(data) / (len(data) - ddof))
+    return var
```
ahartikainen: Can we test this if there are huge and tiny values in the same array

arviz/tests/test_diagnostics.py
```diff
@@ -150,7 +152,7 @@ def test_deterministic(self):
         assert (abs(reference["rhat_rank"] - arviz_data["rhat_rank"]) < 6e-5).all(None)
         assert abs(np.median(reference["rhat_rank"] - arviz_data["rhat_rank"]) < 1e-14).all(None)
         not_rhat = [col for col in reference.columns if col != "rhat_rank"]
-        assert (abs(reference[not_rhat] - arviz_data[not_rhat]) < 1e-11).all(None)
+        assert (abs((reference[not_rhat] - arviz_data[not_rhat])).values < 1e-8).all(None)
```
ahartikainen: Is this just random noise or did the function change?

arviz/tests/test_diagnostics.py
```diff
@@ -433,12 +459,12 @@ def test_multichain_summary_array(self, draws, chains):
...
-            assert ess_mean_hat == ess_mean_hat_
-            assert ess_sd_hat == ess_sd_hat_
-            assert ess_bulk_hat == ess_bulk_hat_
-            assert ess_tail_hat == ess_tail_hat_
+            assert_almost_equal(mcse_mean_hat, mcse_mean_hat_)
+            assert_almost_equal(mcse_sd_hat, mcse_sd_hat_)
+            assert_almost_equal(ess_mean_hat, ess_mean_hat_)
+            assert_almost_equal(ess_sd_hat, ess_sd_hat_)
+            assert_almost_equal(ess_bulk_hat, ess_bulk_hat_)
+            assert_almost_equal(ess_tail_hat, ess_tail_hat_)
```
ahartikainen: Is there some difference between the results?

arviz/utils.py
Ban-zee: Wanted to play safe here but after looking at it again, I think this won't lead to any breakdown.

arviz/stats/diagnostics.py
Ban-zee: I guess we can use _sqr, as you suggested.

arviz/tests/test_diagnostics.py
Ban-zee: one of the results was the rounded version of the other one. This rounding off took place at around 10th or 11th place of the decimal iirc.

arviz/tests/test_diagnostics.py
Ban-zee: There was a small change :( so I had to adjust the tests.

ahartikainen: Ok. (The accuracy for rhat is something like 1e-6, so this change is fine)

arviz/tests/test_diagnostics.py
ahartikainen: Can you check the summary code and see if it needs update? (It's is basically copy from the main functions, just trying to minimize duplicate calculations)

arviz/tests/test_diagnostics.py
```diff
@@ -520,3 +557,75 @@ def test_split_chain_dims(self, chains, draws):
         if chains is None:
             chains = 1
         assert split_data.shape == (chains * 2, draws // 2)
+
+
+def test_numba_bfmi():
+    state = Numba.numba_flag
```
canyon289: All these test methods need docstrings

arviz/stats/diagnostics.py
```diff
@@ -530,7 +543,11 @@ def ks_summary(pareto_tail_indices):
     df_k : dataframe
       Dataframe containing k diagnostic values.
     """
-    kcounts, _ = np.histogram(pareto_tail_indices, bins=[-np.Inf, 0.5, 0.7, 1, np.Inf])
+    _numba_flag = Numba.numba_flag
```
canyon289: Can this be set at the top once instead of being called in every func?

ahartikainen: Would that work?

arviz/stats/diagnostics.py
```diff
@@ -510,13 +514,22 @@ def geweke(ary, first=0.1, last=0.5, intervals=20):
         last_slice = ary[int(end - last * (end - start)) :]
 
         z_score = first_slice.mean() - last_slice.mean()
-        z_score /= np.sqrt(first_slice.var() + last_slice.var())
+        if _numba_flag:
```
canyon289: My next challenge for you is to get rid of all these if else statements.

There might be a pattern where we can do something like

```
z_score = _numba(f1, f2, args)
```

This way all the numba specific logic is contained in the Numba class and all the math logic is left here. What do you think?

Ban-zee: Sure.

arviz/stats/diagnostics.py
Ban-zee: It's only being called twice; will setting this it at the top make any difference?

Ban-zee:
**Screenshot of the benchmark tests**
bench-T

ColCarroll:
This is cool -- just so I'm clear about reading this, `numba` is making `circ_std` ~30% faster, `histogram` ~60% faster, and `variance` ~50% faster?

Ban-zee:
Approximately yes, on my machine.

Ban-zee:
It is a single element array

ColCarroll:
`np.float_` for safety, or are you confident it will always be size 1?

ahartikainen:
But `_var` returns always float, right? then `np.zeros(1, dtype=float)`

Ban-zee:
_var does but _sqrt does not. It returns an nd array so the type conversion. The results will always be a single element as batched_traces is 2 dimensional

arviz/stats/diagnostics.py
canyon289: If its only being called twice then I don't think its worth it

arviz/stats/diagnostics.py
```diff
@@ -621,9 +638,9 @@ def _rhat(ary):
     # Calculate chain mean
     chain_mean = np.mean(ary, axis=1)
     # Calculate chain variance
-    chain_var = np.var(ary, axis=1, ddof=1)
+    chain_var = _numba_var(svar, np.var, ary, axis=1, ddof=1)
```
canyon289: This looks nice. Thank you!

arviz/utils.py
```diff
@@ -100,3 +100,66 @@ def format_sig_figs(value, default=None):
...
+    def enable_numba(cls):
+        """To enable numba."""
+        if numba_check():
+            cls.numba_flag = True
+        else:
+            raise ValueError("Numba is not installed")
+
+
+def _numba_var(function_1, function_2, data, axis=None, ddof=0):
+    """To replace the if-else blocks."""
```
canyon289: Slight nudge here to change the function_1, function_2 to something more human recognizable like "numba_func", "numpy_func" or something like that, and expand out docstring

I would also change the docstring from 

`     """To replace the if-else blocks."""`
because this only makes sense to us from history. It won't make sense to someone in the future because they can't see the if else blocks that were replaced.

Something like 

> Returns appropriate function for numba speedup if Numba is installed or enabled

With argument documentation as well.

arviz/utils.py
```diff
@@ -100,3 +100,66 @@ def format_sig_figs(value, default=None):
...
+        except ImportError:
+            return function
+
+    if function:
+        return wrapper(function)
+    else:
+        return wrapper
+
+
+def numba_check():
```
canyon289: Can this be added to the Numba class to consolidate all the logic?

It doesn't have to be but just asking for stylistic and organization reasons

Ban-zee: In my opinion, putting `numba_check` inside the class would not work as `numba_flag` is a class variable and calling class functions within a class is not possible in python. To get it working, I might have to switch to instance but that might lead to further breakdowns.

Ban-zee: In my opinion, putting `numba_check` inside the class would not work as `numba_flag` is a class variable and calling class functions within a class is not possible in python. To get it working, I might have to switch to instance but that might lead to further breakdowns.

canyon289: Can class functions not be called within a class? The following code works, one class method calls the other. Are we thinking of the same thing?

```
class Test:
    @classmethod
    def say_hi(cls): 
        print("hi")

    @classmethod 
    def say_hi_for_real(cls):
        print("Really gonna say hi")
        cls.say_hi()
    
Test.say_hi_for_real()
```

> Really gonna say hi
> hi

Ban-zee: My aim is to set `numba_flag` to `numba_check()` inside the class itself. 

    @classmethod
    def numba_check(cls):
        """Check if numba is installed."""
        numba = importlib.util.find_spec("numba")
        return numba is not None

    numba_flag = numba_check()

    @classmethod
    def disable_numba(cls):
        """To disable numba."""
        cls.numba_flag = False

    @classmethod
    def enable_numba(cls):
        """To enable numba."""
        if numba_check():
            cls.numba_flag = True
        else:
            raise ValueError("Numba is not installed")

Ban-zee: I was aiming for something like this:
class Numba:


    numba_flag = None

    @classmethod
    def numba_check(cls):
        """Check if numba is installed."""
        numba = importlib.util.find_spec("numba")
        return numba is not None

    numba_flag = Numba.numba_check()

    @classmethod
    def disable_numba(cls):
        """To disable numba."""
        cls.numba_flag = False

    @classmethod
    def enable_numba(cls):
        """To enable numba."""
        if cls.numba_check():
            cls.numba_flag = True
        else:
            raise ValueError("Numba is not installed")


A possible workaround is the following :

    @classmethod
    def numba_check(cls):
        """Check if numba is installed."""
        numba = importlib.util.find_spec("numba")
        return numba is not None

    @classmethod
    def set_flag(cls):
        cls.numba_flag = cls.numba_check()

    @classmethod
    def disable_numba(cls):
        """To disable numba."""
        cls.numba_flag = False

    @classmethod
    def enable_numba(cls):
        """To enable numba."""
        if cls.numba_check():
            cls.numba_flag = True
        else:
            raise ValueError("Numba is not installed")

But this would lead extra code and tests which would defeat the purpose of including `numba_check` inside the class. 
I guess leaving `numba_check` outside the class should be fine.

Ban-zee:
What happened to coverage?

ahartikainen:
It's 96.72%

https://coveralls.io/builds/24205010

I have noticed that the following number is the only number to trust on that page.

image

arviz/utils.py
canyon289: Ah I see. Alright no worries was just curious

arviz/tests/test_utils.py
```diff
@@ -134,3 +151,83 @@ def placeholder_func():
...
+    Numba.disable_numba()
+    val = Numba.numba_flag
+    assert not val
+    Numba.enable_numba()
+    val = Numba.numba_flag
+    assert val
+    assert flag == Numba.numba_flag
+
+
+@pytest.mark.parametrize("axis", (0, 1))
```
canyon289: @Ban-zee As I get time Im going to keep picking with comments :)

Can we move numba disable to a parametrization? That way if the numba fails it's easier to tell because the tests are separated?

canyon289:
It's my opinion that we should merge this sooner rather than later, before there's any merge conflicts.

@Ban-zee
Not required for this PR but I want to ask here

I may have missed this but is there documentation on how to run the speed tests?
Is it possible to add speed tests to ci?

arviz/tests/test_utils.py
Ban-zee: Sure

Ban-zee:
How about if I add this to contributing: 
To run the **benchmark steps** do the following:

- Install airspeed velocity via `pip install asv`

- `cd arviz/`

- `asv run`

CONTRIBUTING.md
```diff
@@ -208,4 +208,14 @@ Alternatively, to start a jupyter notebook, there are two steps, first run:
...
 
+## Running the benchmark tests
+
+To run the **benchmark tests** do the following:
+
+- Install airspeed velocity via `pip install asv`
+
+- `cd arviz/`
+
+- `asv run`
```
OriolAbril: what about?

    $ cd arviz/
    $ asv run

minor comment, it is only to follow the convention from the rest of the contributing file.

arviz/stats/diagnostics.py
```diff
@@ -453,6 +448,12 @@ def mcse(data, *, var_names=None, method="mean", prob=None):
     )
 
 
+@conditional_vect
+def _sqrt(a_a, b_b):
+    return (a_a + b_b) ** 0.5
+
+
+@conditional_jit
 def geweke(ary, first=0.1, last=0.5, intervals=20):
```
OriolAbril: I have no idea why nor how to solve it, I point it out just in case. If you build the docs locally, you will see that the `geweke` link on the api page is not working anymore. It shows the text, but it does not link to the `geweke` page.

doc/index.rst
```diff
@@ -66,6 +66,7 @@ Here is the citation in BibTeX format
     Quickstart<notebooks/Introduction>
     Example Gallery<examples/index>
     Cookbook<notebooks/InferenceDataCookbook>
+    Numba<notebooks/Numba>
```
OriolAbril: Completely unrelated. Should this toctree have the same elements as the sticky nav-bar on top of the website? 

If so this seems like a good ocasion to add `InferenceData<notebooks/XarrayforArviZ>` here and `("Numba", "notebooks/Numba"),` between lines 137-138 of `doc/conf.py`. Adding numba to the nav-bar is probably more important, otherwise it can be hard to find.

arviz/stats/diagnostics.py
canyon289: Oh I think we might need to do a functools wrap. Decorators "eat" the docstrings of their decorated functions

https://www.thecodeship.com/patterns/guide-to-python-function-decorators/

canyon289: I think we should do this on this PR because we don't want to regress the docs for every function that we use `conditional_jit` for

CONTRIBUTING.md
Ban-zee: Sure, and I'll change ELPD data too

Ban-zee:
Strange, the tests work fine.

ahartikainen:
````
with pytest.raises(ValueError) as err:
    get_coords(data, coords)
    
>       assert "Coords {'NOT_A_COORD_NAME'} are invalid coordinate keys" in str(err)

arviz/tests/test_plot_utils.py:112: AssertionError
````

Is says AssertationError is not ValueError. Did pytest update?

Ban-zee:
Sorry for the delay :)

PR #714, opened 2019-06-20
CI broken again

OriolAbril:
tfp 0.7 has just been released, so now tensorflow 1.14 is a requirement


ahartikainen:
Yes, I was waiting this to happen :)

Issue #336, opened 2018-10-08
Add ddof=1 to summary


ahartikainen:
Maybe after https://github.com/pydata/xarray/pull/2447 is in released version.

ahartikainen:
This should be fixed now.

PR #716, opened 2019-06-21
ddof=1 for std

ahartikainen:
Fixes https://github.com/arviz-devs/arviz/issues/336

Minimal changes. I think we can assume that users should use updated xarray?


canyon289:
Yea let's assume they do, and if they don't we can add a warning later

PR #718, opened 2019-06-22
Centralize number formatting.

ColCarroll:
Uses the code from #705, but re-simplifies the API to have only the single formatting option. Right now formatting will be the integer part or `round_to`, whichever is bigger.

@rpgoldman tagging you in case you have opinions.

```python
az.plot_posterior(data, ['mu', 'tau']);
```
image

```python
az.plot_posterior(data, ['mu', 'tau'], round_to=0);
```
image
```python
az.plot_posterior(data, ['mu', 'tau'], round_to=5);
```
image

### DIFF
--- a/arviz/plots/posteriorplot.py
+++ b/arviz/plots/posteriorplot.py
@@ -304,7 +304,9 @@ def _plot_posterior_op(
         elif point_estimate == "median":
             point_value = np.median(values)
         sig_figs = significant_fig_func(point_value)
-        point_text = f"{point_estimate}={point_value:.{sig_figs}g}"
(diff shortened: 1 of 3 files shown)


ahartikainen:
LGTM

ColCarroll:
i'm going to say this was fast enough that we don't need to deprecate `significant_figs` (apologies and kudos to anyone who built a system on the master branch yesterday).

rpgoldman:
The one issue I could see with this is the use of the significant figures to format the _percentage_ range used in the HPD.

I think that this is a very different kind of number from the values of the variable in question, and so should probably just be printed in an unvarying format (I bet just taking the integer part of the percentage would handle 99% of the cases).

PR #724, opened 2019-06-30
Quantile mcse plot [GSoC]

### LINKED ISSUE #621: New R-hat, ESS and quantile-MCSE functions + related plots + rank plots

OriolAbril:
Add quantile MCSE plots to ArviZ and extend plot_ess functionalities. Closes #621. 

### `plot_mcse`
Add `plot_mcse` function, with generates 2 main types of plot: quantile vs quantile MCSE (default) or quantile vs quantile_value +/- quantile MCSE (`errorbar=True`). `divergences` (or other boolean variables like `max_depth` in the paper) locations can be shown as rug plot at the bottom. The default plot also accepts the option `extra_methods` which plots two horizontal lines at `mcse_mean` and `mcse_sd` locations. These lines are labeled inline, via `ax.annotate`, and the position can be customized (the behaviour in plot_ess is the same, see examples there). Below there are some examples:

##### default plot
image
##### default plot with rug and extra_methods
image
##### errorbar plot
image

### `plot_ess`
Added `extra_methods` option (as well as its customization args). It is only compatible with kind local or quantile, in evolution case it is ignored. Below there are some examples of customization of inline labels of extra methods lines:

##### set label outside axes
```python
az.plot_ess(idata, var_names=["mu", "tau"], kind="local", 
                   extra_methods=True, text_kwargs={"ha": "left"})
```
image
##### use dashed lines and annotate labels with arrows
```python
text_kwargs = {"x": 0.1, "xytext": (-5,15), "textcoords": "offset points", 
               "arrowprops": {"arrowstyle": "->"}, "va": "bottom"}
az.plot_ess(idata, var_names=["mu", "theta"], kind="local", coords={"school": ["Choate"]},
            extra_methods=True, extra_kwargs={"ls": ":"}, text_kwargs=text_kwargs)
```
image
Note: I do not think this case will be used much, it is only to highlight it is a possibility.


ColCarroll:
Can you post an example of the plot?

ahartikainen:
I have not yet read the source, but could this plot also include

    quantile_value +/- mcse_q

? 

It would probably look like a transposed s-curve?

OriolAbril:
Currently, the plots look like this for mcse plot and rug values:

image

And for "errorbar" option:

image

ahartikainen:
This is not needed: should we add option to include mcse_mean and mcse_sd as horizontal lines? (Same could go for ess)

OriolAbril:
> This is not needed: should we add option to include mcse_mean and mcse_sd as horizontal lines? (Same could go for ess)

@ahartikainen I have half done these horizontal lines, but I am not sure on whether I should allow personalization for them or not. Two kwargs dict seem to much, as there are various kwargs dicts involved. For ess, one would work because the two values are either equal or sd method yields a lower value, but for mcse I am not sure that sd>mean holds. I don't know either if the difference between ess sd and ess mean can be assumed as known. 

For now, there is no room for customization in any of the two, here is how it looks:

image

ahartikainen:
Could we "inline" names for those extra lines (outside ax, on the right)? they could also be much dimmer (like in background) by default.

OriolAbril:
Modified the default behaviour of the extra methods:

image

How does this look?

ahartikainen:
I like it (is there an option to move those labels outside axis?). 

Are those diverging -locations calculated with quantile function?

How do I transform from quantile values to real values? (0-1 --> min_x, max_x)?

OriolAbril:
> I like it (is there an option to move those labels outside axis?).

Not yet, I am working on it.

> Are those diverging -locations calculated with quantile function?

I take their rank value (with all chains and draws combined) divided by the number of samples. 

> How do I transform from quantile values to real values? (0-1 --> min_x, max_x)?

You would need the cdf of the parameter. 

I can add an option to use parameter values as x values, and thus, use non-uniform "bins". There is one example of each in the [online appendix](https://avehtari.github.io/rhat_ess/rhat_ess.html#22_hierarchical_model:_eight_schools) of Aki's paper (the link goes directly to section 2.2, you should not have to scroll down too much).

OriolAbril:
Ready for review. I think we can leave the option to use parameter values as x axis for another PR.

arviz/plots/essplot.py
```diff
@@ -26,12 +26,14 @@ def plot_ess(
...
     n_points=20,
+    extra_methods=False,
     min_ess=400,
     ax=None,
     extra_kwargs=None,
+    text_kwargs=None,
     hline_kwargs=None,
     rug_kwargs=None,
-    **kwargs
+    **kwargs,
```
canyon289: Is it just the trailing comma that is adding this error?

OriolAbril: Now the error seems due to a missing trailing comma. Any idea on when an extra comma becomes a syntax error and when the syntax error is actually due to the extra comma missing? Are there any docs on this? Is it a black issue?

ahartikainen: https://github.com/python/black/issues/759

> Looks like we improperly merged two different syntax changes: `f(**x,)` (trailing comma in a call) first became legal in 3.5, but `def f(**x,): pass` (trailing comma in a definition) became legal only in 3.6. I'll try to disentangle the two in the code.

OriolAbril: I guess we have to wait until this fix is released then.

ahartikainen:
LGTM

@OriolAbril do you want to resolve the conflicts, after that PR can be merged.

OriolAbril:
I will get right to it.

PR #725, opened 2019-07-04
Fix some sphinx warnings

ColCarroll:
Oddly, this gets all the links on the API working for me.

`hpdplot` had a zero-width space I deleted!


arviz/data/io_emcee.py
```diff
@@ -327,7 +327,7 @@ def from_emcee(
     .. plot::
         :context: close-figs
 
-        >>> az.plot_trace(emcee_data, var_names=["var_2"], coords={"var_2_dim_0": 4})
+        >>> az.plot_trace(emcee_data, var_names=["eta0"])
```
OriolAbril: This is actually creating a new warning and making the 2nd image of the three in `from_emcee` to not be generated. Which warning were you trying to avoid?

From the Travis-doc build:

    /opt/arviz/arviz/data/io_emcee.py:docstring of arviz.from_emcee:126: WARNING: Exception occurred in plotting arviz-from_emcee-5

And then when copying images:

```
copying images... [ 29%] examples/_images/styles.png
copying images... [ 30%] generated/../build/plot_directive/generated/arviz-from_emcee-3.png
copying images... [ 31%] generated/../build/plot_directive/generated/arviz-from_emcee-9.png
copying images... [ 32%] generated/../build/plot_directive/generated/arviz-plot_autocorr-1.png
```

There should be [3 images copied](https://arviz-devs.github.io/arviz/generated/arviz.from_emcee.html#arviz.from_emcee), `arviz-from_emcee-5.png` is missing.

ColCarroll: i wonder if this is due to caching? but docker should take care of that...

OriolAbril: Could be, have you tried removing `build`, `generated` and `examples` folder from inside `doc` before generating the docs?

`eta0` is a variable name in the [first example](https://github.com/arviz-devs/arviz/blob/0779bad9350a9fde6edb74b02b1e4bf9734bd17e/arviz/data/io_emcee.py#L302), but in the [second example](https://github.com/arviz-devs/arviz/blob/0779bad9350a9fde6edb74b02b1e4bf9734bd17e/arviz/data/io_emcee.py#L322), no var names are passed, so ArviZ default names for variables and dims are used. `eta0` should not exist, and `var_2` should.

ColCarroll: Yeesh -- I'd been getting an exception building docs locally saying that this variable does not exist. I'll change this back to just adjust underline lengths, and play around with the cache on travis...

PR #727, opened 2019-07-05
Khat plot: hover tweaking [GSoC]

OriolAbril:
### `plot_khat`
* [x] Allow customization of hover labels
* [x] Add example to gallery

### Context manager
* [x] Create a class in arviz to manage an interactive backend context:
  ```python
  plt.plot(...) # inline
  with interactive_backend():
      plt.plot(...) # interactive
  plt.plot(...) # inine
  ```
  For now, a prototype of this class is present in ArviZ home. It seems to work, I will be on the lookout for issues. 

  I decided not to use CamelCase even though it is implemented as a class because most context managers are in lowercase. One example is `matplotlib.rc_context` itself. `interactive_backend` is based on `rc_context` and in addition, `rc_context` is the reason why it is defined as a class and not with `contextlib`.

### Documentation
* [x] Modify layout of class doc pages on the website. The first image is the current website, the bottom image is the proposed layout.
image
image

### `plot_elpd`
* Add hover labels to `plot_elpd`? If eventually added, they will be added in a following PR.

### DIFF
--- a/doc/_templates/class.rst
+++ b/doc/_templates/class.rst
@@ -1,0 +1,30 @@
+{{ fullname | escape | underline}}
+
+.. currentmodule:: {{ module }}
+
+.. autoclass:: {{ objname }}
+
+   {% block methods %}
+   {% if methods %}
+
+   .. rubric:: Methods
+
+   .. autosummary::
+   {% for item in methods %}
+      {% if item != "__init__" %}
+         {{ item }}
+      {% endif %}
+   {%- endfor %}
+   {% endif %}
+   {% endblock %}
+
+   {% block attributes %}
+   {% if attributes %}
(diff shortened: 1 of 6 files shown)


doc/_templates/class.rst
```diff
@@ -0,0 +1,30 @@
+{{ fullname | escape | underline}}
+
+.. currentmodule:: {{ module }}
+
+.. autoclass:: {{ objname }}
```
Ban-zee: This looks good.

ahartikainen:
Was this ready?

LGTM

OriolAbril:
Yes, it is ready! I will solve conflicts after plot_mcse PR is merged and everything will be set.

PR #728, opened 2019-07-06
Update conditional_jit wrapper

ColCarroll:
This looks a little complicated, but my goals were:

1. Docs should appear on decorated functions
2. Toggling `az.Numba.enable_numba()` and `az.Numba.disable_numba()` should work
3. Limit overhead when using numba (since that's the point)

The first is done with `functools.wraps`
The second is done in `__call__` in `maybe_numba_fn`
The third is done using a lazy property: it compiles the first time it is called, and then is an attribute lookup. This might be overkill, I feel least strongly about that. The alternative would evaluate something like `if self._numba_fn is None:` on every function call, which is not a ton of overhead.

There is one failing test with a lot of monkeypatching!

image

### DIFF
--- a/arviz/utils.py
+++ b/arviz/utils.py
@@ -1,5 +1,6 @@
 """General utilities."""
 import importlib
+import functools
 import warnings
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


ahartikainen:
I think you need to change that test and mock the maybe function? Or then change the test type?

arviz/utils.py
```diff
@@ -50,7 +51,53 @@ def _var_names(var_names, data):
     return var_names
 
 
-def conditional_jit(function=None, **kwargs):  # noqa: D202
+class lazy_property:  # pylint: disable=invalid-name
```
ahartikainen: Is this function really needed? Python does this automatically, right?

Ban-zee:
We can use `numba_flag` in `conditional_jit`. We'll avoid monkeypatching then.

arviz/utils.py
canyon289: When you say function @ahartikainen what do you mean? This comment is on a class definition?

ahartikainen: haha, true

I meant that import numba stuff, but now I read it again, this is loading the wrapper once?

ColCarroll: Yeah -- if you import arviz and immediately disable numba, this will never even try to import it.

ColCarroll: There's a tradeoff the first time you call the function, though: 
- the first numba function used globally will import numba
- the first time each numba-fied function runs, it will get jit-ted first

Note in my screenshot you can see this performance penalty the first time `geweke` is called.

Ban-zee:
To eliminate the warnings use `conditional_vect(forceobj=True)`.

ColCarroll:
Should be all set.

ahartikainen:
LGTM

PR #731, opened 2019-07-08
No observations in MultiObservedRV

ahartikainen:
Fixes the problem, when MultiObservedRV does not contain observations (DensityDist)

See https://discourse.pymc.io/t/pm-traceplot-error/3524

### DIFF
--- a/arviz/tests/test_data_pymc.py
+++ b/arviz/tests/test_data_pymc.py
@@ -120,8 +120,8 @@ class TestDataPyMC3:
 
     def test_multiple_observed_rv_without_observations(self):
         with pm.Model():
-            mu = pm.Normal('mu')
-            x = pm.DensityDist('x', pm.Normal.dist(mu, 1.).logp, observed={'value': .1})
+            mu = pm.Normal("mu")
+            x = pm.DensityDist("x", pm.Normal.dist(mu, 1.0).logp, observed={"value": 0.1})
             trace = pm.sample(100, chains=2)
         inference_data = from_pymc3(trace=trace)
         assert inference_data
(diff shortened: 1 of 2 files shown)


ahartikainen:
If anyone can make a simple model to reproduce this error, then we can add it to the tests. Currently the example is too complicated to be included in tests, and I don't know enough of pymc3 to make it simple.

ColCarroll:
I'll try to get to one tonight.

junpenglao:
This should do:
```python
import pymc3 as pm
with pm.Model():
    mu = pm.Normal('mu')
    x = pm.DensityDist('x', pm.Normal.dist(mu, 1.).logp, observed={'value': .1})
    trace = pm.sample()
pm.traceplot(trace)
```

arviz/tests/test_data_pymc.py
```diff
@@ -118,6 +118,19 @@ def test_multiple_observed_rv(self):
         assert not fails
         assert not hasattr(inference_data.sample_stats, "log_likelihood")
 
+    def test_multiple_observed_rv_without_observations(self):
+        with pm.Model():
+            mu = pm.Normal("mu")
+            x = pm.DensityDist(  # pylint disable=unused-variable
```
OriolAbril: The colon is missing: `pylint: disable=unused-variable`

OriolAbril:
Also, this shows that the lint in azure is not equal to the lint in travis yet, ideas @canyon289?

arviz/tests/test_data_pymc.py
ColCarroll:
```suggestion
            x = pm.DensityDist(  # pylint: disable=unused-variable
```

ahartikainen:
Thanks! LGTM as far as I'm concerned

PR #734, opened 2019-07-10
[WIP] Set up ArviZ rcParams

### LINKED ISSUE #609: Restrict the max number of plots by default

### LINKED ISSUE #706: ArviZ global level options
## Feature Description

There are many options in ArviZ which would be easier to handle as global options. Moreover, having global options in a style similar to matplotlib or pandas can make the experience of using ArviZ easier and better tailored to the end users. These options go from the maximum number of plots (#609) to text formatting options or even the default information criterion.

I believe that this can end up containing many different parameters, so with this in mind, environment variables are probably not ideal on the long run. Having something similar to matplotlib rcParams or pandas options seems more adient.

### pandas-like

The first implementation option is to have some pandas-like API. That is, setting the variables from some `config.py` file (a priori not user friendly) and expose some functions to modify these options.

### matplotlib-like

This second option is similar to the first one, but options are read from a json or json like configuration file, which is though to be used and modified by the end user. In addition to this, there are some functions exposed to modify these options too.

---

I prefer the matplotlib like version, because I think it is more convenient to ArviZ (i.e. someone who prefers LOO over WAIC can modify the rcparams file and forget about it instead of calling the option modifier function at the start of every script or python/ipython startup scripts). 

I propose to start with a simple json file (preferably 2 level) as configuration file and a custom class (children of dict for example) where all the options would be stored once read from the file for easy access.

OriolAbril:
Set ArviZ global options as rcParams, following the behaviour of matplotlib's rcParams. The idea is to have a slightly simplified version of matplotlib's implementation. For now and as an example, only 3 parameters are present. This PR itself can be used to propose and discuss other possible global parameters.

Closes #609 and closes #706.

Below I will write (and keep updating the list) to check that all relevant functions and docs have been updated:
* [x] `data.load`
* [ ] `plot.max_subplots`
* [x] `stats.inference_criterion` 

rcParams modification and loading from different folders and files should already work. However, all of it is still preliminary. The folders do not make much sense yet, the parameters can be modified in `az.rcParams` but the functions have not been updated yet either... Please review and comment.

### DIFF
--- a/arviz/__init__.py
+++ b/arviz/__init__.py
@@ -23,7 +23,7 @@ if not logging.root.handlers:
     _log.setLevel(logging.INFO)
     _log.addHandler(handler)
 
-from .rcparams import rcParams
+from .rcparams import rcParams, rc_context
 from .data import *
 from .plots import *
 from .stats import *
--- a/arviz/__init__.py
+++ b/arviz/__init__.py
@@ -23,6 +23,7 @@ if not logging.root.handlers:
     _log.setLevel(logging.INFO)
     _log.addHandler(handler)
 
+from .rcparams import rcParams
 from .data import *
 from .plots import *
 from .stats import *
--- a/arviz/rcparams.py
+++ b/arviz/rcparams.py
@@ -10,9 +10,8 @@ import locale
 _log = logging.getLogger(__name__)
 
 
-def make_validate_choice(accepted_values):
+def _make_validate_choice(accepted_values):
     """Validate value is in accepted_values."""
-
     def validate_choice(value):
         if value.lower() in accepted_values:
             return value.lower()
@@ -21,7 +20,7 @@ def make_validate_choice(accepted_values):
     return validate_choice
 
 
-def validate_positive_int(value):
+def _validate_positive_int(value):
     """Validate value is a natural number."""
     try:
         value = int(value)
@@ -33,18 +32,18 @@ def validate_positive_int(value):
         raise ValueError("Only positive values are valid")
 
 
-def validate_positive_int_or_none(value):
+def _validate_positive_int_or_none(value):
     """Validate value is a natural number or None."""
     if value is None or isinstance(value, str) and value.lower() == "none":
         return None
     else:
-        return validate_positive_int(value)
+        return _validate_positive_int(value)
 
 
 defaultParams = {
-    "data.load": ("lazy", make_validate_choice(("lazy", "eager"))),
-    "plot.max_subplots": (40, validate_positive_int_or_none),
-    "stats.information_criterion": ("waic", make_validate_choice(("waic", "loo"))),
+    "data.load": ("lazy", _make_validate_choice(("lazy", "eager"))),
+    "plot.max_subplots": (40, _validate_positive_int_or_none),
+    "stats.information_criterion": ("waic", _make_validate_choice(("waic", "loo"))),
 }
 
 
@@ -61,6 +60,7 @@ class RcParams(dict):
         self.update(*args, **kwargs)
 
     def __setitem__(self, key, val):
+        """Add validation to __setitem__ function."""
         try:
             try:
                 cval = self.validate[key](val)
@@ -74,6 +74,7 @@ class RcParams(dict):
             )
 
     def __repr__(self):
+        """Customize repr of RcParams objects."""
         class_name = self.__class__.__name__
         indent = len(class_name) + 1
         repr_split = pprint.pformat(dict(self), indent=1, width=80 - indent).split("\n")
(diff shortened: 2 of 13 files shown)


OriolAbril:
Candidates of possible global parameters:

* `data.stan_log_likelihood`? It would be `None` by default but would allow each user to set one value there. I am guessing most people will probably be quite methodic and always name the log likelyhood the same (i.e. `log_lik`), and would be used in `from_pystan`, `cmdstan` and `cmdpystan`. @ahartikainen 
* `stats.credible_interval`? Set the default credible interval, defaulting to 0.94.
* `stats.pointwise_elpd`? Default to false, but allow to set it to true in order to get pointwise elpd diagnostics by default (it should not add much overhead actually).
* `stats.ic_scale`? `

azure-pipelines.yml
```diff
@@ -51,26 +51,27 @@ jobs:
...
   - script: |
-      if [[ $NAME == LINT ]]; 
+      if [[ $NAME == LINT ]];
       then docker run -t arviz:latest bash -c "./scripts/lint.sh";
         fi
-        
+
+      if [ $? -ne 0 ]; then
+        exit 1
+        fi
```
OriolAbril: @canyon289 I think moving it here will make both lint jobs act the same. At least here they are doing the same. Does it also make sense to you? I will move it to a different PR after some more experiments.

ahartikainen:
LGTM

ahartikainen:
I merged this so we have the basic structure in-place and let's create new PRs to implement more options.

arviz/rcparams.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+defaultParams = {  # pylint: disable=invalid-name
+    "data.load": ("lazy", _make_validate_choice(("lazy", "eager"))),
+    "plot.max_subplots": (40, _validate_positive_int_or_none),
+    "stats.information_criterion": ("waic", _make_validate_choice(("waic", "loo"))),
+}
+
+
+class RcParams(MutableMapping, dict):  # pylint: disable=too-many-ancestors
```
ColCarroll: TIL `MutableMapping`

arviz/rcparams.py
```diff
@@ -0,0 +1,266 @@
...
+        - ``$XDG_CONFIG_HOME/arviz/arvizrc`` (if ``$XDG_CONFIG_HOME``
+          is defined)
+        - or ``$HOME/.config/arviz/arvizrc`` (if ``$XDG_CONFIG_HOME``
+          is not defined)
+    - On other platforms,
+        - ``$HOME/.arviz/arvizrc`` if ``$HOME`` is defined
+
+    Otherwise, the default defined in ``rcparams.py`` file will be used.
+    """
+    # no blank lines allowed after function docstring by pydocstyle,
```
ColCarroll: :joy_cat:

arviz/rcparams.py
```diff
@@ -0,0 +1,266 @@
...
+    # but black requires white line before function
+
+    def gen_candidates():
+        yield os.path.join(os.getcwd(), "arvizrc")
+        arviz_data_dir = os.environ.get("ARVIZ_DATA")
+        if arviz_data_dir:
+            yield os.path.join(arviz_data_dir, "arvizrc")
+        xdg_base = os.environ.get("XDG_CONFIG_HOME", str(Path.home() / ".config"))
```
ColCarroll: I super wish everyone used `$XDG_CONFIG_HOME`

arviz/rcparams.py
```diff
@@ -0,0 +1,266 @@
...
+        if sys.platform.startswith(("linux", "freebsd")):
+            configdir = str(Path(xdg_base, "arviz"))
+        else:
+            configdir = str(Path.home() / ".arviz")
+        yield os.path.join(configdir, "arvizrc")
+
+    for fname in gen_candidates():
```
ColCarroll: i'm a `pathlib` convert since yesterday, but it might be nice for `gen_candidates` to yield `Path` objects so you don't have to cast back and forth, and you can just call `if fname.exists() and not fname.isdir():  return str(fname)`

arviz/rcparams.py
```diff
@@ -0,0 +1,266 @@
...
+def read_rcfile(fname):
+    """Return :class:`arviz.RcParams` from the contents of the given file.
+
+    Unlike `rc_params_from_file`, the configuration class only contains the
+    parameters specified in the file (i.e. default values are not filled in).
+    """
+    _error_details_fmt = 'line #%d\n\t"%s"\n\tin file "%s"'
+
+    config = RcParams()
+    with open(fname, "r") as rcfile:
```
ColCarroll: We may want to just use `pyyaml` or similar here so that we aren't accidentally responsible for maintaining a configuration file spec!

arviz/__init__.py
```diff
@@ -23,6 +23,7 @@
     _log.setLevel(logging.INFO)
     _log.addHandler(handler)
 
+from .rcparams import rcParams, rc_context
```
ColCarroll: `get_arviz_rcfile` might be nice here, too

PR #736, opened 2019-07-11
Fix azure lint job

OriolAbril:
I will inlcude 4 commits in this PR. The first one is the modifications to azure-pipelines.yml, which should pass all checks. The second will have a doc lint error, the third a black error and the third a pylint error. If all 3 fail in the same way as travis I will undo the lint errors to merge the PR.

### DIFF
--- a/arviz/utils.py
+++ b/arviz/utils.py
@@ -2,7 +2,6 @@
 import importlib
 import warnings
 import numpy as np
-import re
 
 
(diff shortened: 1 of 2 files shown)


OriolAbril:
Seems to work properly now. It should be ready to merge.

ColCarroll:
Thanks! Will check it out.

PR #739, opened 2019-07-12
Fix #666

### LINKED ISSUE #666: Add method to delete groups from InferenceData
## Tell us about it
Add a method to delete groups from InferenceData

With many groups InferenceData can get quite large, making it hard to add to VCS or email. Things like sample_stats might not be needed for all results sharing. 

## Thoughts on implementation
1. Delete the object from `az.InferenceData` with `del` method in python
2. Remove string in `az.InferenceData._groups`

OriolAbril:
Add `del` method to inference data and add `groups` argument to `InferenceData.to_netcdf`. Fixes #666

### DIFF
--- a/arviz/tests/test_data.py
+++ b/arviz/tests/test_data.py
@@ -318,6 +318,33 @@ def test_sel_method(inplace):
             assert np.all(dataset.draw.values == np.arange(200, ndraws))
 
 
+@pytest.mark.parametrize("use", ("del", "delattr"))
+def test_del_method(use):
+    data = np.random.normal(size=(4, 500, 8))
+    idata = from_dict(
+        posterior={"a": data[..., 0], "b": data},
+        sample_stats={"a": data[..., 0], "b": data},
+        observed_data={"b": data[0, 0, :]},
+        posterior_predictive={"a": data[..., 0], "b": data},
+    )
+    test_dict = {
+        "posterior": ("a", "b"),
+        "sample_stats": ("a", "b"),
+        "observed_data": ["b"],
+        "posterior_predictive": ("a", "b"),
+    }
+    fails = check_multiple_attrs(test_dict, idata)
+    assert not fails
+    if use == "del":
+        del idata.sample_stats
+    else:
+        delattr(idata, "sample_stats")
+    test_dict.pop("sample_stats")
+    fails = check_multiple_attrs(test_dict, idata)
+    assert not fails
+    assert not hasattr(idata, "sample_stats")
+
+
 class TestNumpyToDataArray:
     def test_1d_dataset(self):
         size = 100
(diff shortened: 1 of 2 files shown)


canyon289:
Looks good. Just could use a couple of tests

OriolAbril:
> Looks good. Just could use a couple of tests

Yes! I just wanted to make sure the implementation was ok before. I'll add them tomorrow.

arviz/tests/test_data.py
```diff
@@ -318,6 +318,33 @@ def test_sel_method(inplace):
...
+    )
+    test_dict = {
+        "posterior": ("a", "b"),
+        "sample_stats": ("a", "b"),
+        "observed_data": ["b"],
+        "posterior_predictive": ("a", "b"),
+    }
+    fails = check_multiple_attrs(test_dict, idata)
+    assert not fails
+    if use == "del":
```
canyon289: This is very nitpicky, some whitespace between "logical blocks" would make this easier to follow

arviz/tests/test_data.py
```diff
@@ -631,7 +658,15 @@ def test_io_function(self, data, eight_schools_params):
         inference_data = self.get_inference_data(  # pylint: disable=W0612
             data, eight_schools_params
         )
-        assert hasattr(inference_data, "posterior")
+        groups = inference_data._groups  # pylint: disable=protected-access
```
canyon289: Comment on what this code block is doing, in particular why there's an if else would be helpful

arviz/tests/test_data.py
```diff
@@ -640,25 +675,41 @@ def test_io_function(self, data, eight_schools_params):
...
         assert not os.path.exists(filepath)
 
-    def test_io_method(self, data, eight_schools_params):
+    @pytest.mark.parametrize("groups_arg", [False, True])
+    def test_io_method(self, data, eight_schools_params, groups_arg):
         inference_data = self.get_inference_data(  # pylint: disable=W0612
             data, eight_schools_params
         )
-        assert hasattr(inference_data, "posterior")
+        if groups_arg:
```
canyon289: Same thing here this test is getting pretty big actually. is there a way to separate the new assert into a new test?

arviz/tests/test_data.py
OriolAbril: I can also write all test dict by hand if it is more clear, I will only have to write it one, so it is not too much work.

arviz/tests/test_data.py
OriolAbril: Yes, I tried to put them toghether because they are both testing the same function with one different argument. They can be divided without much trouble between `test_io_method` and `test_io_method_groups`

OriolAbril: How does it look now? I tried to minimize the parts where using a `groups` list and `None` diverged, because separating them was repeating a lot of equal code.

OriolAbril:
If tests are clear enough, it should be ready to merge.

Issue #740, opened 2019-07-12
az.compare fails for pymc3 mixture models


canyon289:
**Describe the bug**
<img width="840" alt="image" src="https://user-images.githubusercontent.com/7213793/61141152-c5eeaa80-a481-11e9-87f3-03eedfc23b78.png">

I believe it's because of the extra coord in ic
<img width="587" alt="image" src="https://user-images.githubusercontent.com/7213793/61141476-8aa0ab80-a482-11e9-8ad5-6441f7d514e9.png">
**To Reproduce**
```
import arviz as az
import pymc3 as pm
import numpy as np

vals=np.array([1,2,3])
with pm.Model() as m:
    mu = pm.Normal("mu", mu=0, sd=1)
    y = pm.Normal("y", mu=mu, sd=1, observed=vals)
    trace=pm.sample(draws=100, tune=100)

data = az.from_pymc3(trace=trace)
comp = az.compare({"trace": data, "trace_2":data})
print(comp)

# Mixture model
with pm.Model() as two_components:
    p = pm.Dirichlet("p", a=np.ones(2))

    # Each mean of the mixture data has its own estimate of a mean with a fixed SD in this case
    means = pm.Normal("means", mu=vals.mean(), sd=10, shape=2)

    # Estimate of the standard deviation of what?
    sd = pm.HalfCauchy("sd", 1)
    y = pm.NormalMixture("y", w=p, mu=means, sd=sd, observed=vals)
    trace_mixture = pm.sample(draws=10, tune=10)

data_mixture = az.from_pymc3(trace=trace_mixture)

# Failing line
comp = az.compare({"trace": data_mixture, "trace_2":data_mixture})

```

**Expected behavior**
Comparison dataframe generated

**Additional context**
Appear on 0.4.1

OriolAbril:
@canyon289 I was thinking today, and do we really want `from_xyz` methods to return datasets with length 1 dimensions? We could add a squeeze in `from_pymc3` before converting to dataset.

PR #741, opened 2019-07-12
Bugfix for  comparison exception

canyon289:
@ahartikainen Is this the correct fix for the issue below? Shape problems get me everytime

https://github.com/arviz-devs/arviz/issues/740

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -275,7 +275,7 @@ def _ic_matrix(ics, ic_i):
         if len(ic) != rows:
             raise ValueError("The number of observations should be the same across all models")
 
-        ic_i_val[:, idx] = ic
+        ic_i_val[:, idx] = ic.squeeze()
 
     return rows, cols, ic_i_val
 


OriolAbril:
@canyon289 I will take a look at it. It looks like it is probably an issue I created when modifying the `stats.py` functions to use xarray. I say so because in addition to return DataArrays instead of ndarrays, the functions are applied via apply ufunc, instead of reshaping to 2D objects `(nsamples, nobservations)` (like originally) only chain and draws are stacked `(*shape, nsamples)`.

I think that the squeeze will not probably solve the issue, because the problem is likely to be with multidimensional objects (which in general cannot be converted to 1D objects via squeeze). I may not have time until tomorrow though.

canyon289:
No rush! Was just curious. Thanks for helping me with this.

Squeeze solved my specific example but I don't know if its generalizeable

OriolAbril:
I have just checked the behaviour on a [2d model](https://github.com/OriolAbril/Bayesian-Inference-examples/tree/master/Binomial_on_2D_data) and got the following error:

    ValueError: could not broadcast input array from shape (13,8) into shape (13)

In this case the problem cannot be solved with a squeeze. I will send a PR with a possible fix in some minutes, and in the following weeks I will try to test all stats functions on 1D and 2D data (like it is done with plots) in order to catch these issues in CI testing.

canyon289:
Closing in favor of https://github.com/arviz-devs/arviz/pull/748

PR #742, opened 2019-07-12
Numba-plots

Ban-zee:
This optimizes `_fast_kde_2d`. `_fast_kde` was already optimized via the `conditional_jit` on `np.histogram` which greatly reduced my work with plots. This should optimize a lot of other plots as many of them are built on top of `plot_kde`. A few minor additions on the remaining plots and plot_utils need to be added. A caveat though, in most of the cases of plots, the bottlenecks are matplotlib plotting methods which are not compatible with numba `no-python` mode. I am however, trying my best to optimize the plots as much as possible :).

### DIFF
--- a/arviz/plots/essplot.py
+++ b/arviz/plots/essplot.py
@@ -334,7 +334,7 @@ def plot_ess(
             ylabel.format("Relative ESS" if relative else "ESS"), fontsize=ax_labelsize, wrap=True
         )
         if kind == "evolution":
-            ax_.legend(title="Method", fontsize=xt_labelsize)
+            ax_.legend(title="Method", fontsize=xt_labelsize, title_fontsize=xt_labelsize)
         else:
             ax_.set_xlim(0, 1)
         if rug:
--- a/arviz/plots/essplot.py
+++ b/arviz/plots/essplot.py
@@ -334,7 +334,7 @@ def plot_ess(
             ylabel.format("Relative ESS" if relative else "ESS"), fontsize=ax_labelsize, wrap=True
         )
         if kind == "evolution":
-            ax_.legend(title="Method", title_fontsize=xt_labelsize, fontsize=xt_labelsize)
+            ax_.legend(title="Method", fontsize=xt_labelsize)
         else:
             ax_.set_xlim(0, 1)
         if rug:
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -291,7 +291,7 @@ class PlotHandler:
         """Collect labels and ticks from plotters."""
         val = self.plotters.values()
 
-        @conditional_jit
+        @conditional_jit(forceobj=True)
         def label_idxs():
             labels, idxs = [], []
             for plotter in val:
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -298,10 +298,9 @@ class PlotHandler:
                 sub_labels, sub_idxs, _, _ = plotter.labels_ticks_and_vals()
                 labels.append(sub_labels)
                 idxs.append(sub_idxs)
-            return labels, idxs
+            return np.concatenate(labels), np.concatenate(idxs)
 
-        labels, idxs = label_idxs()
-        return np.concatenate(labels), np.concatenate(idxs)
+        return label_idxs()
 
     def display_multiple_ropes(self, rope, ax, y, linewidth, rope_var):
         """Display ROPE when more than one interval is provided."""
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -11,7 +11,7 @@ from ..stats import hpd
 from ..stats.diagnostics import _ess, _rhat
 from .plot_utils import _scale_fig_size, xarray_var_iter, make_label, get_bins
 from .kdeplot import _fast_kde
-from ..utils import _var_names
+from ..utils import _var_names, conditional_jit
 
 
 def pairwise(iterable):
@@ -289,11 +289,18 @@ class PlotHandler:
 
     def labels_and_ticks(self):
         """Collect labels and ticks from plotters."""
-        labels, idxs = [], []
-        for plotter in self.plotters.values():
-            sub_labels, sub_idxs, _, _ = plotter.labels_ticks_and_vals()
-            labels.append(sub_labels)
-            idxs.append(sub_idxs)
+        val = self.plotters.values()
+
+        @conditional_jit
+        def label_idxs():
+            labels, idxs = [], []
+            for plotter in val:
+                sub_labels, sub_idxs, _, _ = plotter.labels_ticks_and_vals()
+                labels.append(sub_labels)
+                idxs.append(sub_idxs)
+            return labels, idxs
+
+        labels, idxs = label_idxs()
         return np.concatenate(labels), np.concatenate(idxs)
 
     def display_multiple_ropes(self, rope, ax, y, linewidth, rope_var):
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -345,9 +345,9 @@ def _cov_1d(x):
 
 
 def _cov(data):
-    if data.ndimn == 1:
+    if data.ndim == 1:
         return _cov_1d(data)
-    elif data.ndimn == 2:
+    elif data.ndim == 2:
         x = data
         avg, _ = np.average(x, axis=1, weights=None, returned=True)
         fact = x.shape[1] - 1
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -338,12 +338,12 @@ def _histogram(x, n_bins, range_hist=None):
 
 
 @conditional_jit(cache=True)
-def cov_(data):
+def _cov(data):
     return np.cov(data)
 
 
 @conditional_jit(cache=True)
-def stack(x, y):
+def _stack(x, y):
     return np.vstack((x, y))
 
 
@@ -384,13 +384,13 @@ def _fast_kde_2d(x, y, gridsize=(128, 128), circular=False):
     d_x = (xmax - xmin) / (n_x - 1)
     d_y = (ymax - ymin) / (n_y - 1)
 
-    xyi = stack((x, y)).T
+    xyi = _stack((x, y)).T
     xyi -= [xmin, ymin]
     xyi /= [d_x, d_y]
     xyi = np.floor(xyi, xyi).T
 
     scotts_factor = len_x ** (-1 / 6)
-    cov = cov_(xyi)
+    cov = _cov(xyi)
     std_devs = np.diag(cov ** 0.5)
     kern_nx, kern_ny = np.round(scotts_factor * 2 * np.pi * std_devs)
 
@@ -400,7 +400,7 @@ def _fast_kde_2d(x, y, gridsize=(128, 128), circular=False):
     y_y = np.arange(kern_ny) - kern_ny / 2
     x_x, y_y = np.meshgrid(x_x, y_y)
 
-    kernel = stack((x_x.flatten(), y_y.flatten()))
+    kernel = _stack((x_x.flatten(), y_y.flatten()))
     kernel = np.dot(inv_cov, kernel) * kernel
     kernel = np.exp(-kernel.sum(axis=0) / 2)
     kernel = kernel.reshape((int(kern_ny), int(kern_nx)))
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -360,7 +360,7 @@ def _cov(data):
         c_c *= np.true_divide(1, fact)
         return c_c.squeeze()
     else:
-        raise ValueError("{} dimension arrays are not supported".format(data.ndimn))
+        raise ValueError("{} dimension arrays are not supported".format(data.ndim))
 
 
 @conditional_jit(cache=True)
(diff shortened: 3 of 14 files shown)


arviz/plots/kdeplot.py
```diff
@@ -331,12 +331,22 @@ def _fast_kde(x, cumulative=False, bw=4.5, xmin=None, xmax=None):
...
     return grid
 
 
+@conditional_jit(cache=True)
+def cov_(data):
+    return np.cov(data)
+
+
+@conditional_jit(cache=True)
+def stack(x, y):
```
ColCarroll: should be `def stack(tple):`, I think, or change wherever it is called to use 2 inputs.

image

Ban-zee:
I'll fix the failing tests by tomorrow

Ban-zee:
Should be up for review once the sphinx issue has been resolved.

Ban-zee:
Up for review.

arviz/plots/essplot.py
```diff
@@ -334,7 +334,7 @@ def plot_ess(
             ylabel.format("Relative ESS" if relative else "ESS"), fontsize=ax_labelsize, wrap=True
         )
         if kind == "evolution":
-            ax_.legend(title="Method", title_fontsize=xt_labelsize, fontsize=xt_labelsize)
+            ax_.legend(title="Method", fontsize=xt_labelsize)
```
Ban-zee: This was causing Sphinx to fail so I had to remove it :(

ahartikainen: Do you know why?

Ban-zee: Encountered unknown parameter. It is a bit strange.

Ban-zee: Reordering fixed it.

OriolAbril:
Would benchmarks of the numbified plotting functions make sense? I think that benckmarking only `_fast_kde` and `fast_kde_2d` should do because the other functions basically end up calling this one.

arviz/plots/khatplot.py
```diff
@@ -302,3 +303,8 @@ def hover(event):
                     fig.canvas.draw_idle()
 
     fig.canvas.mpl_connect("motion_notify_event", hover)
+
+
+@conditional_jit
+def _khat_histogram(data, bin_edges):
```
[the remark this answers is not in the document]
Ban-zee: The numba 0.44 update has thrown up a lot of significant changes and they are planning to remove some functions by 0.46. So I have kept these functions separate to ease the debugging in case something breaks.

Ban-zee:
Sadly yes. However, on datasets smaller than the one mentioned, the performance is much better.

benchmarks/benchmarks.py
```diff
@@ -83,3 +86,248 @@ def _circular_standard_deviation(samples, high=2 * np.pi, low=0, axis=None):
         except ImportError:
             data = np.random.randn(10000, 1000)
             return circstd(data)
+
+
+class Fast_Kde_1d:
```
OriolAbril: Is there a way to do the benchmarks calling arviz functions? Otherwise, if somebody were to change the source of `_fast_kde` this file would also have to be modified for benchmarks to be up to date.

arviz/plots/forestplot.py
```diff
@@ -289,11 +289,18 @@ def make_plotters(self):
...
+        val = self.plotters.values()
+
+        @conditional_jit(forceobj=True)
+        def label_idxs():
+            labels, idxs = [], []
+            for plotter in val:
+                sub_labels, sub_idxs, _, _ = plotter.labels_ticks_and_vals()
+                labels.append(sub_labels)
+                idxs.append(sub_idxs)
+            return labels, idxs
```
OriolAbril: Would `return np.concatenate(labels), np.concatenate(idxs)` here (where is compiled by numba) instead of 4 lines later (where it is not) make any difference?

arviz/plots/plot_utils.py
```diff
@@ -253,6 +256,14 @@ def purge_duplicates(list_in):
     return _list
 
 
+def _dims(data, var_name, skip_dims):
+    return [dim for dim in data[var_name].dims if dim not in skip_dims]
+
+
+def _zip_dims(new_dims, vals):
+    return [{k: v for k, v in zip(new_dims, prod)} for prod in product(*vals)]
```
OriolAbril: Have these functions been created for performance reasons? They are a single line of code and only called once

arviz/stats/diagnostics.py
```diff
@@ -616,6 +617,10 @@ def _z_scale(ary):
     return z
 
 
+def _stack(x, y):
+    return np.vstack((x, y))
```
OriolAbril: Same as `_stack` in `kdeplot.py`?

arviz/tests/test_plots.py
```diff
@@ -375,6 +376,17 @@ def test_plot_kde(continuous_model, kwargs):
...
 
+def test_cov():
+    x = np.random.rand(100, 100)
+    assert np.allclose(_cov(x), np.cov(x))
+
+
+def test_stack():
+    x = np.random.randn(100, 100, 5)
+    y = np.random.randn(100, 100, 5)
+    assert np.allclose(_stack(x, y), np.vstack((x, y)))
```
OriolAbril: `np.allclose` will fail if they both have different shape, but I think it would be more clear if before, the shape was checked.

benchmarks/benchmarks.py
Ban-zee: Build fails :(

arviz/stats/diagnostics.py
Ban-zee: I'll use a common method. :)

arviz/plots/plot_utils.py
Ban-zee: I was aiming for a loop lift via numba but Sphinx was failing. I left these methos as such so as to jit them later with a later, compatible version of Numba. Numba-devs are planning for some major changes with their following versions so I am writing jitted functions separately for debugging.

arviz/tests/test_plots.py
Ban-zee: Thank you for pointing this out.

Ban-zee:
Definitely. Issue is jitting np.cov speeds up well upto a certain size but
once the threshold has been breached it slows down the method without
throwing any error or warning of falling back to object mode.

arviz/plots/kdeplot.py
```diff
@@ -374,13 +379,16 @@ def _fast_kde_2d(x, y, gridsize=(128, 128), circular=False):
...
 
-    xyi = np.vstack((x, y)).T
+    xyi = _stack(x, y).T
     xyi -= [xmin, ymin]
     xyi /= [d_x, d_y]
     xyi = np.floor(xyi, xyi).T
 
     scotts_factor = len_x ** (-1 / 6)
-    cov = np.cov(xyi)
+    if np.shape(xyi)[0] < 10000:
```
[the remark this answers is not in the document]
Ban-zee: I may have found something in `np.cov` source code. I'll update in a few hours.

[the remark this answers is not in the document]
canyon289: I would ask for a docstring for magic numbers like these as well so future developers know why this choice was made

arviz/tests/test_plots.py
```diff
@@ -389,6 +395,17 @@ def test_plot_kde_cumulative(continuous_model, kwargs):
     assert axes
 
 
+def test_plot_kde_2d():
```
canyon289: Can this be two different tests?

canyon289: Parametrizing is fine

benchmarks/benchmarks.py
canyon289: What does build fails means?

Ban-zee:
Fixed the issue with `np.cov`. Will add tests and fix the mess that is `benchmarks.py` by tomorrow.

Ban-zee:
After working out the new cov, I guess it is best to include `fast_kde` and `fast_kde_2d` in the benchmarks.

ahartikainen:
LGTM

Let's tune benchmarks in other PR if needed

arviz/plots/kdeplot.py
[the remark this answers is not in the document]
canyon289: @Ban-zee Asking more explicitly, would like to see a docstring on the magic numbers before merge.

arviz/tests/test_plots.py
canyon289: Blocking merge on this comment as well

arviz/plots/kdeplot.py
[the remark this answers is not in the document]
Ban-zee: I have removed the condition to check for the size of the data.

Ban-zee:
Screenshot from 2019-07-20 18-27-24

arviz/plots/kdeplot.py
```diff
@@ -331,12 +331,43 @@ def _fast_kde(x, cumulative=False, bw=4.5, xmin=None, xmax=None):
...
 def _histogram(x, n_bins, range_hist=None):
     grid, _ = np.histogram(x, bins=n_bins, range=range_hist)
     return grid
 
 
+def _cov_1d(x):
+    x = x - x.mean(axis=0)
+    fact = x.shape[0] - 1
+    by_hand = np.dot(x.T, x.conj()) / fact
+    return np.array(by_hand)
```
ColCarroll: that surprises me that this is not already an array

arviz/plots/kdeplot.py
```diff
@@ -331,12 +331,43 @@ def _fast_kde(x, cumulative=False, bw=4.5, xmin=None, xmax=None):
...
+        x -= avg[:, None]
+        x_t = x.T
+        c_c = _dot(x, x_t.conj())
+        c_c *= np.true_divide(1, fact)
+        return c_c.squeeze()
+    else:
+        raise ValueError("{} dimension arrays are not supported".format(data.ndim))
+
+
+@conditional_jit(cache=True)
```
ColCarroll: something like this could be 
```python
_dot = conditional_jit(cache=True)(np.dot)
```

arviz/plots/kdeplot.py
```diff
@@ -331,12 +331,43 @@ def _fast_kde(x, cumulative=False, bw=4.5, xmin=None, xmax=None):
...
+        return _cov_1d(data)
+    elif data.ndim == 2:
+        x = data.astype(float)
+        avg, _ = np.average(x, axis=1, weights=None, returned=True)
+        fact = x.shape[1] - 1
+        if fact <= 0:
+            warnings.warn("Degrees of freedom <= 0 for slice", RuntimeWarning, stacklevel=2)
+            fact = 0.0
+        x -= avg[:, None]
+        x_t = x.T
```
ColCarroll: i'd not define `x_t` separately, since you only use it once, and `x.T` is syntactically clear.

arviz/plots/kdeplot.py
```diff
@@ -331,12 +331,43 @@ def _fast_kde(x, cumulative=False, bw=4.5, xmin=None, xmax=None):
...
+    return np.array(by_hand)
+
+
+def _cov(data):
+    if data.ndim == 1:
+        return _cov_1d(data)
+    elif data.ndim == 2:
+        x = data.astype(float)
+        avg, _ = np.average(x, axis=1, weights=None, returned=True)
+        fact = x.shape[1] - 1
```
ColCarroll: can you change `fact` to `degrees_of_freedom`?

arviz/plots/kdeplot.py
```diff
@@ -331,12 +331,43 @@ def _fast_kde(x, cumulative=False, bw=4.5, xmin=None, xmax=None):
...
+        if fact <= 0:
+            warnings.warn("Degrees of freedom <= 0 for slice", RuntimeWarning, stacklevel=2)
+            fact = 0.0
+        x -= avg[:, None]
+        x_t = x.T
+        c_c = _dot(x, x_t.conj())
```
ColCarroll: is there a more descriptive name for `c_c`?

arviz/plots/kdeplot.py
Ban-zee: It's easier to debug this way.

Ban-zee:
Done.

canyon289:
> Done.

@Ban-zee Feeling good about this one?

Ban-zee:
Yes.

arviz/plots/forestplot.py
```diff
@@ -289,12 +289,18 @@ def make_plotters(self):
...
+        @conditional_jit(forceobj=True)
+        def label_idxs():
+            labels, idxs = [], []
+            for plotter in val:
+                sub_labels, sub_idxs, _, _ = plotter.labels_ticks_and_vals()
+                labels.append(sub_labels)
+                idxs.append(sub_idxs)
+            return np.concatenate(labels), np.concatenate(idxs)
+
+        return label_idxs()
```
ColCarroll: wow, this looks like a tricky way to optimize!

arviz/stats/diagnostics.py
```diff
@@ -550,7 +550,8 @@ def ks_summary(pareto_tail_indices):
     """
     _numba_flag = Numba.numba_flag
     if _numba_flag:
-        kcounts = histogram(pareto_tail_indices)
+        bins = np.asarray([-np.Inf, 0.5, 0.7, 1, np.Inf])
```
ColCarroll: we *should* move `bins` to line 551, so that we don't repeat ourselves on line 556.

ColCarroll:
Thanks @Ban-zee !

PR #743, opened 2019-07-12
make order of model in compare df explicit

### LINKED ISSUE #733: Add rank in compare dataframe
Compare dataframe should contain rank column, which would make the order or models explicit. 

Deviance vs log -scales have a different sign and it can be confusing to remember which one is the best. And even harder if user resorts the rows.

aloctavodia:
Fix #733

Also in case people do crazy stuff like sorting the compare DataFrame, `plot_compare` will resort the DataFrame by default, so the plot makes sense.

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -88,8 +88,9 @@ def compare(
 
     Returns
     -------
-    A DataFrame, ordered from lowest to highest IC. The index reflects the key with which the
+    A DataFrame, ordered from best to worst model. The index reflects the key with which the
     models are passed to this function. The columns are:
+    rank : The rank-order of the models. 0 is the best.
     IC : Information Criteria (WAIC or LOO).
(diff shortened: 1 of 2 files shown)


arviz/stats/stats.py
```diff
@@ -88,8 +88,9 @@ def compare(
 
     Returns
     -------
-    A DataFrame, ordered from lowest to highest IC. The index reflects the key with which the
+    A DataFrame, ordered from best to worst model. The index reflects the key with which the
```
ColCarroll: lgtm! "best" and "worst" are sometimes loaded words -- is it very accepted that they are appropriate here? Would a note like "ordered from best to worst model  (measured by information criteria)." be more clear?

aloctavodia: Yeah, I share the same concern.

ColCarroll:
LGTM when tests pass.

PR #744, opened 2019-07-12
Add rugby dataset

ColCarroll:
I have verified this works locally. The description of the dataset is in the PR:

>The Six Nations Championship is a yearly rugby competition between Italy, Ireland,
Scotland, England, France and Wales. Fifteen games are played each year, representing
all combinations of the six teams.
 This example uses and includes results from 2014 - 2017, comprising 60 total
games. It models latent parameters for each team's attack and defense, as well
as a parameter for home team advantage.
 See https://docs.pymc.io/notebooks/rugby_analytics.html by Peader Coyle
for more details and references.

I have also updated `load_arviz_data` so calling it with no parameters returns a dictionary, which makes it a little easier to work with.
image

Previous behavior was to throw a `ValueError` with the information in it.

### DIFF
--- a/arviz/data/datasets.py
+++ b/arviz/data/datasets.py
@@ -61,7 +61,24 @@ See Gelman and Hill (2006) for details on the example, or
 https://docs.pymc.io/notebooks/multilevel_modeling.html#Correlations-among-levels
 by Chris Fonnesbeck for details on this implementation.
 """,
-    )
+    ),
+    "rugby": RemoteFileMetadata(
+        filename="rugby.nc",
+        url="https://ndownloader.figshare.com/files/16254359",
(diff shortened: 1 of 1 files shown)


ColCarroll:
I should add https://figshare.com/articles/rugby_nc/8866133 is the figshare page for the dataset, and I would love to merge this before too long if there are no objections.

ColCarroll:
Also, I can't find where we store the code for _building_ the examples. Does this exist? I have a small notebook...

OriolAbril:
I could only find the models for the 8 schools, in the quickstart and the coockbooks notebook when I updated their posterior predictive samples. I could not find the code for the radon model, and it is still due to resample the posterior predictive.

ColCarroll:
I will make a new repo with code for each example, and maybe detailed instructions in case other people want to upload datasets.

PR #746, opened 2019-07-13
Add regression datasets

### LINKED ISSUE #242: Add more fitted models
figshare (https://figshare.com/) seems like a possibility that I think scikit-learn uses. In terms of size, [the radon example here](https://docs.pymc.io/notebooks/multilevel_modeling.html) generates a netCDF file of about 15MB, which is not so bad.

Perhaps it could even be hosted in a separate github repo?

ColCarroll:
This adds two synthetic datasets to figshare, and then to arviz. They are both synthetic datasets, and described in the PR. Fixes #242. 

Will merge before scipy sprints start if there are no objections in the meantime.

image

### DIFF
--- a/arviz/data/datasets.py
+++ b/arviz/data/datasets.py
@@ -102,6 +102,30 @@ fit with PyMC3.
 True weights and intercept are included as deterministic variables.
 """,
     ),
+    "classification1d": RemoteFileMetadata(
+        filename="classification1d.nc",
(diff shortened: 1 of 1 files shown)


ColCarroll:
Ok, I also added two classification datasets along the same lines. I'll push reproducible scripts to build all these after the sprints, especially I might appreciate another look at these :joy: 

image

PR #748, opened 2019-07-13
Fix az.compare for multidimensional obsects

### LINKED ISSUE #740: az.compare fails for pymc3 mixture models

OriolAbril:
Should fix #740 for multidim objects (their log_likelihood has shape `chain, draw, *shape` where `*shape` is `>=2D`.

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -196,6 +196,7 @@ def compare(
         ics = ics.append([ic_func(dataset, pointwise=True, scale=scale)])
     ics.index = names
     ics.sort_values(by=ic, inplace=True, ascending=ascending)
+    ics[ic_i] = ics[ic_i].apply(lambda x: x.values.flatten())
 
     if method.lower() == "stacking":
         rows, cols, ic_i_val = _ic_matrix(ics, ic_i)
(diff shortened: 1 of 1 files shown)


OriolAbril:
Ready for merging. I will add tests on using multidim data in `az.compare` in another PR.

PR #750, opened 2019-07-13
Rugby plot

RaulPL:
Replaces the existing example with a new a new one using the rugby dataset.
image


ColCarroll:
Thanks for the contribution, @RaulPL !

PR #755, opened 2019-07-15
Add multidim tests to test_stats [GSOC]

OriolAbril:
Following #748, this PR adds test on multidimensional objects `shape(log_likelihood) = (chain, draw, dim1, dim2)` to `test_stats.py`, in order to find issues like #740 in CI tests. Thus, `compare`, `waic`, `loo`, and `loo_pit` are now tested on the `centered_eight` data and on the `multidim_models` fixture


OriolAbril:
Ready to merge

PR #764, opened 2019-07-23
Add note on scale use in waic and loo [GSoC]

OriolAbril:
@canyon289 do you think this is clear enough? Or would it be best to add the explanation somewhere else like the notes section?

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -398,7 +398,8 @@ def loo(data, pointwise=False, reff=None, scale="deviance"):
         - `log` : 1 * log-score (after Vehtari et al. (2017))
         - `negative_log` : -1 * (log-score)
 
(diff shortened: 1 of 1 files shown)


arviz/stats/stats.py
```diff
@@ -398,6 +398,8 @@ def loo(data, pointwise=False, reff=None, scale="deviance"):
         - `log` : 1 * log-score (after Vehtari et al. (2017))
         - `negative_log` : -1 * (log-score)
 
+        A high log-score (or a low deviance) indicate a good predictive accuracy of the model.
```
aloctavodia: What about? 

A higher log-score (or a lower deviance) indicates a model with better predictive accuracy.

canyon289:
Looks good to me!

PR #765, opened 2019-07-24
preallocate potentially large arrays

OriolAbril:
As discussed in slack, this PR will try to improve performance (mainly RAM usage, but in some cases, execution time will improve too) by preallocating large arrays instead of having them grow inside loops.


OriolAbril:
Ready to merge

PR #769, opened 2019-07-24
grid lines in plot_forest

OriolAbril:
Fixes https://github.com/pymc-devs/pymc3/issues/3515. I am not sure if just like this is enough or if we have strong aesthetic arguments against grid lines in forestplot. It can always be sub-seeded to a `grid` argument defaulting to false.

### DIFF
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -208,7 +208,6 @@ def plot_forest(
         idx += 1
 
     for ax in axes:
-        ax.grid(False)
         # Remove ticklines on y-axes
         ax.tick_params(axis="y", left=False, right=False)
 
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -208,6 +208,7 @@ def plot_forest(
         idx += 1
 
     for ax in axes:
+        ax.grid(False, axis="x")
         # Remove ticklines on y-axes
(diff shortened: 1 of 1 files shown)


canyon289:
I would say make this an argument, and for this PR would be nice to have a screenshot of the example so we can compare to gallery

canyon289:
But I'm also not very opinionated and would be fine merging

aloctavodia:
I suggest using (by default) grid-lines false for ridgeplots and vertical-only for forestplot, not sure if this should be and argument or something the user should change by modifying the returned axis. Probably I am equally fine with both options.

OriolAbril:
@aloctavodia this sounds like a better default. Now if the matplotlib style has grid lines as default, only the horizontal ones will be removed. To depart from this default, one would have to call `ax.grid` on the returned axes object.

arviz/plots/forestplot.py
```diff
@@ -208,7 +208,7 @@ def plot_forest(
         idx += 1
 
     for ax in axes:
-        ax.grid(False)
+        ax.grid(False, axis="x")
```
aloctavodia: I would suggest to do: 

        if kind == "ridgeplot": 
            ax.grid(False)
        else:
            ax.grid(False, axis="y")

But I understand if we do not want to make a difference between `ridgeplot` and `forestplot`

Issue #220, opened 2018-09-09
Add other data to schema?


ColCarroll:
Per #219, other data may be involved in inference. In the eight schools model, for example, the data block contains:

```
data {
  int<lower=0> J; // number of schools 
  real y[J]; // estimated treatment effects
  real<lower=0> sigma[J]; // s.e. of effect estimates 
}
```

right now, we only extract "observed_data", which are observations of parameters in the model. More specifically, the `posterior_predictive/` group should match with the `observed_data/` group. Open for suggestions on whether and how to store the rest of the data.

ahartikainen:
I think we should have some special group that contains model specific information. 

{library, version, algorithm, init settings}

This is to make sampled data "reproducible" or at least in mcmc sense. So we could have "generic" kw and then special kw for library specific settings.

For Stan this could include stan model + cppcode, init info, warm-up info (inv_metric, stepsize, starting location for sample, seed), thinning info etc.

Do we have some good name for this and can we make it different kind of group than posterior etc are (e.g. in repr it could be shown as "model information, see `model_settings`.

Also, who has better sugggestion for the name than `model_setting`

OriolAbril:
I think there are 2 different kinds of data involved whose possible names could be `model_parameters` and `sampling_parameters`/`sampler_parameters` (or maybe params to make it shorter). In my opinion, both additions to the InferenceData object would be really valuable.

### Model parameters
These would be parameters of the model which is not actually observed data but they are needed to run the model, as @ColCarroll was pointing out. `J` and `sigma` would fall in this category. In some cases the border between observed data and model parameters could be a little blurry, so the best definition would be probably "arguments samples with the `observed` kwarg in pymc3" or what is the same "defined in the data section of the stan code, and sampled with `~` in the model section".

They are currently stored in the `observed_data` group automatically for emcee, and in general for pystan too. In pymc3 however, they are not stored at all for pymc3 objects, as `observed_data` comes from data sampled with `observed=...` kwarg. I think there are two approaches, either storing this info always in `observed_data` or create a new group for them. I prefer the second option, but it may be harder to explain.

### Sampler parameters
This should include the information of sampler algorithm name and parameters like @ahartikainen was saying. Currently, only the library and its version is being stored as attrs of each dataset. That comes to my mind right now, the only complication with this information would be when merging inference data objects.

---

[...]

PR #771, opened 2019-07-27
Sampling wrappers

OriolAbril:
There are some algorithms that require refitting to some extent, some examples are Leave Future Out, reloo and Simulation Based Calibration.

As ArviZ has no sampling capabilities, and in general it is crucial that the likelihood functions used are on the same scale (as they may not be normalized), we thought of creating some wrappers in ArviZ so that functions in ArviZ can call sampling packages (PyMC3, PyStan, emcee...) always using the same syntax. Therefore, they could be used for everybody.

I started working on creating some API for these wrappers (which more or less I think I am starting to converge on an idea both simple and general enough. The preliminary docs of the SamplingWrapper class can be read [here](https://oriolabril.github.io/arviz/api/wrappers.html#wrappers). 

I have been looking into PyStan and PyMC3, but I found some trouble which will require some more time until I completely understand their API and objects. In PyStan I am struggling with calling the pointwise log likelihood from within Python and in PyMC3 I am following @aloctavodia pointer on using the Data class, but I am still working the details. Therefore, to have something to use I also did [one example](https://github.com/OriolAbril/arviz/blob/refitting-experiments/refitting_emcee.ipynb) as a proof of concept using emcee, which is the library I am more familiar with for now and it is also pure python+numpy which in this case has some advantages. The reloo function works properly when calculating `loo`, `loo_se` and `loo_i`, but not to calculate `p_loo` (I will get to it eventually), the important thing is that `reloo` manages to refit the model on different data and work with the result. 

**EDIT1**: I have already uploaded the [PyStan example](https://github.com/OriolAbril/arviz/blob/refitting-experiments/refitting_pystan.ipynb). I have also modified how refitting is tested. Now instead of creating an outiler on the data to force reloo to refit the model, I just modify good pareto_k values to force refitting. Hence, now the original PSIS-LOO should have already worked in the first place, and then the partially refitted loo should give the same value (unless there were an issue with refitting on wrong data or on the lppd-i calculation).

**EDIT2**: [This](https://github.com/OriolAbril/arviz/blob/refitting-experiments/refitting_pymc.ipynb) is how far I have managed to get into refitting with PyMC3. I have one problem for which there should be no problem in finding a solution (it is basically #220) and then a second problem whose solution I do not see at all. You can check the notebook with the example, the problem is basically that the kinds of refits done for reloo of LFO will require the shape of the observations and variables in the model to vary. And this does not seem possible in PyMC3.


ahartikainen:
PyStan: elementwise log likelihood needs to be handcoded and is calculated in generated quantities block.

ahartikainen:
I think the example for PyStan looks good.

Adding some good docs, should make this procedure easier to understan for the users.

Here is some simple example that needs to be make more clear.

## model

```
data {
    good_data
    bad_data
    ...
}
...
model {
    // fit against good_data
   ...
}
generated quantities {
    ....
    log_lik_good for good data
    log_lik_bad for bad data
}
```

Also, the bad and good log_lik names should be optional

OriolAbril:
I have already added the `PyStanSamplingWrapper` skeleton and `reloo` to the main code from the experiments notebook. I will try to implement LFO before Friday.

I was thinking on creating a "refitting" notebook explaining a little why would we want to refit models in ArviZ and afterwards, links to different notebooks with examples of different libraries. This first notebook would be on the headbar like the Cockbook or Numba and the others would be lost in the limbo (aka the notebook not present in any toctree warning will resurface). Does this sound okey?

ahartikainen:
We can keep this still open (and do one release before this is merged)

ahartikainen:
Is this ready? I think we can merge this? It uses log_likelihood group?

OriolAbril:
It is ready-ish, I updated it to use log_likelihoods group, the only thing left is make sure docs show sampling wrappers, the notebook with teh example can be found and other documentation tweaks.

mathDR:
Would this still be considered "experimental"?

OriolAbril:
Yes. It is experimental because it is still untested both in CI and usage testing at least on our side) as we have not had much time to dedicate to it, and therefore, the API could still change (i.e. if the current approach were not flexible enough to implement IWMM for example). But it is not experimental in the sense that the functionality might be removed at any future version.

If you have used it and can help with testing we'd be happy to help and it would go a long way in making that non-experimental.

mathDR:
Okay great.  Will do!

PR #774, opened 2019-07-30
Numba-data

Ban-zee:
The speedup is not large but the changes do save a few microseconds. I'll add the benchmark tests for the same in the upcoming commits.


Ban-zee:
Ready for review

arviz/data/base.py
```diff
@@ -3,9 +3,9 @@
 import datetime
 import warnings
 
-import numpy as np
 import pkg_resources
 import xarray as xr
+from ..utils import two_de, arange
```
canyon289: Reading through this code I have a slight preference that arange should be under `utils` namescope
so this like would be `from .. import utils` and then subsequent calls would be `utils.arange`

The reason for this is that this way its really obvious its not the `np.arange` method but another one of our own design. It does add characters but I think it removes a mental burden that makes it worthwhile

thoughts?

arviz/data/io_cmdstan.py
```diff
@@ -14,7 +14,7 @@
 
 from .inference_data import InferenceData
 from .base import requires, dict_to_dataset, generate_dims_coords
-
+from ..utils import one_de, full
```
canyon289: I have same comment here with full as the one with arange above

arviz/tests/test_utils.py
```diff
@@ -239,3 +248,13 @@ def test_stack():
...
     assert np.allclose(np.vstack((x, y)), _stack(x, y))
     assert _stack
+
+
+@pytest.mark.parametrize("data", [np.random.randn(1000), np.random.randn(1000).tolist()])
+def test_two_de(data):
+    assert np.allclose(two_de(data), np.atleast_2d(data))
+
+
+@pytest.mark.parametrize("data", [np.random.randn(100), np.random.randn(100).tolist()])
```
canyon289: What does the tolist check here? Can docstring be added? Same with the test above?

Ban-zee: tolist() is used to create a non ndarray object, to test the else condition.

arviz/data/base.py
Ban-zee: Sure

arviz/data/base.py
```diff
@@ -2,10 +2,9 @@
 from copy import deepcopy
 import datetime
 import warnings
-
-import numpy as np
 import pkg_resources
 import xarray as xr
+import arviz.utils as utils
```
canyon289: These should still be relative imports. I believe `from . import utils` should work (or `..` )

canyon289: The reason why btw is that in some scenarios `import arviz` could import the sys installed version of arviz rather than the modules that are relative to this package. By specifying relative import were sure of what code is being imported

arviz/tests/test_utils.py
canyon289: Good thinking

Ban-zee:
Any more changes needed?

arviz/data/io_dict.py
```diff
@@ -1,8 +1,7 @@
 """Dictionary specific conversion code."""
 import warnings
-
-import numpy as np
 import xarray as xr
+import arviz.utils as utils
```
canyon289: Should be relative import

Ban-zee: Missed that :)

Ban-zee: Done.

Ban-zee:
Ready for merge

arviz/data/io_cmdstan.py
```diff
@@ -7,15 +7,15 @@
 import logging
 import re
 
-
 import numpy as np
 import pandas as pd
 import xarray as xr
+import arviz.utils as utils
```
canyon289: One more system import :) Since you're importing utils below you should be able to delete this one

arviz/data/io_pyro.py
```diff
@@ -1,8 +1,7 @@
 """pyro-specific conversion code."""
-import numpy as np
-
 from .inference_data import InferenceData
 from .base import dict_to_dataset
+from ..utils import expand_dims
```
canyon289: Gonna be super picky here, it would be great if all imports were all the same

For example in this case its 
`..utils import expand_dims`
but in other modules its

```
from .. import utils
utils.expand_dims
```
I personally prefer the second since it makes it clear the method is defined somewhere else, but consistency matters more than that

arviz/data/io_tfp.py
```diff
@@ -4,7 +4,7 @@
 
 from .inference_data import InferenceData
 from .base import dict_to_dataset, generate_dims_coords, make_attrs
-
+from ..utils import expand_dims, one_de
```
canyon289: One more :)

Ban-zee: :+1:

canyon289:
This LGTM. I'll give it 24 hours in case anyone else wants to review

canyon289:
Alright its now been 168 hours. Merging!

PR #775, opened 2019-07-30
plot_ppc: change mean computation

aloctavodia:
This changes how we compute the posterior predictive mean when `kind='density'`. Currently we do a KDE using all the posterior predictive samples as input. With this PR we average the (binned) densities. In an ideal world this should be equivalent, but in practice there are some generally (small) differences. 

One problem with the current implementation is that we could have something like this

image (2)

The mean is not _inside_ the _envelope_ defined by the individual pp samples. This picture is from Chris Fonnesbeck.

In the following examples I reproduce the same effect, although with a less obvious effect, please zoom-in or use a very wide screen :-). The blue lines are the pp samples, the orange lines are the mean using the KDE (over all the pp samples) and the green lines are the average of the densities.

kde_ppc_N

This other example shows a Student-T distribution, and how the new method is better at capturing the shape of the distribution.
 
kde_ppc_T

notice that when the sample size is _really_ low the KDE (orange) looks like a better approximation and when the sample size is larger the averaged densities looks like a better approximation. Hopefully in the future if we implement an adaptive bandwidth estimation or some other more flexible KDE method this issue will go away.

### DIFF
--- a/arviz/plots/ppcplot.py
+++ b/arviz/plots/ppcplot.py
@@ -304,41 +304,53 @@ def plot_ppc(
                     drawstyle=plot_kwargs["drawstyle"],
                 )
 
+            pp_densities = []
+            pp_xs = []
+            for vals in pp_sampled_vals:
+                vals = np.array([vals]).flatten()
+                if dtype == "f":
+                    pp_density, lower, upper = _fast_kde(vals)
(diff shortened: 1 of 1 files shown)


aloctavodia:
and this is the figure from @fonnesbeck with the new method
lala

PR #776, opened 2019-07-30
Remove effective sample size from docs

OriolAbril:
`az.effective_sample_size` was removed in #770, but it was still listed in the api section of the docs.


ahartikainen:
LGTM

PR #777, opened 2019-07-31
don't run tests twice

ahartikainen:
Skip branch tests (run the tests with PR)


ahartikainen:
This is small change and works, so I will merge.

PR #778, opened 2019-07-31
Fix issue with rug values in plot_ess and plot_mcse

OriolAbril:
I don't know why I thought `argsort` could be used for ranking data. Today I have realized that the rug values were not placed where they were supposed to be. Now they are.

image

image


ahartikainen:
That is common mistake.

ahartikainen:
With all unique values, argsort(argsort(x)) gives the rank. `rankdata` handles duplicates correctly.

PR #780, opened 2019-08-01
fix from_pymc bug for multivariate observations

OriolAbril:
`from_pymc` would overwrite the "log_likelihood" dimensions defined in "dims" parameter, which in multivariate cases where the shape of the observations and of the log likelihood is different gives an error of wrong number of dimensions. 

I have left using the same dims as the observed data as a default, which in most cases is the case, but giving preference to "log_likelihood" dimensions defined in the dims argument in case they are present.

I am not sure if there should be a test for this, I can add one.

### DIFF
--- a/arviz/data/io_pymc3.py
+++ b/arviz/data/io_pymc3.py
@@ -52,7 +52,7 @@ class PyMC3Converter:
             return None, None
         else:
             if self.dims is not None:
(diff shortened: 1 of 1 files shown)


canyon289:
I think there should be a test for this. That's basically my answer for everything though :)

canyon289:
It looks covered already. As is such merging

image

PR #781, opened 2019-08-01
Add var name in LOO-PIT legend [GSOC]

OriolAbril:
At first I decided to not include the variable name in the LOO-PIT plot, but I am changing my mind. I have not followed plot_ppc and used the var_name as xlabel, because unlike there, the x values are not variable values. I am inclined towards adding the name in the legend, but the title could also be an option. Now it looks like this, it is a minor change.

image


ahartikainen:
LGTM

PR #784, opened 2019-08-03
Set Numpy Random Seed in one place for tests

### LINKED ISSUE #661: Set Numpy Random Seed in one place for tests
## Tell us about it
For test suite put `np.random.seed` in conftest.py so its available in all tests, eliminating duplication across various different tests

image

## Thoughts on implementation
See above

sameshl:
set numpy random seed in conftest.py and removed it from other test files.

closes #661

### DIFF
--- a/arviz/tests/conftest.py
+++ b/arviz/tests/conftest.py
@@ -7,10 +7,12 @@ import numpy as np
 
 _log = logging.getLogger(__name__)
 
+
 @pytest.fixture(autouse=True)
 def random_seed():
     """Reset numpy random seed generator."""
     np.random.seed(0)
+
 
 def pytest_addoption(parser):
     """Definition for command line option to save figures from tests."""
--- a/arviz/tests/conftest.py
+++ b/arviz/tests/conftest.py
@@ -9,6 +9,7 @@ _log = logging.getLogger(__name__)
 
 @pytest.fixture(autouse=True)
 def random_seed():
+    """Reset numpy random seed generator."""
     np.random.seed(0)
 
 def pytest_addoption(parser):
--- a/arviz/tests/conftest.py
+++ b/arviz/tests/conftest.py
@@ -3,7 +3,7 @@
 import os
 import logging
 import pytest
-
+import numpy as np
 
 _log = logging.getLogger(__name__)
 
--- a/arviz/tests/conftest.py
+++ b/arviz/tests/conftest.py
@@ -7,6 +7,7 @@ import pytest
 
 _log = logging.getLogger(__name__)
 
+np.random.seed(0)
 
 def pytest_addoption(parser):
     """Definition for command line option to save figures from tests."""
--- a/arviz/tests/conftest.py
+++ b/arviz/tests/conftest.py
@@ -7,7 +7,9 @@ import pytest
 
 _log = logging.getLogger(__name__)
 
-np.random.seed(0)
+@pytest.fixture(autouse=True)
+def random_seed():
+    np.random.seed(0)
 
 def pytest_addoption(parser):
     """Definition for command line option to save figures from tests."""
--- a/arviz/tests/helpers.py
+++ b/arviz/tests/helpers.py
@@ -39,7 +39,6 @@ def chains():
 
 def create_model(seed=10):
     """Create model with fake data."""
-    np.random.seed(seed)
     nchains = 4
     ndraws = 500
     data = {
@@ -240,9 +239,6 @@ def emcee_schools_model(data, draws, chains):
     sigma = data["sigma"]
     J = data["J"]  # pylint: disable=invalid-name
     ndim = J + 2
-
-    # make reproducible
-    np.random.seed(0)
 
     pos = np.random.normal(size=(chains, ndim))
     pos[:, 1] = np.absolute(pos[:, 1])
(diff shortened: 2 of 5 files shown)


arviz/tests/conftest.py
```diff
@@ -7,6 +7,7 @@
 
 _log = logging.getLogger(__name__)
 
+np.random.seed(0)
```
ColCarroll: This should actually be in a fixture so that it runs before every test. something like

```python
@pytest.fixture(autouse=True)
def random_seed():
    np.random.seed(0)
```

sameshl: OK, will make that change!

sameshl:
@ColCarroll Can you help me understand where the error is in the build?

ahartikainen:
You need to import numpy in conftest

Add `import numpy as np` to the imports.

ahartikainen:
You can see the last error here https://travis-ci.org/arviz-devs/arviz/jobs/568008509

Your function is missing docstring.

I think something similar is enough (feel free to change if you want):

    """Reset numpy random number generator."""

sameshl:
I added the missing docstring. But on a side note, can you tell me where exactly was the error [here](https://travis-ci.org/arviz-devs/arviz/jobs/568008509) stating "missing docstring" because I coudn't figure it by myself.

OriolAbril:
> I added the missing docstring. But on a side note, can you tell me where exactly was the error [here](https://travis-ci.org/arviz-devs/arviz/jobs/568008509) stating "missing docstring" because I coudn't figure it by myself.

There are various kind of tests run simultaneously. The first ones are the "unit" tests, which are the first 4. They check that the code _works_ as it should. Then there is the "lint". It checks that the code is written according to some conventions. And finally (and only in travis) there is the "sphinx" test, which tests that the docs can be properly built. 

To check the coding style before pushing, you can run:

    $ ./scripts/lint.sh

Also, to be able to run all tests without installing all dependencies in your computer, you can use docker. There is a guide on developing in docker in the [contributing guide](https://github.com/arviz-devs/arviz/blob/master/CONTRIBUTING.md).

Ban-zee:
Using black should resolve the error.

arviz/tests/helpers.py
```diff
@@ -39,7 +39,6 @@ def chains():
 
 def create_model(seed=10):
```
ahartikainen: this function needs to remove the seed

sameshl:
I am not able to figure out my error at <https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=449>. Can someone help me out?

ColCarroll:
Ah, you should run on the command line

```bash
black arviz/
```

you might need to `pip install black` for this to work.

ColCarroll:
By the way, you can see [details for black here](https://github.com/psf/black), and a pull request checklist (which you are on the last step of) [here](https://github.com/arviz-devs/arviz/blob/master/CONTRIBUTING.md#pull-request-checklist).

sameshl:
> Ah, you should run on the command line
> 
> ```shell
> black arviz/
> ```
> 
> you might need to `pip install black` for this to work.

I ran that. The output I got is 
```
All done! ✨ 🍰 ✨
62 files left unchanged.
```
and there is no change to any file as seen by running `git status` after running `black arviz/`

arviz/tests/helpers.py
OriolAbril: I would leave these seed here, because it is used to create different models. If we decide to remove the seed in `create_model`, we would need to modify also `create_multidimensional_model` **and** all the calls to these 2 functions.

OriolAbril:
It looks like you linked an old build message, where the error was due to black formatting. 

The current error (https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=452) is due to the modifications in `create_model`. As I commented above, as you have modified this function, all the places where it is called must be modified too for the code to work.

arviz/tests/helpers.py
sameshl: So should just leave the seed here or make the required changes to `create_multidimensional_model`?

ahartikainen:
So for now,  only things left are

- adding seed argument back to the test (and for now, you can leave the np.random.seed(seed) call to that function)
- `black arviz`
- `pylint arviz` (sometimes there are some pylint errors that are not seen locally, so remember to check travis / azure after the test have run)

I think tests should pass after it.

sameshl:
> So for now, only things left are
> 
>     * adding seed argument back to the test (and for now, you can leave the np.random.seed(seed) call to that function)
> 
>     * `black arviz`
> 
>     * `pylint arviz` (sometimes there are some pylint errors that are not seen locally, so remember to check travis / azure after the test have run)
> 
> 
> I think tests should pass after it.

I did that and pushed my changes. But I don't understand what is the error here?

arviz/tests/helpers.py
ColCarroll: the error in the logs is from this line: `seed` is unused. i agree with @OriolAbril that for now we should 

- leave the seed as an argument (this requires no action)
- add back the two calls to `np.random.seed` that were removed

OriolAbril: Only one call to `np.random.seed(seed)` was removed and has to be added back though. The one that was in line 42 of `helpers.py`, right after the docstring.

ColCarroll:
Thanks @sameshl!

sameshl:
@ColCarroll Thank you! Looking forward to contributing more here :smiley:

PR #788, opened 2019-08-08
posteriorplot kwargs bug

OriolAbril:
kwargs were only passed to `plt.hist` but not to `plot_kde` this should fix it.

### DIFF
--- a/arviz/tests/test_plots.py
+++ b/arviz/tests/test_plots.py
@@ -646,7 +646,14 @@ def test_plot_rank(models, kwargs):
         {"ref_val": None},
         {"ref_val": {"mu": [{"ref_val": 1}]}},
         {"bins": None, "kind": "hist"},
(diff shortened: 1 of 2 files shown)


arviz/tests/test_plots.py
```diff
@@ -648,7 +648,7 @@ def test_plot_rank(models, kwargs):
         {"ref_val": None},
         {"ref_val": {"mu": [{"ref_val": 1}]}},
         {"bins": None, "kind": "hist"},
-        {"mu": {"ref_val": (-1, 1)}},
+        {"ref_val": {"mu": [{"ref_val": -1}, {"ref_val": 1}]}},
```
OriolAbril: I think this will fix the test (we'll see) but it does not plot 2 reference values, only the first. What was the goal of this test? It used to work because "mu" was a kwarg and kwargs were ignored for kde, they were only taken into account for hist.

OriolAbril: I decided to convert the test to testing multiple ref_vals with coordinates which was not tested before.

OriolAbril:
Ready for review

PR #790, opened 2019-08-09
rename "samples" dim to "sample" for coherence

OriolAbril:
Methods that require stacking chains and draws will now expect a "sample" dim, following "chain" and "draw" dim names.


ahartikainen:
LGTM

PR #791, opened 2019-08-12
Modify pytorch installation in create env script

OriolAbril:
~~This should fix pytorch installation.~~ Problem was not actually with pytorch installation, but I think the changes are still useful.

I have seen that version 1.2.0 has just been released, and with it, recommended method shown in their website has been modified too. In my opinion, now it is quite more sensible and it should install cpu wheel of latest version automatically, so pytorch_version env variable will only be used for old versions and we will always be running tests on latest pytorch, no need to update it manuallly from time to time. We may have the same trouble as with tfp if pyro and pytorch releases are not coordinated though.


OriolAbril:
This should be ready to merge after executing the tests again now that the issue with python 3.7 is solved.

OriolAbril:
Ready to merge

PR #794, opened 2019-08-13
Add more data to InferenceData objects [GSOC]

OriolAbril:
~For now should solve only half of #220, the part related to `constant_data`.~ I would also like to support multiple log_likelihoods somehow (related to #771).

EDIT: constant data support was added in #814 

---

## EDIT
I update the description of what this PR does and what will still be pending. 

#### Does
* Supports storage of multiple observed variables each with its own log likelihood data
  * dict
   * emcee
   * pymc
   * pystan
* Updates loo, waic... to get the log likelihood data from the new group. To do so, a function `get_log_likelihood` has been added in stats_utils. From here on, log likelihood values should be accessed using this helper function (it could even become public or a method of inference data, we have to think about this)

Note: `lp` variable is left in `sample_stats` group, there should be no change involving this.

#### Pending
* support storage of multiple observed values
  * cmdstan
  * cmdstanpy
  * numpyro
  * pyro
* update ic functions to use the data available
  * add a var_name/obs_name to waic, loo, compare...
  * take into account diferent cases such as an ic value per observation, considering all observations a single variable...
* add new group to inference data schema


OriolAbril:
> Do we get the same groups for `from_pymc` / `from_emcee`?

Yes! I am working on it.

OriolAbril:
`from_pymc3`, `from_pystan`, `from_dict` and `from_emcee` should all be properly updated to use `constant_data` and `log_likelihoods`. I may add some more tests but that should be all. I believe the only missing thing is "lp" extraction in pystan3. I also looked at `from_pyro` and nothing needs to be changed for now.

I have also updated the stats functions waic, loo and loo_pit to extract the log_likelihood properly. Storing it in sample_stats still works, it will only send a PendingDeprecationWarning. 

I have rerun the cookbook notebook. Now it shows the updated groups when using `from_pymc`, `from_pystan` of `from_emcee`. 

I tried to make everything backwards compatible and I think I succeded. The only breaking change would be for users accessing directly `idata.sample_stats.log_likelihood` or `lp` as it will now be stored in `log_likelihoods` group.

ahartikainen:
For PyStan3 `lp__` is already taken in the `get_sample_stats_stan3` function.

So let's add a kws `variables` and `ignore`. 

if `variables` is defined, ignore `ignore`.
if `variables` not defined, and ignore is, use ignore.

```
def get_sample_stats_stan3(fit, variables=None, ignore=None):
    """Extract sample stats from PyStan3 fit."""
    dtypes = {"divergent__": bool, "n_leapfrog__": np.int64, "treedepth__": np.int64}

    if isinstance(variables, str):
        variables = [variables]
    if isinstance(ignore, str):
        ignore = [ignore]    

    data = OrderedDict()
    for key in fit.sample_and_sampler_param_names:
        if (variables and key not in variables) or (ignore and key in ignore):
            continue
        new_shape = -1, fit.num_chains
        values = fit._draws[fit._parameter_indexes(key)]  # pylint: disable=protected-access
        values = values.reshape(new_shape, order="F")
        values = np.moveaxis(values, [-2, -1], [1, 0])
        dtype = dtypes.get(key)
        values = values.astype(dtype)
        name = re.sub("__$", "", key)
        data[name] = values

    return data
```

and then for sample_stats:

```
@requires("posterior")
def sample_stats_to_xarray(self):
    """Extract sample_stats from posterior."""
    posterior = self.posterior
    data = get_sample_stats_stan3(posterior, ignore="lp__")
    return dict_to_dataset(data, library=self.stan, coords=self.coords, dims=self.dims)
```

```
@requires("prior")
def sample_stats_prior_to_xarray(self):
    """Extract sample_stats_prior from prior."""
    prior = self.prior
    data = get_sample_stats_stan3(prior, ignore="lp__")
    return dict_to_dataset(data, library=self.stan, coords=self.coords, dims=self.dims)
```

and for log_likelihood

```
@requires("posterior")
def log_likelihoods_to_xarray(self):
    """Store log_likelihoods and lp in log_likelihoods group."""
    fit = self.posterior
    model = self.posterior_model

    # log_likelihood values
    log_likelihood = self.log_likelihood
    if log_likelihood is not None:
        if isinstance(log_likelihood, str):
            log_likelihood = [log_likelihood]
        if isinstance(log_likelihood, (list, tuple)):
            log_likelihood = {name: name for name in log_likelihood}
        data = {
            obs_var_name: get_draws_stan3(fit, model=model, variables=log_like_name)[
                log_like_name
            ]
            for obs_var_name, log_like_name in log_likelihood.items()
        }

       data["lp"] = get_sample_stats_stan3(fit, variables="lp__")["lp"]

        return dict_to_dataset(data, library=self.stan, coords=self.coords, dims=self.dims)
    return None
```

OriolAbril:
Ready for review! I will create an issue to keep in mind to update cmdstan and cmdstanpy too in a following PR.

OriolAbril:
Normal parameter will work for sure, I will check in detail what happens with log likelihoods named lp, I think that currently, the log likelihood data would be overwritten with the lp data.

OriolAbril:
Now all three should work in the same way if a log likelihood is named "lp": raise a warning and overwriting this log likelihood with actual lp data.

canyon289:
@ColCarroll @junpenglao @aseyboldt 

If any of you get the chance could you review `io_pymc3.py`. I'm less familiar with pymc3 internals so could we could use your double check!
https://github.com/arviz-devs/arviz/pull/794/files#diff-77d75b54eeb8138e75e51d211dd3f7c4

arviz/data/base.py
```diff
@@ -20,7 +26,8 @@ def __call__(self, func):  # noqa: D202
         def wrapped(cls, *args, **kwargs):
```
aseyboldt: Add `functools.wrap`?

arviz/data/io_dict.py
```diff
@@ -44,7 +49,7 @@ def posterior_to_xarray(self):
         if "log_likelihood" in data:
             warnings.warn(
                 "log_likelihood found in posterior."
-                " For stats functions log_likelihood needs to be in sample_stats.",
+                " For stats functions log_likelihood needs to be in log_likelihoods.",
```
aseyboldt: Strange error message, I might be confused if I read that.

arviz/data/io_pymc3.py
```diff
@@ -34,47 +42,41 @@ def __init__(
             )
             else None
         )
+        if self.observations is not None:
```
ColCarroll: this should replace line 33 above, but with `if self.model is not None: ...`

arviz/data/io_dict.py
```diff
@@ -115,6 +137,26 @@ def observed_data_to_xarray(self):
...
+        data = self.constant_data
+        if not isinstance(data, dict):
+            raise TypeError("DictConverter.constant_data is not a dictionary")
+        if self.dims is None:
+            dims = {}
+        else:
+            dims = self.dims
+        constant_data = dict()
+        for key, vals in data.items():
+            vals = np.atleast_1d(vals)
```
aseyboldt: Why `atleast_1d`? Woudn't a scalar also be reasonable?

arviz/data/io_pymc3.py
```diff
@@ -125,27 +129,38 @@ def posterior_predictive_to_xarray(self):
...
-        return dict_to_dataset(
-            {k: np.expand_dims(v, 0) for k, v in self.prior.items()},
-            library=self.pymc3,
-            coords=self.coords,
-            dims=self.dims,
-        )
+        # workaround due to pymc#3588, to do, divide between prior and prior predictive
+        # probably after solving this issue or while solving it.
+        model_vars = []
+        if self.trace is not None:
```
ColCarroll: it looks like you construct `model_vars` here and in `constant_data_to_xarray`, but protect against different things in each (for example, I am nervous that we check whether `self.trace is not None` here, but not there).  It is also a little hard to figure out that `constant_data_to_xarray` wants all the elements *not* in `model_vars`.

Maybe factor out a small helper function to construct that list?

arviz/data/io_pymc3.py
```diff
@@ -34,47 +42,41 @@ def __init__(
...
+            chain_likelihoods = []
+            for chain in self.trace.chains:
+                log_like_chain = [
+                    self.log_likelihood_vals_point(point, var, log_like_fun)
+                    for point in self.trace.points([chain])
+                ]
+                chain_likelihoods.append(np.stack(log_like_chain))
+            if var.name == "lp":
+                warnings.warn(
+                    "Found variable named 'lp'. Its likelihood will be overwritten by the "
```
aseyboldt: Can't we avoid that?

arviz/data/io_pymc3.py
```diff
@@ -34,47 +42,41 @@ def __init__(
...
+        if self.observations is not None:
+            self.observations = {obs.name: obs.observations for obs in self.model.observed_RVs}
 
         import pymc3
 
         self.pymc3 = pymc3
 
+    def log_likelihood_vals_point(self, point, var, log_like_fun):
+        """Compute log likelihood for each observed point."""
+        log_like_val = np.atleast_1d(log_like_fun(point))
```
aseyboldt: Why atleast_1d? If the variable is a scalar, then so should the logp. We shouldn't just invent dimensions.

arviz/data/io_pymc3.py
```diff
@@ -34,47 +42,41 @@ def __init__(
...
-        # This next line is brittle and may not work forever, but is a secret
-        # way to access the model from the trace.
-        model = self.trace._straces[0].model  # pylint: disable=protected-access
-        if len(model.observed_RVs) != 1:
-            return None, None
-        else:
-            if self.dims is not None:
-                coord_name = self.dims.get(
-                    "log_likelihood", self.dims.get(model.observed_RVs[0].name)
+        """Compute log likelihood of each observation."""
```
aseyboldt: Where does the final combined logp come from? Usually that should be in the sampler stats already.

aseyboldt:
Oh, and stan drops constants in the `lp__`, right?

ahartikainen:
Stan might drop constants, depending how ones writes  the program

Drops: `y ~ normal(0,1);`
Keeps: `target += normal_lpdf(y | 0,1);`

arviz/data/base.py
OriolAbril: I only modified slightly the behaviour of `requres` class and tried to modify it as little as possible. I can definitely change it to use `functools.wrap` unless @aloctavodia, @ColCarroll, @ahartikainen, @canyon289 have some reason against it. 

For example, in `az.interactive_backend`, `contextlib` cannot be used as it does not work properly with matplotlib backend changes, thus, I followed [`mpl.rc_context`](https://matplotlib.org/3.1.1/_modules/matplotlib.html#rc_context). I can't tell if there is some reason that I cannot see behind `requires` implementation.

arviz/data/io_dict.py
OriolAbril: using `be in 'log_likelihoods' group` would be clearer?

arviz/data/io_dict.py
OriolAbril: I copied the code from `observed_data_to_xarray` above without thinking too much about it. I will check how are scalar values stored in `observed_data` and `constant_data` groups and if there are extra dimension I will modify both.

arviz/data/io_pymc3.py
OriolAbril: I'll check the behaviour too.

arviz/data/io_pymc3.py
OriolAbril: `log_likelihoods` group contains the `logp_elemwise` of each observed variable stored with the variable name (like `posterior_predictive` and `observed_data` groups that have matching variables. And in addition, it stores the `model_logp` stored as `lp`.

arviz/data/io_pymc3.py
OriolAbril: Definitely, how do you propose we handle it? Maybe we can discuss it in slack too, whatever you prefer.

arviz/data/io_dict.py
OriolAbril: Or whatever group name we agree on eventually

OriolAbril:
> Isn't the name `log_likelihoods` confusing? It isn't really the "likelihood" P(y|phi) after all. Is it just per variable conditioned on all others? 

In the case of a single value, it would store the pointwise log likelihood, so that log(P(y|phi)) is the sum of the values stored here. If there are several variables, it stored the pointwise values of each variable independently (i.e. the variable `x` would be `log(P(x_i|phi_s))` and the variable `z` would be `log(P(z_j|phi_s))` with `s` indicating the sample) so that log(P(y|phi)) would be the sum of all the variables stored there but `lp`. And then `lp` would be the sum of the the prior log(P(phi)) and log(P(y|phi)). 

We did some brainstorming in slack to choose the name, but everything is still open to ideas, so feel free to propose another name. I hope the explanation was clear enough.

> How does it handle transformations?

Transformations are not handled by ArviZ at all, the `from_xyz` methods only retrieve data from the sampling library which is where transformations would be handled.

ahartikainen:
I think we can keep this still open and let's do some research what would be the best structure.

ahartikainen:
Should we close this PR and create another one for the likelihood/probability/density/whatever new group we are planning?

OriolAbril:
I was planning on rebasing the PR to remove commits related to `constant_data` groups and use this same PR.

ahartikainen:
That sounds good too. Then we will have this discussion in one place.

rpgoldman:
@OriolAbril 

If you see my comments on the gitter, you will see that I had trouble with this patch when importing a trace from pymc3 that had observed variables with missing values.  The problem was that the log likelihood Data Array squeezed out the missing data, causing dimension mismatch.  I think the following tweak will fix this problem:

```
    def log_likelihood_vals_point(self, point, var, log_like_fun):
        """Compute log likelihood for each observed point."""
        log_like_val = utils.one_de(log_like_fun(point))
        if var.missing_values:
            log_like_val = np.where(var.observations.mask, 0.0, log_like_val)
        return log_like_val
```

I replaced the *removal* of missing observations with zero entries.  I *think* this is right, because these are logs and will be added, so zero is the identity element. But please have a look.

arviz/stats/stats.py
```diff
@@ -439,15 +440,11 @@ def loo(data, pointwise=False, reff=None, scale="deviance"):
...
-            raise TypeError(
-                "Must be able to extract a {group} group from data!".format(group=group)
-            )
-    if "log_likelihood" not in inference_data.sample_stats:
-        raise TypeError("Data must include log_likelihood in sample_stats")
+    if not hasattr(inference_data, "posterior"):
+        raise TypeError("Must be able to extract a posterior group from data!")
     posterior = inference_data.posterior
-    log_likelihood = inference_data.sample_stats.log_likelihood
+    log_likelihood = _get_log_likelihood(inference_data)
```
rpgoldman: I believe that this must be changed to 
```
log_likelihood = _get_log_likelihood(inference_data, "lp")
```
Otherwise I get an error doing `az.compare` or `az.waic` on a dataset with multiple observed variables -- get log likelihood complains that its `var_name` argument cannot be none if there are multiple variables in the `log_likelihood` group of the `InferenceData`.  I believe in this case one wants the composite log-likelihood from the special `lp` variable.

OriolAbril: We do have to look into this, I am not sure what would be correct in these cases with multiple variables though. Another possibility would be to calculate loo/waic for each variable or simply to stack together all variables in `log_likelihoods` except `lp`. Intuitively, using `lp` feels weird as you have the same value for all observations, it only depends on the draw and chain, so I would not know how to interpret the pointwise loo/waic values.

I remember discussing this at some point with @ahartikainen and @aloctavodia, maybe you have had time to look into this? I still have not been able to work out the maths nor experiment at all.

OriolAbril:
Thanks for catching this! I would have surely missed this issue.

OriolAbril:
> I replaced the _removal_ of missing observations with zero entries. I _think_ this is right, because these are logs and will be added, so zero is the identity element. But please have a look.

I have decided on `nan` because assigning a float to them would mean assigning the observation a likelihood/probability. Moreover, in general I expect all other values to be negative (they would all be if the likelihood was normalized) so using 0 would actually consider the missing values the most probable ones. I haven't checked but it looks like it would end up in weird results of loo/waic.

rpgoldman:
@OriolAbril I'm hoping someone with more of a statistics background than me can speak to this.  McElreath's book defines

_WAIC = -2(lppd - p<sub>WAIC</sub>)_

He defines _lppd_ and _p<sub>WAIC</sub>_ separately, but not in a way that seems to agree with the textual presentation:

> The log-pointwise-predictive-density is the total across observations of the logarithm of the average likelihood of each observation. (p. 191)

I think for observations  _o<sub>i</sub>_ from _1..n_ and samples _s<sub>j</sub>_ from _1..s_ this should be:

_lppd = ∑<sub>1 ≤ i ≤ n</sub> 1/s ( ∑<sub>1 ≤ j ≤ s</sub> log P(o<sub>i</sub>|s<sub>j</sub>)  )_

My concern about this is that the observations of different variables are not independent.  So if _o<sub>i</sub> = o<sub>i</sub><sup>1</sup> ... o<sub>i</sub><sup>k</sup>_ then I believe the conditional log prob above should be summed across these _o<sub>i</sub><sup>1</sup> ... o<sub>i</sub><sup>k</sup>_.  This makes me worried, because what do we do if the different variables making up _o_ have different shapes?  This is what I meant earlier, when I talked about the log prob of a cross product of the observed variables.

Then he defines _p<sub>WAIC</sub>_ as the variance in log probability over the set of observations.  It's too hard to fight with Markdown, so I'm not going to spell out this equation, but I think it's a straightforward variant of the previous one.

Does this seem right? And if so, what do we do about the fact that it's not obvious how to take the cross-product of observations?

rpgoldman:
@OriolAbril Given the issues with multi-variate model comparison, how about we add arguments to `compare` and `waic` so that users can compare models with respect to a particular data variable?

It's frustrating to get this error message:
```
TypeError: Found several log likelihood arrays ['z(Gini index)', 'per minor edu spend', 'z(OECD social spending pct)', 'Educational Outcomes', 'WVS: Educating Women'], var_name cannot be None
```
...without being able to pass `var_name`!

Alternatively, one might intercept this error in `compare()` and raise an error with a more helpful error message, E.g., 
```
TypeError: Found several log likelihood arrays ['z(Gini index)', 'per minor edu spend', 'z(OECD social spending pct)', 'Educational Outcomes', 'WVS: Educating Women'].  Cannot compare models with multi-variate observations.
```

or 

```
Cannot compute WAIC for models with multi-variate observations.
```

At least it's better to tell the user they are lost, instead of leaving them thinking they can solve the problem.

OriolAbril:
Ready for review (tests should pass too). I update the description of what this PR does and what will still be pending. 

#### Does
* Supports storage of multiple observed variables each with its own log likelihood data
  * dict
   * emcee
   * pymc
   * pystan
* Updates loo, waic... to get the log likelihood data from the new group. To do so, a function `get_log_likelihood` has been added in stats_utils. From here on, log likelihood values should be accessed using this helper function (it could even become public or a method of inference data, we have to think about this)

#### Pending
* support storage of multiple observed values
  * cmdstan
  * cmdstanpy
  * numpyro
  * pyro
* update ic functions to use the data available
  * add a var_name/obs_name to waic, loo, compare...
  * take into account diferent cases such as an ic value per observation, considering all observations a single variable...
* add new group to inference data schema

rpgoldman:
**Question**: what's the new group you plan to add to the `InferenceData` schema? (in "Pending")

rpgoldman:
Is there some way to make the group usage less error-prone? In particular, I am worried that we have  to hard-code string constants everywhere. That seems very likely to cause bugs.

Is there some way to limit group usage to known group names to avoid this kind of error?

OriolAbril:
We could add some kind of control over group names in the InferenceData init method and in io_netcdf. All the other functions (even from_dict) do not allow group names outside the ones in the inference data schema.

rpgoldman:
Should we have some kind of `ID_GROUPS` constant or something that we could use for checking?

arviz/data/io_pymc3.py
```diff
@@ -92,40 +93,35 @@ def arbitrary_element(dct: Dict[Any, np.ndarray]) -> np.ndarray:
...
+        log_likelihood_dict = {}
+        for var, log_like_fun in cached:
+            chain_likelihoods = []
+            for chain in self.trace.chains:
+                log_like_chain = [
+                    self.log_likelihood_vals_point(point, var, log_like_fun)
+                    for point in self.trace.points([chain])
+                ]
+                chain_likelihoods.append(np.stack(log_like_chain))
+            if var.name == "lp":
```
ahartikainen: Is this a problem anymore?

arviz/data/io_pystan.py
```diff
@@ -67,19 +68,32 @@ def sample_stats_to_xarray(self):
...
+        # log_likelihood values
+        log_likelihood = self.log_likelihood
+        if isinstance(log_likelihood, str):
+            log_likelihood = [log_likelihood]
+        if isinstance(log_likelihood, (list, tuple)):
+            log_likelihood = {name: name for name in log_likelihood}
+        data = {
+            obs_var_name: get_draws(fit, variables=log_like_name)[log_like_name]
+            for obs_var_name, log_like_name in log_likelihood.items()
+        }
```
ahartikainen: I don't know if it really matters, but we could call  `get_draws` "once"

```
data = {
            obs_var_name: get_draws(fit, variables=log_like_name)[log_like_name]
            for obs_var_name, log_like_name in log_likelihood.items()
        }
```

---> 

```
log_likelihood_draws = get_draws(fit, variables=list(log_likelihood.keys())
data = {
            obs_var_name: log_likelihood_draws [log_like_name]
            for obs_var_name, log_like_name in log_likelihood.items()
        }
```

?

arviz/data/io_pystan.py
```diff
@@ -233,22 +250,28 @@ def posterior_to_xarray(self):
...
+        log_likelihood = self.log_likelihood
+        model = self.posterior_model
+        if isinstance(log_likelihood, str):
+            log_likelihood = [log_likelihood]
+        if isinstance(log_likelihood, (list, tuple)):
+            log_likelihood = {name: name for name in log_likelihood}
+        data = {
+            obs_var_name: get_draws_stan3(fit, model=model, variables=log_like_name)[log_like_name]
+            for obs_var_name, log_like_name in log_likelihood.items()
+        }
```
ahartikainen: Same here as above, not sure if it really matters

arviz/data/io_pystan.py
```diff
@@ -451,15 +474,6 @@ def get_sample_stats(fit, log_likelihood=None):
...
         data[name] = values
 
-    # log_likelihood
-    if log_likelihood is not None:
-        log_likelihood_data = get_draws(fit, variables=log_likelihood)
-        data["log_likelihood"] = log_likelihood_data[log_likelihood]
-
-    # lp__
-    stat_lp = get_draws(fit, variables="lp__")
-    data["lp"] = stat_lp["lp__"]
```
ahartikainen: Are we putting `lp` back to `posterior` or keeping here? It should not go to `log_likelihood`

arviz/stats/stats_utils.py
```diff
@@ -386,6 +386,33 @@ def not_valid(ary, check_nan=True, check_shape=True, nan_kwargs=None, shape_kwar
...
 
+def get_log_likelihood(idata, var_name=None):
+    """Retrieve the log likelihood dataarray of a given variable."""
+    if hasattr(idata, "sample_stats") and hasattr(idata.sample_stats, "log_likelihood"):
+        warnings.warn(
+            "Storing the log_likelihood in sample_stats groups will be deprecated",
+            PendingDeprecationWarning,
+        )
+        return idata.sample_stats.log_likelihood
+    if not hasattr(idata, "log_likelihoods"):
```
ahartikainen: `log_likelihoods` or `log_likelihood`

arviz/data/io_pymc3.py
OriolAbril: not a problem anymore, fixed it

arviz/data/io_pystan.py
OriolAbril: I moved it to sample stats

arviz/stats/stats_utils.py
OriolAbril: good catch, thanks!

OriolAbril:
Ready to merge!!

VincentBt:
>     take into account different cases such as an ic value per observation, considering all observations a single variable...

How would we deal with the case where all observations are considered as a single variable? 
I understand from [this comment](https://github.com/arviz-devs/arviz/pull/794#issuecomment-522533923) and #987 that the likelihoods of each observed variable should simply be summed? 

That would mean replacing in the function `_get_log_likelihood` in `arviz/stats/stats_utils.py` the following part:
```
if len(var_names) > 1:
    raise TypeError(
        "Found several log likelihood arrays {}, var_name cannot be None".format(var_names)
    )
```
by the new code:
```
if len(var_names) > 1:  
    log_likelihood_sum = 0
    for var_name in var_names:
        log_likelihood_sum += idata.log_likelihood[var_name]
    return log_likelihood_sum
```

but I get with my model example with 2 observed variables the following warning:
```
UserWarning: For one or more samples the posterior variance of the log predictive densities exceeds 0.4. 
This could be indication of WAIC starting to fail. 
See http://arxiv.org/abs/1507.04544 for details
```
so that would mean that WAIC is not good anymore for more than one observed variable?

OriolAbril:
@VincentBt both waic and loo still have to be updated to use the new data they have available in log_likelihood group, they work only for single observations because their code has not been modified. There is a Google Summer of Code [project to fix this](https://github.com/arviz-devs/arviz/wiki/GSoC-2020-projects#support-multiple-observations), some arguments will probably be added to them such as var_names to calculate the value for different observations independently, a method argument to indicate that all observations must be merged in a single array and get its IC or to indicate that the model is non factorizable in which case https://arxiv.org/abs/1810.10559 should be followed... It is a work in progress :)

PR #804, opened 2019-09-02
Add minimal package versions for requirements.txt

ahartikainen:
Added minimal requirements for the packages 

Tested with python 3.6, some older packages are missing wheels for 3.7, but I don't think it is our problem).

I did not try older netcdf4, but I don't think we have had any problems with it.


ahartikainen:
@aloctavodia @ColCarroll or @canyon289 if you are ok with these, we can merge.

Here are the tests run with the minimal versions.

https://travis-ci.org/arviz-devs/arviz/builds/580166084

I still think we should test against the newest and let's fix these if we have issues with the versions.

aloctavodia:
LGTM

canyon289:
This PR seems to be breaking pypi upload.  Should I remove dev requirements in another pr?

image

https://stackoverflow.com/questions/54887301/how-can-i-use-git-repos-as-dependencies-for-my-pypi-package

ahartikainen:
Can you remove test from setup.py?

Issue #810, opened 2019-09-07
[FR] Support data from NumPyro 


fehiepsi:
## Tell us about it

Currently, we developed [numpyro](https://github.com/pyro-ppl/numpyro) as an alternative back-end to Pyro. The output of MCMC in numpyro is a numpy array so we can use `from_dict` api to convert to arviz inference data. However, we also allow users collect other fields such as `potential_energy`, `step_size`, `num_steps`,... which have different names in arviz. So it would be convenient for users to have a utility `from_numpyro`, which closes that gap.

I can work on a PR if arviz devs find this FR useful. :)

## Thoughts on implementation

I guess the implementation should be straightforward (because all objects from numpyro are numpy arrays) and similar to other `from_...` utilities .

cc @neerajprad

ahartikainen:
Yes, this sounds great.

fehiepsi:
Update TODO tasks from #811 

- [ ] compute obs's log_likelihood from posterior
- [ ] get observed data from posterior
- [ ] improve the schools test (the fix is available at https://github.com/pyro-ppl/numpyro/pull/332)

All of these are blocked by the next NumPyro release.

canyon289:
I think this one has been addressed. Thanks @fehiepsi

PR #811, opened 2019-09-09
Implement from_numpyro

fehiepsi:
This PR addresses #810.

I marked some TODOs for future PRs (when NumPyro MCMC object stores model arguments/keywords for the purpose of extracting log_likelihoods, observed_data). Currently, users need to use numpyro.handlers, which is not handy, to obtain these informations.

Currently, the names of [sample stats in NumPyro](https://numpyro.readthedocs.io/en/stable/mcmc.html#numpyro.mcmc.HMCState) are:
+ ``potential_energy``
+ ``num_steps``  (which can be converted to tree depth by using np.log2 function)
+ ``accept_prob``
+ ``mean_accept_prob`` (mean acceptance probability until current iteration during warmup adaptation or sampling)
+ ``diverging``
+ ``adapt_state.step_size``
+ ``adapt_state.inverse_mass_matrix``

Do I need to rename those fields in `sample_stats_to_xarray` function to match arviz conventions?

### DIFF
--- a/arviz/data/__init__.py
+++ b/arviz/data/__init__.py
@@ -11,6 +11,7 @@ from .io_pymc3 import from_pymc3
 from .io_pystan import from_pystan
 from .io_emcee import from_emcee
 from .io_pyro import from_pyro
+from .io_numpyro import from_numpyro
 from .io_tfp import from_tfp
 
 __all__ = [
@@ -30,6 +31,7 @@ __all__ = [
     "from_cmdstanpy",
     "from_dict",
     "from_pyro",
+    "from_numpyro",
     "from_tfp",
     "from_netcdf",
     "to_netcdf",
--- a/arviz/data/converters.py
+++ b/arviz/data/converters.py
@@ -7,6 +7,7 @@ from .base import dict_to_dataset
 from .io_cmdstan import from_cmdstan
 from .io_cmdstanpy import from_cmdstanpy
 from .io_emcee import from_emcee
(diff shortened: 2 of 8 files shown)


ahartikainen:
Maybe you could rename the variables that are common between stan and pymc3 so when we get more diagnostic plots we can have good defaults.

Also, you might add original names to attrs?

Maybe also have both tree depth and num_steps?

fehiepsi:
> rename the variables that are common between stan and pymc3

Good idea! Thanks @ahartikainen , I'll make corresponding changes.

canyon289:
Wow this is great. I accidentally created a mild conflict when I merged in another docs pr but it should be a straightforward fix :) Let me know if you need any help

canyon289:
One question I do have, Are you planning on fixing the todos in this pr or in a later pr?

fehiepsi:
Hi @canyon289 , currently, supporting these TODOs (extracting log_likelihood and observed_rvs) is not possible with NumPyro MCMC object (because we don't store model's args, kwargs). Though it is an easy fix, we have to wait for the next NumPyro release to solve these TODOs in arviz. I think I should mark this PR as WIP and wait for that release. :)

fehiepsi:
> make Github issues to track the todos

Hi @canyon289 , that sounds a good plan to me because these TODOs are just for additional features (the API will not change). It has taken us a few months just for the last release but we hope that from now on, we will have more frequent release cycles. ^^

PR #816, opened 2019-09-15
Notebooks fixup

canyon289:
Pyro example needed some stuff wrapped in `torch.Tensor` to work


canyon289:
Ack let me fix version number commit included in here by mistake

PR #817, opened 2019-09-15
Fix pyro example in notebook

canyon289:
Trying again

https://github.com/arviz-devs/arviz/pull/816


canyon289:
Self merging to keep the release train rolling. If I messed something up let me know

Issue #313, opened 2018-10-01
timeseries / regression plot


ahartikainen:
I think we need timeseries / regression plot.

Should it go under ppc plot?

We accept x and y 
x:
- ndarray (1D) = user provided numpy array
- str = user provided parameter (data, posterior or posterior predictive)

y:
- ndarray (1D --> (chain, draw, N) 3D) = user provided numpy array
- str = user provided parameter (data, posterior or posterior predictive)

There are multiple ways to visualize uncertainty:
- scatter (error-bars, violin-points)
- line (area fill, percentile lines)

ahartikainen:
Also, random draws from posterior are one good way to visualize the uncertainty. At least for static images.

Here was some discussion about the quantiles https://github.com/arviz-devs/arviz/issues/2

canyon289:
This visualization from Stitch Fix is a nice example in my opinion
https://multithreaded.stitchfix.com/assets/posts/2016-04-21-forget-arima/fig2.png
https://multithreaded.stitchfix.com/blog/2016/04/21/forget-arima/

utkarsh-maheshwari:
I believe line plots are good representation but the best representation would obviously depend on the type of data. I propose that in short term, we should focus on integrating line plots for time series analysis and later we can add more plots to the library. I would love to work on this feature.

OriolAbril:
I think that the best way to begin is probably by creating a small database of regression and timeseries models, maybe take examples from ROS https://avehtari.github.io/ROS-Examples/examples.html (it whould not be too much work to port to cmdstanpy or pystan) or using https://github.com/bambinos/Bambi_resources/tree/master/ROS and then see how they could be reproduced from ArviZ and InferenceData objects. There are many things to take into account for the plots and I think it will be useful to get a better picture of what could be supported to decide what will be supported and how.

utkarsh-maheshwari:
Sure. 
By now, I am not very familiar with R. I'll try to re-implement some examples of bambi.

OriolAbril:
You probably won't need to reimplement them, bambi already uses ArviZ, it is more than anything to get an idea of the different possibilities regarding regression and timeseries plots and to get familiar with ArviZ+xarray usage which can be quite different from ArviZ development where xarray does not play such an important role

utkarsh-maheshwari:
[...]

PR #823, opened 2019-09-17
Add tests for adding posterior_predictive without trace

### LINKED ISSUE #822: from_pymc3 errors out when given no trace
**Describe the bug**

If I invoke `az.from_pymc3(None, prior=prior_trace)` arviz raises an error.

**To Reproduce**
Here's a simple way to reproduce this:
```
#!/usr/bin/env python
# coding: utf-8

import pymc3 as pm
import arviz as az
import numpy as np
with pm.Model() as model:
    mu = pm.Normal('mu', mu=0, sigma=1)
    obs = pm.Normal('obs', mu=mu, sigma=1, observed=np.random.randn(100))

data = az.from_pymc3(None, prior=prior)
```

Here's the backtrace:
```
TypeError                                 Traceback (most recent call last)
<ipython-input-7-4df8674953f8> in <module>
----> 1 data = az.from_pymc3(None, prior=prior)

~/.virtualenvs/xplan-experiment-analysis/lib/python3.7/site-packages/arviz/data/io_pymc3.py in from_pymc3(trace, prior, posterior_predictive, coords, dims)
    224         posterior_predictive=posterior_predictive,
    225         coords=coords,
--> 226         dims=dims,
    227     ).to_inference_data()

~/.virtualenvs/xplan-experiment-analysis/lib/python3.7/site-packages/arviz/data/io_pymc3.py in __init__(self, trace, prior, posterior_predictive, coords, dims)
     26         )
     27         self.nchains = trace.nchains if hasattr(trace, "nchains") else 1
---> 28         self.ndraws = len(trace)
     29         self.prior = prior
     30         self.posterior_predictive = posterior_predictive

TypeError: object of type 'NoneType' has no len()
```

**Expected behavior**
Arviz should have created and returned an `InferenceData` object with only a `prior` group in it.

**Additional context**
Arviz 0.5.1

OriolAbril:
~Fixes #822~ and adds a test to check behaviour of `from_pymc3` when trace is `None`.

EDIT: Only adds the test as a follow-up to #833.

### DIFF
--- a/arviz/tests/test_data_pymc.py
+++ b/arviz/tests/test_data_pymc.py
@@ -152,3 +152,29 @@ class TestDataPyMC3:
         test_dict = {"posterior": ["beta"], "observed_data": ["obs"], "constant_data": ["x"]}
         fails = check_multiple_attrs(test_dict, inference_data)
         assert not fails
+
+    def test_no_trace(self):
+        with pm.Model():
+            x = pm.Data("x", [1.0, 2.0, 3.0])
+            y = pm.Data("y", [1.0, 2.0, 3.0])
+            beta = pm.Normal("beta", 0, 1)
+            obs = pm.Normal("obs", x * beta, 1, observed=y)  # pylint: disable=unused-variable
+            trace = pm.sample(100, tune=100)
+            prior = pm.sample_prior_predictive()
+            posterior_predictive = pm.sample_posterior_predictive(trace)
+
+        # Only prior
+        inference_data = from_pymc3(prior=prior)
+        test_dict = {"prior": ["beta", "obs"]}
+        fails = check_multiple_attrs(test_dict, inference_data)
+        assert not fails
+        # Only posterior_predictive
+        inference_data = from_pymc3(posterior_predictive=posterior_predictive)
+        test_dict = {"posterior_predictive": ["obs"]}
+        fails = check_multiple_attrs(test_dict, inference_data)
+        assert not fails
+        # Prior and posterior_predictive but no trace
+        inference_data = from_pymc3(prior=prior, posterior_predictive=posterior_predictive)
+        test_dict = {"prior": ["beta", "obs"], "posterior_predictive": ["obs"]}
+        fails = check_multiple_attrs(test_dict, inference_data)
+        assert not fails


rpgoldman:
I think this may still have issues.  I tried using it and then converting a prior trace, and I get an error where arviz tries to have the dimensions of a prior trace be `chains` and `draws`, but the only *actual* dimensions in the trace are the draws and the dimension of the variable in question.

I have looked and the problem seems to arise in the call to `numpy_to_data_array()`, which is in `data/base.py`, and automatically adds `chains` and `draws` to the dimensions of every variable.

The variable in question is a predictor (a function of an input variable), so is not replicated through the trace.

Here is the line of code that defines the variable whose dimensions cause the error:
```
ztemp = Deterministic("ztemp", inc_temp - 30)
```
The variable `inc_temp` is defined as:
```
inc_temp = Data("inc_temp", self.df["inc_temp"].to_numpy())
```
... so you can see that it does not vary over the course of the trace. I suppose I could suppress this variable, and just make it a `SharedVariable`, but that seems undesirable.

OriolAbril:
@rpgoldman I fear I do not know pymc3 well enough to follow you. Am I right to assume that the issue is that either `inc_temp` or `z_temp` are present in the prior dict resulting from `pm.sample_prior_predictive`? And that then ArviZ tries to assign them draw amd chain dimensions they do not have?

rpgoldman:
@OriolAbril I'm not sure if this is a problem with Arviz, or PyMC3.

The issue is that my prior trace contains within it some sampled values, with shape `(500, x)` for `x` the shape of the sampled variable.

The prior trace **also** contains some *un-sampled* predictor variables, with shape `x`, where as above `x` is the shape of the sampled variable.

I don't know whether this is a bug in PyMC3, or in the way I am *using* PyMC3, but it causes `from_pymc3()` to error out.

ahartikainen:
I think we should somehow "ignore" errors due to edgecases and maybe throw a warning?

rpgoldman:
@ahartikainen @OriolAbril  I just posted something on the pymc3 slack channel about this, because I'm not sure *what* is going wrong, and if it is my fault.  Having a deterministic function of a predictor variable in a pymc3 model seems uncertain in its effect.

rpgoldman:
Actually, this suggests a question about the general use of Arviz -- if predictors must be kept out of the trace objects, doesn't this hamper plotting with predictors and predictions?

OriolAbril:
Could this be related to https://github.com/pymc-devs/pymc3/issues/3588?

rpgoldman:
> @rpgoldman I fear I do not know pymc3 well enough to follow you. Am I right to assume that the issue is that either `inc_temp` or `z_temp` are present in the prior dict resulting from `pm.sample_prior_predictive`? And that then ArviZ tries to assign them draw amd chain dimensions they do not have?

Yes, that is exactly what is happening.

rpgoldman:
> Could this be related to [pymc-devs/pymc3#3588](https://github.com/pymc-devs/pymc3/issues/3588)?

Yes: the deterministic that is being included in the trace is not a `pm.Data`, but it *is* a simple function of a `pm.Data`.

pymc-devs/pymc3#3588 suggests that it's not sufficient to omit `pm.Data`: we must also omit all *functions of* `pm.Data` alone.  I.e., we should look at everything that has only `pm.Data` in its parents and omit those variables, as well as the `pm.Data` alone.

ahartikainen:
That sounds like a good solution.

rpgoldman:
It also sounds to me like I should have been using `from_dict` instead of `from_pymc3`, but that definitely didn't jump out at me as the right option.  Seems like`from_pymc3` should just farm out processing of posterior predictive and prior predictive to `from_dict`, and `from_pymc3` should just merge the functionality of the MultiTrace translator with `from_dict`.

Does that make sense?

It also suggests that there should be a `merge_inference_data()` function or a `merge()` method defined on `InferenceData` so that one could, for example, load an `InferenceData` from netcdf and then add a prior trace.

rpgoldman:
> Actually, this suggests a question about the general use of Arviz -- if predictors must be kept out of the trace objects, doesn't this hamper plotting with predictors and predictions?

Is this related to #313 ?  It seems like `InferenceData` doesn't have any way of incorporating predictors, and that my attempt to do so using `pm.Data` and `pm.Deterministic` is what's causing this pain.

ahartikainen:
>It also suggests that there should be a merge_inference_data() function or a merge() method defined on InferenceData so that one could, for example, load an InferenceData from netcdf and then add a prior trace. Maybe it should be merge_groups, instead.


See concat function that can join independent groups

https://github.com/arviz-devs/arviz/blob/master/arviz/data/inference_data.py

We need to add it to the docs

rpgoldman:
@ahartikainen Have a look at #825 ?

OriolAbril:
This does not fix #822 as it was already fixed in #833. I was going to close this PR but I realized that I had included a test to detect the issue in #822 so we can include this PR as a test for #833.

OriolAbril:
Ready to merge

ahartikainen:
LGTM

PR #828, opened 2019-09-18
WIP Add grid option to plotting functions.

rpgoldman:


rpgoldman:
Could someone help me with this?  I have just formatted with `black` on my home machine, and it made no changes, but the azure test is failing because the file is rejected as not black-formatted.

aloctavodia:
Hi @rpgoldman notice that you are adding a bunch of files that should not be added.

OriolAbril:
How would you feel about adding the `ax` argument present in most functions to `plot_density` and then either make public the default grid utilities of create an extra utility to generate axes? It would not be as straightforward as adding grid but I think it gives more options like overlaying plots or having a figure with different ArviZ plots in each subplot.

aloctavodia:
I think @OriolAbril proposal is a very good one.

rpgoldman:
I'm not exactly sure how @OriolAbril proposal would work.  We have a set of subplots, correct?  So wouldn't we need to pass an `Iterable` of `Axes` instead of a single one? This may be a naive question because I am not a `matplotlib` expert!

And I assume we should scan the other plotting functions in the API and decide whether to change them, as well. Yes?

Follow-up suggestion: maybe we could pull out  the code that builds the axis map into a separate function?  That might make the code easier to read -- right now I think there's a big bunch of code that mostly has the effect of building the `axis_map` variable.  For someone reading the code, if they could just skip over that and say "OK, now I have an axis map," it might be easier to maintain.  But that's a matter of individual preference.

OriolAbril:
I did a quick look to plotting functions, and only `plot_density`, `plot_forest`, `plot_joint` and `plot_trace` do not have an `ax` argument. We may be lacking some documentation on `ax` argument usage, but I think all the other functions which do have the `ax` argument iterate over it with `np.ravel(ax)`. The only exceptions would be functions that only accept one axes like `plot_kde`.

I also think we should try to use the same approach on all functions, unless someone does this before, I will work on the other 3 functions whenever possible.

rpgoldman:
> Hi @rpgoldman notice that you are adding a bunch of files that should not be added.

@aloctavodia Thank you for catching that!  I got rid of the `.mypy_cache` and put it in `.gitignore` to avoid this in the future.

I don't understand how I ended up with `pip-wheel-metadata`, though.  But I've removed them.

rpgoldman:
> I did a quick look to plotting functions, and only `plot_density`, `plot_forest`, `plot_joint` and `plot_trace` do not have an `ax` argument. We may be lacking some documentation on `ax` argument usage, but I think all the other functions which do have the `ax` argument iterate over it with `np.ravel(ax)`. The only exceptions would be functions that only accept one axes like `plot_kde`.
> 
> I also think we should try to use the same approach on all functions, unless someone does this before, I will work on the other 3 functions whenever possible.

If that's the case, do you want to drop this MR, and start a new one with a branch where we add axes arguments? (maybe `ax` for Axes and `ax_it` for Axes Iterable to avoid confusion?)

Also, if we are doing that, it would be nice to support `coords` here so that we can plot a subset of the distributions.  I have some regression models that are built along the lines of @junpenglao 's Schizophrenia model, which means I end up with variables with lots of rows, and it would be nice to be able to subset them.

OriolAbril:
> Also, if we are doing that, it would be nice to support `coords` here so that we can plot a subset of the distributions.

That would be great, I actually had no idea the `coords` argument was missing. In the meantime you can use `az.plot_density(idata.sel(**coords, inplace=False), ...)`.

About dropping the PR, whatever you prefer. 

Regarding the name, personally, I would not change it, as it would break a lot of function calls of users using `ax` argument, and in addition, the cases that accept an iterable also work with an axes instance when there is only one variable to plot, so there may still be some confusion. I would rather modify its docstring.

rpgoldman:
@OriolAbril I was thinking of doing a new PR because the name of this one would no longer be representative.  Also someone else might have an easier time than me working with the axes. But if no one has the time, LMK and I will give it a whack.

OriolAbril:
I have worked with axes before, so I can work on this whenever I have time (probably at some point this weekend).

rpgoldman:
If I get a chance, I'll start (and be happy for you to critique and improve!), but I'm pretty busy right now.

rpgoldman:
@OriolAbril I don't know what is the status of your alternative to this PR.  Would you mind letting me know?  And I could try to make some of the changes you suggest above, unless you are in the middle of it.

Also, what about the idea I floated above, that we might provide a general way to generate subplots that either all these display functions could use or (probably better) that the *user* could use to send a set of axes to arviz display functions?

If we can do one of these things, then this PR will be obsolete, and we can drop it.  But I'd rather be be sure we don't lose functionality before I do.

OriolAbril:
I did some similar changes in #835, but I did not modify `plot_density` as it was half done here, some utilities can probably be used here too though. The changes I made are still using array of axes. I think we have to give some more thought to returning xarrays of axes or using helper functions to ease overlaying plots.

rpgoldman:
Should I merge my changes into your #835 branch and we could work on it jointly there?

I don't actually know how that works on GitHub -- will it let me push to a branch on your repo?

OriolAbril:
It was already merged to master, so I think updating this branch with master is probably the best option

rpgoldman:
Will do, thanks.  One more request: will you please point me at a plot function that you think is a good example of the way things *should* be done, and I will try to conform to that pattern when I update this?

OriolAbril:
I think the most similar function to `plot_density` is `plot_forest` in the sense that both can take lists of inference data objects which makes axes and coords handling somewhat tricky.

aloctavodia:
@rpgoldman any update on this?

rpgoldman:
> @rpgoldman any update on this?

I thought this was dead, because the plan was to have the caller assemble the axes as desired, and pass them in, instead of controlling the way the plotting function generated axes.

I'm a little uncertain about what the plans are for making that happen.  There was some talk about using an xarray of axes for the purpose.  The issue as I see it is how to allow the caller to control how axes will be allocated to variables (`var_names`) and dimensions.  `xarray` could do that for us, or some more conventional Python associative data structure.

Has there been any discussion of such a general mechanism that you can point me to?

ahartikainen:
Hi, remember to `git pull` before working with the branch (I fixed merge conflicts)

OriolAbril:
The grid argument was eventually added in https://github.com/arviz-devs/arviz/pull/1459. We should probably still consider exposing axes creation helper functions specific to arviz/xarray objects.

PR #832, opened 2019-09-27
WIP: Add coords argument to plot_density

rpgoldman:
Not sure that this works properly yet. I'm kind of concerned about letting people specify coords to a function that accepts an *Iterable* of InferenceData, instead of just a single InferenceData.

Are we really sure we are confident in plotting multiple InferenceData? It seems like it might make the meaning of various parameters uncertain.  Mightn't it be better to just have the caller loop over the data sets instead of having arviz do it autonomously?

### DIFF
--- a/arviz/plots/densityplot.py
+++ b/arviz/plots/densityplot.py
@@ -1,8 +1,12 @@
 """KDE and histogram plots for multiple variables."""
 import warnings
+from typing import Dict, List, Any, Optional, Tuple, TYPE_CHECKING
 import numpy as np
 
 from ..data import convert_to_dataset
+
+if TYPE_CHECKING:
+    from ..data import InferenceData
 from ..stats import hpd
 from .kdeplot import _fast_kde
 from .plot_utils import (
@@ -11,6 +15,7 @@ from .plot_utils import (
     xarray_var_iter,
     default_grid,
     _create_axes_grid,
+    get_coords,
 )
 from ..utils import _var_names
 from ..rcparams import rcParams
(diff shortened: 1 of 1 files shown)


arviz/plots/densityplot.py
```diff
@@ -135,9 +144,13 @@ def plot_density(
...
     if not isinstance(data, (list, tuple)):
-        datasets = [convert_to_dataset(data, group=group)]
-    else:
-        datasets = [convert_to_dataset(datum, group=group) for datum in data]
+        data = [data]
+
+    if coords is None:
+        coords = {}
+    datasets = [
+        get_coords(convert_to_dataset(datum, group=group), coords) for datum in data
```
OriolAbril: I think this will probably break if one of the inference data objects has a coordinate the other doesn't, which currently is one possibility in `plot_density`. Using inference data [`sel`](https://github.com/arviz-devs/arviz/blob/master/arviz/data/inference_data.py#L122) method instead of dataset `sel` will solve the issue with coordinate names by applying only the existing ones, but the case of mismatching coordinate values would not be solved. 

As I do not think this will be a common case, maybe the best option is to create a helper function in `densityplot.py` itself in order to take this 2 possibilities into account?

rpgoldman: I'm not sure what to do about this.  Maybe it would help to articulate what limits we put on the set of `InferenceData` objects we accept here?  Without doing something like that, I don't see how we can say whether or not this function will plot correctly.

To be honest, I'm not sure what is gained by permitting a set of `InferenceData` here instead of requiring the user to invoke this function multiple times, once per `InferenceData` object.  Presumably the issue is for model comparison, but I'm afraid I'm just guessing.

Because I don't know what is the answer to that, I don't know how to think about what we need to support here.

canyon289:

canyon289:
Closing this for now. If its still active and needs to be reopened feel free to do so!

Issue #829, opened 2019-09-19
Can't change predictors in posterior_predictive trace


rpgoldman:
**Describe the bug**
A common thing to want to do is to train a model, and then use posterior predictive sampling to predict responses to new predictor values.
As far as I can tell, this cannot be done with the present version of Arviz.

Why is this?  Because when one builds an `InferenceData` object with a posterior trace and a posterior predictive trace that has a different set of predictors, the dimensions of the variables in the posterior predictive trace, in general, will not agree with the dimensions of the same variables in the posterior trace, because the predictor values are different.

One could imagine building an `InferenceData` object with *only* the posterior predictive trace, and no posterior trace.  This seems not to work, either.  The reason is that if one looks at the `posterior_predictive_to_xarray` method in `io_pymc3.py`, one sees that the interpretation of the posterior predictive trace argument is based on the `nchains` and `draws` properties of the translator, which are harvested from the posterior trace.

I will try to scratch together a reproducible test case, but the above should be enough to get started -- the issues are clear from looking at the source code.

## Big design issue

I believe that the issues in dealing with models that contain external values/predictors display at least two fundamental problems with `InferenceData`:

1. If we are to make predictions outside the training set, we cannot assume that the coordinates for the posterior trace group will be the same as the coordinates for the prior or posterior predictive groups, and equivalently, the shapes of the random variables may be different across the different groups.

2. An alternative would be to make it possible to create `InferenceData` objects that **only** contain prior predictive or posterior predictive traces.  But to do this,  `from_pymc3()` needs to be overhauled, because it makes many assumptions about the availability of the posterior trace, and relationships between it and the predictive traces.

ahartikainen:
Can you give an example?

What I think, is that there should be 1to1 relation between `posterior` draws and `posterior_predictive` "draws".

And similar thing applies to `prior` and `prior_predictive`

Currently we also have "1to1" relation with `posterior_predictive` and `observed_data`.

What you want is to also have other "unseen" `posterior_predictive` values --> yes I think we are missing this, and we should support it. Then we still should have 1to1 relation with `posterior` and we should have some convenient way to separate values related to `observed_data` and "missing_data".

rpgoldman:
Sure, here's my exact example:

I have data from synthetic bio experiments where different strains of cells are grown under different growth conditions.  The output measure is fluorescence, measured by flow cytometry.

So the predictors/independent variables are strain, growth conditions (temperature, growth medium, and population density).

[...]

PR #833, opened 2019-09-27
Improve import of posterior and prior predictive traces from PyMC3

rpgoldman:
Importing from PyMC3 in the absence of a posterior trace works  poorly -- `from_pymc3` uses its posterior trace argument, and a backdoor into the PyMC3 model object to inform its translation.  If the posterior trace is missing, it doesn't have enough information to interpret the posterior or prior trace dictionaries.

But one *definitely* needs these because out-of-sample prediction from a model may yield traces of different shape than the posterior trace.

This attempts to fix by allowing the user to specify a pymc3 model to `from_pymc3`.

### DIFF
--- a/arviz/data/io_pymc3.py
+++ b/arviz/data/io_pymc3.py
@@ -19,13 +19,12 @@ class PyMC3Converter:
         self.trace = trace
         # This next line is brittle and may not work forever, but is a secret
         # way to access the model from the trace.
-        self.model = (
-            None
-            if trace is None
-            else self.trace._straces[0].model  # pylint: disable=protected-access
-        )
-        self.nchains = trace.nchains if hasattr(trace, "nchains") else 1
-        self.ndraws = len(trace)
+        if trace is not None:
+            self.model = self.trace._straces[0].model  # pylint: disable=protected-access
+            self.nchains = trace.nchains if hasattr(trace, "nchains") else 1
+            self.ndraws = len(trace)
+        else:
+            self.model = self.nchains = self.ndraws = 0
         self.prior = prior
         self.posterior_predictive = posterior_predictive
         self.coords = coords
--- a/arviz/data/io_pymc3.py
+++ b/arviz/data/io_pymc3.py
@@ -1,22 +1,47 @@
 """PyMC3-specific conversion code."""
 import logging
+from typing import Dict, List, Any, Optional, TYPE_CHECKING
+
 import numpy as np
 import xarray as xr
-
 from .. import utils
 from .inference_data import InferenceData
 from .base import requires, dict_to_dataset, generate_dims_coords, make_attrs
 
+if TYPE_CHECKING:
+    import pymc3 as pm
+
 _log = logging.getLogger(__name__)
+
+Coords = Dict[str, List[Any]]
+Dims = Dict[str, List[str]]
 
 
 class PyMC3Converter:
     """Encapsulate PyMC3 specific logic."""
 
+    model = None  # type: Optional[pm.Model]
+    nchains = None  # type: int
+    ndraws = None  # type: int
+    posterior_predictive = None  # Type: Optional[Dict[str, np.ndarray]]
+    prior = None  # Type: Optional[Dict[str, np.ndarray]]
+
     def __init__(
-        self, *, trace=None, prior=None, posterior_predictive=None, coords=None, dims=None
+        self,
+        *,
+        trace=None,
+        prior=None,
+        posterior_predictive=None,
+        coords: Optional[Coords] = None,
+        dims: Optional[Dims] = None,
+        model=None
     ):
+        import pymc3
+
+        self.pymc3 = pymc3
+
         self.trace = trace
+
         # This next line is brittle and may not work forever, but is a secret
         # way to access the model from the trace.
         if trace is not None:
@@ -24,9 +49,32 @@ class PyMC3Converter:
             self.nchains = trace.nchains if hasattr(trace, "nchains") else 1
             self.ndraws = len(trace)
         else:
-            self.model = self.nchains = self.ndraws = 0
+            self.model = None
+            self.nchains = self.ndraws = 0
+
+        # this permits us to get the model from command-line argument or from with model:
+        try:
+            self.model = self.pymc3.modelcontext(model or self.model)
+        except TypeError:
+            self.model = None
+
         self.prior = prior
         self.posterior_predictive = posterior_predictive
+
+        def arbitrary_element(dct: Dict[Any, np.ndarray]) -> np.ndarray:
+            return next(iter(dct.values()))
+
+        if trace is None:
+            # if you have a posterior_predictive built with keep_dims,
+            # you'll lose here, but there's nothing I can do about that.
+            self.nchains = 1
+            aelem = (
+                arbitrary_element(prior)
+                if posterior_predictive is None
+                else arbitrary_element(posterior_predictive)
+            )
+            self.ndraws = aelem.shape[0]
+
         self.coords = coords
         self.dims = dims
         self.observations = (
@@ -43,10 +91,6 @@ class PyMC3Converter:
         )
         if self.observations is not None:
             self.observations = {obs.name: obs.observations for obs in self.model.observed_RVs}
-
-        import pymc3
-
-        self.pymc3 = pymc3
 
     @requires("trace")
     @requires("model")
@@ -215,7 +259,9 @@ class PyMC3Converter:
         )
 
 
-def from_pymc3(trace=None, *, prior=None, posterior_predictive=None, coords=None, dims=None):
+def from_pymc3(
+    trace=None, *, prior=None, posterior_predictive=None, coords=None, dims=None, model=None
+):
     """Convert pymc3 data into an InferenceData object."""
     return PyMC3Converter(
         trace=trace,
@@ -223,4 +269,5 @@ def from_pymc3(trace=None, *, prior=None, posterior_predictive=None, coords=None
         posterior_predictive=posterior_predictive,
         coords=coords,
         dims=dims,
+        model=model,
     ).to_inference_data()


arviz/data/io_pymc3.py
```diff
@@ -1,5 +1,8 @@
 """PyMC3-specific conversion code."""
 import logging
+from typing import Dict, List, Any, Optional, TYPE_CHECKING
+if TYPE_CHECKING:
+    import pymc3 as pm
```
OriolAbril: won't this break ArviZ for users without PyMC3 installed? Maybe it would be better a try/except?

rpgoldman: No, this will only break ArviZ for users without PyMC3 installed *who try to use mypy*.  

If you don't use mypy for type checking, then the `if TYPE_CHECKING` block never runs.

rpgoldman:
Now that this has passed, it would be great to get it merged, but also maybe put some discussion in a notebook.

arviz/data/io_pymc3.py
ahartikainen: Is it mypy which needs this import?

Is there a reason for typing stuff (At some point we should add it, but it is not currently implemented?)

Also this needs try except because we might run mypy without pymc3 installed

OriolAbril:
I think we should discuss how to deal with constant data (particularly how to deal with it when modified to get predictions) and with these predictions (which I am guessing is the main goal of this PR). 

I am sorry if I am repeating myself too much, but I see various drawbacks to the approach of storing the predictions in the posterior predictive group. 

To begin with, the posterior predictive group will stop matching the observed data group, thus preventing the inference data group from being used in many model checking scenarios. I fear that if the posterior predictive has to store either posterior predictive samples on observed data or on predictions, ArviZ may actually be making posterior predictive checks difficult instead of simple (in order to encourage a robust Bayesian workflow which should be one of our goals). I would set different constraints on them too, posterior predictive samples should have a 1on1 relation to posterior samples (which is why there is a warning when their shapes do not match) whereas this is not needed with predictions, with predictions we only need enough samples so that the kdes make sense.

Moreover, ArviZ should also ease and allow reproducibility of MCMC runs (even though we are not there yet, we are trying, i.e. #220). Currently, `constant_data` group is thought to store the constant data used (together with the observed data) to obtain the posterior samples, if the posterior predictive has been generated using different data, it should be stored too. In this area though, I think this proposes a very interesting approach, which would be to create a function taking an `InferenceData` object, a posterior predictive dict and a PyMC3 model and add 2 new groups to the inference data object _predictors_ and _predictions_ (with better names). `from_pymc3` could be updated as well to take one dict as `posterior_predictive` and another as `predictions` to generate the inference data object in one go, but we must think about handling of constant data. 

I know this probably looks more complicated on the user, but I think it will only be so at first.

arviz/data/io_pymc3.py
rpgoldman: Yes, it is mypy. The type  checking variable will be false when running.

I don't believe this needs try except because 

1. it seems to me to be fine if one cannot run the type checker without PyMC3 installed: we aren't requiring even developers to use mypy. 
2. I'm pretty sure that the error would be in mypy, not in arviz. And I don't think one should try to type check without PyMC3 installed.

I am typing stuff because it helps me to understand the code - it makes it more self documented, and because it helps me avoid introducing errors. It's gradual typing, so does not demand that all the code be annotated. I typically add annotations locally when trying to add a fix or enhancement. No one but me has  to change their coding practices, but I'd appreciate it if my annotations were retained  as long as they don't cause problems.

rpgoldman:
> I think we should discuss how to deal with constant data (particularly how to deal with it when modified to get predictions) and with these predictions (which I am guessing is the main goal of this PR).
> 
> I am sorry if I am repeating myself too much, but I see various drawbacks to the approach of storing the predictions in the posterior predictive group.

Minor note: I think the above should have "predictors," not "predictions."

Also agreed - is probably not workable to include traces with  different predictors in the same Inference data object (or the constant data would need to be held in a group, not at the top level).

I think my changes here support this: for out of sample predictions, we cannot store the predictive trace in the same Inference data object as the posterior trace, and obey the constraints you suggest.

That means we must have a way to populate inference data * without * a posterior trace. Since arviz uses the posterior trace to get at the model, which it needs to interpret any traces, arviz needs a different way to get a model for posterior predictive and prior predictive traces. That's what I'm trying to do here.

I agree about constant data, but I  think that's a long discussion, and I'd rather get this issue fixed right away, rather than wait for that whole issue to be resolved. But all your points below seem like good ones to me.

The one thing is that I don't know how we get observed data into an ID object, or how the plotters are expected to use it.

Maybe we could have an arviz discussion during one of the lab meetings?
> 
> To begin with, the posterior predictive group will stop matching the observed data group, thus preventing the inference data group from being used in many model checking scenarios. I fear that if the posterior predictive has to store either posterior predictive samples on observed data or on predictions, ArviZ may actually be making posterior predictive checks difficult instead of simple (in order to encourage a robust Bayesian workflow which should be one of our goals). I would set different constraints on them too, posterior predictive samples should have a 1on1 relation to posterior samples (which is why there is a warning when their shapes do not match) whereas this is not needed with predictions, with predictions we only need enough samples so that the kdes make sense.
> 
> Moreover, ArviZ should also ease and allow reproducibility of MCMC runs (even though we are not there yet, we are trying, i.e. #220). Currently, `constant_data` group is thought to store the constant data used (together with the observed data) to obtain the posterior samples, if the posterior predictive has been generated using different data, it should be stored too. In this area though, I think this proposes a very interesting approach, which would be to create a function taking an `InferenceData` object, a posterior predictive dict and a PyMC3 model and add 2 new groups to the inference data object _predictors_ and _predictions_ (with better names). `from_pymc3` could be updated as well to take one dict as `posterior_predictive` and another as `predictions` to generate the inference data object in one go, but we must think about handling of constant data.
> 
> I know this probably looks more complicated on the user, but I think it will only be so at first.

OriolAbril:
> Minor note: I think the above should have "predictors," not "predictions."

Sorry about that, I expressed myself incorrectly. First I meant "modified constant data"/predictors and then "out of sample posterior predictive samples"/predictions. 

> I think my changes here support this: for out of sample predictions, we cannot store the predictive trace in the same Inference data object as the posterior trace, and obey the constraints you suggest.

I agree too that it would be interesting to get some more or less working version of this sooner than later. My main comment for the patch is that I would rather add a new group (even though it may be renamed or modified in some time) than having 2 inference data objects whose groups store different information (hence my proposal in #829). 

This is basically for two reasons. Inference data groups are in great part what allow ArviZ to be backend agnostic, each `from_xyz` handles the storing of the information available in the correct category; having different data in the same group feels really wrong to me. Moreover, we will probably move data between groups or even rename some groups in order to properly handle refitting and log likelihood factors, so adding one more to the mix does not worry so much. I really think that one way or another predictions will have its own group, so its preliminary implementation should hint at that.

> Also agreed - is probably not workable to include traces with different predictors in the same Inference data object (or the constant data would need to be held in a group, not at the top level).

I actually see this as `sample_stats` or `observed_data`. By default, they refer to the `posterior` group. We have a `sample_stats_prior` to account for this, and also, `observed_data` has no relation whatsoever with prior samples even though it is at the top level. I see no problem in having a `predictions`/... group and then a `predictors`/`constant_data_predictions`/... group, both of them at the top level. 

> Maybe we could have an arviz discussion during one of the lab meetings?

I am totally onboard with that :)

rpgoldman:
Just a quick follow-up to this:

> I am sorry if I am repeating myself too much, but I see various drawbacks to the approach of storing the predictors in the posterior predictive group.

This will raise problems in the interaction with PyMC3, I believe, because *as a programming language*, PyMC3 does not have constructs that correspond to "predictors".  So it's not obvious how one pulls this information out of a model.  Maybe it would be a good feature to add to PyMC -- to clearly flag entities as predictors, and observations?

Other possible issues:

* There might be actuals that one wants to compare with predictions.  These might be observations, or they might  *not* be observations.

* The PyMC3 `Data` objects, IMO, do not work properly.  They can be used to change predictors or observations (note that this means you can't easily tell what they are), *but* if shapes are affected at all, everything breaks.  This seems fundamentally unsatisfactory.

I'm not certain I understand your proposal for additional groups inside `InferenceData`.  Would you consider writing this up in a little more structured way -- i.e., describing what your new `InferenceData` would look like -- specifically in terms of what groups are available, and how constant data, dims, and coords interact with these different groups, specifically in terms of in and out of sample predictions?

Thanks!

OriolAbril:
What feels natural to me would be the following (pseudocode):

```python
# Get posterior samples plus posterior predictive, prior and 
# prior predictive for model checking
with pm.Model() as model:
    # define model
    ...

    # run inference
    trace = pm.sample()
    post_pred = pm.sample_posterior_predictive(trace)

# Store posterior inferece results
idata = az.from_pymc3(trace, post_pred, prior, ...)
# Output
# InferenceData with groups
#    > posterior
#    > posterior_predictive
#    > prior
#    > sample_stats
#    > observed_data
#    > constant_data

# Use posterior samples for prediction
with model:
    pm.set_data(...)
    predictions = pm.sample_posterior_predictive(trace)

# Up until this point everything already works in ArviZ,  you can use the model 
# defined in #829 to get a working example of this groups and differently shaped 
# predictions.
# Below is my proposal of utilities to add

# Add predictions to inference data
idata.add_pymc3_predictions(model, predictions)

# Output
# InferenceData with the previous groups and:
#    > constant_data_predictions/predictors
#    > predictions
```

Using something like this approach allows ArviZ to store the predictions using nearly the same code as the one used for posterior predictive (except the shape mismatch warning) and it could also use `model` arg to get `constant_data_predictions` using the same code in `constant_data` (that is, storing there all `pm.Data` which are not observed)

> * The PyMC3 `Data` objects, IMO, do not work properly.  They can be used to change predictors or observations (note that this means you can't easily tell what they are), _but_ if shapes are affected at all, everything breaks.  This seems fundamentally unsatisfactory.

While experimenting with refitting, I actually had no issues with shape modifications _iff_ only `sample_posterior_predictive` is used. I only (and always) had issues trying to use `sample`.

rpgoldman:
Slightly OT (but relevant to the challenges of posterior predictive sampling:

> > The PyMC3 Data objects, IMO, do not work properly. They can be used to change predictors or observations (note that this means you can't easily tell what they are), but if shapes are affected at all, everything breaks. This seems fundamentally unsatisfactory.
> While experimenting with refitting, I actually had no issues with shape modifications iff only sample_posterior_predictive is used. I only (and always) had issues trying to use sample.

I ended up with troubles because of dependencies between the predictor shapes and observed variable shapes.  I think there is an issue about this on PyMC3 (not submitted by me). This occurred even when doing posterior predictive sampling.  I *think* it's because the shape of the predictors got baked into the theano graph.

To present a new set of predictors I had to build a new `pm.Model`, but after that I was able to use a saved trace from the old model, because it had only internal parameter nodes.

Ideally, one would be able to have the predictors and observations both in `pm.Data` and just use `set_data`, but this only works in some circumstances.

arviz/data/io_pymc3.py
```diff
@@ -1,33 +1,80 @@
...
         # way to access the model from the trace.
-        self.model = (
-            None
-            if trace is None
-            else self.trace._straces[0].model  # pylint: disable=protected-access
-        )
-        self.nchains = trace.nchains if hasattr(trace, "nchains") else 1
-        self.ndraws = len(trace)
+        if trace is not None:
+            self.model = self.trace._straces[0].model  # pylint: disable=protected-access
```
ColCarroll: should line 56 - 59 live up here, so that if `model` is provided, we do not use protected access?

ColCarroll:
Merging given the good point that this improves the codebase, and it seems like everyone is in agreement to continue this discussion at lab meeting.

arviz/data/io_pymc3.py
rpgoldman: I wasn't sure, because I thought that might break previous code that was carelessly written inside a `with Model():` construct (which used to be useless, but also harmless).
I think it's the right thing, but didn't want to make that change without approval.
I'd encourage you to make that patch. If you prefer for me to do it, LMK.

PR #834, opened 2019-09-27
Ignore mypy's cache directories.

rpgoldman:
Please merge this sooner, rather than later! It's benign and these cache directories are driving me crazy! Thanks!


rpgoldman:
Thank you!!!

PR #835, opened 2019-09-29
Add coords and ax arguments to all plotting functions

### LINKED ISSUE #813: [FR] Add ax argument to plot_forest
## Tell us about it
I've seen a couple of pull requests adding an `ax` argument to various `plot_X` functions (#773, #653, #162). This enables plotting on existing axes, giving users greater control of the plots (e.g., putting multiple plots in the same figure). Would it be possible to do this for `plot_forest` as well?

## Thoughts on implementation
I don't have much here. I don't think that I would be able to create PR for this myself, since I'm pretty unfamiliar with the arviz codebase, but I'd greatly appreciate this if this were eventually added.

edit: a complication might be that `plot_forest` can create several axes, not just one, but it's possible that other plots deal with this as well.

OriolAbril:
Most `plot_xyz` have `coords` and `ax` arguments, but there are some exceptions. This PR intends to add these arguments in all functions (if they make sense, `plot_compare` will continue not having `coords` argument). 

I have started with `plot_forest`, fixes #813. Similarly to what happened in `plot_density`, it can also accept a list of inferencedata objects, so I modified `get_coords`. I set it to accept 3 combinations: 

* dataset and coords dict, 
* list of datasets and coords dict
* list of dataset and list of coords dicts

This allows to handle idata objects with different dimensions, but I am not totally convinced by this approach. Any ideas will be welcome.

`ax` argument added to `plot_joint`

### DIFF
--- a/arviz/plots/jointplot.py
+++ b/arviz/plots/jointplot.py
@@ -20,6 +20,7 @@ def plot_joint(
     fill_last=True,
     joint_kwargs=None,
     marginal_kwargs=None,
+    ax_tuple=None,
 ):
     """
     Plot a scatter or hexbin of two variables with their respective marginals distributions.
@@ -51,6 +52,9 @@ def plot_joint(
         Additional keywords modifying the join distribution (central subplot)
     marginal_kwargs : dicts, optional
         Additional keywords modifying the marginals distributions (top and right subplot)
+    ax_tuple : tuple of axes, optional
+        Tuple containing (axjoin, ax_hist_x, ax_hist_y). If None, a new figure and axes
+        will be created.
 
     Returns
     -------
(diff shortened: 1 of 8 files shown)


arviz/plots/jointplot.py
```diff
@@ -20,6 +20,7 @@ def plot_joint(
     fill_last=True,
     joint_kwargs=None,
     marginal_kwargs=None,
+    ax_tuple=None,
```
ahartikainen: Maybe `axes`?

OriolAbril: I'd rather `ax` like the other functions or something clearly different like I tried. It it is okey I'll change it by `ax`

OriolAbril: It is conceptually different than the other functions but you can always overlay plots doing:

    ax = plot_xyz(data1)
    az.plot_xyz(data2, ax=ax)

so maybe for the end users there is not much difference

aloctavodia:
Is this ready to merge?

OriolAbril:
Yes! I was originally planning on addind `ax` argument to `plot_trace` but now it would be best to wait to do this. Therefore, unless there are any comments or reviews it is ready.

PR #838, opened 2019-10-05
Hpd intervals

aloctavodia:
Introduces a new argument `multimodal` to `az.hpd`, if `True` (defaults) it may compute more than 1 hpd interval if the distribution is multimodal and the modes are _well separated_. For `plot_posterior` `multimodal = True` by default and we may get something like:
lala

if `multimodal = False` (or the older behaviour)

lele

For all other functions/plots, like `summary`,`plot_forest` etc  `multimodal = False`, mainly because at the moment I am not sure how to handle more than one interval.

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -315,7 +315,7 @@ def hpd(ary, credible_interval=0.94, circular=False, multimodal=False):
 
     Parameters
     ----------
-    x : Numpy array
+    ary : Numpy array
         An array containing posterior samples
     credible_interval : float, optional
         Credible interval to compute. Defaults to 0.94.
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -12,6 +12,8 @@ from scipy.optimize import minimize
 import xarray as xr
 
 from ..data import convert_to_inference_data, convert_to_dataset, InferenceData
+from ..plots.kdeplot import _fast_kde
+from ..plots.plot_utils import get_bins
 from .diagnostics import _multichain_statistics, _mc_error, ess, _circular_standard_deviation
 from .stats_utils import (
     make_ufunc as _make_ufunc,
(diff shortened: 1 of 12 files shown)


arviz/stats/stats.py
```diff
@@ -305,27 +307,26 @@ def _ic_matrix(ics, ic_i):
     return rows, cols, ic_i_val
 
 
-def hpd(ary, credible_interval=0.94, circular=False):
```
junpenglao: I think it is still useful to have circular karg, it is different than multimodal right?

arviz/stats/stats.py
```diff
@@ -305,27 +307,26 @@ def _ic_matrix(ics, ic_i):
...
+def hpd(ary, credible_interval=0.94, multimodal=True):
```
junpenglao: I think it is still useful to have circular karg, it is different than multimodal right?

junpenglao:
Looking at the implementation a bit more, I suggest that we leave the hpd function as before, but create a new function just for multiple mode density region. And in plot_posterior if `multimodal = True` we call a different function. 
We can then also keep the circular kwarg as before from hpd function.
What do you think?

ahartikainen:
Do we have some way to address the uncertainty in kde?

Like this algo to give correct results we should have dynamic bw etc?

arviz/stats/stats.py
aloctavodia: The circular correction will work with a unimodal distribution, but not necessary with a multimodal one. The circular correction shifts the distribution by its mean in an attempt to move the interval away from the "borders" . Anyway I can add it again with a warning, or think of a less trickier way to compute it.

aloctavodia:
Why a new function instead of just an argument? Not sure how to estimate the uncertainty of the KDE. Dynamic bw is not necessary, but it could help increase the accuracy of the estimate. Tomás has a candidate to replace our kde implementation, that on average provides more accurate results and it seems to be fast, still we need to runbenchmark more benchmarks

junpenglao:
My reasoning of implementing a different function is that the two is based on different theory - unimodal is using CDF, where multimodal is based on a dynamic threshold on histogram (kind of remind me slice sampling).

PR #839, opened 2019-10-06
Add InferenceData class to docs

### LINKED ISSUE #711: Add InferenceData class to docs
Add the InferenceData class to the docs (in addition to the jupyter notebook tutorials on xarray, concatenation and variable transformation).

This will allow to easily access the docs of InferenceData related methods like `sel` or `concat` and it could also contain examples on common InferenceData object operations (i.e. concatenation of results from multiple runs of the same model, deleting a group or modifying/adding variables to a group).

I have no strong opinions on what should go there and what should go directly in the API page, for instance, `concat` could go directly in the API page, but with `sel` (and the future `del` may be similar) it would be strange to be outside.

OriolAbril:
Fixes #711. 

This PR adds an entrance to the API section named InferenceData which links to the following page:

image

At the same time, each entrance listed in the Methods section links to a page for each method. The one for `sel` looks like this:

image

This is achieved using the `class_members` template (as opposite to the `class` template used for `Numba` or `interactive_backend` classes). It also extends the documentation on `InferenceData` and its `__init__` and `sel` mehtods.

### DIFF
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -13,9 +13,9 @@ from ..rcparams import rcParams
 class InferenceData:
     """Container for inference data storage using xarray.
 
-    For a detailed introduction on ``InferenceData`` objects and its usage, see
-    :doc:`/notebooks/XarrayforArviZ`. This page is provides help and documentation
-    on ``InferenceData`` methods and its low level implementation.
+    For a detailed introduction to ``InferenceData`` objects and their usage, see
+    :doc:`/notebooks/XarrayforArviZ`. This page provides help and documentation
+    on ``InferenceData`` methods and their low level implementation.
     """
 
     def __init__(self, **kwargs):
@@ -174,7 +174,7 @@ class InferenceData:
         -------
         InferenceData
             A new InferenceData object by default.
-            When `inplace==True` perform selection inplace and return `None`
+            When `inplace==True` perform selection in place and return `None`
 
         Examples
         --------
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -11,19 +11,53 @@ from ..rcparams import rcParams
 
 
 class InferenceData:
-    """Container for accessing netCDF files using xarray."""
+    """Container for inference data storage using xarray.
+
+    For a detailed introduction on ``InferenceData`` objects and its usage, see
+    :doc:`/notebooks/XarrayforArviZ`. This page is provides help and documentation
+    on ``InferenceData`` methods and its low level implementation.
+    """
 
     def __init__(self, **kwargs):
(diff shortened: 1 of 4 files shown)


arviz/data/inference_data.py
```diff
@@ -11,19 +11,53 @@
 
 
 class InferenceData:
-    """Container for accessing netCDF files using xarray."""
+    """Container for inference data storage using xarray.
+
+    For a detailed introduction on ``InferenceData`` objects and its usage, see
```
rpgoldman: For a detailed introduction to ``InferenceData`` objects and their usage, see

arviz/data/inference_data.py
```diff
@@ -11,19 +11,53 @@
...
+    :doc:`/notebooks/XarrayforArviZ`. This page is provides help and documentation
```
rpgoldman: "This page provides help and documentation" (drop "is")

arviz/data/inference_data.py
```diff
@@ -11,19 +11,53 @@
...
+    on ``InferenceData`` methods and its low level implementation.
```
rpgoldman: its -> their

arviz/data/inference_data.py
```diff
@@ -119,19 +153,52 @@ def __add__(self, other):
...
         inplace : bool
             If True, modify the InferenceData object inplace, otherwise, return the modified copy.
         **kwargs : mapping
             It must be accepted by Dataset.sel()
+
+        Returns
+        -------
+        InferenceData
+            A new InferenceData object by default.
+            When `inplace==True` perform selection inplace and return `None`
```
rpgoldman: nit: "perform selection in place"

arviz/data/inference_data.py
```diff
@@ -11,19 +11,53 @@
 
 
 class InferenceData:
```
rpgoldman: It would be nice to add the `Attributes` section to the docstring for this class, since they are so important.

Am I right in thinking that the "standard" groups are `posterior`, `prior`, and `posterior_predictive`, but that the caller can add whatever groups they want?  Maybe explain this in the docstring (if you are documenting the `groups` attribute).

OriolAbril: I was intentionally leaving the attributes and groups out until we have had the ArviZ devs meeting and discussed about them, I'd rather add them in another PR (hopefully already including some way to store multiple log likelihood, out of sample predictions and the predictors used to get these out of sample predictions and any other modification to the schema).

arviz/data/inference_data.py
OriolAbril: Thanks for all the corrections!

rpgoldman:
I think this is ready to go, and urge you to merge it.

PR #840, opened 2019-10-07
Update build versions

OriolAbril:
Pull request trying to both fix failures with tensorflow and cmdstanpy and to update emcee versions used in tests.

@ahartikainen you should be able to work on cmdstanpy on this same branch so it is merged after all tests pass again.


ahartikainen:
there was some hidden char which failed the import. I rebase and (re-organized) some commits, so remember to pull if you still need to edit the files.

ahartikainen:
Ok, 0.5.0 did not yet update the StanFit class to StanMCMC, but the next version will contain it, so I will keep the docstrings as is and update the code when we update pypi. (that is before the next release for arviz)

ahartikainen:
LGTM

ahartikainen:
LGTM

PR #844, opened 2019-10-23
Update CmdStanPy class

ahartikainen:
Small fix.


mitzimorris:
re CmdStanGQ - method `sample_plus_quantities` returns a pandas DataFrame with columns for original sample, including sampler state, and new generated quantities variables, and method `generated_quantities` and `gnerated_quantities ` give you just the generated quantities values with no sampler state info - is this useful to Arviz?

CmdStanVB can produce a sample, but I don't think anyone uses that sample for anything.

ahartikainen:
For now, we can let user to handle GQ with `from_dict` if needed.

I need to think about what would be suitable to add GQ and how to merge it.

PR #846, opened 2019-10-28
fix bug in emcee blobs handling

OriolAbril:
emcee blobs generally have as dimensions `draw, chain, blob_dim1...` however, in the case of a single scalar blob it has shape `draw, chain`, this PR takes into account this possibility.


OriolAbril:
Should be ready to merge. Given that fixes for the ci are underway, I say we merge it right after them.

PR #847, opened 2019-10-28
Add chain_dim as from_tfp argument

### LINKED ISSUE #456: from_tfp with multiple chains?
How should `from_tfp` be used with multiple chains? It looks like it is only compatible with one chain, and it treats the different chains as different variables.

## Example

    import tensorflow as tf
    import tensorflow_probability as tfp
    import tensorflow_probability.python.edward2 as ed
    import numpy as np
    import arviz as az
    dtype = np.float32

    def unnormalized_log_prob(x):
        return -x**2.

    samples_gauss, _ = tfp.mcmc.sample_chain(
        num_results=1000,
        current_state=np.ones(20, dtype=dtype),
        kernel=tfp.mcmc.HamiltonianMonteCarlo(
            unnormalized_log_prob,
            step_size=1.0,
            num_leapfrog_steps=3),
        num_burnin_steps=500,
        parallel_iterations=4) 

    with tf.Session() as sess:
        [samples_gauss_, ] = sess.run([samples_gauss, ])

    az.from_tfp([samples_gauss_]).posterior 
    # It is expected to be a list with a variable (array) at each position

#### Output
    <xarray.Dataset>
    Dimensions:      (chain: 1, draw: 1000, var_0_dim_0: 20)
    Coordinates:
      * chain        (chain) int64 0
      * draw         (draw) int64 0 1 2 3 4 5 6 7 ... 993 994 995 996 997 998 999
      * var_0_dim_0  (var_0_dim_0) int64 0 1 2 3 4 5 6 7 ... 12 13 14 15 16 17 18 19
    Data variables:
        var_0        (chain, draw, var_0_dim_0) float32 0.38988703 ... -0.04516393
    Attributes:
    ...

### Workaround
My array has dimensions (ndraws, nchains), therefore simply transposing the array allows to call:

    from_array = az.convert_to_inference_data(samples_gauss_.T)
    az.plot_trace(from_array);

#### Output
Which as it can be seen, works properly.

image

I don't know if I am missing how to use the coords or dims parameter to somehow sort this out or if it is an implementation detail.

OriolAbril:
I a basically pushing this PR to see if it would fix #456, I have not had time to do much testing, I'll try to in the following days. If somebody feels like giving it a try in the meantime feel free!

### DIFF
--- a/arviz/data/io_tfp.py
+++ b/arviz/data/io_tfp.py
@@ -58,9 +58,13 @@ class TfpConverter:
             self.tf = tf  # pylint: disable=invalid-name
 
     def handle_chain_location(self, ary):
+        """Move the axis corresponding to the chain to first position.
+
+        If there is only one chain which has no axis, add it.
+        """
         if self.chain_dim is None:
             return utils.expand_dims(ary)
-        return ary.swapaxes(0, chain_dim)
+        return ary.swapaxes(0, self.chain_dim)
 
     def posterior_to_xarray(self):
         """Convert the posterior to an xarray dataset."""
@@ -128,7 +132,9 @@ class TfpConverter:
 
         data = {}
         with self.tf.Session() as sess:
-            data["obs"] = self.handle_chain_location(sess.run(posterior_preds, feed_dict=self.feed_dict))
+            data["obs"] = self.handle_chain_location(
+                sess.run(posterior_preds, feed_dict=self.feed_dict)
+            )
         return dict_to_dataset(data, library=self.tfp, coords=self.coords, dims=self.dims)
 
     def sample_stats_to_xarray(self):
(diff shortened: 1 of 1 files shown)


OriolAbril:
I have tried a couple of simple models and it seems to work. @junpenglao @jeffpollock9 it would be great if you could review it or simply try the code on some of your models. Possible changes that come to my mind, if they seem interesting to you please say so:
* accept a tuple as input of `chain_dim` with the same length as `var_names` so that each variable can have the chain dim at different positions
* add a `draw_dim` in addition for cases where the draw dimension is not the first

I'd like to point out that I have set the default `chain_dim` to `None` which assumes all data comes from single chain, so that it mimcs the current behaviour of `from_tfp`, however, maybe it would be better to have `chain_dim=1` as default? I assume the first dimension is always the draw dimension and in general the second would be the chain.

Thanks a lot!

OriolAbril:
This should be ready to merge too. I think it will be a useful feature so we will get people testing it. The default is set to `None` so it is not a breaking change, everything will work the same unless `chain_dim` is specifically used.

PR #848, opened 2019-10-28
Fix pystan adapt_engaged

ahartikainen:
Added `control=dict(adapt_engaged=False)`


ahartikainen:
Going with merge now.

PR #849, opened 2019-10-28
Skip Pyro tests temporarily

ahartikainen:
Until fix is provided for`from_pyro`.


canyon289:
Can marker be changed to `xfail` instead? That way if it passes somehow we get a test failure?

ahartikainen:
Yes, can you edit the branch?

Issue #845, opened 2019-10-24
ArviZ incompatible with pyro 0.5


OriolAbril:
**Describe the bug**
The most recent CI builds fail when using pyro 0.5. Here is what I think is the relevant part of the code to run the model for testing:

```python
...
nuts_kernel = NUTS(_pyro_conditioned_model, adapt_step_size=True)
posterior = MCMC(  # pylint:disable=not-callable
    nuts_kernel, num_samples=draws, warmup_steps=500
).run(_pyro_centered_model, sigma, y)

# This block lets the posterior be pickled
for trace in posterior.exec_traces:
    for node in trace.nodes.values():
        node.pop("fn", None)
posterior.kernel = None
...
```

I have seen that in the latest release `.run` does not return anything, however, moving it to the next line in the form of `posterior.run()` does not solve the issue as `exec_traces` is not an attribute of MCMC objects. Moreover, `exec_traces` is also used in `from_pyro`. Any ideas? i.e. @fehiepsi ?

fehiepsi:
@OriolAbril The new version of MCMC does not hold `exec_traces` anymore. Instead, all posterior samples can be obtained with [get_samples](http://docs.pyro.ai/en/stable/mcmc.html#pyro.infer.mcmc.api.MCMC.get_samples) method. So the `posterior_to_xarray` method will be simpler than before. But we will not be able to access `observed_data` using the latest Pyro version (because no trace is stored). We will make sure that it is doable in Pyro 1.0 (which is scheduled next month). cc @neerajprad

Do you want me to make a PR to update `from_pyro` for the latest version? If you are willing to do, I am happy to review it. :)

OriolAbril:
I don't have much time right now, so feel free to make a PR if you do, I will try to review it as soon as possible. 

Moreover, given that `io_pyro` has to be modified, it is probably time to set a minimum pyro version and make ArviZ only support recent ones? It may be too much work and not very useful to try to support everything given the differences in retrieving the data.

Also it we should update the CI that runs with an old version of pyro because it is a really old version.

fehiepsi:
Yeah, I'll make a PR soon. :)

fehiepsi:
I think we can close this issue.

PR #853, opened 2019-10-30
Update for pyro 0.5

fehiepsi:
Addresses #845. In this PR, I added
+ posterior_predictive
+ diverging stat
+ prior + prior_predictive
and remove `observed_data` because it is not accessible for now (will support it for the next Pyro release).

It is pretty complicated to make it compatible with the old Pyro API so I also drop that support.

cc @OriolAbril

### DIFF
--- a/arviz/data/io_pyro.py
+++ b/arviz/data/io_pyro.py
@@ -59,7 +59,7 @@ class PyroConverter:
         divergences = self.posterior.diagnostics()["divergences"]
         diverging = np.zeros((self.nchains, self.ndraws), dtype=np.bool)
         for i, k in enumerate(sorted(divergences)):
-            diverging[i, divergences[k]] = 1
+            diverging[i, divergences[k]] = True
         data = {"diverging": diverging}
         return dict_to_dataset(data, library=self.pyro, coords=self.coords, dims=self.dims)
 
@@ -115,8 +115,6 @@ def from_pyro(posterior=None, *, prior=None, posterior_predictive=None, coords=N
         Prior samples from a Pyro model
     posterior_predictive : dict
         Posterior predictive samples for the posterior
-    observed_data : dict
-        Observed data used in the sampling.
     coords : dict[str] -> list[str]
         Map of dimensions to coordinates
     dims : dict[str] -> list[str]
--- a/arviz/data/io_pyro.py
+++ b/arviz/data/io_pyro.py
@@ -4,43 +4,31 @@ from .base import dict_to_dataset
 from .. import utils
 
 
-def _get_var_names(posterior):
-    """Extract latent and observed variable names from pyro.MCMC.
-
-    Parameters
-    ----------
-    posterior : pyro.MCMC
-        Fitted MCMC object from Pyro
-
-    Returns
-    -------
-    list[str], list[str]
-    observed and latent variable names from the MCMC trace.
-    """
-    sample_point = posterior.exec_traces[0]
-    nodes = [node for node in sample_point.nodes.values() if node["type"] == "sample"]
-    observed = [node["name"] for node in nodes if node["is_observed"]]
-    latent = [node["name"] for node in nodes if not node["is_observed"]]
-    return observed, latent
-
-
 class PyroConverter:
     """Encapsulate Pyro specific logic."""
 
-    def __init__(self, *, posterior, coords=None, dims=None):
+    def __init__(self, *, posterior, prior=None, posterior_predictive=None, observed_data=None, coords=None, dims=None):
         """Convert pyro data into an InferenceData object.
 
         Parameters
         ----------
-        posterior : pyro.MCMC
+        posterior : pyro.infer.MCMC
             Fitted MCMC object from Pyro
+        prior: dict
+            Prior samples from a Pyro model
+        posterior_predictive : dict
+            Posterior predictive samples for the posterior
+        observed_data : dict
+            Observed data used in the sampling.
         coords : dict[str] -> list[str]
             Map of dimensions to coordinates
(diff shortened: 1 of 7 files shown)


arviz/tests/test_data_pyro.py
```diff
@@ -10,7 +10,6 @@
 )
 
 
-@pytest.mark.xfail(reason="Temporary xfail for Pyro tests.")
 class TestDataPyro:
```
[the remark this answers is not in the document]
fehiepsi: Yeah, it is possible. I’ll add it.

arviz/data/io_pyro.py
```diff
@@ -1,119 +1,123 @@
...
+                "sample_stats": self.sample_stats_to_xarray(),
+                "posterior_predictive": self.posterior_predictive_to_xarray(),
+                "prior": self.prior_to_xarray(),
             }
         )
 
 
-def from_pyro(posterior=None, *, coords=None, dims=None):
-    """Convert pyro data into an InferenceData object.
+def from_pyro(posterior=None, *, prior=None, posterior_predictive=None, coords=None, dims=None):
```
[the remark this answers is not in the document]
fehiepsi: Thanks for explaining! I understand how arviz works better now. Will address this...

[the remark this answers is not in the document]
fehiepsi: I decide to follow io_pymc and set nchains, ndraws to 0 if posterior=None. Is it reasonable to do so?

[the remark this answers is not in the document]
OriolAbril: It is, these values are only stored in the converter to check posterior predictive shape in io_pymc. I don't know if maybe in pyro it would make sense to check/reshape prior samples too?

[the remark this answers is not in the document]
fehiepsi: Thanks! Then I will stick with this pattern.

arviz/data/io_pyro.py
```diff
@@ -1,119 +1,131 @@
...
+_log = logging.getLogger(__name__)
 
 
 class PyroConverter:
     """Encapsulate Pyro specific logic."""
 
-    def __init__(self, *, posterior, coords=None, dims=None):
-        """Convert pyro data into an InferenceData object.
+    nchains = None  # type: int
+    ndraws = None  # type: int
```
OriolAbril: It looks like this is not used, is it for type hints?

arviz/data/io_pyro.py
```diff
@@ -1,119 +1,131 @@
...
+
+    def __init__(
+        self, *, posterior=None, prior=None, posterior_predictive=None, coords=None, dims=None
```
OriolAbril: I would allow `posterior` to be a positional argument. The * shoild go between posterior and prior

arviz/data/io_pyro.py
```diff
@@ -1,119 +1,131 @@
...
-    def observed_data_to_xarray(self):
-        """Convert observed data to xarray."""
-        from pyro.infer import EmpiricalMarginal
+    @requires("posterior")
+    def sample_stats_to_xarray(self):
+        """Extract sample_stats from Pyro posterior."""
+        divergences = self.posterior.diagnostics()["divergences"]
+        diverging = np.zeros((self.nchains, self.ndraws), dtype=np.bool)
+        for i, k in enumerate(sorted(divergences)):
+            diverging[i, divergences[k]] = 1
```
OriolAbril: Use `True` preferably

arviz/data/io_pyro.py
```diff
@@ -1,119 +1,131 @@
...
-            data = {}
-            for var_name in self.latent_vars:
-                # pylint: disable=no-member
-                samples = EmpiricalMarginal(
-                    self.posterior, sites=var_name
-                ).get_samples_and_weights()[0]
-                samples = samples.numpy().squeeze()
-                data[var_name] = utils.expand_dims(samples)
+        data = self.posterior.get_samples(group_by_chain=True)
+        data = {k: v.detach().cpu().numpy() for k, v in data.items()}
```
OriolAbril: If only one chain is used, would this work?

arviz/data/io_pyro.py
```diff
@@ -1,119 +1,131 @@
...
     Parameters
     ----------
-    posterior : pyro.MCMC
+    posterior : pyro.infer.MCMC
         Fitted MCMC object from Pyro
+    prior: dict
+        Prior samples from a Pyro model
+    posterior_predictive : dict
+        Posterior predictive samples for the posterior
+    observed_data : dict
```
OriolAbril: There is no observed_data argument, are you planning on adding it?

arviz/data/io_pyro.py
fehiepsi: I think so. I just mimic io_pymc3 file. :D

arviz/data/io_pyro.py
fehiepsi: I think so too. But this is a convention of `io_pystan` and `io_pymc3`. This is not important I guess.

arviz/data/io_pyro.py
fehiepsi: yes, this only converts pytorch tensor to numpy. The `group_by_chain=True` will make sure that the sample shape is nchains x ndraws x shape, even with 1 chain.

arviz/data/io_pyro.py
fehiepsi: Nice catch! I should remove it.

arviz/data/io_pyro.py
ahartikainen: Yeah, for the class init this is ok (no one really use that class directly)

PR #854, opened 2019-10-31
start updating and rewriting schema.md

OriolAbril:
This is still very preliminary, but please do comment and review. [Here](https://github.com/OriolAbril/arviz/blob/schema_update/schema.md) is the link to see the rendered version.


schema.md
```diff
@@ -1,82 +1,71 @@
...
 
-/initial_point
-    One point in parameter space for each chain, where that chain started.
-    TODO We could also store that as an attribute to each var in /trace
+#### `posterior`
+Samples from the posterior distribution p(theta|y).
 
-/tuning_trace?
-    Same as /trace, but optional. It contains the trace during tuning. Optional. (only if store_tune=True or so?)
+#### `sample_stats`
```
sethaxen: Do all existing supported packages conform to these conventions?

schema.md
```diff
@@ -1,82 +1,71 @@
...
 
-/divergences
-    Points in parameter space where the leapfrog starts that lead to a divergence (excluding tuning).
-    Does stan have access to that info? We could also just store the accepted point of a divergent trajectory.
+#### `posterior_predictive`
+Posterior predictive samples p(y|y) corresponding to the posterior predictive pdf evaluated at the `observed_data`. Samples should match with `posterior` ones and each variable should have a counterpart in `observed_data`. The `observed_data` counterpart variable may have a different name.
 
-/observed_data
-    All data that is used in observed variables (or the data (or transformed?) data sections in stan.)
+#### `constant_data`
```
sethaxen: I don't really follow this. An example shared between this and `constant_data` would be helpful.

schema.md
```diff
@@ -1,82 +1,71 @@
...
-/tuning_trace?
-    Same as /trace, but optional. It contains the trace during tuning. Optional. (only if store_tune=True or so?)
+#### `sample_stats`
+Information and diagnostics about each `posterior` sample, provided by the inference backend. It may vary depending on the algorithm used by the backend (i.e. an affine invariant sampler has no energy associated). The name convention used for `sample_stats` variables is the following:
+* `lp`: unnormalized log probability of the sample
+* `step_size`
+* `step_size_bar`
+* `tune`: boolean variable indicating if the sampler is tuning or sampling
+* `depth`:
+* `tree_size`:
```
sethaxen: I assume this is the same as PyMC3's stat of the same name. Is this equivalent to the number of steps taken?

schema.md
```diff
@@ -1,82 +1,71 @@
...
-    One point in parameter space for each chain, where that chain started.
-    TODO We could also store that as an attribute to each var in /trace
+#### `posterior`
+Samples from the posterior distribution p(theta|y).
 
-/tuning_trace?
-    Same as /trace, but optional. It contains the trace during tuning. Optional. (only if store_tune=True or so?)
+#### `sample_stats`
+Information and diagnostics about each `posterior` sample, provided by the inference backend. It may vary depending on the algorithm used by the backend (i.e. an affine invariant sampler has no energy associated). The name convention used for `sample_stats` variables is the following:
+* `lp`: unnormalized log probability of the sample
```
ahartikainen: `(unnormalized) joint log probability of the sample`

Also, is it `for sample` or `for model`

schema.md
```diff
@@ -1,82 +1,71 @@
...
+* [ArviZ](https://arviz-devs.github.io/arviz/) in Python which integrates with:
+  - [emcee](https://emcee.readthedocs.io/en/stable/)
+  - [PyMC3](https://docs.pymc.io)
+  - [pyro](https://pyro.ai/)
+      and [numpyro](https://pyro.ai/numpyro/)
+  - [PyStan](https://pystan.readthedocs.io/en/latest/index.html),
+      [CmdStan](https://mc-stan.org/users/interfaces/cmdstan)
+      and [CmdStanPy](https://cmdstanpy.readthedocs.io/en/latest/index.html)
+  - [tensorflow-probability](https://www.tensorflow.org/probability)
+* [ArviZ.jl](https://github.com/sethaxen/ArviZ.jl) in julia which integrates with:
```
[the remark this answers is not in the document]
aloctavodia: We have plans to add support for Variational Inference, also I would like to add support for Sequential Monte Carlo.

schema.md
sethaxen: e.g. CmdStan has the following stats: lp, accept_stat, stepsize, treedepth, n_leapfrog, diverging, energy. To conform to this schema, shouldn't stepsize=>step_size, treedepth=>depth, n_leapfrog=>tree_size, accept_stat=>mean_tree_accept?

schema.md
```diff
@@ -1,82 +1,71 @@
...
+  - [Turing](https://turing.ml/dev/).
```
sethaxen: ArviZ.jl also supports directly the Stan wrappers [CmdStan.jl](https://github.com/StanJulia/CmdStan.jl), [StanSample.jl](https://github.com/StanJulia/StanSample.jl), and [Stan.jl](https://github.com/StanJulia/Stan.jl), as well as indirectly any package that uses [MCMCChains.jl](https://github.com/TuringLang/MCMCChains.jl) to store chains.

schema.md
ColCarroll:
```suggestion
* [ArviZ.jl](https://github.com/sethaxen/ArviZ.jl) in Julia which integrates with:
```

schema.md
ColCarroll: this one is particular to NUTS: there is this doubling routine, followed by a sample from the tree once the stopping criteria is reached. I believe this number is `2 ** tree_depth`.

schema.md
ColCarroll: yep -- i believe we rename certain statistics from certain libraries to conform. for example, pymc3 (unfortunately) uses the overloaded term `sample` instead of `draw` to describe draws from the posterior distribution...

schema.md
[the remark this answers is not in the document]
OriolAbril: Maybe "implementation in progress" is better? I wasn't very strict as ArviZ itself is still beta

schema.md
OriolAbril: I started copying names there from categories in InferenceData examples. I was planning in listing everything that comes to our mind so we have a list of our name convention. 

Currently pymc3 and stan rename some quantities internally in `from_xyz` so that ArviZ always uses the same names, but it is not clearly explained anywhere (at least anywhere I could find)

schema.md
OriolAbril: I think it is for sample and model?

schema.md
OriolAbril: I'll add them!!

OriolAbril:
> My main feedback is that some of these will be easier to understand with examples.

Yep, I feel like this should probably added to the documentation too.

schema.md
rpgoldman: It seems odd to me that the information in the Arviz `sample_stats` does not contain the information in `pymc3.stats.summary()`, which is variable-wise. ,instead of just sample-wise.

Shouldn't we pull that information into `sample_stats`, as well?

schema.md
rpgoldman: I agree: I don't think that the distinction between observed and constant data is clear.  I *suspect* that what is intended is that observed data be "evidence" and constant data be "predictors."  I.e., if we look at the model as a graph, and interpret it causally, the `constant_data` are at the top, and the `observed_data` are at the bottom.

But consider the possibility of a model graph where there are *middle* nodes that are (partially) observed.  (E.g., there is a causal influence that may be observed, but may only be indirectly observed.)  In this case, I believe that the distinction breaks down -- or I am not understanding it right.

schema.md
sethaxen: In ArviZ.jl, I'm putting Turing's summary information into the attributes section, but if there's a better place for it, I'll put it there.

schema.md
```diff
@@ -1,82 +1,71 @@
...
-comment?: str
-model_version?: str
-timestamp: int
-author?: str
+## Current design
+`InferenceData` stores all quantities relevant in order to fulfill its goals in different groups. Each group, described below, stores a conceptually different quantity generally represented by several multidimensional labeled variables.
 
-/coords
-For each dimension, we store a corresponding variable here, that contains the labels for that dimension.
+Each group should have one entry per variable, with the first two dimensions of each variable should be the sample identifier (`chain`, `draw`). Dimensions must be named and explicit their index values, called coordinates. Coordinates can have repeated identifiers and may not be numerical. Variable names must not share names with dimensions.
```
rpgoldman: I'm being picky, but for some of these groups (e.g., `observed_data` and `constant_data`), chain and draw don't make sense.

Also, in the following sentence, I think that you mean "specify" instead of "explicit."

One last point: how can coordinates have repeated identifiers?  Wouldn't that break indexing by coordinate?

schema.md
```diff
@@ -1,82 +1,71 @@
...
-comment?: str
-model_version?: str
-timestamp: int
-author?: str
+## Current design
+`InferenceData` stores all quantites relevant in order to fulfill its goals in different groups. Each group, described below, stores a conceptually different quantity generally represented by several multidimensional labeled variables.
 
-/coords
-For each dimension, we store a corresponding variable here, that contains the labels for that dimension.
+Each group should have one entry per variable, with the first two dimensions of each variable should be the sample identifier (`chain`, `draw`). Dimensions must be named and explicit their index values, called coordinates. Coordinates can have repeated identifiers and may not be numerical. Variable names must not share names with dimensions.
```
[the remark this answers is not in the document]
rpgoldman: Yes, I find that I simply cannot get my memory to keep "dimension" and "coordinate" straight, because I don't think xarray's use of these terms lines up with how they are used when we talk.

schema.md
```diff
@@ -1,82 +1,71 @@
...
-                sym_pos_axis: [dim_idx1, dim_idx2]
-                interval_lower:
-                interval_upper
-                TODO How would this deal with cases where the lower and upper bounds depend on the index?
-
-
-
-TODO: In order to reproduce the run, it may make sense to also store some data on the random state (in numpy, this is a tuple of arrays), as
-well as some version info.  Hopefully just `PyMC3, (3, 4, 1)` or similar works.
+#### `constant_data_predictions`
```
rpgoldman: Minor point: suggest that we consistently name all of these attributes of `predictions` with `predictions_` as prefix.  So change this to `predictions_constant_data`.

schema.md
OriolAbril: I will correct the wording, thanks!

Regarding repeated identifiers, the radon example has repeated identifiers. This has some issues when using `sel` and `where` should be used instead generally, but for most plots it is not an issue

schema.md
OriolAbril: I like it

schema.md
rpgoldman: Thanks, I didn't know that.

schema.md
OriolAbril: Yes, it goes with the "Later" comment. What I understand as observed data (which basically comes from how is it currently implemented in ArviZ), all `data` which is used as `pm.<Distribution>(..., observed=data)` in PyMC3 or as `data ~ <distribution>(...)` in Stan goes to observed variable group, as they are modeled as random variables (without taking into account at all if they are then used as parameters of another random variable). Any other relevant data (relevant in the sense of needed to reproduce the analysis) should go to constant data.

schema.md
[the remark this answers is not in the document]
OriolAbril: Does the example solve the doubts about dimensions and coordinates? I am trying to follow xarray notation to the letter, I hope I have not changed concepts of place. Maybe it would also help an example here: 
> ...called coordinates. For example, a possible dimension would be `chain`, whose coordinates would be `[0, 1, 2, 3]`. Coordinates can have...

OriolAbril:
What I see missing before merging this PR is: adding a PyMC3 example and modifying the documentation to include (or link to) the schema description which I can do myself. Also properly defining the quantities stored in `sample_stats` which would be great if someone who already knows how to describe them steps in (I barely understand some of them so it's not ideal).

What I would leave for future PRs: trying to be consequent with `prior` and `prior_predictive` (i.e. PyMC3 or pyro sample both distributions at the same time and `from_pymc3/pyro` currently store all variables in `prior`), take care of the log_likelihood in `sample_stats_prior` (it is currently omitted and ends up stored in `prior`), add `predictions` and `predictions_constant_data` groups (which will require to update the schema examples and maybe the example datasets)

ahartikainen:
I think prior code should be updated a bit (also let's add prior for the sigma, so that there is sane prior). Also log_lik doesn't make any sense with prior.

```
linreg_prior_code = """
data {
  int<lower=0> N;
  real time_since_joined[N];
}

generated quantities {
    real b0;
    real b1;
    real log_sigma;
    real<lower=0> sigma;
    vector[N] slack_comments_hat;
    
    b0 = normal_rng(0,10);
    b1 = normal_rng(0,10);
    sigma = abs(normal_rng(0,10));
    log_sigma = log(sigma);
    for (i in 1:N) {
        slack_comments_hat[i] = normal_rng(b0 + b1 * time_since_joined[i], sigma);
    }
}
"""
linreg_prior_data_dict = {"N": N, "time_since_joined": time_since_joined}
sm_prior = pystan.StanModel(model_code=linreg_prior_code)
prior = sm_prior.sampling(data=linreg_data_dict, iter=200, chains=1, algorithm='Fixed_param', warmup=0)
```

OriolAbril:
I forgot about the prior on sigma, my bad. 

What would the difference with the current implementation be regarding this new code for the prior? I would rather keep the code as simple as possible for the example adding a second piece of Stan code may be confusing.

ahartikainen:
Parameters are not sampled from prior with fixed_param, see values

OriolAbril:
I have reorganized the `schema.md` document and the example notebooks and added them do the documentation. Links to the schema section have also been added to quickstart, cookbook and xarrayforarviz notebooks and the index section. 

For now this updated version of the docs can be checked in my fork: https://oriolabril.github.io/arviz/schema/schema.html.

Example notebooks have been set to use xarray html rich rendering which seems to work with sphinx.

OriolAbril:
I think this is ready to merge, we will continue improving it in following PRs.

PR #857, opened 2019-11-03
Bokeh plot_trace

ahartikainen:
WIP. There is so many things that need to be fixed. Also more tools etc.

In notebook / script:
```
import arviz as az
import bokeh.plotting as bkp

# if using notebook, else output is in temp html
bkp.output_notebook()

axes = az.plot_trace(idata, backend="bokeh")
```

Please try, comment, fix/edit (feel free to commit to the branch)


ahartikainen:
https://docs.bokeh.org/en/latest/

https://docs.bokeh.org/en/latest/docs/gallery.html

https://docs.bokeh.org/en/latest/docs/user_guide.html

https://docs.bokeh.org/en/latest/docs/dev_guide/models.html # see the structure

arviz/plots/distplot.py
```diff
@@ -83,6 +84,78 @@ def plot_dist(
     -------
     ax : matplotlib axes
     """
+    if backend is None or backend.lower() in ("mpl", "matplotlib"):
+        ax = _plot_dist_mpl(
```
canyon289: Being picky here, and maybe commenting too soon. I bet a lot of these keyword arguments could be constructed in a dict as theyre the same between both boken and mpl and just passed onto bokeh or mpl

canyon289:
I see arviz looking like this, and added more modifications to plot_kde (that are probably broken) to broken.

I think benefit of this is two fold

* Makes it easier to test math independently from plotting
* Makes it easier to read code as it separates math concern from plotting concern

image

ahartikainen:
That sounds good.

ahartikainen:
@canyon289  Some functions are currently missing (kde stuff), are they in your local codebase? (I can wait)

arviz/plots/distplot.py
```diff
@@ -83,73 +84,41 @@ def plot_dist(
...
-            fill_last=fill_last,
-            textsize=textsize,
-            plot_kwargs=plot_kwargs,
-            fill_kwargs=fill_kwargs,
-            rug_kwargs=rug_kwargs,
-            contour_kwargs=contour_kwargs,
-            ax=ax,
-        )
-    return ax
+    if backend is None or backend.lower() in ("mpl", "matplotlib"):
```
OriolAbril: How about reading `rcParams["plot.backend"]` when it is `None`?

ahartikainen: Yes, this sound good. Also there are now many things we should define for rcParams

ahartikainen:
I added points and lines for traceplot and also bloated the tools. Added also webgl.

ahartikainen:
We also need to define Linked selection

https://docs.bokeh.org/en/latest/docs/user_guide/data.html

Also I would want to create a "shadow" that show which parth of the kde of the data is shown when zooming into trace. Also selections from kde to trace and somehow trace to kde would be nice.

I don't mean we are keeping all of these, but let's explore.

ahartikainen:
code is currently broken, needs update for the cds (and some other problem also?)

ahartikainen:
Ok, I think it "works".

It really needs still warnings that something doesn't work.

Also minor tweaks, but maybe in another PR
(linked brushing, default tools, implement ColumnDataSource for linked updates, selection etc)
(better interface to interact) etc

Cleanup for commented out code

Update docstrings

canyon289:
@ahartikainen 
Once tests pass I think we should do a little bit of cleanup, add warnings to bokeh, and merge. Because this PR introduces the structure we can go one by one one the remaining plots to refactor into new backend format.

It also allows us to do the same with tests, where we can break `test_plots.py` into just tests of matplotlib, bokeh, and computation

ahartikainen:
Latest black 19.10b0 has fixed the trailing comma error.

ahartikainen:
This is ready, @canyon289 can you check/edit the exception messages?

canyon289:
@ahartikainen 
Since tests are passing I say we merge so we can work on the tickets here in parallel. There might be some bugs here and there that are untested but I consider the repo to be a "dev" state and we make no guarantees about stability there.

Once we merge I'll convert all these todos into issues.

image

PR #859, opened 2019-11-05
[WIP] simple mpld3 demo

aloctavodia:
This is an example of using mpld3, so far it seems simple things are simple to do. Like writing in a notebook `mpld3.display()` after a plot gives you a simple interactive plot (just zooming and panning) similar to what you get with `%matplotlib notebook`. You can also (un)comment `mpld3.display()` to switch between the interactive and static version of the plot. 

Here I am using a mpld3 plugin `InteractiveLegendPlugin` this requires just a few changes to the code, but surely I am breaking stuff so more changes are needed. It seems plugins are easy to write assuming you know javascript and D3. 

Still not sure about how easy will be to do more complex stuff like choosing a portion of the trace and getting the density/histogram for that portion.

Attached you will find a html file showing an example of a interactive traceplot.

[lolo.zip](https://github.com/arviz-devs/arviz/files/3809568/lolo.zip)

### DIFF
--- a/arviz/plots/distplot.py
+++ b/arviz/plots/distplot.py
@@ -115,7 +115,7 @@ def plot_dist(
         plot_kwargs.setdefault("color", color)
         legend = label is not None
 
-        plot_kde(
+        _, line = plot_kde(
             values,
             values2,
             cumulative=cumulative,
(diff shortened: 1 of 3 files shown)


canyon289:
Soooooooo adding this to the backends

aflaxman:
Hi! Cool to see `mpld3` is still useful.  It has become a bit abandoned in the last five years, though, and the lead developer has shifted to `altair` https://github.com/altair-viz/altair  It looks like you are aware of this alternative way to get to d3 through python (#92 ), but I thought I'd chime in to make sure the dots get connected.

aloctavodia:
Closing, as using mpld3, will most likely than note require us to maintain that library.

PR #860, opened 2019-11-08
use jitted histogram everywhere

aloctavodia:


canyon289:
LGTM

OriolAbril:
I have just seen many deprecation warnings when building the docs locally (they also happen in travis here but everything seems to work). It looks like they were introduced here. @Ban-zee are these deprecation warnings something to worry about?

Ban-zee:
A deprecation might denote that the particular function is not compatible
with the no python mode anymore which can lead to slowdown of the jitted
function

aloctavodia:
I still not find this when working with ArviZ, I will keep an eye on this and see if this need some change

PR #862, opened 2019-11-10
Feature/parallel coverals

ahartikainen:
https://docs.coveralls.io/parallel-build-webhook


ahartikainen:
Small change, but makes it easier to read coveralls results

PR #863, opened 2019-11-11
Add tests for bokeh

ahartikainen:
Bokeh tests and general fixes for bokeh.


arviz/tests/test_plots_bokeh.py
```diff
@@ -0,0 +1,93 @@
...
+        {"var_names": ["mu", "tau"]},
+        {"combined": True},
+        {"compact": True},
+        {"combined": True, "compact": True, "legend": True},
+        {"lines": [("mu", {}, [1, 2])]},
+        {"lines": [("mu", {}, 8)]},
+    ],
+)
+def test_plot_trace(models, kwargs):
+    axes = plot_trace(models.model_1, backend="bokeh", show=False, **kwargs)
```
canyon289: Aw I missed chance to comment but in future tests I dont think we should test plotting at the user interface anymore. We should import the bokeh plot function and test against those. That way we're speed up tests and making tests more unit like.

This is small amount of tests so no big deal, Ill draw a picture and/or make a PR later to refactor :)

Thank you for adding!

ahartikainen: Yeah we can change this, I just wanted tests quickly to master and there was also small bugs, same with the current PR for rug.

Let's clean these after initial implementation.

PR #865, opened 2019-11-11
[Bokeh] rug and divergences to traceplot

ahartikainen:
Rug + divergence

also fixes multidim input.


ahartikainen:
This is ready after the tests are run.

It is still quite ugly code and the defaults are a "mess", but plots look good. So I think we should move forward and after bokeh implementation is finished, let's rethink our bokeh UI.

@canyon289 and others, feel free to edit the branch as you see fit.

PR #866, opened 2019-11-12
Don't require posterior in loo if not used. Also make errors more dispassionate.

sethaxen:
`loo` currently checks if `posterior` is in the `InferenceData` near the beginning of the method, even though it is only needed if `reff` is not provided. This PR makes it so `posterior` is only required if `reff` is not provided.

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -484,14 +484,12 @@ def loo(data, pointwise=False, reff=None, scale="deviance"):
 
     """
     inference_data = convert_to_inference_data(data)
-    for group in ("posterior", "sample_stats"):
-        if not hasattr(inference_data, group):
-            raise TypeError(
-                "Must be able to extract a {group} group from data!".format(group=group)
-            )
+    if not hasattr(inference_data, "sample_stats"):
+        raise TypeError(
+            "Must be able to extract a sample_stats group from data!"
+        )
     if "log_likelihood" not in inference_data.sample_stats:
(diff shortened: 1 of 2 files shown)


arviz/stats/stats.py
```diff
@@ -484,14 +484,10 @@ def loo(data, pointwise=False, reff=None, scale="deviance"):
 
     """
     inference_data = convert_to_inference_data(data)
-    for group in ("posterior", "sample_stats"):
-        if not hasattr(inference_data, group):
-            raise TypeError(
-                "Must be able to extract a {group} group from data!".format(group=group)
-            )
+    if not hasattr(inference_data, "sample_stats"):
+        raise TypeError("Must be able to extract a sample_stats group from data!")
```
canyon289: Im fine with these except the exclamation marks. I know its picky but i dont think they add anything. Would rather leave the statement as all fact devoid of emotional interpretation

sethaxen: I'm happy to remove them, but they were already there and also appear elsewhere even in the same file.

aloctavodia: In that case we should remove all exclamations marks to be consistent

canyon289: Oops I missed that. Thanks @sethaxen for doing some cleanup and pointing that out.

canyon289:
Please add some tests, after that it lgtm! 

If you have questions about how to add tests too let me know! We're all here to help

PR #867, opened 2019-11-12
add vlines to rank plot

aloctavodia:
... and few other minor changes.

`kind = "bars"`

`kind = "vlines"`

When using `vlines` we have `ax.set_ylim(0, all_counts.mean() * 2)`, that is the height is 2 times the value of `ref_line`. This makes easier to see if a deviation is "large" or not.

in a next PR I would like to add the ranks (probably with the vlines as default) to trace plot, something like this:

non-centered

or with `bars`

non-centered

### DIFF
--- a/arviz/plots/plot_utils.py
+++ b/arviz/plots/plot_utils.py
@@ -130,6 +130,28 @@ def get_bins(values):
     return np.arange(x_min, x_max + width + 1, width)
 
 
+def _sturges_formula(dataset, mult=1):
+    """Use Sturges' formula to determine number of bins.
+
+    See https://en.wikipedia.org/wiki/Histogram#Sturges'_formula
+    or https://doi.org/10.1080%2F01621459.1926.10502161
+
+    Parameters
+    ----------
+    dataset: xarray.DataSet
+        Must have the `draw` dimension
+
+    mult: float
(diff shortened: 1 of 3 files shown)


arviz/plots/plot_utils.py
```diff
@@ -130,6 +130,28 @@ def get_bins(values):
...
+
+    See https://en.wikipedia.org/wiki/Histogram#Sturges'_formula
+    or https://doi.org/10.1080%2F01621459.1926.10502161
+
+    Parameters
+    ----------
+    dataset: xarray.DataSet
+        Must have the `draw` dimension
+
+    mult: float
```
canyon289: I would mak this as optional in docstring

canyon289: I guess maybe we shouldnt to be consistent. Nevermind!

arviz/plots/rankplot.py
```diff
@@ -110,7 +105,6 @@ def plot_rank(data, var_names=None, coords=None, bins=None, ref_line=True, figsi
     length_plotters = len(plotters)
```
canyon289: Maybe add one example in the docs above so people can see what it looks like

canyon289:
LGTM

PR #868, opened 2019-11-12
Extend the summary function.

rpgoldman:
- Revise the documentation string to explain that it returns either a
DataFrame or a DataSet.

- Add some type information.

- Add coords and dims arguments.


rpgoldman:
Still needs some tests I haven't had time to add, then I will strip off the "WIP."

rpgoldman:
I think that this could be merged at this point, unless someone would like to further review it.

PR #894, opened 2019-11-14
Add ArviZ.jl to readme and docs

OriolAbril:
It also fixes the example of plot_rank with vlines which I saw was not working

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,9 @@
 ArviZ (pronounced "AR-_vees_") is a Python package for exploratory analysis of Bayesian models.
 Includes functions for posterior analysis, model checking, comparison and diagnostics.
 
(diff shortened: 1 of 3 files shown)


sethaxen:
Maybe it's better to link to ArviZ.jl's documentation: https://arviz-devs.github.io/ArviZ.jl/dev/ since there will ultimately be much more information there. We can update that to the stable docs once the first version is released.

OriolAbril:
Will https://arviz-devs.github.io/ArviZ.jl/latest point to the stable version once present? It seems to already work even though it points to dev version now.

sethaxen:
`latest` always points to `dev` and is now deprecated.

sethaxen:
LGTM

Issue #877, opened 2019-11-13
Add backend plot defaulting into rcparams


canyon289:

ahartikainen:
Did we already implement this or do we need to update the plotting code? Maybe some util function

    def sanitize_backend(backend=None):
        if backend is None:
            backend  = rcParams["backend"]
        elif backend not in supported_backends:
            raise TypeError('Backend not supported {}. Select backend from {{"matplotlib", "bokeh"}}'.format(str(backend))
        return backend

OriolAbril:
I think @canyon289 was working on this in #949

ahartikainen:
Yes,  of course :laughing:

canyon289:
I think this is good now between Aris PR and mine.

Issue #792, opened 2019-08-12
Integrate new parameters as az.rcParams


OriolAbril:
## Extend `az.rcParams`

Global options were recently added to ArviZ (see #734) following matplotlib's implementation. Currently though, only 3 parameters are integrated with ArviZ code. This issue can serve two purposes, first to list and keep track of possible rcParams ideas, and second point out that many of them are up for the grabs and to give some basic directions on implementation of new rcParams so that it becomes easier for the community to add new parameters. 

## How to extend rcParams

rcParams implementation is basically a python dictionary with 2 basic peculiarities. The first one is that keys are fixed; the dictionary is initialized when importing ArviZ and loads all parameters. Once this is done, new keys cannot be added and present keys cannot be deleted. The second one is that when trying to modify the value of one of the parameters, the new value is first validated before updating rcParams. This prevents errors in the ArviZ functions that use rcParams, as they know what to expect.

To implement new parameters, in addition to giving a look to the already implemented params, the following checklist should be followed:

* Create one validation function (or use one of the already present) 
* Add the parameter to the `defaultParams` dict.
* Add the parameter to the arvizrc.template (with a short description of what the param does)
* Modify the corresponding functions so that they use rcParams
* Document the new rcParam where appropriate (i.e. `data.load` is documented in `from_netcdf` methods)
* If a new validation function was created, add a test for it
* If possible, add a test for the new rcParam integration using `az.rc_context`

## Ideas

Here I leave an unfinished list with some ideas, please comment on possible additions to rcParams:

#### Data
* `data.stan_log_likelihood` allow to set a default name (like `log_lik`) for the log likelihood variable in Stan.

#### Stats
* ~`stats.credible_interval` Set the default credible interval, defaulting to 0.94.~
* `stats.pointwise_elpd` Default to false, but allow to set it to true in order to get pointwise elpd diagnostics by default (pointwise values are always calculated, so the only overhead would be in the creation of the `ELPDData` object)
* ~`stats.ic_scale`~
* `stats.compare_method` set the default for `method` argument in `az.compare`. Low priority, its main use would be backwards compatibility if we ever modified the default.

#### Plot
* ~`plot.point_estimate` set the default point estimate in `plot_density` and `plot_posterior`.~
* `plot.bandwidth` or `plot.bw` default bandwidth for kde

ahartikainen:
`data.stan_log_likelihood` -> `data.elementwise_log_likelihood`/ `data.elem_log_likelihood`

[...]

PR #895, opened 2019-11-15
Add more parameters to rcparams

OriolAbril:
Related to #792 and to #877.

Adds parameters `plot.backend`, `plot.point_estimate`, `stats.credible_interval` and `stats.ic_scale` to rcparams dict. However it does not modify the functions to use them. I think that now than most if not all plots will be tinkered with it is a good moment to get everything ready so that integrating rcparms is only calling them.

Handling of none is tricky. For instance, `None` is a valid value of `plot.point_estimate` meaning no point estimate. In this cases, I propose setting the default to `"auto"` so that auto means look up the value in rcparams. In general, using `None` as look up rcparams should be fine.

### DIFF
--- a/arvizrc.template
+++ b/arvizrc.template
@@ -8,6 +8,7 @@ data.load                    : lazy  # Sets the default data loading mode.
 ### PLOT  ###
 # rcParams related with plotting functions
 plot.max_subplots            : 40    # Maximum number of subplots.
+plot.backend                 : matplotlib  # Plotting backend.
 
 ### STATS ###
 # rcParams related with statistical and diagnostic functions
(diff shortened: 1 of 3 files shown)


arvizrc.template
```diff
@@ -7,8 +7,12 @@ data.load                    : lazy  # Sets the default data loading mode.
...
+plot.backend                 : matplotlib  # Plotting backend.
 plot.max_subplots            : 40    # Maximum number of subplots.
+plot.point_estimate          : mean
 
 ### STATS ###
 # rcParams related with statistical and diagnostic functions
-stats.information_criterion  : waic  # Default information criterion
+stats.credible_interval      : 0.94
+stats.information_criterion  : waic
+stats.ic_scale               : deviance
```
ahartikainen: could we add all the options in the comment (for those which have options)

PR #897, opened 2019-11-16
Add quantiles for bokeh traceplot

ahartikainen:
Add quantiles to bokeh traceplot.


ahartikainen:
This is ready after the tests are passed.

PR #900, opened 2019-11-17
Expand MCSE acronym

canyon289:
Make it easier for people to understand what MCSE is

### DIFF
--- a/arviz/plots/mcseplot.py
+++ b/arviz/plots/mcseplot.py
@@ -35,7 +35,7 @@ def plot_mcse(
     text_kwargs=None,
     **kwargs
 ):
-    """Plot Monte Carlo Standard Error
+    """Plot quantile or local Monte Carlo Standard Error.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


arviz/plots/mcseplot.py
```diff
@@ -36,9 +34,8 @@ def plot_mcse(
     extra_kwargs=None,
     text_kwargs=None,
     **kwargs
-    # fmt: on
 ):
-    """Plot quantile, local or evolution of effective sample sizes (ESS).
+    """Plot Monte Carlo Standard Error
```
canyon289: @OriolAbril Did I do this right?

OriolAbril: The plot is actually quantile or local MCSE, to get the mean MCSE the `mcse` function should be used. I would write `Plot quantile or local Monte Carlo Standard Error.` 

Also, thanks for noticing! This is me copy-pasting too agressively from `plot_ess`

OriolAbril:
LGTM

PR #906, opened 2019-11-19
split autocorrplot

ahartikainen:
~Currently, bokeh just crashes :)~

Uses bokeh.figure.segment plot. Looking good.

### DIFF
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -2,11 +2,9 @@
 import numpy as np
 
 from ..data import convert_to_dataset
-from ..stats import autocorr
 from .plot_utils import (
     _scale_fig_size,
     default_grid,
-    make_label,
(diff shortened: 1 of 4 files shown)


arviz/plots/autocorrplot.py
```diff
@@ -100,27 +102,34 @@ def plot_autocorr(
 
     if ax is None:
         _, axes = _create_axes_grid(
-            length_plotters, rows, cols, figsize=figsize, squeeze=False, sharex=True, sharey=True
+            length_plotters, rows, cols, figsize=figsize, squeeze=False, sharex=True, sharey=True, backend=backend,
```
OriolAbril: Won't this pass `backend` to `plt.subplots`?

ahartikainen:
This is ready now. The figure becomes a bit large, but I think that is fine for now.

PR #908, opened 2019-11-21
Add bokeh autocorr tests

ahartikainen:
Add tests.


ahartikainen:
Bokeh code uses a bit different structure for grid than matplotlib (grid is created at the end of the code, not when axes are created)

ahartikainen:
Can use also None values, so at least currently, they are kept

PR #909, opened 2019-11-21
separate prior from prior predictive variables

OriolAbril:
Currently, pymc3 and pyro store both prior and prior predictive in the `prior` group. This aims to fix the issue when possible.

Only ~numpyro~ and ~updating docs (cookbook and schema example) is missing.~

EDIT: 

numpyro is done and the schema is still in a too early phase, I don't think we need to update the examples every PR


OriolAbril:
Should be ready to merge.

PR #912, opened 2019-11-23
Bokeh elpdplot

ahartikainen:
Add minimal support for elpdplot.

Multidim stuff tripped me, not sure what is in what dimension (bokeh glyph methods don't handle multidim stuff automatically).

Code style is not pretty.

### DIFF
--- a/arviz/plots/backends/bokeh/bokeh_elpdplot.py
+++ b/arviz/plots/backends/bokeh/bokeh_elpdplot.py
@@ -9,7 +9,6 @@ import numpy as np
 
 from ...plot_utils import (
     _scale_fig_size,
-    format_coords_as_labels,
 )
 from ....rcparams import rcParams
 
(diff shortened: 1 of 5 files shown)


arviz/plots/backends/bokeh/bokeh_elpdplot.py
```diff
@@ -173,16 +173,20 @@ def _plot_atomic_elpd(
...
+    else:
+        ax_.yaxis.minor_tick_line_color = None
+        ax_.yaxis.major_label_text_font_size = "0pt"
+
     if xlabels_shown:
         if xlabels:
             ax_.xaxis.ticker = np.arange(0, len(coord_labels))
             ax_.xaxis.major_label_overrides = {
-                key: str(value)
+                str(key): str(value)
```
ahartikainen: Does this work?

Because ticker value is transformed to string value?

OriolAbril: I changed it because otherwise it did not work. It seems like it works, but when you try to show the plot you get a typeerror saying dict keys must be strings. Doing this I was able to generate and see the plot with the labels

PR #916, opened 2019-11-24
Proposal for PyMC3 predictions support

OriolAbril:
Proposal on how to add data to `predictions` and `constant_data_predictions` with PyMC3.

One example usage (borrowing the model and code from @corriebar in https://github.com/pymc-devs/pymc3/issues/3640) could be the following:

```python
import numpy as np
import pymc3 as pm
import pandas as pd
import arviz as az

size = 200
true_intercept = 1
true_slope = 2

x = np.linspace(0, 1, size)
# y = a + b*x
true_regression_line = true_intercept + true_slope * x
# add noise
y = true_regression_line + np.random.normal(scale=.5, size=size)

data = pd.DataFrame(data={"x":x, "y":y})
del x
del y

with pm.Model() as model:
    x = pm.Data("x", data["x"])
    # Define priors
    sigma = pm.HalfCauchy('sigma', beta=10, testval=1.)
    intercept = pm.Normal('Intercept', 0, sigma=20)
    x_coeff = pm.Normal('x_coeff', 0, sigma=20)

    # Define likelihood
    likelihood = pm.Normal('y', mu=intercept + x_coeff * x,
                        sigma=sigma, observed=data["y"])

    # Inference!
    trace = pm.sample(1000, cores=2) # draw 3000 posterior samples using NUTS sampling

idata = az.from_pymc3(trace, dims={"y": ["custom_dim"]})

print(idata)
#Inference data with groups:
#        > posterior
#        > sample_stats
#        > observed_data
#        > constant_data

more_data = np.linspace(1, 1.2, num=3)

with model:
    pm.set_data({"x": more_data})
    post_pred = pm.sample_posterior_predictive(trace, samples=1000)
   
    idata_all = az.add_predictions_pymc3(idata, predictions=post_pred) 

print(idata_all)
#Inference data with groups:
#        > posterior
#        > sample_stats
#        > observed_data
#        > constant_data
#        > predictions
#        > constant_data_predictions
print(idata_all.predictions)
#<xarray.Dataset>
#Dimensions:     (chain: 1, custom_dim: 3, draw: 1000)
#Coordinates:
#  * chain       (chain) int64 0
#  * draw        (draw) int64 0 1 2 3 4 5 6 7 ... 992 993 994 995 996 997 998 999
#  * custom_dim  (custom_dim) int64 0 1 2
#Data variables:
#    y           (chain, draw, custom_dim) float64 2.59 4.246 ... 2.909 4.701
#Attributes:
#    created_at:                 2019-11-24T22:24:18.516325
#    inference_library:          pymc3
#    inference_library_version:  3.7
```

The default is set to returning a new inferenceData object for now. Also, to get `constant_data_predictions` `add_predictions_pymc3` must be called within a `Model` context or a model passed to it. As it can be seen in the example, when possible, it uses the dimensions from related groups.

### DIFF
--- a/arviz/data/__init__.py
+++ b/arviz/data/__init__.py
@@ -7,7 +7,7 @@ from .converters import convert_to_dataset, convert_to_inference_data
 from .io_cmdstan import from_cmdstan
 from .io_cmdstanpy import from_cmdstanpy
 from .io_dict import from_dict
-from .io_pymc3 import from_pymc3
+from .io_pymc3 import from_pymc3, add_predictions_pymc3
 from .io_pystan import from_pystan
 from .io_emcee import from_emcee
 from .io_pyro import from_pyro
@@ -24,6 +24,7 @@ __all__ = [
     "dict_to_dataset",
     "convert_to_dataset",
     "convert_to_inference_data",
+    "add_predictions_pymc3",
     "from_pymc3",
     "from_pystan",
     "from_emcee",
--- a/arviz/data/base.py
+++ b/arviz/data/base.py
@@ -230,3 +230,54 @@ def make_attrs(attrs=None, library=None):
     if attrs is not None:
         default_attrs.update(attrs)
     return default_attrs
+
+
+def get_predictions_dims(idata, dims=None):
+    """Get the dimensions for the predictions and constant data predictions."""
+    if dims is not None:
+        return dims
+    dims = {}
+    if hasattr(idata, "posterior_predictive"):
+        dataset = idata.posterior_predictive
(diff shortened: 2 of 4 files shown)


rpgoldman:
Alternative approach to out-of-sample prediction is given in my PR #983 

I made this implementation to clarify and make concrete some issues I see with the new schema design (#913), and that I have posted in comments.

The approach taken here is to require an entirely distinct `InferenceData` object for out of sample predictions, rather than attempting to fit predictions in the same object as the posterior trace, etc.  See #913 discussion for my rationale.

PR #932, opened 2019-12-02
Experimenting with adding bokeh plots to documentation

OriolAbril:


ahartikainen:
Very cool. Is there a preview somewhere?

OriolAbril:
I have just pushed to my fork the docs generated with the current code. It's not pretty yet, for now it just works.

The [example gallery](https://oriolabril.github.io/arviz/examples/index.html) can be generated automatically by `sphinx-build` (given that [bokeh's export_png](https://docs.bokeh.org/en/latest/docs/user_guide/export.html) works). There are two important workarounds, one is that the plot must be saved as `ax = az.plot_xyz()` and the other is that `show=True` must be explicitly on the function call (the code replaces it with `show=False`, otherwise every example generated a plot in the browser when creating the docs locally). 

Bokeh plots can be embedded in the same way as matplotlib plot, simply changing `.. plot::` by `.. bokeh-plot::`. [One example](https://oriolabril.github.io/arviz/generated/arviz.plot_elpd.html#arviz.plot_elpd)

ahartikainen:
show True can be fixed with rcparams.

scripts/create_testenv.sh
```diff
@@ -86,3 +87,7 @@ fi
 #  Install editable using the setup.py
 pip install  --no-cache-dir -r requirements.txt
 pip install --no-cache-dir -r requirements-dev.txt
+
+if [ "$NAME" = "SPHINX" ]; then
+    conda install -y -c conda-forge selenium phantomjs
```
canyon289: At this point we might want to split requirements into `requirements-dev,txt` and `requirements-docs.txt` so people can build docs locally. No need to do on this pr if you dont want to. Just something to think about

OriolAbril: I thought about this, but only selenuim can be pip installed, phantomjs has to be installed with conda or npm

OriolAbril:
Ready for review! Check the rendered docs in my fork gh-pages to see how will it look like: https://oriolabril.github.io/arviz/index.html

OriolAbril:
Issues yet to be solved:
* Add more bokeh examples in gallery
* Add some bokeh examples in api
* Building docs locally without docker. Right now building the docs requires both matplotlib, bokeh but also selenium and phantomjs (these last two for png generation in bokeh, see more in [bokeh_docs](http://docs.bokeh.org/en/1.3.2/docs/user_guide/export.html)). The problem comes with phantomjs because it cannot be installed using pip.

ahartikainen:
Can we use svg output?

OriolAbril:
Latest pydocstyle has some issues https://github.com/PyCQA/pydocstyle/issues/434, for now we should use previous version until fixed.

PR #934, opened 2019-12-03
InferenceData keeps all groups

### LINKED ISSUE #907: Concat fails with empty InferenceData
**Describe the bug**
`arviz.concat` produces incorrect outputs when one of the arguments is an `InferenceData` with no groups.

**To Reproduce**
```python
>>> import arviz as az

>>> idata = az.load_arviz_data("centered_eight")

>>> idata
Inference data with groups:
    > posterior
    > sample_stats
    > posterior_predictive
    > prior
    > observed_data

>>> az.concat(idata, az.InferenceData())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/arviz/data/inference_data.py", line 307, in concat
    group_data = getattr(arg, group)
UnboundLocalError: local variable 'group' referenced before assignment

>>> az.concat(az.InferenceData(), idata)
Inference data with groups:
    > observed_data
```

**Expected behavior**
Both `concat` commands should have produced an `InferenceData` with the same groups and data as `idata`.

**Additional context**
Using master branch on OS X.

anzelpwj:
Fixes #907. I have added a unit test which catches this error (as the concat would only catch the last group it iterated over).

### DIFF
--- a/arviz/tests/test_data.py
+++ b/arviz/tests/test_data.py
@@ -548,6 +548,24 @@ def test_bad_inference_data():
         InferenceData(posterior=[1, 2, 3])
 
 
+def test_inference_concat_keeps_all_fields():
+    """From failures observed in issue #907"""
+    idata1 = from_dict(posterior={'A': [1, 2, 3, 4]}, sample_stats={"B": [2, 3, 4, 5]})
+    idata2 = from_dict(prior={'C': [1, 2, 3, 4]}, observed_data={"D": [2, 3, 4, 5]})
+
+    idata_c1 = concat(idata1, idata2)
+    idata_c2 = concat(idata2, idata1)
+
+    assert hasattr(idata_c1, "prior")
+    assert hasattr(idata_c2, "prior")
+    assert hasattr(idata_c1, "observed_data")
+    assert hasattr(idata_c2, "observed_data")
+    assert hasattr(idata_c1, "posterior")
+    assert hasattr(idata_c2, "posterior")
+    assert hasattr(idata_c1, "sample_stats")
+    assert hasattr(idata_c2, "sample_stats")
+
+
 class TestDataConvert:
     @pytest.fixture(scope="class")
     def data(self, draws, chains):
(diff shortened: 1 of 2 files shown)


arviz/tests/test_data.py
```diff
@@ -548,6 +548,24 @@ def test_bad_inference_data():
         InferenceData(posterior=[1, 2, 3])
 
 
+def test_inference_concat_keeps_all_fields():
```
OriolAbril: I hope I am not too picky, it would be great to move the function below the other functions testing concat behaviour , around line [303](https://github.com/arviz-devs/arviz/blob/66bc210aa68efff1bb43a0821c6232826aaa90a6/arviz/tests/test_data.py#L303)

arviz/tests/test_data.py
```diff
@@ -548,6 +548,24 @@ def test_bad_inference_data():
...
+    idata_c2 = concat(idata2, idata1)
+
+    assert hasattr(idata_c1, "prior")
+    assert hasattr(idata_c2, "prior")
+    assert hasattr(idata_c1, "observed_data")
+    assert hasattr(idata_c2, "observed_data")
+    assert hasattr(idata_c1, "posterior")
+    assert hasattr(idata_c2, "posterior")
+    assert hasattr(idata_c1, "sample_stats")
+    assert hasattr(idata_c2, "sample_stats")
```
OriolAbril: Also very small comment, the helper function `check_multiple_attrs` can be used, it tries to simplify multiple checks on groups (keys of test_dict) and their variables (values of test_dict). I think the test dict would be something like:

```
test_dict = {
    "posterior": ["A"], 
    "sample_stats": ["B"],
    "prior": ["C"], 
    "observed_data": ["D"],
}
fails_c1 = check_multiple_attrs(test_dict, idata_c1)
assert not fails_c1
fails_c2 = check_multiple_attrs(test_dict, idata_c2)
assert not fails_c2
```

anzelpwj: Looks like a good improvement, will implement.

arviz/tests/test_data.py
anzelpwj: Not a problem.

canyon289:
Thanks @anzelpwj!

PR #935, opened 2019-12-03
Add RcParam for HTTP protocol for remote dataset fetching

canyon289:


OriolAbril:
How about using a rcparam `data.allow_http` or `data.use_https` so that the default keeps being using https? 

It could be handled when loading instead of at url definition (I think a `str.replace("https", "http")` should do)

canyon289:
I'm good with that! I'll update PR later this week

arviz/data/datasets.py
```diff
@@ -213,8 +215,14 @@ def load_arviz_data(dataset=None, data_home=None):
         remote = REMOTE_DATASETS[dataset]
         home_dir = get_data_home(data_home=data_home)
         file_path = os.path.join(home_dir, remote.filename)
+
         if not os.path.exists(file_path):
-            urlretrieve(remote.url, file_path)
+            http_type = RcParams["data.http_protocol"]
+
+            # Replaces http type. Redundant if http_type is http, useful if http_type is https
+            url = remote.url.replace("http", http_type)
```
canyon289: @OriolAbril Let me know if this string replace method is fine by you. I can always do the string substitution method as well. Picked current pattern so urls would be valid if anyone wanted to copy and paste but I'm not tied to the pattern

arviz/data/datasets.py
```diff
@@ -7,6 +7,8 @@
 from urllib.request import urlretrieve
 
 from .io_netcdf import from_netcdf
+from ..rcparams import RcParams
```
OriolAbril: RcParams i camel case is the name of the class defined, the specific instance to customize arviz is `rcParams` (using the same names as matplotlib which I figured would be less confusing)

arvizrc.template
```diff
@@ -1,6 +1,8 @@
 #### ArviZ rcParams template ####
 ### DATA  ###
 # rcParams related with data loading related functions
+data.http_protocol           : http  # protocol for remote dataset load,
```
OriolAbril: This should be https, to match the default in rcparams.py

PR #938, opened 2019-12-04
Swap out docker jupyter commands, fixes #936

### LINKED ISSUE #936: Docker image is missing some important packages
Hi all,

I'm trying to open a Jupyter notebook in the Docker image as per `CONTRIBUTING.md`, and when running from a (relatively new) Docker image I am finding that I cannot run `import arviz as az` without finding that we are missing the following packages:
- `scipy`
- `xarray`
- `matplotlib`
- `netCDF4`

Once I installed these for packages, I was finally able to import the package.

**To Reproduce**
- Download docker image.
- Open Docker Jupyter notebook.
- Try to import `arviz` package.
- For whatever module it complains its missing, go back, pip install in the Docker image, and then restart.
- Repeat this and you get the 4 missing packages above.

**Expected behavior**
I should be able to import the package without any trouble from a clean Docker image.

**Additional context**
MacOSX 10.14 computer, Docker image downloaded on 2019-12-03 at around 3 PM PST.

anzelpwj:
We found in testing that you need to slightly modify the commands used for opening a Jupyter notebook hosted in Docker so that it loads the correct version of Python (with all the packages from `requirements.txt`). This does the proper opening. Fixes #936 .

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -203,8 +203,8 @@ To start a bash shell inside Docker, run:
 Alternatively, to start a jupyter notebook, there are two steps, first run:
 
     $ docker run --mount type=bind,source="$(pwd)",target=/opt/arviz/ --name jupyter-dock -it -d -p 8888:8888 arviz
-    $ docker exec -it jupyter-dock pip install jupyter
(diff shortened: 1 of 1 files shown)


OriolAbril:
LGTM

canyon289:
LGTM as well

Issue #941, opened 2019-12-05
Numpyro pickle error


ahartikainen:
Tests are failing with pickle error

in tests/helpers

    >                   pickle.dump(func(eight_schools_data, draws, chains), buff)
    E                   AttributeError: Can't pickle local object 'hmc.<locals>.init_kernel'

https://travis-ci.org/arviz-devs/arviz/jobs/621006910?utm_medium=notification&utm_source=github_status

cc @fehiepsi

ahartikainen:
Is there some lambda somewhere in numpyro code?

ahartikainen:
Ok, there are functions inside functions

https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/mcmc.py

So, either hmc should be a class with `__call__` implemented or then we need to do something clever.

fehiepsi:
Hi @ahartikainen, thanks for the help diagnosing!

I agree that making hmc a class can solve the error. However, we want to keep the functional nature of JAX as much as possible while still provides some convenient classes for the user interface. This is helpful for reducing unexpected behaviors when using JAX tracing mechanism with classes. Recently, we add a caching mechanism to avoid recompiling the sampler when running mcmc on new data. And it has caused the pickle error. Sorry for the regression, I'll make a fix soon.

canyon289:
@fehiepsi Was this problem fixed? Just tidying things up :)

fehiepsi:
Yup! :)

PR #944, opened 2019-12-05
fix numpyro pickle error

fehiepsi:
Addresses #941.


OriolAbril:
This was extremely fast! Thanks a lot!  Do not worry about the test failure, it has been solved in another PR.

PR #948, opened 2019-12-07
Clean bokeh code

ahartikainen:
Docstring and some simple changes.

### DIFF
--- a/arviz/plots/backends/bokeh/__init__.py
+++ b/arviz/plots/backends/bokeh/__init__.py
@@ -1,5 +1,6 @@
 # pylint: disable=no-member
 """Bokeh Plotting Backend."""
+import packaging
 
 
 def output_notebook(*args, **kwargs):
@@ -34,7 +35,7 @@ def check_bokeh_version():
     try:
         import bokeh
 
-        assert bokeh.__version__ >= "1.4.0"
+        assert packaging.version.parse(bokeh.__version__) >= packaging.version.parse("1.4.0")
     except (ImportError, AssertionError):
         raise ImportError("'bokeh' backend needs Bokeh (1.4.0+) installed.")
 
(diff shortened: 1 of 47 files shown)


arviz/plots/backends/bokeh/__init__.py
```diff
@@ -30,5 +31,22 @@ def copy_docstring(lib, function):
     return doc
 
 
+def check_bokeh_version():
```
canyon289: In new PR we can add this into the new method added if that sounds good

arviz/plots/backends/bokeh/bokeh_parallelplot.py
```diff
@@ -3,26 +3,18 @@
...
-                "poly_select",
-                "undo",
-                "redo",
-                "reset",
-                "save,hover",
-            ]
-        )
+        tools = rcParams["plot.bokeh.tools"]
+        output_backend = rcParams["plot.bokeh.output_backend"]
+        dpi = rcParams["plot.bokeh.figure.dpi"]
```
canyon289: This pattern means if people want to change things they need to both modify RcParams or arg, so api is sort of split. Is my understanding correct?

canyon289:
Scanning over this looks good to me!

arviz/plots/backends/bokeh/bokeh_parallelplot.py
ahartikainen: Api is split now, until we fix the api

arviz/plots/backends/bokeh/__init__.py
ahartikainen: Sounds ok

arviz/plots/backends/bokeh/bokeh_parallelplot.py
canyon289: Works for me

PR #963, opened 2019-12-15
Backend kwargs

ahartikainen:
Implement backend_kwargs for all plots.

### DIFF
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -111,6 +111,7 @@ def plot_autocorr(
     )
 
     if ax is None:
+        show = backend_kwargs.pop("show", None) if backend_kwargs is not None else None
         _, axes = _create_axes_grid(
             length_plotters,
             rows,
@@ -122,6 +123,8 @@ def plot_autocorr(
             backend=backend,
             backend_kwargs=backend_kwargs,
         )
+        if show is not None:
+            backend_kwargs["show"] = show
     else:
         axes = ax
 
@@ -143,10 +146,6 @@ def plot_autocorr(
         autocorr_plot_args.pop("xt_labelsize")
         autocorr_plot_args.pop("titlesize")
         autocorr_plot_args["line_width"] = autocorr_plot_args.pop("linewidth")
-
-    else:
-
-        autocorr_plot_args.pop("backend_kwargs")
 
     # TODO: Add backend kwargs
     plot = get_plotting_function("plot_autocorr", "autocorrplot", backend)
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -144,6 +144,10 @@ def plot_autocorr(
         autocorr_plot_args.pop("titlesize")
         autocorr_plot_args["line_width"] = autocorr_plot_args.pop("linewidth")
 
+    else:
+
+        autocorr_plot_args.pop("backend_kwargs")
+
     # TODO: Add backend kwargs
     plot = get_plotting_function("plot_autocorr", "autocorrplot", backend)
     axes = plot(**autocorr_plot_args)
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -8,7 +8,7 @@ from .plot_utils import (
     xarray_var_iter,
     _create_axes_grid,
     filter_plotters_list,
-    get_plotting_method,
+    get_plotting_function,
 )
 from ..utils import _var_names
 
@@ -144,7 +144,7 @@ def plot_autocorr(
         autocorr_plot_args["line_width"] = autocorr_plot_args.pop("linewidth")
 
     # TODO: Add backend kwargs
-    method = get_plotting_method("plot_autocorr", "autocorrplot", backend)
-    axes = method(**autocorr_plot_args)
+    plot = get_plotting_function("plot_autocorr", "autocorrplot", backend)
+    axes = plot(**autocorr_plot_args)
 
     return axes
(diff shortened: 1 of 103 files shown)


arviz/rcparams.py
```diff
@@ -96,6 +96,7 @@ def _validate_probability(value):
     "plot.bokeh.figure.dpi": (60, _validate_positive_int),
     "plot.bokeh.figure.width": (500, _validate_positive_int),
     "plot.bokeh.figure.height": (500, _validate_positive_int),
+    "plot.bokeh.show": (True, _make_validate_choice({True, False}, typeof=bool)),
```
OriolAbril: I think it would be best to create a new validation function for booleans. The validate choice relies on calling `cls(value)` which in the case of booleans would mean accepting nearly everything mapping everyhting to `True`. In fact, the string `"false"` is mapped to `True` doing this. 

I think that only booleans `True`, `False` and strings `true` and `false` (ignoring cases) should be accepted. The string will be what is generally used (i.e. when reading an arvizrc file), and the booleans will be used when using `rc_context`.

canyon289: Ok I'll change this to a string!

OriolAbril: using `make_validate_choice` either with boolean or with string won't work, as you say at some point is has to turn into boolean. I am proposing to create a new `_validate_boolean` that returns `True` (as boolean) if value is `True` or `"true"`, returns `False` (as boolean) if value is `False` or `"false"` and raises an error otherwise.

arviz/rcparams.py
```diff
@@ -35,7 +35,10 @@ def validate_choice(value):
             raise ValueError("Could not convert to {}".format(typeof.__name__))
         if isinstance(value, str):
             value = value.lower()
+
         if value in accepted_values:
+            # Convert value to python boolean if string matches
+            value = {"true": True, "false": False}.get(value, value)
```
canyon289: @OriolAbril Is this a good idea?

OriolAbril: I think this will do the trick, but I find it quite confusing. For example, if someone where to use:

```
with az.rc_context({"plot.bokeh.show": True}):
    az.plot_xyz()
```

It would convert `True` to a string, check whether it was in accepted values and then convert it back to a boolean. Moreover, even though the `typeof` would be string and the accepted values would be strings too, the return value would be of boolean type.

ahartikainen: What if user uses "True" or "False"

canyon289: It would work because RcParams lowercases the params before checking.

@OriolAbril I'd agree it'd be confusing if someone checks under the hood. I'll make another proposal

canyon289: Actually I'm not sure what do here. Because in RcParams file the input will be a string, but in context manager the input will be string but it needs to be converted to a bool. But if it comes from context manager it'll be a bool and need to ultimately return a bool.

Since `validate_choice` doesn't know the source not sure what to do that is least confusing

OriolAbril: My proposal was to create a new `_validate_boolean` that returns `True` (as boolean) if value is `True` or `"true"`, returns `False` (as boolean) if value is `False` or `"false"` and raises an error otherwise.

arviz/plots/backends/bokeh/__init__.py
```diff
@@ -26,6 +23,22 @@
 from .rankplot import plot_rank
 from .traceplot import plot_trace
 from .violinplot import plot_violin
+from ....rcparams import rcParams
+
+
+def backend_kwarg_defaults(add_default=True, *args, **kwargs):
```
canyon289: Move this up to top. That way its instantiated before the imports

arviz/plots/backends/bokeh/autocorrplot.py
```diff
@@ -4,14 +4,22 @@
 from bokeh.layouts import gridplot
 from bokeh.models.annotations import Title
 
+from . import backend_kwarg_defaults
```
canyon289: Because this is run when autocorr is imported, and because in __init__ autocorr is above definition, python can't import. Like other comment move `backend_kwarg_defaults` above imports in the __init__ files and everything should be good

arviz/plots/backends/bokeh/distplot.py
```diff
@@ -36,17 +36,18 @@ def plot_dist(
     if backend_kwargs is None:
         backend_kwargs = {}
 
-    backend_kwargs = {**BACKEND_KWARG_DEFAULTS, **backend_kwargs}
-
+    backend_kwargs = {
+        **backend_kwarg_defaults(
```
canyon289: This is nice! Thanks!

arviz/plots/autocorrplot.py
```diff
@@ -110,6 +111,7 @@ def plot_autocorr(
     )
 
     if ax is None:
+        show = backend_kwargs.pop("show", None) if backend_kwargs is not None else None
```
canyon289: I agree with @OriolAbril .All this show logic is more confusing to me. Can we just default show in the backend_kwargs and avoid all these conditionals and nones?

ahartikainen: No, because mol does not have show and this could be None.

ahartikainen:
So each upper-level plot should call `_create_axis_grid` which should always return 2D array of plots. 

Also `backend_kwargs` should be used all kwargs going for `figure`.

We would also need a way to define which plots are empty (axis of for plt and `None` for bokeh)

Also, we need a way to define special size figures (`plot_density`).

ahartikainen:
Need to remove some comments before merge.

Let's fox the logic in another PR.

ahartikainen:
Also, another PR can handle how can we create docs with bokeh without calling show

PR #964, opened 2019-12-18
Summary: add option to return diagnostic, stats or both (default)

aloctavodia:
Sometimes I do not want summary to return diagnostics, one use case is when teaching. Showing diagnostics but deferring explanations generates anxiety. I guess there are many other use cases so the PR :-)

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -827,7 +827,7 @@ def summary(
     fmt : {'wide', 'long', 'xarray'}
         Return format is either pandas.DataFrame {'wide', 'long'} or xarray.Dataset {'xarray'}.
     kind : {'all', 'stats', 'diagnostics'}
-        Whether to include the `stats` `mean`, `sd`, `hpd_3%`, `hpd_97%`, or the diagnostics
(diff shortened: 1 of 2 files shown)


arviz/stats/stats.py
```diff
@@ -823,12 +824,16 @@ def summary(
...
     var_names : list
         Names of variables to include in summary
-    include_circ : bool
-        Whether to include circular statistics
     fmt : {'wide', 'long', 'xarray'}
         Return format is either pandas.DataFrame {'wide', 'long'} or xarray.Dataset {'xarray'}.
+    kind : {'all', 'stats', 'diagnostics'}
+        Whether to include the `stats` `mean`, `sd`, `hpd_3%`, `hpd_97%`, or the diagnostics
+        `mcse_mean`, `mcse_sd`, `ess_bulk`, `ess_tail`, and `r_hat`. Default to include `all` of
+        them.
```
OriolAbril: Very minor formatting comment. I would use the same format for stats and for diagnostics, maybe a colon after them too.

PR #967, opened 2019-12-21
Move axes creation to backend code

ahartikainen:
Move axes creation from frontend to backend. (two plots: autocorr, density).


OriolAbril:
LGTM

PR #968, opened 2019-12-22
Support log_likelihood and obs data for pyro io

fehiepsi:
This PR addresses the remaining issue of #853. Users can get information criterions with from_pyro now.

### DIFF
--- a/arviz/tests/test_data_pyro.py
+++ b/arviz/tests/test_data_pyro.py
@@ -2,6 +2,7 @@
 import numpy as np
 import pytest
 import torch
+import pyro
 from pyro.infer import Predictive
 
 from ..data.io_pyro import from_pyro
@@ -43,12 +44,14 @@ class TestDataPyro:
         inference_data = self.get_inference_data(data, eight_schools_params)
         test_dict = {
             "posterior": ["mu", "tau", "eta"],
-            "sample_stats": ["diverging", "log_likelihood"],
+            "sample_stats": ["diverging"],
             "posterior_predictive": ["obs"],
             "prior": ["mu", "tau", "eta"],
             "prior_predictive": ["obs"],
-            "observed_data": ["obs"],
         }
+        if pyro.__version__.startswith("1"):
+            test_dict["sample_stats"].append("log_likelihood")
+            test_dict["observed_data"] = ["obs"]
         fails = check_multiple_attrs(test_dict, inference_data)
         assert not fails
 
@@ -75,7 +78,9 @@ class TestDataPyro:
         idata = from_pyro(data.obj)
         test_dict = {
             "posterior": ["mu", "tau", "eta"],
-            "sample_stats": ["diverging", "log_likelihood"],
+            "sample_stats": ["diverging"],
         }
+        if pyro.__version__.startswith("1"):
+            test_dict["sample_stats"].append("log_likelihood")
         fails = check_multiple_attrs(test_dict, idata)
         assert not fails
--- a/arviz/tests/test_data_pyro.py
+++ b/arviz/tests/test_data_pyro.py
@@ -73,6 +73,9 @@ class TestDataPyro:
 
     def test_inference_data_only_posterior(self, data):
         idata = from_pyro(data.obj)
-        test_dict = {"posterior": ["mu", "tau", "eta"], "sample_stats": ["diverging"]}
+        test_dict = {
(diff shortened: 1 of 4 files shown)


ahartikainen:
How do we want to support old pyro versions? Should we drop the support?

fehiepsi:
Oppss... sorry I need to skip those stats in tests for the old Pyro. We don’t need to drop support for the old Pyro version.

arviz/tests/test_data_pyro.py
```diff
@@ -48,6 +49,9 @@ def test_inference_data(self, data, eight_schools_params):
             "prior": ["mu", "tau", "eta"],
             "prior_predictive": ["obs"],
         }
+        if pyro.__version__.startswith("1"):
```
canyon289: What are the other pyro version where this won't work?

OriolAbril: I think what is happening is that pyro <1.0.0 (i.e. 0.5.1) does not store the observed data information. Therefore, even though `from_pyro` works in both versions, it will only contain observed data and log likelihood data if the version is >=1. Both versions need to be tested but using slightly different tests.

fehiepsi: Thank @OriolAbril! You have pointed out the whole story for this line of code.

ahartikainen: Hi, now reading again this, I think this actually needs a bit more robust check.

    import packaging

    ....

    packaging.version.parse(pyro.__version__) >= packaging.version.parse("1.0.0")

I mean I guess this will still work with 2.0 etc

fehiepsi: Thank @ahartikainen! I just made a change with your suggestion. It is better.

arviz/tests/test_data_pyro.py
```diff
@@ -48,6 +50,9 @@ def test_inference_data(self, data, eight_schools_params):
             "prior": ["mu", "tau", "eta"],
             "prior_predictive": ["obs"],
         }
+        if packaging.version.parse(pyro.__version__) >= packaging.version.parse("1.0.0"):
```
canyon289: @fehiepsi Thanks for making change. 
I would still ask that this if statement is split out into another test sort of like this below. This way its more obvious that portions of test are being skipped if someone tests on old env.

`@pytest.mark.skipif(packaging.version.parse(pyro.__version__) >= packaging.version.parse("1.0.0"), reason="requires pyro 2.0.0 or higher")
def test_inference_data(self, data, eight_schools_params):
        inference_data = self.get_inference_data(data, eight_schools_params)
        test_dict = {
            "posterior": ["sample_stats"],
            "sample_stats": ["observed_data"],
            }
```
`

**Comment by canyon289** [comment]
Approving but still ask that the if statements be removed in favor of splitting tests apart and marking with `pytest.skip`

File: arviz/tests/test_data_pyro.py (hunk 2 above)
Comment (fehiepsi) [reply arviz/tests/test_data_pyro.py]: It is fine for me to split the tests. I'll do that.

File: arviz/tests/test_data_pyro.py [hunk 3]
```diff
@@ -51,6 +53,16 @@ def test_inference_data(self, data, eight_schools_params):
         fails = check_multiple_attrs(test_dict, inference_data)
         assert not fails
 
+    @pytest.skipif(
```
ahartikainen: `pytest.mark.skipif`

arviz/tests/test_data_pyro.py
```diff
@@ -72,9 +84,16 @@ def test_inference_data_no_posterior(self, data, eight_schools_params):
...
         idata = from_pyro(data.obj)
-        test_dict = {
-            "posterior": ["mu", "tau", "eta"],
-            "sample_stats": ["diverging"],
-        }
+        test_dict = {"posterior": ["mu", "tau", "eta"], "sample_stats": ["diverging"]}
+        fails = check_multiple_attrs(test_dict, idata)
+        assert not fails
+
+    @pytest.skipif(
```
ahartikainen: `pytest.mark.skipif`

arviz/tests/test_data_pyro.py
fehiepsi: Oppsss, thanks, @ahartikainen !

ahartikainen:
Thanks for the PR, this is great!

PR #969, opened 2019-12-22
move backend show to args

ahartikainen:
This changes the logic, show backend `show` is not in `backend_kwargs`. This also implement this for matplotlib.


arviz/rcparams.py
```diff
@@ -108,6 +108,7 @@ def _validate_boolean(value):
     "plot.bokeh.figure.height": (500, _validate_positive_int),
     "plot.bokeh.show": (True, _validate_boolean),
     "plot.matplotlib.constrained_layout": (True, _validate_boolean),
+    "plot.matplotlib.show": (True, _validate_boolean),
```
OriolAbril: I think I prefer false as default for matplotlib, because it is "backward compatible" and showing plots is easier than with bokeh, so calling `plt.show` from ArviZ is not that important.

PR #970, opened 2019-12-23
constrain numpy version temporarily

ahartikainen:
Python 3.7 + jax + numpy 1.18 have some problems, let's constrain numpy version until this bug has been fixed by jax -package.


fehiepsi:
Yeah, thanks for detecting the issue! This should be the reason. I have raised the issue [upstream](https://github.com/google/jax/issues/1912).

PR #976, opened 2019-12-26
Bokeh interactive legend for pair_plot 2 variable case

ahartikainen:
Enable hide-show for legend on 2 variable case.

For n-var case it would cause too much of a clutter. We can use widgets for them.

### DIFF
--- a/arviz/plots/backends/__init__.py
+++ b/arviz/plots/backends/__init__.py
@@ -1,3 +1,4 @@
+# pylint: disable=no-member,invalid-name,redefined-outer-name
 """ArviZ plotting backends."""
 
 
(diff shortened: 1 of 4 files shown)


ahartikainen:
@canyon289 do you think importing two functions from backends.bokeh init doesn't trigger bokeh import?

ahartikainen:
If this procedure imports bokeh, maybe we could move these helper functions up one level?

ahartikainen:
Ok, I moved the user facing helper codes up one level. I think this is fine.

instead of

    import arviz as az
    import bokeh.plotting as bkp
    bkp.output_notebook()

one can now do

    import arviz as az
    az.backends.output_notebook()

The first option is a bit annoying in a long run, so having this wrapper really helps to fix this problem.

PR #979, opened 2019-12-29
Fixed bug in extracting prior samples for cmdstanpy.

justinbois:
When extracting *prior* samples from cmdstanpy output, the call to `_unpack_frame()`, *posterior* samples were being unpacked. I fixed this so that prior samples are unpacked.


ahartikainen:
Thanks for correcting this bug. 

Totally my fault.

justinbois:
My pleasure! I just wish I had more time so that I could contribute to this very useful project. Thanks  all for your hard work!

PR #981, opened 2019-12-30
infrerence_data to ColumnDataSource (bokeh)

ahartikainen:
This PR creates ColumnDataSource from the inference_data.

utils:
   function to transform inference data groups to dictionary (flattened against selected dimensions)

backends:
    function to create CDS object. Data as DataFrame.

The main idea is to create CDS either from `posterior_groups` or `prior_groups` so a user can plot multiple bokeh plots with the same CDS, which enables linked select.

### DIFF
--- a/arviz/utils.py
+++ b/arviz/utils.py
@@ -372,7 +372,7 @@ def flat_inference_data_to_dict(
                                    str: dim_separator_start,
                                    str: dim_separator_end)
                         group_info: (str: group separator start, str: group separator end)
-                Example: ((",", "[", "]"), "_", "")
+                Example: ((",", "[", "]"), ("_", ""))
                     - add_group_info == False: theta[0,0]
                     - add_group_info == True: theta_posterior[0,0]
     index_origin : int, optional
@@ -398,11 +398,15 @@ def flat_inference_data_to_dict(
 
     if dimensions is None:
         dimensions = "chain", "draw"
+    elif isinstance(dimensions, str):
+        dimensions = (dimensions,)
 
     if var_name_format is None:
         var_name_format = "brackets"
 
-    var_name_format = var_name_format.lower()
+    if isinstance(var_name_format, str):
+        var_name_format = var_name_format.lower()
+
     if var_name_format == "brackets":
         dim_join_separator, dim_separator_start, dim_separator_end = ",", "[", "]"
         group_separator_start, group_separator_end = "_", ""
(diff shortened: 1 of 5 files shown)


ahartikainen:
We still need to update backend code to enable reading CDS and add functionality to the backend which will handle our CDS format.

The variables are named as

    {var_name}_ARVIZ_CDS_GROUP_{group_name}_ARVIZ_CDS_SELECTION_{dims}

where `_ARVIZ_CDS_SELECTION_{dims}` is used on for nD variables (has dims more than `chain` and `draw`)

ahartikainen:
Currently can not eat `observed_data` or any other group which does not have `chain` and `draw` dimensions.

ahartikainen:
I think we could split that function so one creates dictionary --> as.utils

Function that transform dict to cds. Here.

It just need some possible group and dim formatting options.

arviz/utils.py
```diff
@@ -320,3 +323,167 @@ def expand_dims(x):
...
+                    tuple: (dim_info, group_info)
+                        dim_info: (str: `.join` separator,
+                                   str: dim_separator_start,
+                                   str: dim_separator_end)
+                        group_info: (str: group separator start, str: group separator end)
+                Example: ((",", "[", "]"), ("_", ""))
+                    - add_group_info == False: theta[0,0]
+                    - add_group_info == True: theta_posterior[0,0]
+    index_origin : int, optional
+        Start parameter indeces from `index_origin`. Either 0 or 1.
```
OriolAbril: I think it would be better to only use this when creating the inference data. From what I understand it is happening here (I have had no time to try it, only read the code) this assumes the index origin in InferenceData is always 0 and it adds one if index origin is 1. However, there is the possibility of the InferenceData index origin being 2000 if `.sel` has been used to discard some burn-in, so the index origin 1 would change the start from 2000 to 2001.

ahartikainen: Yeah, this makes sense

ahartikainen: No, I think index origin is only an artifact for flattened names. InferenceData is always 0-based.

ahartikainen: But I need to make syre that selection works as intended

OriolAbril: I have been thinking a little bit more about indexing, and it does make sense to only use it for visualization, not in Inference Data. We should probably create one issue with what is missing to integrate the rcparam in ArviZ.

The function in plot utils to make labels should probably modified, summary should also use the rcparam value as default, and maybe `sel` should also have an index origin argument? Otherwise the index shown in the plots cannot be used for selection right away. Would this be an issue?

ahartikainen: Yeah, I think those are currently only locations which needs a change.

ahartikainen: I changed this so it currently uses numpy indices, so it should work in all the cases for this function.

I also added rcparam for summary.

.sel thing needs some thinking.

PR #983, opened 2019-12-30
Populate InferenceData with out-of-sample prediction results from PyMC3 predictive samples

rpgoldman:
In this approach, unlike the one outlined in the schema, the `constant_data` contains the constant data used to generate the *predictions* not to generate the posterior_trace.
This could be modified, but I took this decision because the posterior trace used to generate the predictions in general CANNOT be the same as the posterior trace created by pymc3.sample() -- variables whose shape depends on the shape of the constant data or the observations must be removed.

### DIFF
--- a/arviz/data/io_pymc3.py
+++ b/arviz/data/io_pymc3.py
@@ -10,8 +10,12 @@ from .base import requires, dict_to_dataset, generate_dims_coords, make_attrs
 
 if TYPE_CHECKING:
     import pymc3 as pm
+    from pymc3 import MultiTrace, Model
     import theano
     from typing import Set
+else:
+    MultiTrace = Any
+    Model = Any
 
 _log = logging.getLogger(__name__)
 
@@ -270,6 +274,7 @@ class PyMC3Converter:  # pylint: disable=too-many-instance-attributes
                 constant_data_vars[var.name] = var
 
         def is_data(name, var) -> bool:
+            assert self.model is not None
             return var not in self.model.deterministics and var not in self.model.observed_RVs \
                and var not in self.model.free_RVs and \
                (self.observations is None or name not in self.observations)
--- a/arviz/data/io_pymc3.py
+++ b/arviz/data/io_pymc3.py
@@ -410,7 +410,9 @@ def predictions_from_pymc3(
     if idata_orig is None:
         return new_idata
     if inplace:
-        return cast(InferenceData, concat([idata_orig, new_idata], dim=None, inplace=True))
+        cast(InferenceData, concat([idata_orig, new_idata], dim=None, inplace=True))
+        return idata_orig
     # if we are not returning in place, then merge the old groups into the new inference
     # data and return that.
-    return cast(InferenceData, concat([new_idata, idata_orig], dim=None, copy=True, inplace=True))
+    concat([new_idata, idata_orig], dim=None, copy=True, inplace=True)
+    return new_idata
--- a/arviz/data/io_pymc3.py
+++ b/arviz/data/io_pymc3.py
@@ -17,7 +17,7 @@ Coords = Dict[str, List[Any]]
 Dims = Dict[str, List[str]]
 
 
-class PyMC3Converter: # pylint: disable=too-many-instance-attributes
+class PyMC3Converter:  # pylint: disable=too-many-instance-attributes
     """Encapsulate PyMC3 specific logic."""
 
     model = None  # type: Optional[pm.Model]
@@ -80,8 +80,10 @@ class PyMC3Converter: # pylint: disable=too-many-instance-attributes
                 get_from = posterior_predictive
             if get_from is None:
                 # pylint: disable=line-too-long
-                raise ValueError("""When constructing InferenceData must have at least
-                                    one of trace, prior, posterior_predictive or predictions.""")
+                raise ValueError(
+                    """When constructing InferenceData must have at least
+                                    one of trace, prior, posterior_predictive or predictions."""
+                )
 
             aelem = arbitrary_element(get_from)
             self.ndraws = aelem.shape[0]
@@ -182,10 +184,10 @@ class PyMC3Converter: # pylint: disable=too-many-instance-attributes
                 # pylint: disable=line-too-long
                 _log.warning(
                     "posterior predictive variable %s's shape not compatible with number of chains and draws. "
-                    "This can mean that some draws or even whole chains are not represented.", k
+                    "This can mean that some draws or even whole chains are not represented.",
+                    k,
                 )
         return dict_to_dataset(data, library=self.pymc3, coords=self.coords, dims=self.dims)
-
 
     @requires(["posterior_predictive"])
     def posterior_predictive_to_xarray(self):
@@ -255,24 +257,28 @@ class PyMC3Converter: # pylint: disable=too-many-instance-attributes
         # if both predictions AND trace are supplied, then the trace should be the
         # *thinned* trace used in prediction and NOT the full posterior trace. Hence we
         # give precedence here to checking the predictions.
-        model_vars = None             # type: Set[str]
+        model_vars = None  # type: Set[str]
         if self.predictions is not None:
-            model_vars = set(self.pymc3.util.get_default_varnames(
-                self.predictions.keys(),
-                include_transformed=True
-            ))
+            model_vars = set(
+                self.pymc3.util.get_default_varnames(
+                    self.predictions.keys(), include_transformed=True
+                )
+            )
         else:
             model_vars = set()
         if self.trace is not None:
-            model_vars = model_vars | \
-                         set(self.pymc3.util.get_default_varnames(  # pylint: disable=no-member
-                             self.trace.varnames, include_transformed=True
-                         ))
+            model_vars = model_vars | set(
+                self.pymc3.util.get_default_varnames(  # pylint: disable=no-member
+                    self.trace.varnames, include_transformed=True
+                )
+            )
         if self.observations is not None:
             # pylint: disable=line-too-long
-            model_vars = model_vars | \
-                         {obs.name for obs in self.observations.values() if hasattr(obs, "name")} | \
-                         set(self.observations.keys())
+            model_vars = (
+                model_vars
+                | {obs.name for obs in self.observations.values() if hasattr(obs, "name")}
+                | set(self.observations.keys())
+            )
 
         # this check is necessary in filtering constant variables because I found that some still
         # slipped through, notably the bounding transformed variables introduced by bounded and
@@ -283,7 +289,8 @@ class PyMC3Converter: # pylint: disable=too-many-instance-attributes
             return name
 
         constant_data_vars = {
-            name: var for name, var in self.model.named_vars.items() \
+            name: var
+            for name, var in self.model.named_vars.items()
             if untransformed_name(name) not in model_vars
         }
         if not constant_data_vars:
@@ -305,8 +312,8 @@ class PyMC3Converter: # pylint: disable=too-many-instance-attributes
             coords = {key: xr.IndexVariable((key,), data=coords[key]) for key in val_dims}
             try:
                 constant_data[name] = xr.DataArray(vals, dims=val_dims, coords=coords)
-            except ValueError as e: # pylint: disable=invalid-name
-                raise ValueError("Error translating constant_data variable %s: %s"%(name, e))
+            except ValueError as e:  # pylint: disable=invalid-name
+                raise ValueError("Error translating constant_data variable %s: %s" % (name, e))
         return xr.Dataset(data_vars=constant_data, attrs=make_attrs(library=self.pymc3))
 
     def to_inference_data(self):
@@ -317,13 +324,13 @@ class PyMC3Converter: # pylint: disable=too-many-instance-attributes
         will not have those groups.
         """
         id_dict = {
-                "posterior": self.posterior_to_xarray(),
-                "sample_stats": self.sample_stats_to_xarray(),
-                "posterior_predictive": self.posterior_predictive_to_xarray(),
-                "predictions": self.predictions_to_xarray(),
-                **self.priors_to_xarray(),
-                "observed_data": self.observed_data_to_xarray(),
-            }
+            "posterior": self.posterior_to_xarray(),
+            "sample_stats": self.sample_stats_to_xarray(),
+            "posterior_predictive": self.posterior_predictive_to_xarray(),
+            "predictions": self.predictions_to_xarray(),
+            **self.priors_to_xarray(),
+            "observed_data": self.observed_data_to_xarray(),
+        }
         if self.predictions:
             id_dict["predictions_constant_data"] = self.constant_data_to_xarray()
         else:
@@ -332,8 +339,14 @@ class PyMC3Converter: # pylint: disable=too-many-instance-attributes
 
 
 def from_pymc3(
-    trace=None, *, prior=None, posterior_predictive=None, predictions=None,
-    coords=None, dims=None, model=None
+    trace=None,
+    *,
+    prior=None,
+    posterior_predictive=None,
+    predictions=None,
+    coords=None,
+    dims=None,
+    model=None
 ):
     """Convert pymc3 data into an InferenceData object."""
     return PyMC3Converter(
@@ -346,8 +359,10 @@ def from_pymc3(
         model=model,
     ).to_inference_data()
 
-def predictions_from_pymc3(predictions, posterior_trace, model,
-                           coords=None, dims=None) -> InferenceData:
+
+def predictions_from_pymc3(
+    predictions, posterior_trace, model, coords=None, dims=None
+) -> InferenceData:
     """Translate out-of-sample predictions into ``InferenceData`` using ``from_pymc3``.
 
     Parameters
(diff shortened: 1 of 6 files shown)


rpgoldman:
A problem I am having with this is the division of variables into `predictions_constant_data` versus `predictions` in PyMC3.

Currently, what we do is we try to guess what's a constant value by looking at the variables and seeing if they are in the predictive trace (not constant), or if they are in the posterior trace or if they are in the observations.  If not, then we assume that they are constant data.

This really does not work, because the user, in PyMC3, can specify a set of variables of interest to be in the predictive trace (this defaults to the set of observed random variables). So membership in (or absence from) the trace is not a good way to determine anything about a random variable.

This means, if the user is not interested in a random variable, so omits it from the predictive trace, `constant_data_to_xarray` will misinterpret those omitted variables as being constant data.

I really think it would be better to simply *tell* the translator what is constant data, but I don't have a great plan for how to do this.  So if anyone does, please LMK.

rpgoldman:
Huge help from Brandon Willard of PyMC3 and Symbolic PyMC3 enabled me to more accurately determine which variables belong in `constant_data` or `predictions_constant_data`. I believe this is close to being ready to merge, if reviewed.

rpgoldman:
This won't pass tests until PyMC3 is fixed.  See https://github.com/pymc-devs/pymc3/pull/3763

arviz/data/io_pymc3.py
```diff
@@ -68,29 +77,36 @@ def arbitrary_element(dct: Dict[Any, np.ndarray]) -> np.ndarray:
...
-                if posterior_predictive is None
-                else arbitrary_element(posterior_predictive)
-            )
+            get_from = None
+            if predictions is not None:
+                get_from = predictions
+            elif prior is not None:
+                get_from = prior
+            elif posterior_predictive is not None:
+                get_from = posterior_predictive
```
OriolAbril: I would keep the priority order with posterior predictive before prior: `predictions`, `posterior_predictive`, `prior`. 

This is because the default in `pm.sample_posterior_predictive` is `samples=None` which will end up with ndraw samples, however, for `pm.sample_prior_predictive` the default is `samples=500` which will generally not be equal to ndraws.

arviz/data/io_pymc3.py
```diff
@@ -249,44 +287,94 @@ def constant_data_to_xarray(self):
...
+        return InferenceData(**id_dict)
 
 
 def from_pymc3(
-    trace=None, *, prior=None, posterior_predictive=None, coords=None, dims=None, model=None
+    trace=None,
+    *,
+    prior=None,
+    posterior_predictive=None,
+    predictions=None,
```
OriolAbril: I think having a `predictions` argument here will be quite confusing. 

I would only use `predictions` argument in `PyMC3Converter` class. Therefore it will not be visible to users in the `from_pymc3` docs but it will still be available for internal usage in `predictions_from_pymc3` (which would then use the class instead of `from_pymc3`, the code is basically the same in both cases).

arviz/data/io_pymc3.py
```diff
@@ -249,44 +287,94 @@ def constant_data_to_xarray(self):
...
         posterior_predictive=posterior_predictive,
+        predictions=predictions,
         coords=coords,
         dims=dims,
         model=model,
     ).to_inference_data()
+
+
+def predictions_from_pymc3(
+    predictions, posterior_trace, model, coords=None, dims=None
```
OriolAbril: I think that `posterior_trace` and `model` should default to `None`. 

I think that the most general use case will be calling this from inside the model context and without any thinning, and in this case, the model can be obtained from the context and the trace is not really needed.

rpgoldman:
@OriolAbril Could you LMK how you see this lining up with your work on #794?

Is #794 going to be be merged soon, so that I can rebase on top of it?  Should this MR branch of off yours (I don't like that idea, because I think it will be error-prone)?  Or do you think we could come up with some subset of #794 that could be merged to master that would be enough to harmonize my work with yours, but still allow you to keep working on yours separately?

OriolAbril:
I think both need to be thoroughly reviewed, and #794 has been waiting for quite long so it can wait until this has been merged. 

I don't mind rebasing on top of this once merged, and I actually think that they will merge quite well, here there are nearly no modifications to sample stats/likelihood handling. To be extra sure though, are you planning on adding some tests here (once the functionality is decided) or in a future PR? We could also wait for the tests PR if it is not too long.

Does this sound good @rpgoldman ?

rpgoldman:
I'm not sure exactly how to improve the tests.  In particular:

1. We test on python 3.5, but I believe PyMC3 no longer supports anything before 3.6.  Should we drop the `io_pymc3` tests on 3.5?
2. Here's a test failure that requires a bug-fix on PyMC3:
```
=================================== FAILURES ===================================
_________ TestDataPyMC3.test_multiple_observed_rv_without_observations _________

self = <arviz.tests.test_data_pymc.TestDataPyMC3 object at 0x7f034d8d2198>

    def test_multiple_observed_rv_without_observations(self):
        with pm.Model():
            mu = pm.Normal("mu")
            x = pm.DensityDist(  # pylint: disable=unused-variable
                "x", pm.Normal.dist(mu, 1.0).logp, observed={"value": 0.1}
            )
>           trace = pm.sample(100, chains=2)

arviz/tests/test_data_pymc.py:128: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/local/envs/testenv_3.5_PYSTAN_latest_PYRO_latest_EMCEE_latest_TF_latest/lib/python3.5/site-packages/pymc3/sampling.py:498: in sample
    trace.report._run_convergence_checks(trace, model)
/usr/local/envs/testenv_3.5_PYSTAN_latest_PYRO_latest_EMCEE_latest_TF_latest/lib/python3.5/site-packages/pymc3/backends/report.py:84: in _run_convergence_checks
    self._ess = ess = ess(trace, var_names=varnames)
/usr/local/envs/testenv_3.5_PYSTAN_latest_PYRO_latest_EMCEE_latest_TF_latest/lib/python3.5/site-packages/pymc3/stats/__init__.py:24: in wrapped
    return func(*args, **kwargs)
arviz/stats/diagnostics.py:187: in ess
    dataset = convert_to_dataset(data, group="posterior")
arviz/data/converters.py:168: in convert_to_dataset
    inference_data = convert_to_inference_data(obj, group=group, coords=coords, dims=dims)
arviz/data/converters.py:89: in convert_to_inference_data
    return from_pymc3(trace=kwargs.pop(group), **kwargs)
arviz/data/io_pymc3.py:346: in from_pymc3
    model=model,
arviz/data/io_pymc3.py:324: in to_inference_data
    id_dict["constant_data"] = self.constant_data_to_xarray()
arviz/data/base.py:36: in wrapped
    return func(cls, *args, **kwargs)
arviz/data/base.py:36: in wrapped
    return func(cls, *args, **kwargs)
arviz/data/io_pymc3.py:277: in constant_data_to_xarray
    if is_data(name, var):
arviz/data/io_pymc3.py:270: in is_data
    return var not in self.model.deterministics and var not in self.model.observed_RVs \
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pymc3.model.MultiObservedRV object at 0x7f038c06cac8>, other = mu

    def __eq__(self, other):
>       return self.id == other.id
E       AttributeError: 'MultiObservedRV' object has no attribute 'id'
```
This was only fixed on 7 Jan 2020, and is not in the released version of PyMC3.

arviz/data/io_pymc3.py
rpgoldman: I have a tentative version of the above and will push it momentarily. Note that I have augmented the `orig_idata` by hand instead of using `concat`.  LMK if you think `concat` would be better.

arviz/data/io_pymc3.py
rpgoldman: Done.

rpgoldman:
@OriolAbril OK, this should have all the changes you requested.

I'll check in with PyMC3 about maybe cutting a bug-fix release.

arviz/data/io_pymc3.py
```diff
@@ -370,11 +385,20 @@ def predictions_from_pymc3(
...
     """
-    return from_pymc3(
+    new_idata = PyMC3Converter(
         trace=posterior_trace, predictions=predictions, model=model, coords=coords, dims=dims
-    )
+    ).to_inference_data()
+    if idata_orig is None:
+        return new_idata
+    idata_orig.predictions = new_idata.predictions
+    idata_orig.predictions_constant_data = new_idata.predictions_constant_data
```
OriolAbril: I would use concat for two reasons: the first is that this would allow to use `**kwargs` and pass them to `concat` (which already has both inplace and copy arguments), the second is that this does not update the `_groups` attribute (used to iterate over groups in several cases such as saving or printing)

OriolAbril:
> I'll check in with PyMC3 about maybe cutting a bug-fix release.

@canyon289 is the expert, but maybe we can mark the failing test with some fixture in python 3.5 so CI build passes.

arviz/data/io_pymc3.py
rpgoldman: OK, I will change that and push.

rpgoldman:
@OriolAbril Why do you think the `.pyi` file should be excluded from the repository?  According to the various PEPs about type hinting, this is the right way to make type hints for a library, and in the case of Arviz, this stub file "explains" to users of `InferenceData` how the groups work, since otherwise, users of the API cannot be confident that, for example, the `posterior` attribute is legitimate to use.

OriolAbril:
> @OriolAbril Why do you think the `.pyi` file should be excluded from the repository? According to the various PEPs about type hinting, this is the right way to make type hints for a library, and in the case of Arviz, this stub file "explains" to users of `InferenceData` how the groups work, since otherwise, users of the API cannot be confident that, for example, the `posterior` attribute is legitimate to use.

Basically ignorance about type hinting. I think I had never seen a `.pyi` file before, so my first thought was "this has somehow avoided the gitignore"

rpgoldman:
> > @OriolAbril Why do you think the `.pyi` file should be excluded from the repository? According to the various PEPs about type hinting, this is the right way to make type hints for a library, and in the case of Arviz, this stub file "explains" to users of `InferenceData` how the groups work, since otherwise, users of the API cannot be confident that, for example, the `posterior` attribute is legitimate to use.
> 
> Basically ignorance about type hinting. I think I had never seen a `.pyi` file before, so my first thought was "this has somehow avoided the gitignore"

In that case, I will leave it.  If anyone else uses `mypy`, it might help them, and if they don't it will do no harm.

rpgoldman:
By the way, I was not able to run black on `io_pymc3.py` for some reason.  It failed to parse the `PyMC3Converter` `__init__()` method. No idea why.

OriolAbril:
I fetched the code from this PR and managed to run black on it. Maybe you have an old version?

rpgoldman:
> I fetched the code from this PR and managed to run black on it. Maybe you have an old version?

Pretty sure not: I updated it after the first failure.

I *think* you might be able to commit the reformatted file to this MR.  If you can figure out how, please do.  If you can't, please either email it to me, or message it to me on Slack, and I will.

OriolAbril:
Managed to push to your branch the black formatting

rpgoldman:
@OriolAbril 

1.  I discovered an issue in log_likelihood handling in `predictions_from_pymc3()`.  If we are building `InferenceData` from pymc3 predictions out of sample, then we may have to have deformed the trace to a point where it does not support extraction of the log likelihood.

I'm going to assume that it's OK not to have log likelihoods in this case.  The reasoning would be that if you have done out of sample predictions that require modification of the trace, then you should have *another* `InferenceData` from which the log likelihood can be extracted and investigated.  I will modify the tests so that they regard not having log likelihood in this case as being OK, unless you have some idea about how I should recompute them (presumably from the modified model).  I think that should be left for a further MR if it's required.

I just wanted to explain why this is happening.

2. I think for uniformity in naming, I should rename `predictions_from_pymc3()` to `from_pymc3_predictions()`, and if you have any ideas for a better name than that, please suggest.

3.  I expect a little more activity to add tests for `predictions_constant_data`, and when those pass, I hope we can merge (I think I'll compress the history a bit first).

OriolAbril:
> 1. I discovered an issue in log_likelihood handling in `predictions_from_pymc3()`.  If we are building `InferenceData` from pymc3 predictions out of sample, then we may have to have deformed the trace to a point where it does not support extraction of the log likelihood

I thought the only constraint is that if there are `predictions` (hence the model has been modified for out of sample posterior predictive sampling) log likelihood cannot be extracted because it would raise an error when calling theano. Is this right? Are there other exceptions?

> 2. I think for uniformity in naming, I should rename `predictions_from_pymc3()` to `from_pymc3_predictions()`, and if you have any ideas for a better name than that, please suggest.

I like both names, I don't have any strong opinion on this. Like this looks fine.

rpgoldman:
> > 1. I discovered an issue in log_likelihood handling in `predictions_from_pymc3()`.  If we are building `InferenceData` from pymc3 predictions out of sample, then we may have to have deformed the trace to a point where it does not support extraction of the log likelihood
> 
> I thought the only constraint is that if there are `predictions` (hence the model has been modified for out of sample posterior predictive sampling) log likelihood cannot be extracted because it would raise an error when calling theano. Is this right? Are there other exceptions?

No other exceptions, and yes, this is the limitation.  It seems regrettable, but I don't see any way to avoid it.
> 
> > 1. I think for uniformity in naming, I should rename `predictions_from_pymc3()` to `from_pymc3_predictions()`, and if you have any ideas for a better name than that, please suggest.
> 
> I like both names, I don't have any strong opinion on this. Like this looks fine.

I guess I thought having everything start with `from_` made it more clear.

rpgoldman:
OK, this all works for me, and if you check out PyMC3 at `master`, it should work for you, too.  Unfortunately, right now this is broken working against PyMC3 as released because of a bug in the released version.
I will look into monkey-patching PyMC3 to make it work with the current version, too....

arviz/data/io_pymc3.py
```diff
@@ -1,29 +1,54 @@
...
 from .base import requires, dict_to_dataset, generate_dims_coords, make_attrs
 
 if TYPE_CHECKING:
     import pymc3 as pm
+    from pymc3 import MultiTrace, Model  # pylint: disable=invalid-name
+    import theano
+    from typing import Set  # pylint: disable=ungrouped-imports
+else:
+    MultiTrace = Any  # pylint: disable=invalid-name
+    Model = Any  # pylint: disable=invalid-name
```
OriolAbril: I was just thinking, once we start adding type hinting to ArviZ, the use of `Any` may become quite common, maybe it is worth it to include `Any` as a good name in pylintrc

arviz/data/io_pymc3.py
```diff
@@ -1,29 +1,54 @@
...
+___all__ = [""]
 
 _log = logging.getLogger(__name__)
 
 Coords = Dict[str, List[Any]]
 Dims = Dict[str, List[str]]
+# random variable object ...
+Var = Any  # pylint: disable=invalid-name
+
+# pylint: disable=line-too-long
```
OriolAbril: What is the scope of this diasable? (mainly curiosity)

arviz/data/io_pymc3.py
rpgoldman: Actually, it was `Model` and `MultiTrace` that pylint didn't like, because they look like class names not variable names -- which is what they are!  I'm pretty sure the `Any` was fine.

arviz/data/io_pymc3.py
rpgoldman: > What is the scope of this disable? (mainly curiosity)

I'm afraid I don't know as well as I should.  I'm going to try to leave it out and see what happens.  If it's not necessary (after black reformatting), or if it can be more tightly scoped, I will cut or mve it.

arviz/data/io_pymc3.py
OriolAbril: Ohhh, yeah, that makes much more sense

arviz/data/io_pymc3.py
rpgoldman: I cut it out -- I think it was there because my editor didn't look at the right .pylintrc, and was holding me to 80 characters.  Passes pylint here without it.  Assuming it passes in CI, I think we are good.

rpgoldman:
@OriolAbril Assuming that this passes all the tests again, how should we merge it? The git history is cluttered, and I was thinking of, for example, using rebasing to merge together a bunch of the commits for lint and black compliance.

How about I wait for this bout of tests to pass, then clean up the history, and then if it's still passing tomorrow with the cleaned up history, you or I can merge it?

OriolAbril:
All merges in ArviZ are squash merges, so the PR history does not really matter. I can merge afterwards

rpgoldman:
> All merges in ArviZ are squash merges, so the PR history does not really matter. I can merge afterwards

Oh, great.  Then I'll skip the history editing.  I will push my changelog entry then, and we're done!

rpgoldman:
@OriolAbril Thanks!

PR #989, opened 2020-01-08
Fix erroneous warning in traceplot

canyon289:
Fixes erroneous warning reported by @fonnesbeck


ahartikainen:
Line 290 too

ahartikainen:
I think we can remove keywords and use the default

PR #994, opened 2020-01-13
Solve issue #992 : Integrate point_estimate with rcParam

### LINKED ISSUE #992: integrate point_estimate rcParam
## Tell us about it

`plot.point_estimate` has a default value in rcParams which is not yet used. Some plotting functions include a `point_estimate` kwarg to summarize the distribution and help its interpretation. `point_estimate` can take 4 possible values: mean, median, mode or `None`. 

Part of #792 

## Thoughts on implementation

`point_estimate` is currently hardcoded to have mean as its default, however, it would be interesting to change this default on a user or project basis. 

Appearances of `point_estimate` as argument should be set to have `"auto"` as a default, and in this case, to get the value from rcParams. This would allow users to still use `point_estimate=None` to indicate no point is to be plotted.

percygautam:
Fixes #992 . The following changes are made:

- [x] Added "auto" as default-type for point_estimate param for `densityplot.py` and `posteriorplot.py`

- [x] Added the docstring for changes.

- [x] Passes the pytest.

### DIFF
--- a/arviz/plots/__init__.py
+++ b/arviz/plots/__init__.py
@@ -9,7 +9,7 @@ from .essplot import plot_ess
 from .forestplot import plot_forest
 from .hpdplot import plot_hpd
 from .jointplot import plot_joint
-from .kdeplot import plot_kde, _fast_kde, _fast_kde_2d
+from .kdeplot import plot_kde, _fast_kde_2d
 from .khatplot import plot_khat
 from .loopitplot import plot_loo_pit
 from .mcseplot import plot_mcse
@@ -20,6 +20,7 @@ from .ppcplot import plot_ppc
 from .rankplot import plot_rank
 from .traceplot import plot_trace
 from .violinplot import plot_violin
+from .plot_utils import _fast_kde
 
 
 __all__ = [
--- a/arviz/plots/__init__.py
+++ b/arviz/plots/__init__.py
@@ -9,7 +9,7 @@ from .essplot import plot_ess
 from .forestplot import plot_forest
 from .hpdplot import plot_hpd
 from .jointplot import plot_joint
-from .kdeplot import plot_kde, _fast_kde_2d
+from .kdeplot import plot_kde
 from .khatplot import plot_khat
 from .loopitplot import plot_loo_pit
 from .mcseplot import plot_mcse
@@ -20,7 +20,7 @@ from .ppcplot import plot_ppc
 from .rankplot import plot_rank
 from .traceplot import plot_trace
 from .violinplot import plot_violin
-from .plot_utils import _fast_kde
+from .plot_utils import _fast_kde, _fast_kde_2d
 
 
 __all__ = [
--- a/arviz/plots/backends/bokeh/densityplot.py
+++ b/arviz/plots/backends/bokeh/densityplot.py
@@ -5,11 +5,11 @@ from bokeh.layouts import gridplot
 from bokeh.models.annotations import Title
 
 from . import backend_kwarg_defaults, backend_show
-from ...kdeplot import _fast_kde
 from ...plot_utils import (
     make_label,
     _create_axes_grid,
     calculate_point_estimate,
+    _fast_kde,
 )
 from ....stats import hpd
 from ....stats.stats_utils import histogram
--- a/arviz/plots/backends/bokeh/densityplot.py
+++ b/arviz/plots/backends/bokeh/densityplot.py
@@ -6,7 +6,11 @@ from bokeh.models.annotations import Title
 
 from . import backend_kwarg_defaults, backend_show
 from ...kdeplot import _fast_kde
-from ...plot_utils import _create_axes_grid, make_label
+from ...plot_utils import (
+    make_label,
+    _create_axes_grid,
+    calculate_point_estimate,
+)
 from ....stats import hpd
 from ....stats.stats_utils import histogram
 
@@ -185,11 +189,8 @@ def _d_helper(
         ax.diamond(xmin, 0, line_color="black", fill_color=color, size=markersize)
         ax.diamond(xmax, 0, line_color="black", fill_color=color, size=markersize)
 
+    est = calculate_point_estimate(point_estimate, vec, bw)
     if point_estimate is not None:
-        if point_estimate == "mean":
-            est = np.mean(vec)
-        elif point_estimate == "median":
-            est = np.median(vec)
         ax.circle(est, 0, fill_color=color, line_color="black", size=markersize)
 
     _title = Title()
--- a/arviz/plots/backends/bokeh/densityplot.py
+++ b/arviz/plots/backends/bokeh/densityplot.py
@@ -189,8 +189,8 @@ def _d_helper(
         ax.diamond(xmin, 0, line_color="black", fill_color=color, size=markersize)
         ax.diamond(xmax, 0, line_color="black", fill_color=color, size=markersize)
 
-    est = calculate_point_estimate(point_estimate, vec, bw)
     if point_estimate is not None:
+        est = calculate_point_estimate(point_estimate, vec, bw)
         ax.circle(est, 0, fill_color=color, line_color="black", size=markersize)
 
     _title = Title()
--- a/arviz/plots/backends/bokeh/posteriorplot.py
+++ b/arviz/plots/backends/bokeh/posteriorplot.py
@@ -187,9 +187,9 @@ def _plot_posterior_op(
         ax.text(x=vals, y=[max_data * 0.2, max_data * 0.2], text=list(map(str, vals)), **text_props)
 
     def display_point_estimate(max_data):
-        point_value = calculate_point_estimate(point_estimate, values, bw)
         if not point_estimate:
             return
+        point_value = calculate_point_estimate(point_estimate, values, bw)
         sig_figs = format_sig_figs(point_value, round_to)
         point_text = "{point_estimate}={point_value:.{sig_figs}g}".format(
             point_estimate=point_estimate, point_value=point_value, sig_figs=sig_figs
--- a/arviz/plots/backends/bokeh/posteriorplot.py
+++ b/arviz/plots/backends/bokeh/posteriorplot.py
@@ -9,12 +9,13 @@ from bokeh.models.annotations import Title
 from scipy.stats import mode
 
 from . import backend_kwarg_defaults, backend_show
-from ...kdeplot import plot_kde, _fast_kde
+from ...kdeplot import plot_kde
 from ...plot_utils import (
     make_label,
     _create_axes_grid,
     format_sig_figs,
     round_num,
+    calculate_point_estimate,
 )
 from ....stats import hpd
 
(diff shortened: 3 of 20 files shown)


arviz/plots/densityplot.py
```diff
@@ -61,8 +61,8 @@ def plot_density(
     credible_interval : float
         Credible intervals. Should be in the interval (0, 1]. Defaults to 0.94.
     point_estimate : Optional[str]
-        Plot point estimate per variable. Values should be 'mean', 'median' or None.
-        Defaults to 'mean'.
+        Plot point estimate per variable. Values should be 'mean', 'median', 'mode' or None.
+        Defaults to 'auto'.
```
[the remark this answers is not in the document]
percygautam: Okay, makes sense. I'll add it right away.

arviz/plots/posteriorplot.py
```diff
@@ -59,7 +60,7 @@ def plot_posterior(
     round_to : int, optional
         Controls formatting of floats. Defaults to 2 or the integer part, whichever is bigger.
     point_estimate: str
-        Must be in ('mode', 'mean', 'median', None)
+        Must be in ('mode', 'mean', 'median', None). Defaults to 'auto'.
```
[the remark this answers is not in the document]
percygautam: Will do.

arviz/tests/test_plots_matplotlib.py
```diff
@@ -657,7 +657,7 @@ def test_plot_rank(models, kwargs):
         {"rope": {"mu": [{"rope": (-2, 2)}], "theta": [{"school": "Choate", "rope": (2, 4)}]}},
         {"point_estimate": "mode"},
         {"point_estimate": "median"},
-        {"point_estimate": False},
+        {"point_estimate": "mean"},
```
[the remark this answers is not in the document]
percygautam: Yes `None` makes more sense as rest three are tested many times. Previously, the tests are failing in pytest. I was getting the following ValueError:
`
ValueError: Point estimate should be 'mean', 'median', 'mode' or None, not False
`
Can you explain this behavior?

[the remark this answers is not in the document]
percygautam: I'll make it `None`.

arviz/tests/test_plots_bokeh.py
```diff
@@ -883,7 +883,7 @@ def test_plot_ppc_ax(models, kind):
         {"rope": {"mu": [{"rope": (-2, 2)}], "theta": [{"school": "Choate", "rope": (2, 4)}]}},
         {"point_estimate": "mode"},
         {"point_estimate": "median"},
-        {"point_estimate": False},
+        {"point_estimate": "mean"},
```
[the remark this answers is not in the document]
percygautam: Will do.

percygautam:
@OriolAbril I read the matplotlib files of `plot_density` and `plot_posterior`. I would like to add mode option to `plot_density` in this PR itself. We already got so many issues, I don't see the point to make another one if it can be solved. Please guide me through it!

OriolAbril:
Great! The currently currently has some duplication when calculating the point estimates. The skeleton of the code is basically some ifs clauses and inside the calculation of the point estimate using a `vec`/`values` variable (that is a 1d array). 

It would be great to create a function in `plot_utils` similar to this: 

```
def calculate_point_estimate(point_estimate, values, bw):
    # point_estimate validation and raise error here
    # if mean
        calculations
    # if median
        ...
    return point_value
```

The code inside the function would be more or less the same as the current code in `posteriorplot`. Then, when drawing the point estimate in `plot_posterior` and `plot_density` (both bokeh and matplotlib, so 4 files) the code would be simplified to calling this function (instead of the ifs) and then plotting (which I think won't need to be modified).

Also, both the check for good values plus error raising and getting the default from rcParams can also be moved to this new internal function.

percygautam:
@OriolAbril I have done the changes you asked. Some tests were failing. Kindly review the work. I'll do black formatting as soon as tests passes

ahartikainen:
Can you move `_fast_kde` functionality to `plot_utils`. There is an error in import order.

percygautam:
@OriolAbril  Okay I'll do that.

percygautam:
@ahartikainen Okay I'll do that but a fourier transform method in plot utilities doesn't make sense. Isn't there any workaround to the problem.

ahartikainen:
I think it can go there until we update our KDE code and move it to another location (outside plotting).

percygautam:
@ahartikainen @OriolAbril , I have done the changes asked.

percygautam:
@OriolAbril @ahartikainen  The _fast_kde function is shoeing to be protected and can't be used outside directory `plots`.

arviz/stats/stats.py
```diff
@@ -12,9 +12,8 @@
 from scipy.optimize import minimize
 import xarray as xr
 
+from ..plots.plot_utils import *
```
OriolAbril: Make the import explicit `from ..plots.plot_utils import _fast_kde, get_bins`

percygautam: @OriolAbril I already tried that, it is showing import error and the tests are not running

OriolAbril: `*` will surely not import `_fast_kde` because `*` only imports public functions. Using explicit import should work though, just like it worked with `from ..plots.kdeplot import _fast_kde` which is also on another folder). Could there have been a typo? I currently cannot test it nor work on this locally, this is as much as I can do for now.

percygautam: @OriolAbril I'll explicit import the function, so you can review!

percygautam:
@OriolAbril @ahartikainen There was some the dependency issue which was causing the import error. All set now!

percygautam:
@OriolAbril @ahartikainen There were linting issues. Should be okay now.

percygautam:
@OriolAbril  I am getting error with pylint `No name 'gaussian' in module 'scipy.signal'`.
I guess this is problem with pylint with `gausian` not registered.

arviz/plots/backends/bokeh/densityplot.py
```diff
@@ -185,11 +189,8 @@ def _d_helper(
         ax.diamond(xmin, 0, line_color="black", fill_color=color, size=markersize)
         ax.diamond(xmax, 0, line_color="black", fill_color=color, size=markersize)
 
+    est = calculate_point_estimate(point_estimate, vec, bw)
```
OriolAbril: I would move that inside the if, there is no need to calculate the point value if it will not be plotted

arviz/plots/backends/bokeh/posteriorplot.py
```diff
@@ -187,21 +187,9 @@ def display_rope(max_data):
         ax.text(x=vals, y=[max_data * 0.2, max_data * 0.2], text=list(map(str, vals)), **text_props)
 
     def display_point_estimate(max_data):
+        point_value = calculate_point_estimate(point_estimate, values, bw)
```
OriolAbril: same comment, here it means _after_ the if instead of inside though

arviz/plots/backends/matplotlib/densityplot.py
```diff
@@ -155,11 +160,8 @@ def _d_helper(
         ax.plot(xmin, 0, hpd_markers, color=color, markeredgecolor="k", markersize=markersize)
         ax.plot(xmax, 0, hpd_markers, color=color, markeredgecolor="k", markersize=markersize)
 
+    est = calculate_point_estimate(point_estimate, vec, bw)
```
OriolAbril: inside the if

arviz/plots/backends/matplotlib/posteriorplot.py
```diff
@@ -179,21 +179,9 @@ def display_rope():
         ax.text(vals[1], plot_height * 0.2, vals[1], weight="semibold", **text_props)
 
     def display_point_estimate():
+        point_value = calculate_point_estimate(point_estimate, values, bw)
```
OriolAbril: after the if

arviz/stats/stats.py
```diff
@@ -12,9 +12,8 @@
 from scipy.optimize import minimize
 import xarray as xr
 
+from ..plots.plot_utils import _fast_kde, get_bins
 from ..data import convert_to_inference_data, convert_to_dataset, InferenceData, CoordSpec, DimSpec
-from ..plots.kdeplot import _fast_kde
-from ..plots.plot_utils import get_bins
 from .diagnostics import _multichain_statistics, _mc_error, ess, _circular_standard_deviation
 from .stats_utils import (
     make_ufunc as _make_ufunc,
```
OriolAbril: below on line 27, histogram is imported from stats_utils in a weird way, it can be moved to this import.

I think this triggers some kind of circular import and is the root of the import problems

percygautam: @OriolAbril I noticed this problem earlier too and has removed the import problem with `b2f600d` commit.

arviz/plots/backends/bokeh/densityplot.py
percygautam: Yeah, makes sense. Will do.

arviz/stats/stats.py
OriolAbril: Yes, but the code should use the `histogram` function, not rewrite it here from scratch.

percygautam:
@OriolAbril Where do we need to add warning in the docstring? I have added it in the function `_fast_kde` as it is using gaussian function. The other changes you asked are done.

OriolAbril:
The warning is in the docstring of `scipy.signal.gaussian`, I was thinking we could follow its advise and start using `scipy.signal.windows.gaussian` instead. 

Also, now tests seem to pass and ArviZ can be imported, but we should move the [from ..stats.stats_utils import histogram](https://github.com/percygautam/arviz/blob/ec3374bb7bc9b3d1a42923ec7a24866bbc5ca583/arviz/stats/stats.py#L26) to the imports right above:

```
from .stats_utils import (
    make_ufunc as _make_ufunc,
    wrap_xarray_ufunc as _wrap_xarray_ufunc,
    logsumexp as _logsumexp,
    ELPDData,
    stats_variance_2d as svar,
    histogram  # here is its proper place
)
```

percygautam:
@OriolAbril Thanks for the clarification. I didn't understand the problem earlier. I did the changes asked.

arviz/plots/__init__.py
```diff
@@ -9,7 +9,7 @@
 from .forestplot import plot_forest
 from .hpdplot import plot_hpd
 from .jointplot import plot_joint
-from .kdeplot import plot_kde, _fast_kde, _fast_kde_2d
+from .kdeplot import plot_kde, _fast_kde_2d
```
ahartikainen: You can also move `_fast_kde_2d` next to `_fast_kde`, sorry for not saying it explicitly before

percygautam:
@ahartikainen Should I do these changes in this PR itself or create another one after this is merged. Transferring `_fast_kde_2d` to `plot_utils.py` will again cause dependency issues as they did earlier with `_fast_kde`.

OriolAbril:
After modifying the import of `histogram` from stats_utils should be safe to move `fast_kde_2d`, we can do it here.

percygautam:
@OriolAbril I am moving all derived private functions `_cov_1d`, `_cov` and `_dot` in `_fast_kde_2d` to `plot_utils.py`, as it would make sense. Is it okay?

OriolAbril:
LGTM

PR #997, opened 2020-01-14
Violinplot:  fix histogram, add rug

aloctavodia:
The "violin" plot had problem for discrete variables. This fix it and also add the option to display a jittered "rug" plot (defaults to False).

lala

I still needs to add the "rug" for bokeh


ahartikainen:
I can help.

PR #1001, opened 2020-01-14
[WIP] Solve issue #821 : Clarify the usage of "plot_joint"

### LINKED ISSUE #821: Clarify usage of `plot_joint`
## Short Description

The description of how to use `plot_joint` is not very clear about the use and meaning of parameters.

This is partly a usage question and partly a request for documentation improvement.

## Code Example or link

Looking at the docs for `arviz.plot_joint`, I see this: 
```
var_names : Iter of 2 e.g. (var_1, var_2)
    Variables to be plotted, two variables are required.
```
But then when I look at the example, I see this:
```
az.plot_joint(data,
>>>                 var_names=['theta'],
....
```
So … what is an “iter of 2”? `['theta']` is not obviously one, and the example gives a tuple of length 2.
I think what happens in this example is that you are plotting the *single* variable `theta` across a *coordinate* of dimension 2, but this is really quite confusing, and could use a much better explanation.

arviz 0.4.1 (will update!)

percygautam:
Resolves #821 . The following changes are made:

- Changed the docstring for the use of `plot_joint`.


arviz/plots/jointplot.py
```diff
@@ -29,8 +29,9 @@ def plot_joint(
     data : obj
         Any object that can be converted to an az.InferenceData object
         Refer to documentation of az.convert_to_dataset for details
-    var_names : Iter of 2 e.g. (var_1, var_2)
-        Variables to be plotted, two variables are required.
+    var_names : tuple
+        Variables to be plotted.
+        Two variables(or one variable with multiple dimensions) are required.
```
OriolAbril: There are actually 2 cases only: either 2 variables or 1 variable (in which case coords must select a subset of exactly 2 dimensions). The type could show this with something like ` var_names : str or iterable of str` and the description explain these 2 cases.

percygautam: @OriolAbril Okay I'll do these changes.

rpgoldman:
It would be worth mentioning something about the groups in the data argument. I recall being surprised that I couldn't plot a prior, for example, or constant data against samples. 

This can be done, of course, but only by pulling things out of the inference data by hand.

OriolAbril:
There are several plots that could benefit from adding a `group` argument, I'll open some issues, feel free to add some more too.

PR #1008, opened 2020-01-15
[WIP] Add group argument in plot_ppc (fix #1002)

### LINKED ISSUE #1002: Add group keyword to plot_ppc
In addition to posterior predictive checks, prior predictive checks are also a way to check if the desired regions of the parameter space are explored. Adding a group keyword (taking as values either `posterior` or `prior`) would allow to use data from the posterior predictive or prior predictive without much changes to the plotting code.

nitishp25:
- [x] Add ```group``` argument in ```plot_ppc``` in ```ppcplot.py```

- [x] Update bokeh backend of ```plot_ppc```

- [x] Update matplotlib backend of ```plot_ppc```

fixes #1002

### DIFF
--- a/arviz/plots/ppcplot.py
+++ b/arviz/plots/ppcplot.py
@@ -38,34 +38,34 @@ def plot_ppc(
     backend=None,
     backend_kwargs=None,
     show=None,
+    group="posterior_predictive",
 ):
     """
-    Plot for posterior predictive checks.
+    Plot for posterior/prior predictive checks.
 
     Parameters
     ----------
     data : az.InferenceData object
-        InferenceData object containing the observed and posterior
-        predictive data.
+        InferenceData object containing the observed and posterior/prior predictive data.
     kind : str
         Type of plot to display (kde, cumulative, or scatter). Defaults to kde.
     alpha : float
-        Opacity of posterior predictive density curves. Defaults to 0.2 for kind = kde
-        and cumulative, for scatter defaults to 0.7
+        Opacity of posterior/prior predictive density curves.
+        Defaults to 0.2 for kind = kde and cumulative, for scatter defaults to 0.7
     mean : bool
-        Whether or not to plot the mean posterior predictive distribution. Defaults to True
+        Whether or not to plot the mean posterior/prior predictive distribution. Defaults to True
     figsize : tuple
         Figure size. If None it will be defined automatically.
     textsize: float
         Text size scaling factor for labels, titles and lines. If None it will be
         autoscaled based on figsize.
     data_pairs : dict
-        Dictionary containing relations between observed data and posterior predictive data.
+        Dictionary containing relations between observed data and posterior/prior predictive data.
         Dictionary structure:
         Key = data var_name
-        Value = posterior predictive var_name
+        Value = posterior/prior predictive var_name
         For example, `data_pairs = {'y' : 'y_hat'}`
-        If None, it will assume that the observed data and the posterior
+        If None, it will assume that the observed data and the posterior/prior
         predictive data have the same variable name.
     var_names : list
         List of variables to be plotted. Defaults to all observed variables in the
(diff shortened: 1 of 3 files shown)


arviz/plots/ppcplot.py
```diff
@@ -38,34 +38,34 @@ def plot_ppc(
     backend=None,
     backend_kwargs=None,
     show=None,
```
OriolAbril: `group` must be added here somewhere.

arviz/plots/ppcplot.py
```diff
@@ -110,6 +110,9 @@ def plot_ppc(
         check the plotting method of the backend.
     show : bool, optional
         Call backend show function.
+    group : str, optional
```
OriolAbril: I would use `group : {"prior", "posterior"}, optional` trying to follow numpydoc style. Also, just to be extra clear, I would only allow two options, prior or posterior.

arviz/plots/ppcplot.py
```diff
@@ -157,11 +160,10 @@ def plot_ppc(
...
     """
-    for group in ("posterior_predictive", "observed_data"):
-        if not hasattr(data, group):
-            raise TypeError(
-                '`data` argument must have the group "{group}" for ppcplot'.format(group=group)
-            )
+    if group not in ("posterior_predictive", "prior_predictive", "observed_data"):
+        raise TypeError(
+            '`group` argument must be `posterior_predictive`, `prior_predictive` or `observed_data`'
+        )
```
OriolAbril: I would divide the argument checks in 2 parts. First check that the `group` is one of the valid options, and then make sure that the 2 needed groups are present in the inference data object (like it was done before with the hasattr). Also, note that if `group="posterior"` then the inference data must have posterior predictive and observed data groups but _prior predictive is not needed_

OriolAbril: Also, make sure to not use `group` as argument and loop variable at the same time ;)

arviz/plots/ppcplot.py
nitishp25: Ohh, what is the reason for that?

Should I add it below ```backend_kwargs```?

arviz/plots/ppcplot.py
nitishp25: Yeah that makes sense

arviz/plots/ppcplot.py
nitishp25: Yeah actually I was confused earlier about how to check for the group and display error, now I got it. Thanks :)

arviz/plots/ppcplot.py
```diff
@@ -157,11 +160,20 @@ def plot_ppc(
 
         >>> az.plot_ppc(data, num_pp_samples=30, random_seed=7)
     """
-    for group in ("posterior_predictive", "observed_data"):
-        if not hasattr(data, group):
-            raise TypeError(
-                '`data` argument must have the group "{group}" for ppcplot'.format(group=group)
-            )
+    if group == "posterior":
```
OriolAbril: We should start by checking `group in ("prior", "posterior"):` and raise an error if it is not. 

This would also allow to simplify the attribute checking by using `for groups in ("{}_predictive".format(group), "observed_data"):...`

nitishp25: Yeah and it looks much cleaner. Thanks!

OriolAbril:
LGTM

nitishp25:
Was there a specific reason for moving the ```group``` argument from the end to somewhere in between the others?

OriolAbril:
In the first commits group was only in the docstring, it was not an argument of the function

nitishp25:
I just checked the first commit again and it's the last argument there, though it doen't matter now :)

OriolAbril:
I must have missed it then, sorry.

OriolAbril:
Thanks for the PR!

nitishp25:
No problem :)

PR #1009, opened 2020-01-15
Solve issue #1005. Add group argument to plot pair.

### LINKED ISSUE #1005: Add group argument to plot pair
## Tell us about it
Add a group argument (like the one in plot density) to plot pair.

Handling on divergences in this case may require extra modifications to plotting code.

percygautam:
Added group argument to `pairplot.py`. Fixes #1005 .


arviz/plots/pairplot.py
```diff
@@ -34,6 +35,10 @@ def plot_pair(
     data : obj
         Any object that can be converted to an az.InferenceData object
         Refer to documentation of az.convert_to_dataset for details
+    group: Optional[str]
+        Specifies which InferenceData group should be plotted.  Defaults to 'posterior'.
+        Alternative values include 'prior' and any other strings used as dataset keys in the
+        InferenceData.
```
OriolAbril: I think the first line is enough, this last part may even confuse new users not fully understanding the relation between inference data and xarray datasets

arviz/plots/pairplot.py
```diff
@@ -158,13 +163,19 @@ def plot_pair(
                 divergent_data, var_names=("diverging",), combined=True
             )
             diverging_mask = np.squeeze(diverging_mask)
+        elif hasattr(data, "sample_stats_prior") and hasattr(data.sample_stats_prior, "diverging"):
+            divergent_data = convert_to_dataset(data, group="sample_stats_prior")
+            _, diverging_mask = xarray_to_ndarray(
+                divergent_data, var_names=("diverging",), combined=True
+            )
+            diverging_mask = np.squeeze(diverging_mask)
```
OriolAbril: The elif will not achieve the desired behaviour. The cases to take into account are:

* `group="posterior"` --> divergent_data is taken from `sample_stats`
* `group="prior"` --> divergent_data is taken from `sample_stats_prior`
* group in anything else --> no divergent data is plotted

This can use `sample_stats` when group is prior or sample_stats_prior when group is posterior (if there is no sample stats group)

percygautam: @OriolAbril Yes, I realized my mistake and rectified in the new commit just now. Kindly review that solution.

arviz/plots/pairplot.py
percygautam: Okay, I'll do that in the next commit.

percygautam:
@OriolAbril Do I need to change the `_posterior` param in backend `pairplot.py` file also?

arviz/plots/pairplot.py
```diff
@@ -34,6 +35,8 @@ def plot_pair(
     data : obj
         Any object that can be converted to an az.InferenceData object
         Refer to documentation of az.convert_to_dataset for details
+    group : {"prior", "posterior"}, optional
```
OriolAbril: In this case I would not restrict group to be either prior or posterior, in some cases it may be useful to plot other groups such as the posterior predictive.

arviz/plots/pairplot.py
```diff
@@ -142,31 +145,50 @@ def plot_pair(
...
+                diverging_mask = np.squeeze(diverging_mask)
+            else:
+                divergences = False
+                warnings.warn(
+                    "Divergences data not found, plotting without divergences. "
+                    "Make sure the sample method provides divergences data and "
+                    "that it is present in the `diverging` field of `sample_stats` "
+                    "or `sample_stats_prior` or set divergences=False",
+                    SyntaxWarning,
+                )
```
OriolAbril: I would try to avoid duplicating all this code. How does somethig like this look like? (pseudocode alert)

```
if group=posterior:
    divergent_group = sample_stats
elif group=prior:
    divergent_group = sample_stats_prior
else:
    divergences = False
# Get diverging draws and combine chains
    if divergences:
        if hasattr(data, divergent_group) and hasattr(getattr(data, divergent_group), "diverging"):
...  # divergent_group can also be used to format the warning so it is more informative
```

OriolAbril:
> @OriolAbril Do I need to change the `_posterior` param in backend `pairplot.py` file also?

Yes please, this will make the code much more clear

arviz/plots/pairplot.py
percygautam: Okay, will do.

arviz/plots/pairplot.py
percygautam: Okay, will do.

percygautam: @OriolAbril I think the code's fine but we should also add some warning if neither `posterior` or `prior` is passed to `group` param when we are setting `divergences=False`. Warning may be something like `Group assigned not in the divergent group --> plotting without divergence`. What do you think about this?

percygautam:
@OriolAbril I have done the changes but not yet added the warning I was discussing earlier. Also changed the name of `_posterior` to infdata_group` in backend files. Is naming convention right?

percygautam:
I will remove these linting errors in the next commit.

percygautam:
@OriolAbril I have done the linting changes. What are your thoughts about [this](https://github.com/arviz-devs/arviz/pull/1009#discussion_r367387556)?

OriolAbril:
> What are your thoughts about this]?
>
> > I think the code's fine but we should also add some warning if neither posterior or prior is passed to group param when we are setting divergences=False. Warning may be something like Group assigned not in the divergent group --> plotting without divergence.

I am not really sure, maybe @ahartikainen or @aloctavodia have some ideas on how to handle this. I would probably add a line in the description of `divergences` argument explaining that it is only taken into account if group is prior or posterior, in which case no warning is needed.

ahartikainen:
I think it could be added to docstring.

Warnings are great, but people really hate them.

I think we could have divergences also for `posterior_predictive`, `prior_predictive` and `log_likelihood`, `sample_stats` and `sample_stats_prior`, but maybe this can be done in another PR so we have also an option to mix different groups (e.g. plotting `treedepth` against variables etc)

percygautam:
@ahartikainen @OriolAbril Thankyou, for clarification. I'll add the docstring.

percygautam:
@OriolAbril The linting tests seem to fail due to some docker issue. Can you help?

PR #990, opened 2020-01-10
Remove numpy restriction

ahartikainen:
Jax was updated 2020-01-08.

PR #1013, opened 2020-01-19
[WIP] Added API link in examples

### LINKED ISSUE #1000: Show API link en examples
## Tell us about it

Currently each example in the gallery contains: an image, the code used to generate it, and a link to download the code. It would be great to add a link to the API section of the docs corresponding to the function used.

## Thoughts on implementation

This should happen automatically. A new feature should be added to `gallery_generator.py` using a reular expression to get the function name and then writing the corresponding link.

percygautam:
Fixes #1000 . Added API link in the examples.

### DIFF
--- a/doc/sphinxext/gallery_generator.py
+++ b/doc/sphinxext/gallery_generator.py
@@ -245,7 +245,7 @@ class ExampleGenerator:
             name = re.findall(regex, file.read())
         apitext = name[0] if len(name) > 0 else ""
         return (
-            "`" + apitext + " <../../generated/arviz." + apitext + ">`_"
+            "`{apitext} <../../generated/arviz.{apitext}>`_".format(apitext=apitext)
             if apitext
             else "No API Documentation available"
         )
--- a/doc/sphinxext/gallery_generator.py
+++ b/doc/sphinxext/gallery_generator.py
@@ -240,9 +240,13 @@ class ExampleGenerator:
 
     @property
     def apiname(self):
-        name = self.modulename.split("_")
-        name = name[1::]
-        return "_".join(name)
+        name=""
+        with open(op.join(self.target_dir, self.pyfilename), "r") as file:
+            regex = r"az\.(plot\_[a-z_]+)\("
+            matches = re.finditer(regex, file.read(), re.MULTILINE)
+            for matchNum, match in enumerate(matches, start=1):
+                name = match.group(1)
+        return name
 
     @property
     def sphinxtag(self):
--- a/doc/sphinxext/gallery_generator.py
+++ b/doc/sphinxext/gallery_generator.py
@@ -243,7 +243,7 @@ class ExampleGenerator:
         with open(op.join(self.target_dir, self.pyfilename), "r") as file:
             regex = r"az\.(plot\_[a-z_]+)\("
             name = re.findall(regex, file.read())
-        apitext = name[0] if len(name) > 0 else ""
+        apitext = name[0] if name else ""
         return (
             "`{apitext} <../../generated/arviz.{apitext}>`_".format(apitext=apitext)
             if apitext
--- a/doc/sphinxext/gallery_generator.py
+++ b/doc/sphinxext/gallery_generator.py
@@ -240,13 +240,10 @@ class ExampleGenerator:
 
     @property
     def apiname(self):
-        name=""
         with open(op.join(self.target_dir, self.pyfilename), "r") as file:
             regex = r"az\.(plot\_[a-z_]+)\("
-            matches = re.finditer(regex, file.read(), re.MULTILINE)
-            for matchNum, match in enumerate(matches, start=1):
-                name = match.group(1)
-        return name
+            name = re.findall(regex, file.read())
+        return name[0] if len(name) > 0 else None
 
     @property
     def sphinxtag(self):
--- a/doc/sphinxext/gallery_generator.py
+++ b/doc/sphinxext/gallery_generator.py
@@ -328,7 +328,6 @@ class ExampleGenerator:
             ".. raw:: html\n\n"
             "    <div class='figure align-center'>\n"
             "    <a href=./{0}/{1}>\n"
-            
             "    <img src=../_static/{2}>\n"
             "    <span class='figure-label'>\n"
             "    <p>{3}</p>\n"
--- a/doc/sphinxext/gallery_generator.py
+++ b/doc/sphinxext/gallery_generator.py
@@ -40,7 +40,7 @@ MPL_RST_TEMPLATE = """
 .. image:: {img_file}
 
 **Python source code:** :download:`[download source: {fname}]<{fname}>`
-**API documentation:** `{api_name} <../../generated/arviz.{api_name}>`_ 
+**API documentation:** `{api_name} <{api_link}>`_ 
 
 .. literalinclude:: {fname}
     :lines: {end_line}-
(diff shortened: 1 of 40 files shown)


percygautam:
@OriolAbril The link is showing in generated docs when I use `make html`. But I don't how to check whether link will work in github or not?

doc/sphinxext/gallery_generator.py
```diff
@@ -236,6 +238,12 @@ def thumbfilename(self):
         pngfile = self.modulename + "_thumb.png"
         return pngfile
 
+    @property
+    def apiname(self):
+        name = self.modulename.split("_")
+        name = name[1::]
+        return "_".join(name)
```
OriolAbril: I think this will fail in many cases such as `{backend}_plot_ess_evolution` (or quantile or local), `{backend}_plot_loo_pit_ecdf`, `{backend}_plot_pair_hex` (or kde) and so on. There are several plots with more than one image in the gallery were some extra description is needed.

percygautam: @OriolAbril I'll try to come up with a solution.

doc/sphinxext/gallery_generator.py
```diff
@@ -240,9 +240,13 @@ def thumbfilename(self):
 
     @property
     def apiname(self):
-        name = self.modulename.split("_")
-        name = name[1::]
-        return "_".join(name)
+        name=""
+        with open(op.join(self.target_dir, self.pyfilename), "r") as file:
+            regex = r"az\.(plot\_[a-z_]+)\("
```
ahartikainen: This sounds ok. Do we have code that calls 

    from arviz import plot_xyz

doc/sphinxext/gallery_generator.py
```diff
@@ -240,9 +240,13 @@ def thumbfilename(self):
 
     @property
     def apiname(self):
-        name = self.modulename.split("_")
-        name = name[1::]
-        return "_".join(name)
+        name=""
```
ahartikainen: Maybe just return None if nothing is found?

doc/sphinxext/gallery_generator.py
```diff
@@ -240,9 +240,13 @@ def thumbfilename(self):
...
+        with open(op.join(self.target_dir, self.pyfilename), "r") as file:
+            regex = r"az\.(plot\_[a-z_]+)\("
+            matches = re.finditer(regex, file.read(), re.MULTILINE)
+            for matchNum, match in enumerate(matches, start=1):
+                name = match.group(1)
```
ahartikainen: will this return the last found `plot_xyz`? Should it return all the matches?

Do you think `re.findall` would work here? No need for looping?

OriolAbril: I think the first is enough, there is only a couple of cases where more than a function is called, and in these cases the function called is always the same.

doc/sphinxext/gallery_generator.py
percygautam: @ahartikainen Currently in all the example files, we are doing `import arviz as az` and then using the plot function `az.plot_xyz`.

doc/sphinxext/gallery_generator.py
percygautam: Okay, makes sense.

doc/sphinxext/gallery_generator.py
percygautam: @ahartikainen @OriolAbril I worked some examples and let's say we have a hypothetical code as:
```
az.plot_compare(model_compare, figsize=(12, 4))
az.plot_forest(model_compare, figsize=(12, 4))
az.plot_density(model_compare, figsize=(12, 4))
```
If we run our code, it will return the last found match as 'plot_density'.  As in our example files, we are only calling the function one time, so I don't think it'll be a problem.
As a possible solution for, when (in future) we have multiple plots in a single example, we can do something like:
```
    for matchNum, match in enumerate(matches, start=1):
        api_name.append(match.group(1))
    return api_name #returns the list
```
And then maybe check the length when rendering it in the template.
As suggested by @ahartikainen , `re.findall()` will work best for such case. For example, above code can be simplified as
```
with open("mpl_plot_compare.py", "r") as f:
    regex = r"az\.(plot\_[a-z_]+)\("
    api_name = re.findall(regex, f.read())
return api_name  #api_name will contain list
```
How should we do it?

OriolAbril: I think using only the first function is good enough for that. Returning the list with all occurrences seems like an overkill which complicates the code without much gain

percygautam:
@OriolAbril It would return `API Documentation: None`, with a link to `az.None` function which will give file not found an error. I tried to give some condition to rst template so that it does not show the API documentation line, but supposedly rst doesn't support inline code snippets

doc/sphinxext/gallery_generator.py
```diff
@@ -40,6 +40,7 @@ def execfile(filename, globals=None, locals=None):
 .. image:: {img_file}
 
 **Python source code:** :download:`[download source: {fname}]<{fname}>`
+**API documentation:** `{api_name} <../../generated/arviz.{api_name}>`_
```
OriolAbril: Maybe this could be changed to `{api_text}` so that below, if api_name is None nothing is written?

percygautam: @OriolAbril , I have done the changes, now it'll show `None <None>` and not the directory link.

OriolAbril: I think it would be more clear replacing the whole line, so that when no match is fount (e.g. styles case) instead of `**API doc...` a white line is written. Or alternatively something like: 

> **API Documentation**: No API Documentation available for {api_name}

I see myself some months from now confusing the `None <None>` with a doc generation error.

percygautam: @OriolAbril  I understand your point. I have done the changes. Now, it'll show something like:

> **API Documentation**: No API Documentation available

We can't specify the name of documentation available because `api_name` would be an open string and other params like `modulename` will be in the form of `backend_xyz`. If we must do this then we have to use regex again or split the name and then pass to the new variable.

doc/sphinxext/gallery_generator.py
```diff
@@ -236,6 +238,18 @@ def thumbfilename(self):
         pngfile = self.modulename + "_thumb.png"
         return pngfile
 
+    @property
+    def apiname(self):
+        with open(op.join(self.target_dir, self.pyfilename), "r") as file:
+            regex = r"az\.(plot\_[a-z_]+)\("
+            name = re.findall(regex, file.read())
+        apitext = name[0] if len(name) > 0 else ""
```
OriolAbril: I think `if name` acts the same as `if len(name) > 0`

doc/sphinxext/gallery_generator.py
```diff
@@ -236,6 +238,18 @@ def thumbfilename(self):
...
+        return (
+            "`" + apitext + " <../../generated/arviz." + apitext + ">`_"
```
OriolAbril: .format can be used here instead of adding strings

doc/sphinxext/gallery_generator.py
percygautam: @OriolAbril It may cause problem if there are multiple plot functions in the example.

doc/sphinxext/gallery_generator.py
percygautam: Okay, I'll add it.

OriolAbril:
@percygautam #990 has already been merged, so rebasing/merging the latest changes in master should fix the failing tests. After this it can be merged

percygautam:
Okay, Thanks.

PR #1017, opened 2020-01-21
Integrated credible_interval with rcParam

### LINKED ISSUE #986: integrate credible_interval rcParam
## Tell us about it

`stats.credible_interval` has a default value in rcParams which is not yet used. 

Part of #792 

## Thoughts on implementation

All `credible_interval` arguments should default to the value in rcParams. This affects both stats functions like `hpd` or `summary` and plotting functions. However, as most of them call `hpd` eventually, it may be possible to use `None` as default and let `hpd` handle the rcParam defaulting.

Also, tangentially related to #659

percygautam:
Fixes #986 . The param credible_interval takes default value from rcParam.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### New features
 * Add out-of-sample predictions (`predictions` and  `predictions_constant_data` groups) to pymc3 translations. (#983)
 * Violinplot: rug-plot option (#997)
-* Integrated rcParams `plot.point_estimate` (#994) and `stats.ic_scale` (#993)
+* Integrated rcParams `plot.point_estimate` (#994), `stats.ic_scale` (#993) and `stats.credible_interval` (#986)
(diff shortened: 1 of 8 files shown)


aloctavodia:
LGTM!

OriolAbril:
LGTM, can you add it to the changelog? There already is a point about rcParams integration

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 ### New features
 * Add out-of-sample predictions (`predictions` and  `predictions_constant_data` groups) to pymc3 translations. (#983)
 * Violinplot: rug-plot option (#997)
-* Integrated rcParams `plot.point_estimate` (#994) and `stats.ic_scale` (#993)
+* Integrated rcParams `plot.point_estimate` (#994), `stats.ic_scale` (#993) and `stats.credible_interval` (#986)
```
OriolAbril: The number should be this PR instead of the issue

PR #1019, opened 2020-01-22
[WIP] Add pan limit in bokeh

### LINKED ISSUE #959: Add (pan/zoom) limits for Bokeh
Some plots might need suitable panning / zoom limits.

Also we can restrict movement on an axis. (Move only on x axis)

I'm not sure can we limit zooming.

nitishp25:
Adding pan limits in the following plots in bokeh:

- autocorrplot
- forestplot
- parallelplot
- traceplot (y axis)

I'm not sure if this is the right way though.
fixes #959

### DIFF
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -52,7 +52,7 @@ def plot_autocorr(
         Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
     backend_config: bool, optional
         Currently specifies the bounds to use for bokeh axes. Defaults to value set in rcParams.
-    backend_kwargs: bool, optional
+    backend_kwargs: dict, optional
         These are kwargs specific to the backend being used. For additional documentation
         check the plotting method of the backend.
     show : bool, optional
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -50,7 +50,7 @@ def plot_autocorr(
         Matplotlib axes or bokeh figures.
     backend: str, optional
         Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
-    backend_config: bool, optional
+    backend_config: dict, optional
         Currently specifies the bounds to use for bokeh axes. Defaults to value set in rcParams.
     backend_kwargs: dict, optional
         These are kwargs specific to the backend being used. For additional documentation
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -19,6 +19,7 @@ def plot_autocorr(
     textsize=None,
     ax=None,
     backend=None,
+    backend_config=None,
     backend_kwargs=None,
     show=None,
 ):
@@ -49,6 +50,8 @@ def plot_autocorr(
         Matplotlib axes or bokeh figures.
     backend: str, optional
         Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
+    backend_config: bool, optional
+        Currently specifies the bounds to use for bokeh axes. Defaults to value set in rcParams.
     backend_kwargs: bool, optional
         These are kwargs specific to the backend being used. For additional documentation
         check the plotting method of the backend.
@@ -129,6 +132,7 @@ def plot_autocorr(
         autocorr_plot_args.pop("xt_labelsize")
         autocorr_plot_args.pop("titlesize")
         autocorr_plot_args["line_width"] = autocorr_plot_args.pop("linewidth")
+        autocorr_plot_args.update(backend_config=backend_config)
 
     # TODO: Add backend kwargs
     plot = get_plotting_function("plot_autocorr", "autocorrplot", backend)
--- a/arviz/plots/backends/bokeh/autocorrplot.py
+++ b/arviz/plots/backends/bokeh/autocorrplot.py
@@ -25,6 +25,14 @@ def plot_autocorr(
     show,
 ):
     """Bokeh autocorrelation plot."""
+    if backend_config is None:
+        backend_config = {}
+
+    backend_config = {
+        **backend_kwarg_defaults(("bounds_y_range", "plots.bokeh.bounds_y_range"),),
+        **backend_config,
+    }
+
     if backend_kwargs is None:
         backend_kwargs = {}
 
@@ -48,7 +56,7 @@ def plot_autocorr(
     else:
         axes = np.atleast_2d(axes)
 
-    if backend_config is not None and "bounds_x_range" in backend_config:
+    if "bounds_x_range" in backend_config:
         data_range_x = DataRange1d(
             start=0, end=max_lag, bounds=backend_config["bounds_x_range"], min_interval=5
         )
@@ -56,14 +64,9 @@ def plot_autocorr(
         len_y = plotters[0][2].size
         data_range_x = DataRange1d(start=0, end=max_lag, bounds=(0, len_y), min_interval=5)
 
-    if backend_config is not None and "bounds_y_range" in backend_config:
-        data_range_y = DataRange1d(
-            start=-1, end=1, bounds=backend_config["bounds_y_range"], min_interval=0.1
-        )
-    else:
-        data_range_y = DataRange1d(
-            start=-1, end=1, bounds=(rcParams["plot.bokeh.bounds_y_range"]), min_interval=0.1
-        )
+    data_range_y = DataRange1d(
+        start=-1, end=1, bounds=backend_config["bounds_y_range"], min_interval=0.1
+    )
 
     for (var_name, selection, x), ax in zip(
         plotters, (item for item in axes.flatten() if item is not None)
--- a/arviz/plots/backends/bokeh/autocorrplot.py
+++ b/arviz/plots/backends/bokeh/autocorrplot.py
@@ -59,10 +59,16 @@ def plot_autocorr(
         title = Title()
         title.text = make_label(var_name, selection)
         ax.title = title
+        ax.x_range = DataRange1d(
+            start=0, end=max_lag, bounds=(0, len(np.arange(len(y)))), min_interval=5
+        )
+        ax.y_range = DataRange1d(start=-1, end=1, bounds=rcParams["plot.bokeh.bounds"])
 
-    if axes.size > 0:
-        axes[0, 0].x_range = DataRange1d(bounds=rcParams["plot.bokeh.bounds"], start=0, end=max_lag)
-        axes[0, 0].y_range = DataRange1d(bounds=rcParams["plot.bokeh.bounds"], start=-1, end=1)
+    for _, ax in zip(
+        plotters, (item for item in axes.flatten() if item is not None)
+    ):
+        ax.x_range = axes[0, 0].x_range
+        ax.y_range = axes[0, 0].y_range
 
     if backend_show(show):
         bkp.show(gridplot(axes.tolist(), toolbar_location="above"))
--- a/arviz/plots/backends/bokeh/autocorrplot.py
+++ b/arviz/plots/backends/bokeh/autocorrplot.py
@@ -2,6 +2,7 @@
 import bokeh.plotting as bkp
 import numpy as np
 from bokeh.layouts import gridplot
+from bokeh.models import DataRange1d
 from bokeh.models.annotations import Title
 
 from . import backend_kwarg_defaults, backend_show
@@ -59,10 +60,8 @@ def plot_autocorr(
         ax.title = title
 
     if axes.size > 0:
-        axes[0, 0].x_range._property_values["start"] = 0  # pylint: disable=protected-access
-        axes[0, 0].x_range._property_values["end"] = max_lag  # pylint: disable=protected-access
-        axes[0, 0].y_range._property_values["start"] = -1  # pylint: disable=protected-access
-        axes[0, 0].y_range._property_values["end"] = 1  # pylint: disable=protected-access
+        axes[0, 0].x_range = DataRange1d(bounds="auto")  # pylint: disable=protected-access
+        axes[0, 0].y_range = DataRange1d(bounds="auto")  # pylint: disable=protected-access
 
     if backend_show(show):
         bkp.show(gridplot(axes.tolist(), toolbar_location="above"))
--- a/arviz/plots/backends/bokeh/autocorrplot.py
+++ b/arviz/plots/backends/bokeh/autocorrplot.py
@@ -12,7 +12,17 @@ from ....rcparams import rcParams
 
 
 def plot_autocorr(
-    axes, plotters, max_lag, figsize, rows, cols, line_width, combined, backend_kwargs, show,
+    axes,
+    plotters,
+    max_lag,
+    figsize,
+    rows,
+    cols,
+    line_width,
+    combined,
+    backend_config,
+    backend_kwargs,
+    show,
 ):
     """Bokeh autocorrelation plot."""
     if backend_kwargs is None:
@@ -38,11 +48,22 @@ def plot_autocorr(
     else:
         axes = np.atleast_2d(axes)
 
-    len_y = plotters[0][2].size
-    data_range_x = DataRange1d(start=0, end=max_lag, bounds=(0, len_y), min_interval=5)
-    data_range_y = DataRange1d(
-        start=-1, end=1, bounds=(rcParams["plot.bokeh.bounds_y_range"]), min_interval=0.1
-    )
+    if "bounds_x_range" in backend_config:
+        data_range_x = DataRange1d(
+            start=0, end=max_lag, bounds=backend_config["bounds_x_range"], min_interval=5
+        )
+    else:
+        len_y = plotters[0][2].size
+        data_range_x = DataRange1d(start=0, end=max_lag, bounds=(0, len_y), min_interval=5)
+
+    if "bounds_y_range" in backend_config:
+        data_range_y = DataRange1d(
+            start=-1, end=1, bounds=backend_config["bounds_y_range"], min_interval=0.1
+        )
+    else:
+        data_range_y = DataRange1d(
+            start=-1, end=1, bounds=(rcParams["plot.bokeh.bounds_y_range"]), min_interval=0.1
+        )
 
     for (var_name, selection, x), ax in zip(
         plotters, (item for item in axes.flatten() if item is not None)
--- a/arviz/plots/backends/bokeh/autocorrplot.py
+++ b/arviz/plots/backends/bokeh/autocorrplot.py
@@ -61,8 +61,8 @@ def plot_autocorr(
         ax.title = title
 
     if axes.size > 0:
-        axes[0, 0].x_range = DataRange1d(bounds=rcParams["bokeh.bounds"], start=0, end=max_lag)
-        axes[0, 0].y_range = DataRange1d(bounds=rcParams["bokeh.bounds"], start=-1, end=1)
+        axes[0, 0].x_range = DataRange1d(bounds=rcParams["plot.bokeh.bounds"], start=0, end=max_lag)
+        axes[0, 0].y_range = DataRange1d(bounds=rcParams["plot.bokeh.bounds"], start=-1, end=1)
 
     if backend_show(show):
         bkp.show(gridplot(axes.tolist(), toolbar_location="above"))
--- a/arviz/plots/backends/bokeh/autocorrplot.py
+++ b/arviz/plots/backends/bokeh/autocorrplot.py
@@ -48,7 +48,7 @@ def plot_autocorr(
     else:
         axes = np.atleast_2d(axes)
 
-    if "bounds_x_range" in backend_config:
+    if backend_config is not None and "bounds_x_range" in backend_config:
         data_range_x = DataRange1d(
             start=0, end=max_lag, bounds=backend_config["bounds_x_range"], min_interval=5
         )
@@ -56,7 +56,7 @@ def plot_autocorr(
         len_y = plotters[0][2].size
         data_range_x = DataRange1d(start=0, end=max_lag, bounds=(0, len_y), min_interval=5)
 
-    if "bounds_y_range" in backend_config:
+    if backend_config is not None and "bounds_y_range" in backend_config:
         data_range_y = DataRange1d(
             start=-1, end=1, bounds=backend_config["bounds_y_range"], min_interval=0.1
         )
--- a/arviz/plots/backends/bokeh/autocorrplot.py
+++ b/arviz/plots/backends/bokeh/autocorrplot.py
@@ -28,8 +28,11 @@ def plot_autocorr(
     if backend_config is None:
         backend_config = {}
 
+    len_y = plotters[0][2].size
+    backend_config.setdefault("bounds_x_range", (0, len_y))
+
     backend_config = {
-        **backend_kwarg_defaults(("bounds_y_range", "plots.bokeh.bounds_y_range"),),
+        **backend_kwarg_defaults(("bounds_y_range", "plot.bokeh.bounds_y_range"),),
         **backend_config,
     }
 
@@ -56,14 +59,9 @@ def plot_autocorr(
     else:
         axes = np.atleast_2d(axes)
 
-    if "bounds_x_range" in backend_config:
-        data_range_x = DataRange1d(
-            start=0, end=max_lag, bounds=backend_config["bounds_x_range"], min_interval=5
-        )
-    else:
-        len_y = plotters[0][2].size
-        data_range_x = DataRange1d(start=0, end=max_lag, bounds=(0, len_y), min_interval=5)
-
+    data_range_x = DataRange1d(
+        start=0, end=max_lag, bounds=backend_config["bounds_x_range"], min_interval=5
+    )
     data_range_y = DataRange1d(
         start=-1, end=1, bounds=backend_config["bounds_y_range"], min_interval=0.1
     )
--- a/arviz/plots/backends/bokeh/autocorrplot.py
+++ b/arviz/plots/backends/bokeh/autocorrplot.py
@@ -38,6 +38,12 @@ def plot_autocorr(
     else:
         axes = np.atleast_2d(axes)
 
+    len_y = len(plotters[0][2])
+    data_range_x = DataRange1d(start=0, end=max_lag, bounds=(0, len_y), min_interval=5)
+    data_range_y = DataRange1d(
+        start=-1, end=1, bounds=(rcParams["plot.bokeh.bounds_y_range"]), min_interval=0.1
+    )
+
     for (var_name, selection, x), ax in zip(
         plotters, (item for item in axes.flatten() if item is not None)
     ):
@@ -59,16 +65,8 @@ def plot_autocorr(
         title = Title()
         title.text = make_label(var_name, selection)
         ax.title = title
-        ax.x_range = DataRange1d(
-            start=0, end=max_lag, bounds=(0, len(np.arange(len(y)))), min_interval=5
-        )
-        ax.y_range = DataRange1d(start=-1, end=1, bounds=rcParams["plot.bokeh.bounds"])
-
-    for _, ax in zip(
-        plotters, (item for item in axes.flatten() if item is not None)
-    ):
-        ax.x_range = axes[0, 0].x_range
-        ax.y_range = axes[0, 0].y_range
+        ax.x_range = data_range_x
+        ax.y_range = data_range_y
 
     if backend_show(show):
         bkp.show(gridplot(axes.tolist(), toolbar_location="above"))
--- a/arviz/plots/backends/bokeh/autocorrplot.py
+++ b/arviz/plots/backends/bokeh/autocorrplot.py
@@ -38,7 +38,7 @@ def plot_autocorr(
     else:
         axes = np.atleast_2d(axes)
 
-    len_y = len(plotters[0][2])
+    len_y = plotters[0][2].size
     data_range_x = DataRange1d(start=0, end=max_lag, bounds=(0, len_y), min_interval=5)
     data_range_y = DataRange1d(
         start=-1, end=1, bounds=(rcParams["plot.bokeh.bounds_y_range"]), min_interval=0.1
--- a/arviz/plots/backends/bokeh/autocorrplot.py
+++ b/arviz/plots/backends/bokeh/autocorrplot.py
@@ -8,7 +8,6 @@ from bokeh.models.annotations import Title
 from . import backend_kwarg_defaults, backend_show
 from ...plot_utils import _create_axes_grid, make_label
 from ....stats import autocorr
-from ....rcparams import rcParams
 
 
 def plot_autocorr(
--- a/arviz/plots/backends/bokeh/autocorrplot.py
+++ b/arviz/plots/backends/bokeh/autocorrplot.py
@@ -8,6 +8,7 @@ from bokeh.models.annotations import Title
 from . import backend_kwarg_defaults, backend_show
 from ...plot_utils import _create_axes_grid, make_label
 from ....stats import autocorr
+from ....rcparams import rcParams
 
 
 def plot_autocorr(
@@ -60,8 +61,8 @@ def plot_autocorr(
         ax.title = title
 
     if axes.size > 0:
-        axes[0, 0].x_range = DataRange1d(bounds="auto")  # pylint: disable=protected-access
-        axes[0, 0].y_range = DataRange1d(bounds="auto")  # pylint: disable=protected-access
+        axes[0, 0].x_range = DataRange1d(bounds=rcParams["bokeh.bounds"], start=0, end=max_lag)
+        axes[0, 0].y_range = DataRange1d(bounds=rcParams["bokeh.bounds"], start=-1, end=1)
 
     if backend_show(show):
         bkp.show(gridplot(axes.tolist(), toolbar_location="above"))
--- a/arviz/plots/backends/bokeh/forestplot.py
+++ b/arviz/plots/backends/bokeh/forestplot.py
@@ -139,12 +139,12 @@ def plot_forest(
             ax_.yaxis.visible = False
 
         ax_.outline_line_color = None
-        if "bounds_x_range" in backend_config:
+        if backend_config is not None and "bounds_x_range" in backend_config:
             ax_.x_range = DataRange1d(bounds=backend_config["bounds_x_range"], min_interval=1)
         else:
             ax_.x_range = DataRange1d(bounds=rcParams["plot.bokeh.bounds_x_range"], min_interval=1)
 
-        if "bounds_y_range" in backend_config:
+        if backend_config is not None and "bounds_y_range" in backend_config:
             ax_.y_range = DataRange1d(bounds=backend_config["bounds_y_range"], min_interval=2)
         else:
             ax_.y_range = DataRange1d(bounds=rcParams["plot.bokeh.bounds_y_range"], min_interval=2)
--- a/arviz/plots/backends/bokeh/forestplot.py
+++ b/arviz/plots/backends/bokeh/forestplot.py
@@ -138,8 +138,8 @@ def plot_forest(
             ax_.yaxis.visible = False
 
         ax_.outline_line_color = None
-        ax_.x_range = DataRange1d(bounds=rcParams["bokeh.bounds"])
-        ax_.y_range = DataRange1d(bounds=rcParams["bokeh.bounds"])
+        ax_.x_range = DataRange1d(bounds=rcParams["plot.bokeh.bounds"])
+        ax_.y_range = DataRange1d(bounds=rcParams["plot.bokeh.bounds"])
 
     labels, ticks = plot_handler.labels_and_ticks()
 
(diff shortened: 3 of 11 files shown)


nitishp25:
What do you suggest @OriolAbril @ahartikainen ?

arviz/plots/backends/bokeh/autocorrplot.py
```diff
@@ -59,10 +60,8 @@ def plot_autocorr(
         ax.title = title
 
     if axes.size > 0:
-        axes[0, 0].x_range._property_values["start"] = 0  # pylint: disable=protected-access
-        axes[0, 0].x_range._property_values["end"] = max_lag  # pylint: disable=protected-access
-        axes[0, 0].y_range._property_values["start"] = -1  # pylint: disable=protected-access
-        axes[0, 0].y_range._property_values["end"] = 1  # pylint: disable=protected-access
+        axes[0, 0].x_range = DataRange1d(bounds="auto")  # pylint: disable=protected-access
+        axes[0, 0].y_range = DataRange1d(bounds="auto")  # pylint: disable=protected-access
```
ahartikainen: add `start` and `end` values to `DataRange1d` function.

arviz/plots/backends/bokeh/traceplot.py
```diff
@@ -179,6 +179,7 @@ def plot_trace(
             _title = Title()
             _title.text = make_label(var_name, selection)
             axes[idx, col].title = _title
+            axes[idx, col].y_range = DataRange1d(bounds="auto")  # pylint: disable=protected-access
```
ahartikainen: I don't think you need # pylint code here. It was for the `_...` parameter.

arviz/plots/backends/bokeh/autocorrplot.py
```diff
@@ -59,10 +60,8 @@ def plot_autocorr(
         ax.title = title
 
     if axes.size > 0:
-        axes[0, 0].x_range._property_values["start"] = 0  # pylint: disable=protected-access
-        axes[0, 0].x_range._property_values["end"] = max_lag  # pylint: disable=protected-access
-        axes[0, 0].y_range._property_values["start"] = -1  # pylint: disable=protected-access
-        axes[0, 0].y_range._property_values["end"] = 1  # pylint: disable=protected-access
+        axes[0, 0].x_range = DataRange1d(bounds="auto")  # pylint: disable=protected-access
```
ahartikainen: Hi, instead of using `bounds="auto"` maybe use `rcParam.bokeh.bounds` and set it to `auto`. This way user can change this manually by changing that rcParam. E.g. `(0, None)`.

nitishp25: Yeah that makes sense. Should I also add a ```_validate_tuple``` function to check for valid tuple value?

ahartikainen: That sounds reasonable

arviz/rcparams.py
```diff
@@ -93,6 +93,18 @@ def _validate_boolean(value):
...
+def _validate_tuple(value):
+    """Validate value is a tuple of size 2."""
+    if value is None:
+        return value
+    elif value == "auto":
+        return value
+    elif isinstance(value, tuple) and len(value) == 2:
+        return value
+    else:
+        raise ValueError("Only tuple of size 2 is valid.")
```
OriolAbril: Hi, I understand we need to validate a tuple of floats of length 2 right? There are many extra things to take into account. 

To begin with, the validation function must work with python objects (used at initialization from `defaultParams` and while using `rc` argument in `rc_context`) but also with _strings_ (used at initializaiton from arvizrc files and while using the `filename` argument in `rc_context`). 

I also think that the validation functions should be modularized as much as possible.

My suggestion is to divide the function into several smaller pieces and to make it much more restrictive. I would create some function like [`_listify_validator`](https://github.com/matplotlib/matplotlib/blob/c58ca603189f45b7e927e72b1ca6bb640bb9d376/lib/matplotlib/rcsetup.py#L74) from matplotlib. This function could be combined with `validate_float` to validate tuples/lists of floats of any length, creating a `validate_float_iterable` function. Eventually, a `validate_float_pair` could be created, or maybe to be more general a `length` argument can be added to `validate_float_iterable`.

nitishp25: Yes this will make it much more sustainable. I'll make the changes, thanks!

nitishp25: > Hi, I understand we need to validate a tuple of floats of length 2 right? There are many extra things to take into account.
> 
> To begin with, the validation function must work with python objects (used at initialization from `defaultParams` and while using `rc` argument in `rc_context`) but also with _strings_ (used at initializaiton from arvizrc files and while using the `filename` argument in `rc_context`).

How should I deal with compatibility with objects?

OriolAbril: The current function only works with objects. One example of working with objects and strings is [here](https://github.com/arviz-devs/arviz/blob/eb45dfb54d03da5e64690ac2cf0cb63983030377/arviz/rcparams.py#L30), where both `None` and `"none"` are accepted. In general, this is achieved by converting to the desired `type` and then make any extra validation assuming/knowing the type of the value (a good example is [`validate_probability`](https://github.com/arviz-devs/arviz/blob/eb45dfb54d03da5e64690ac2cf0cb63983030377/arviz/rcparams.py#L81), first we make sure the value is compatible with being a float and make sure its type is float and then its range is checked. I hope this answers the question.

nitishp25: Yeah, thanks for the help! Actually I thought that this function wasn't compatible with objects. Now the confusion is cleared.

arviz/rcparams.py
```diff
@@ -93,23 +102,37 @@ def _validate_boolean(value):
...
+            else:
+                raise ValueError("Iterable must be of length: {}".format(length))
+        except Exception:
+            raise
+    elif np.iterable(value) and not isinstance(value, (set, frozenset)):
+        val = [_validate_float_or_none(v) for v in value]
+        if len(val) == length:
+            return tuple(val)
+        else:
+            raise ValueError("Iterable must be of length: {}".format(length))
```
OriolAbril: As rcParams are stll quite new, I would try to divide the code in as many potentially useful building blocks as possible. For example, a function similar to matplotlib's `listify_validator` will probably be used in the future, so it would be interesting to have it in an independent function that can then be called from `validate_float_iterable` and any other function to come.

arviz/rcparams.py
```diff
@@ -93,23 +102,37 @@ def _validate_boolean(value):
...
     else:
-        raise ValueError("Only tuple of size 2 is valid.")
+        raise ValueError("Only float iterable values are valid")
```
OriolAbril: It would be great to include `length` in the error message. a `.format` should be enough

arviz/rcparams.py
```diff
@@ -93,23 +102,37 @@ def _validate_boolean(value):
     return value is True or value == "true"
 
 
-def _validate_tuple(value):
-    """Validate value is a tuple of size 2."""
-    if value is None:
-        return value
+def _validate_float_iterable(value, length):
```
OriolAbril: the `_validate` functions should only have one argument, otherwise, in cases like this one the pattern to follow is the `make_validator` example, where the return value is the validator function

arviz/rcparams.py
nitishp25: Got it, I'll make a more general ```_validate_iterable``` function to validate any iterable data types and then just call it in ```_validate_float_iterable``` in this case. Am I right?

arviz/rcparams.py
nitishp25: Alright

arviz/rcparams.py
nitishp25: Okay, I'll make the changes.

arviz/rcparams.py
```diff
@@ -93,11 +102,64 @@ def _validate_boolean(value):
...
+                val = [validate_scalar(v.strip("([ ])")) for v in value.split(",") if v.strip()]
+                if length is not None:
+                    if len(val) == length:
+                        return val
+                    else:
+                        raise ValueError("Iterable must be of length: {}".format(length))
+                else:
+                    return val
+            except Exception:
+                raise
```
nitishp25: @OriolAbril, should I add a ```allow_stringlist``` argument to handle this exception like ```_listify_validator``` or do we not need it right now (we may need it in future) and find another way of handling this exception?

arviz/plots/backends/bokeh/autocorrplot.py
nitishp25: @ahartikainen, setting ```start``` and ```end``` values of ```DataRange1d``` for ```x_range``` to 0 and ```max_lag``` and for ```y_range``` to -1 and 1 and then setting bounds to ```auto``` restricts the plot to these values only and the entire plot cannot be panned.

So, in order to keep the given ranges to viewport while limiting pan values to the entire range I need to set the value of ```bounds``` to the entire range. How do I get this value of entire ```x_range``` and ```y_range```?

ahartikainen: for x_range values are from 0 --> `np.arange(len(y))`
fir y_range values are from -1 --> 1

nitishp25: Thanks for the help! Also, adding ```min_interval``` to ```DataRange1d``` limits the zoom in value, I've set it to ```5``` (float) for ```plot_autocorr```. Can you suggest the values for other plots?

arviz/plots/backends/bokeh/autocorrplot.py
```diff
@@ -59,10 +59,16 @@ def plot_autocorr(
         title = Title()
         title.text = make_label(var_name, selection)
         ax.title = title
+        ax.x_range = DataRange1d(
+            start=0, end=max_lag, bounds=(0, len(np.arange(len(y)))), min_interval=5
```
ahartikainen: len(np.arange(len(y))) --> len(y)

ahartikainen: or actually `len(y) - 1`

nitishp25: Yeah makes sense. I'll make the changes.

Also, what do you suggest the zoom in limit should be for these plots? I've set ```min_interval=5``` (float) for x-axis of ```autocorrplot```. This restricts zooming to a minimum interval of length 5.

ahartikainen: That sounds good

nitishp25: Do you think it's suitable for other plots as well?

For examples, the y axes of ```traceplot``` and ```parallelplot``` are very detailed so I think they should be left unrestricted?

And x-axis of ```forestplot``` could use ```min_interval=1```?

arviz/rcparams.py
OriolAbril: No need for `allow_stringlist`. I don't think the case it covers (interpreting `"rgbky"` as a list `["r", "g", "b", "k", "y"]`) will be used in ArviZ

arviz/rcparams.py
```diff
@@ -93,11 +102,64 @@ def _validate_boolean(value):
     return value is True or value == "true"
 
 
+def iterable_validator(validate_scalar, length=None):
```
OriolAbril: Mainly for convention, I would name the function argument `scalar_validator` and the function itself `make_iterable_validator`.

arviz/rcparams.py
```diff
@@ -93,11 +102,64 @@ def _validate_boolean(value):
...
+    """Validate value is an iterable datatype."""
+    # no blank lines allowed after function docstring by pydocstyle,
+    # but black requires white line before function
```
OriolAbril: I would use the comments to give attribution to matplotlib's listify validator

arviz/plots/backends/bokeh/autocorrplot.py
```diff
@@ -57,12 +59,16 @@ def plot_autocorr(
...
-    if axes.size > 0:
-        axes[0, 0].x_range._property_values["start"] = 0  # pylint: disable=protected-access
-        axes[0, 0].x_range._property_values["end"] = max_lag  # pylint: disable=protected-access
-        axes[0, 0].y_range._property_values["start"] = -1  # pylint: disable=protected-access
-        axes[0, 0].y_range._property_values["end"] = 1  # pylint: disable=protected-access
+    for _, ax in zip(
+        plotters, (item for item in axes.flatten() if item is not None)
+    ):
+        ax.x_range = axes[0, 0].x_range
+        ax.y_range = axes[0, 0].y_range
```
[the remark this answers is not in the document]
nitishp25: I think in [this](https://arviz-devs.github.io/arviz/examples/bokeh/bokeh_plot_autocorr.html) case, ```len(y)``` will return 500 but it has values from 0 to 499 only right?



> I think this is not necessary now, the range has already been set above.

I did this to ensure link panning. Without this, the plots were not panning all together like they used to. Should I remove this?

arviz/plots/backends/bokeh/forestplot.py
```diff
@@ -138,6 +138,8 @@ def plot_forest(
             ax_.yaxis.visible = False
 
         ax_.outline_line_color = None
+        ax_.x_range = DataRange1d(bounds=rcParams["plot.bokeh.bounds"])
+        ax_.y_range = DataRange1d(bounds=rcParams["plot.bokeh.bounds"])
```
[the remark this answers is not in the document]
nitishp25: Actually, ```bounds="auto"``` by default sets the range automatically to the data range of the plot. So, even though they have different ranges, it deals with them separately.

arviz/rcparams.py
nitishp25: Alright, I'll make the changes.

arviz/rcparams.py
nitishp25: Makes sense. Thanks!

arviz/plots/backends/bokeh/autocorrplot.py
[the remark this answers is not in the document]
OriolAbril: Yes, values will go between 0 and 499, what I meant is that `len(np.arange(len(y))))` should be equal to `len(y)`

arviz/plots/backends/bokeh/forestplot.py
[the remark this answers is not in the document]
OriolAbril: In the case this is `auto` yes, but what if the user sets a tuple?

arviz/plots/backends/bokeh/autocorrplot.py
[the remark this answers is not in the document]
OriolAbril: Regarding panning, I am not sure I understand the issue, I am not extremely familiar with bokeh yet. One difference between adding the extra loop or not is that with extra loop, all plots use the same `DataRange1d` object, whereas with only the first loop, each uses a different `DataRange1d` object with the same parameters. Could this make a difference?

arviz/plots/backends/bokeh/forestplot.py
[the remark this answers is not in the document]
nitishp25: Right. Should I make seperate bounds like ```plot.bokeh.bounds_x_range``` and ```plot.bokeh.bounds_y_range```?

arviz/rcparams.py
```diff
@@ -93,11 +102,64 @@ def _validate_boolean(value):
...
+
+    def validate_float_iterable(value):
+        if value is None or isinstance(value, str) and value.lower() == "none":
+            return None
+        elif value == "auto":
+            return value
+        try:
+            validate_floatlist = iterable_validator(_validate_float_or_none, length)
+            return tuple(validate_floatlist(value))
```
OriolAbril: Instead of converting to tuple here, it is probably better to make `validate_iterable` always return tuples. Otherwise, the following pattern is possible:

```
In [2]: plt.rcParams["lines.dashdot_pattern"]
Out[2]: [6.4, 1.6, 1.0, 1.6]

In [3]: plt.rcParams["lines.dashdot_pattern"].append("unvalidated_value")

In [4]: plt.rcParams["lines.dashdot_pattern"]
Out[4]: [6.4, 1.6, 1.0, 1.6, 'unvalidated_value']
```

And I do not think it should be. It allows to set rcParams to unvalidated values which eventually makes `plt.plot(x,y , "-.")` fail with a weird error message

arviz/plots/backends/bokeh/autocorrplot.py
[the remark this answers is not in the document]
nitishp25: > Regarding panning, I am not sure I understand the issue, I am not extremely familiar with bokeh yet. One difference between adding the extra loop or not is that with extra loop, all plots use the same `DataRange1d` object, whereas with only the first loop, each uses a different `DataRange1d` object with the same parameters. Could this make a difference?

I tested it with [this](https://arviz-devs.github.io/arviz/examples/bokeh/bokeh_plot_autocorr.html) example code (which already supports linked panning i.e. panning any one plot will pan all others in the same way) and didn't find any issues. I looked up the solution for linked panning and [bokeh docs](https://docs.bokeh.org/en/latest/docs/user_guide/interaction/linking.html) suggested to provide the range of the first plot as the range to other plots. That's why I though this might be the solution.

What do you think? @OriolAbril @ahartikainen?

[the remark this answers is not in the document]
OriolAbril: That sounds reasonable. If possible, I would like to avoid two loops over the same set of axes. 

Does defining a `data_range_x` and `data_range_y` **before** the loop and using always this same object (as opposed to creating a new one every iteration as it is currently done) solve the issue?

arviz/rcparams.py
nitishp25: Sorry, I don't understand. Can you please elaborate how unvalidated values are allowed?

OriolAbril: The main difference between a python list and a tuple is that lists are mutable (it can be modified after creation, is elements can be changed, or it can have new elements appended) whereas tuples are not. Therefore, in the above example, the append is modifying the list instead of modifying the `lines.dashdot_pattern` rcParam which is what runs the validation function when setting values. 

Note that in the example we have ended up with a string in the list, whereas trying to set this same list via rcParams (with the validation function) will give the following output:

```
In [7]: plt.rcParams["lines.dashdot_pattern"] = [6.4, 1.6, 1.0, 1.6, 'unvalidated_value']
ValueError                                Traceback (most recent call last)
~/venvs/random/lib/python3.6/site-packages/matplotlib/rcsetup.py in __call__(self, s)
    318                     else val
--> 319                     for val in s]
    320         except ValueError:

~/venvs/random/lib/python3.6/site-packages/matplotlib/rcsetup.py in <listcomp>(.0)
    318                     else val
--> 319                     for val in s]
    320         except ValueError:

ValueError: could not convert string to float: 'unvalidated_value'

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
~/venvs/random/lib/python3.6/site-packages/matplotlib/__init__.py in __setitem__(self, key, val)
    801             try:
--> 802                 cval = self.validate[key](val)
    803             except ValueError as ve:

~/venvs/random/lib/python3.6/site-packages/matplotlib/rcsetup.py in __call__(self, s)
    320         except ValueError:
--> 321             raise ValueError('Could not convert all entries to floats')
    322 

ValueError: Could not convert all entries to floats

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-7-cb71fbf16588> in <module>
----> 1 plt.rcParams["lines.dashdot_pattern"] = [6.4, 1.6, 1.0, 1.6, 'unvalidated_value']

~/venvs/random/lib/python3.6/site-packages/matplotlib/__init__.py in __setitem__(self, key, val)
    802                 cval = self.validate[key](val)
    803             except ValueError as ve:
--> 804                 raise ValueError("Key %s: %s" % (key, str(ve)))
    805             dict.__setitem__(self, key, cval)
    806         except KeyError:

ValueError: Key lines.dashdot_pattern: Could not convert all entries to floats
```

OriolAbril: If tuples were used, no values can be appended, so the only way of updating it is to create a new tuple with the new element and make `rcParams["key"] = new_tuple` which will then always be validated

arviz/plots/backends/bokeh/autocorrplot.py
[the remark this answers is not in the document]
nitishp25: I thought of that too but ```x_range``` needs ```len(y)``` and ```y``` is present in the loop.

Maybe inside the first loop we can check if the ```axes``` is the first one i.e. the first plot and assign the ranges for that plot only and for every other plot we assign the range equal to the range of the first plot.

For example:
```
for loop:
    -----
    -----
    if ax is first:
         axes[0, 0].x_range = DataRange1d(
            start=0, end=max_lag, bounds=(0, len(np.arange(len(y)))), min_interval=5
        )
         axes[0, 0].y_range = DataRange1d(start=-1, end=1, bounds=rcParams["plot.bokeh.bounds_y_range"])
    else:
         ax.x_range = axes[0, 0].x_range
         ax.y_range = axes[0, 0].y_range
```

What do you think?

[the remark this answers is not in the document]
OriolAbril: I think `y` should have length `n_draws` or `n_draws * n_chains` depending on the value of `combine`.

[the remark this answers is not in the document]
nitishp25: Sorry now I got your point! Yes, this will solve the panning issue. How do I access the ```n_draws``` and ```n_chains``` values?

arviz/rcparams.py
nitishp25: Got it, I'll make the changes!

arviz/plots/backends/bokeh/autocorrplot.py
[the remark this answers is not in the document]
nitishp25: Do I need to pass the data from ```plots/autocorrplot.py``` to ```plots/backends/bokeh/autocorrplot.py``` i.e. from frontend to backend to access the ```data["draw"]``` and ```data["chain"]```?

arviz/plots/backends/bokeh/forestplot.py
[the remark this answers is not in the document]
ahartikainen: Yes, that sounds logical

arviz/plots/backends/bokeh/autocorrplot.py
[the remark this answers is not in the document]
OriolAbril: something like:

    chain, draw, *_ = plotters[0][2].shape

should work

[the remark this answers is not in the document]
nitishp25: This gives the following error:

```ValueError: not enough values to unpack (expected at least 2, got 1)```

Should we check in single dimension only like ```plotters[0].shape```?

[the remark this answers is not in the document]
Comment (ahartikainen) [reply arviz/plots/backends/bokeh/autocorrplot.py]: Maybe we just go with two loop solution?

No need to create extra code and num of axes is usually (always) low.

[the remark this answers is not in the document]
Comment (OriolAbril) [reply arviz/plots/backends/bokeh/autocorrplot.py]: I gave a quick look to the code creating plotters, and the combination of chains is handled there. Therefore `len(plotters[0][2])` is equal to `len(y)`, and it will always be a 1d array. The flatten on `x` there there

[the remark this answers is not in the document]
Comment (nitishp25) [reply arviz/plots/backends/bokeh/autocorrplot.py]: So should I remove the ```combined``` argument from both the backends too?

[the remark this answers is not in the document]
Comment (OriolAbril) [reply arviz/plots/backends/bokeh/autocorrplot.py]: My bad, the combined boolean is still needed, I mistook the helper function used by another. If we call `x = plotters[0][2]` there are 2 cases: 
* combined=True -> x.shape = (chain, draw) -> flatten is needed
* combined=False -> x.shape = (draw) -> no flatten
However, as the data used is always the flattened version, what is true is `len(y) = x.size`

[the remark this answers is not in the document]
Comment (nitishp25) [reply arviz/plots/backends/bokeh/autocorrplot.py]: I see. That's why using length as ```plotters[0][2]``` worked for me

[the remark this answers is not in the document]
Comment (OriolAbril) [reply arviz/plots/backends/bokeh/autocorrplot.py]: It should be `.size` otherwise it won't give the correct value with combined=True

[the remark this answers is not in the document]
Comment (nitishp25) [reply arviz/plots/backends/bokeh/autocorrplot.py]: And do I need to flatten it as well for ```combined=True``` or is it already flattened because I'm calling ```len_y = plotters[0][2].size``` **before** the loop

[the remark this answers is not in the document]
Comment (OriolAbril) [reply arviz/plots/backends/bokeh/autocorrplot.py]: [`.size`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.size.html) only returns an integer, `x_prime` still has to be flattened if combined=True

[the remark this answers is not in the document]
Comment (nitishp25) [reply arviz/plots/backends/bokeh/autocorrplot.py]: Got it. I don't know why it slipped from my mind :)

File: arviz/rcparams.py [hunk 15]
```diff
@@ -93,11 +102,60 @@ def _validate_boolean(value):
     return value is True or value == "true"
 
 
+def make_iterable_validator(scalar_validator, length=None):
```
OriolAbril: I was thinking of ways to make this a little shorter and not have repeated code, then I thought about adding the `allow_none` arg and missed myself in the process. How does this look? I think it will do the same:

```
def make_iterable_validator(scalar_validator, length=None, allow_none=False, allow_auto=False):
    """Validate value is an iterable datatype."""
    # based on matplotlib's _listify_validator function

    def validate_iterable(value):
        if allow_none and (value is None or isinstance(value, str) and value.lower() == "none"):
            return None
        if isinstance(value, str):
            if allow_auto and value.lower() == "auto":
                return "auto"
            val = tuple(v.strip("([ ])") for v in value.split(",") if v.strip())  
        if np.iterable(value) and not isinstance(value, (set, frozenset)):  # note the elif changed to if
            val = tuple(scalar_validator(v) for v in value)
            if length is not None and len(val) != length:
                raise ValueError("Iterable must be of length: {}".format(length))
            return val
        raise ValueError("Only iterable values are valid")

    return validate_iterable
```

arviz/rcparams.py
```diff
@@ -93,11 +102,60 @@ def _validate_boolean(value):
...
+            return None
+        elif value == "auto":
+            return value
+        try:
+            validate_floatlist = make_iterable_validator(_validate_float_or_none, length)
+            return validate_floatlist(value)
+        except:
+            raise ValueError("Only float iterable values are valid.")
+
+    return validate_float_iterable
```
OriolAbril: Using the code above we could define here:

    _validate_bokeh_bounds = make_iterable_validator(
        _validate_float_or_none, length=2, allow_none=True, allow_auto=True
    )

arviz/rcparams.py
```diff
@@ -93,11 +102,60 @@ def _validate_boolean(value):
...
+
+
 defaultParams = {  # pylint: disable=invalid-name
     "data.http_protocol": ("https", _make_validate_choice({"https", "http"})),
     "data.load": ("lazy", _make_validate_choice({"lazy", "eager"})),
     "data.index_origin": (0, _make_validate_choice({0, 1}, typeof=int)),
     "plot.backend": ("matplotlib", _make_validate_choice({"matplotlib", "bokeh"})),
+    "plot.bokeh.bounds_x_range": ("auto", _make_validate_float_iterable(length=2)),
+    "plot.bokeh.bounds_y_range": ("auto", _make_validate_float_iterable(length=2)),
```
OriolAbril: And here use the `_validate_bokeh_bounds`

nitishp25: Yes this is way more optimized! I'll definitely make the changes.

Also, how can I add a ```backend_config``` option to enable/disable these limits as suggested in #959 by @ahartikainen. Is it like an argument that the user can pass to ```plot_autocorr```?

OriolAbril: Yes, it would be a new argument to plot_autocorr so that the user can pass to it a dictionary with keys like `bounds_x_range`. The preference would then be to use the value in `backend_config` if present, otherwise use the value in rcParams

nitishp25: Isn't it better to include it in ```backend_kwargs``` so that other plots can also use it?

OriolAbril: `backend_kwargs` is currently used only at axes/figure creation (some PRs ago we decided on not including `show` there because show is not relevant on creation but after plotting). I think this is similar to `show` case? Or can this be defined at creation? (I'm still getting up to date with bokeh)

Also, do not worry too much about other plots for now. Adding this in `backend_kwargs` will mean having to modify their code and use this new key whereas using `backend_config` is adding a new argument and using it.

nitishp25: I don't have much knowledge either but for simple plots (single plot) I've seen that these bounds are passed at the time of figure creation only. But for gridplots I don't think these are needed at the time of axes creation and can be set separately (as done here).

By other plots I meant only those plots for which these bounds have been added:
- autocorrplot
- traceplot
- parallelplot
- forestplot
I should add the ```backend_config``` to these plots only for now right?

nitishp25:
@OriolAbril Please suggest ways to optimize the new ```backend_config``` code

OriolAbril:
backend_config should be like bakend_kwargs/plot_kwargs... , default to None, if None set to empty dict and then use setdefaults to fill default values.

I have just seen that docs on backend kwargs are wrong

nitishp25:
> backend_config should be like bakend_kwargs/plot_kwargs... , default to None, if None set to empty dict and then use setdefaults to fill default values.

And when the user will choose to enter ```bounds_x_range``` and/or ```bounds_y_range``` will this automatically use these values instead of the default ones or do we need to separately overwrite these values over the default ones?

arviz/plots/autocorrplot.py
```diff
@@ -49,7 +50,9 @@ def plot_autocorr(
         Matplotlib axes or bokeh figures.
     backend: str, optional
         Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
-    backend_kwargs: bool, optional
+    backend_config: bool, optional
```
OriolAbril: dict

arviz/plots/backends/bokeh/autocorrplot.py
```diff
@@ -2,17 +2,37 @@
...
 ):
     """Bokeh autocorrelation plot."""
+    if backend_config is None:
+        backend_config = {}
+
+    backend_config = {
+        **backend_kwarg_defaults(("bounds_y_range", "plots.bokeh.bounds_y_range"),),
+        **backend_config,
+    }
+
```
OriolAbril: I think in this case, the priority order should be: user introduced argument (backend_config), autocorr specific default, rcParam value. How does this sound? @ahartikainen @canyon289 

Therefore, in this case I would do:

```
if backend_config is None:
    backend_config = {}

backend_config.setdefault("bounds_range_x", (0, len_y))
```

Then the if is not necessary, now the value in backend_config is the one to be used always.

OriolAbril:
could you add some tests to test the `make_iterable_validator` function? The current [rcparams tests](https://github.com/arviz-devs/arviz/blob/master/arviz/tests/test_rcparams.py#L112) may help

nitishp25:
Sorry for the delay

arviz/tests/test_rcparams.py
```diff
@@ -135,6 +137,42 @@ def test_make_validate_choice(args, allow_none, typeof):
...
+    if (
+        value not in {None, "auto"}
+        and not isinstance(value, (set, frozenset))
+        and len(value) != length
+    ):
+        raise_error = "Iterable must be of length"
+    if value is None and not allow_none:
+        raise_error = "Only ordered iterable"
+    if value == "auto" and not allow_auto:
+        raise_error = "Could not convert"
```
OriolAbril: I was quite puzzled by this, I expected the only ordered iterable error in this case, turns out `np.iterable` returns true for strings, therefore, `"14"` would be accepted as `(1, 4)`.

arviz/tests/test_rcparams.py
```diff
@@ -135,6 +137,42 @@ def test_make_validate_choice(args, allow_none, typeof):
...
+@pytest.mark.parametrize("length", (2, 3))
+@pytest.mark.parametrize(
+    "args",
+    [
+        (False, (1, 2)),
+        ("Only ordered iterable", set(["a", "b", "c"])),
+        (False, None),
+        (False, "auto"),
+    ],
+)
```
OriolAbril: Given the complexity of the validator and how many cases it can take into account, I think it would be more clear to divide all these tests into several smaller test functions. For example: 
* one to test allow_none and allow_auto (testing with same parametrize True, False for both args and against a tuple, auto, None and a tuple-like string); 
* a second one to test the length (fixed a length vs None and against a tuple proper length, and tuple + tuple-like string of wrong lengths)
* a third one to test wrong inputs like set, non tuple-like string, a scalar...

arviz/rcparams.py
```diff
@@ -93,11 +102,39 @@ def _validate_boolean(value):
...
+    # based on matplotlib's _listify_validator function
+
+    def validate_iterable(value):
+        if allow_none and (value is None or isinstance(value, str) and value.lower() == "none"):
+            return None
+        if isinstance(value, str):
+            if allow_auto and value.lower() == "auto":
+                return "auto"
+            value = tuple(v.strip("([ ])") for v in value.split(",") if v.strip())
+        if np.iterable(value) and not isinstance(value, (set, frozenset)):
```
OriolAbril: `not isinstance(value, (str, set, frozenset)):` see other comment. In matplotlib's function there was an elif here, so strings did not get to this point. I think adding this here will keep the desired behaviour.

nitishp25: Does this function need to work with a string value like ```"(1, 4)"```? If not, then for string we can only keep the check for ```"auto"``` and remove line 115 that makes the tuple right? While also adding the str check along with set and frozen set.

arviz/tests/test_rcparams.py
nitishp25: Thanks, didn't think of that!

arviz/rcparams.py
OriolAbril: Yes, it should work with strings like `"1,4"`, `"(1, 4)"` or with square brackets. This strings are what would be read if the user set a tuple in the rcparams file

arviz/tests/test_rcparams.py
nitishp25: Actually ```"14"``` after conversion is stored in value as ```('14', )``` and I don't think we need iterables of length one anyway so should I change ```if np.iterable(value) and not isinstance(value, (set, frozenset)):``` to ```if np.iterable(value) and len(val) > 1 and not isinstance(value, (set, frozenset)):```? And we won't need to exclude strings here since ```"(1, 4)"``` after conversion to iterable will automatically be evaluated by ```scalar_validator```

OriolAbril: Ohh, true, my bad. as `tuple` is called, strings like `"auto"` or `"14"` are converted to a one element tuple, it is not possible for value to still be of type string at this point of the function. No need to modify the code (in a general case where length is None a 1 item iterable should be accepted)

arviz/tests/test_rcparams.py
```diff
@@ -135,6 +137,62 @@ def test_make_validate_choice(args, allow_none, typeof):
         assert value in accepted_values or value is None
 
 
+@pytest.mark.parametrize("allow_none", (True, False))
+@pytest.mark.parametrize("allow_auto", (True, False))
+@pytest.mark.parametrize(
+    "args", [(False, (1, 2)), (False, "auto"), (False, None), (False, "(1, 4)")]
```
OriolAbril: Very minor nit, no need to change it actually, there is no need to pass `raise_error` as it is always the same.

arviz/tests/test_rcparams.py
```diff
@@ -135,6 +137,62 @@ def test_make_validate_choice(args, allow_none, typeof):
...
+    if raise_error:
+        with pytest.raises(ValueError, match=raise_error):
+            validate_iterable(value)
+    else:
+        value = validate_iterable(value)
+        assert np.iterable(value) or value is None or value == "auto"
+
+
+@pytest.mark.parametrize("length", (2, None))
+@pytest.mark.parametrize("args", [(False, (1, 5)), (False, (1, 3, 5)), (False, "(3, 4, 5)")])
```
OriolAbril: same here

arviz/tests/test_rcparams.py
nitishp25: Well, now that I'm aware of it, I feel like changing it. Can I? :)

arviz/tests/test_rcparams.py
```diff
@@ -140,14 +140,15 @@ def test_make_validate_choice(args, allow_none, typeof):
 @pytest.mark.parametrize("allow_none", (True, False))
 @pytest.mark.parametrize("allow_auto", (True, False))
 @pytest.mark.parametrize(
-    "args", [(False, (1, 2)), (False, "auto"), (False, None), (False, "(1, 4)")]
+    "args", [(1, 2), "auto", None, "(1, 4)"]
```
OriolAbril: You can name it value here too

OriolAbril:
LGTM, I'll probably merge later today. Thanks!

PR #1020, opened 2020-01-22
not compute log_likelihood if vectorized trace cannot be obtained in Pyro

fehiepsi:
This resolves the issue raized at pyro [forum](https://forum.pyro.ai/t/the-arviz-function-from-pyro-has-mismatching-dimensions/1515). In Pyro, each variable dimension should be declared as `independent` or `dependent`. MCMC class relaxes this restriction a bit so some incorrect Pyro models still work. However, utilities such as `get_vectorized_trace` (used in `sample_stats_to_xarray`) require that declaration. This PR disables log_likelihood computation for those incorrect Pyro models. Because there are many reasons to trigger the issue, I  add a universal `try ... except ...` here to capture all of them.

### DIFF
--- a/arviz/data/io_pyro.py
+++ b/arviz/data/io_pyro.py
@@ -89,18 +89,23 @@ class PyroConverter:
         # extract log_likelihood
         dims = None
         if self.observations is not None and len(self.observations) == 1:
-            obs_name = list(self.observations.keys())[0]
-            samples = self.posterior.get_samples(group_by_chain=False)
-            predictive = self.pyro.infer.Predictive(self.model, samples)
-            obs_site = predictive.get_vectorized_trace(*self._args, **self._kwargs).nodes[obs_name]
-            log_likelihood = obs_site["fn"].log_prob(obs_site["value"]).detach().cpu().numpy()
-            if self.dims is not None:
-                coord_name = self.dims.get("log_likelihood", self.dims.get(obs_name))
-            else:
-                coord_name = None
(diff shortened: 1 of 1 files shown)


arviz/data/io_pyro.py
```diff
@@ -89,18 +89,23 @@ def sample_stats_to_xarray(self):
...
+                obs_site = vectorized_trace.nodes[obs_name]
+                log_likelihood = obs_site["fn"].log_prob(obs_site["value"]).detach().cpu().numpy()
+                if self.dims is not None:
+                    coord_name = self.dims.get("log_likelihood", self.dims.get(obs_name))
+                else:
+                    coord_name = None
+                shape = (self.nchains, self.ndraws) + log_likelihood.shape[1:]
+                data["log_likelihood"] = np.reshape(log_likelihood, shape)
+                dims = {"log_likelihood": coord_name}
+            except:
```
OriolAbril: Can the exception type vary? If possible, it would be great to explicit it, otherwise a pylint exception is neede

fehiepsi: Thanks! I thought that the tests have passed... my bad

fehiepsi: @OriolAbril The broadcasting issue can happen at multiple paths of Pyro code base so I choose the easier way: to add pylint exception here. Thanks for your suggestion!

VincentBt:
#1044 changes `pass` into `return None` in the except.

PR #1021, opened 2020-01-23
Fix arviz conversion of incomplete pymc3 traces

aseyboldt:
If sampling in pymc3 is aborted with Ctrl-C, it might create a trace that does not include chain 0.


OriolAbril:
LGTM, should this be added to the changelog?

Issue #953, opened 2019-12-08
Bokeh legends


ahartikainen:
Add (interactivate) legend (intuitive interactivity)

https://docs.bokeh.org/en/1.4.0/docs/reference/models/annotations.html#bokeh.models.annotations.Legend

https://docs.bokeh.org/en/1.4.0/docs/user_guide/interaction/legends.html

OriolAbril:
I'll reopen in hopes of reminding myself some day to try interactive legend in `plot_trace`

Issue #954, opened 2019-12-08
Bokeh Selection with subgroups


ahartikainen:
Add interactivity to select subgroups (divergences) and somehow show / hidden this or other groups. Maybe some bokeh widget or interactive legend?

https://docs.bokeh.org/en/1.4.0/docs/user_guide/interaction/widgets.html

percygautam:
@ahartikainen I have started learning about the bokeh plots. Point me if I'm wrong, If we take the example of densityplot, we have eight subplots for different schools. So, we have make only single plot by using interactive legends or bokeh widget( I think dropdown menu works best in this case) to alternate between school choices. 
What exactly do you mean by subgroups? Can you explain further the task?

PR #1024, opened 2020-01-23
Make the legends interactive in bokeh

percygautam:
Fixes part of issue #953 . Make the legends interactive in bokeh.


percygautam:
@ahartikainen @OriolAbril I have started working on the issue and have added the interactive legend feature to `densityplot.py`. Is this the right way? I'll add legends to other plots in the meantime.

ahartikainen:
It is basically this way. The problem currently is that we probably don't want to have legends on top of the figure, so we would really need a clever way to position the legend outside the plot. Also for multiple figures should share one legend.

Here should be one working example

https://gist.github.com/ahartikainen/506c23b58db06e162da9c188ceab146c

E.g. location probably can take also tuple, but I'm not sure.

percygautam:
@ahartikainen  I looked about it and found this open issue [Single Legend for GlyphRenderers on Multiple Plots](https://github.com/bokeh/bokeh/issues/7607) in bokeh. There's a workaround given in the end which we can look for.
There's also a method specified [here](https://stackoverflow.com/questions/46730609/position-the-legend-outside-the-plot-area-with-bokeh), which could work for placing legend outside the plot. But the location of where should we put it needs to be discussed.

percygautam:
@ahartikainen , I have tried to adjust the position of legends and this is what I've come up with.
Screenshot 2020-01-28 at 2 46 37 AM

Also, I've tried to optimize the code for better plotting. We just have to fix the position of title, and the legend position looks good to me.

percygautam:
About the other task of multiple figures sharing one legend, I've tried some methods (like passing glyph renders of one plot to another) but it didn't work. Also, in official bokeh, there is an open issue about the same. I'm thinking of trying to use the `grid` properties for this. Can you please provide any reference for this?
Also, as described in issue #954 , I think we can use subgroups(using dropdown) to select a school and render only one graph at a time. This will also solve the above problem.

ahartikainen:
Looks great.

I think that one-legend-thing might be hard to do. Yeah I think dropdown should be quite "easy" (yeah, they never are) to implement.

Maybe something like dropdown + multiple selections (right)

arviz/plots/backends/bokeh/densityplot.py
```diff
@@ -91,6 +90,17 @@ def plot_density(
...
             )
+            if data_label is not None:
+                if axis_map[label] in legend_items.keys():
+                    legend_items[axis_map[label]].append((data_label, plotted))
+                else:
+                    legend_items[axis_map[label]] = []
+                    legend_items[axis_map[label]].append((data_label, plotted))
+
+    for ax1, lg in legend_items.items():
+        legend = Legend(items=lg, location="center_right", orientation="horizontal",)
```
ahartikainen: I think this is a good compromise.

I don't think there is any "universal" way to position legend so it looks good

percygautam:
@ahartikainen So, I'll start working on other plots' legends. Also, I noticed the interactive legends od [EnergyPlot](https://arviz-devs.github.io/arviz/examples/bokeh/bokeh_plot_energy.html) are not working. I'll start working on this plot along with other plots having legend( I guess one or two more).

percygautam:
> Looks great.
> 
> I think that one-legend-thing might be hard to do. Yeah I think dropdown should be quite "easy" (yeah, they never are) to implement.
> 
> Maybe something like dropdown + multiple selections (right)

I am thinking of working on dropdowns under the `subgroup` issue in another PR, as we have to implement those subsections to lot of plots, unlike this `legend` issue. 
The dropdown with single selection will be doable but for multiple selections, I'll try to do but I'm not sure.

ahartikainen:
Sounds good.

No worries if something doesn't work.

percygautam:
@ahartikainen , I have added interactive legends to all other plots (having legends). The discrepancy of `energyplot` has also been solved and its working fine now.
I have changed the return value of `plot_kde`, in order to correct `energyplot`. This change is not affecting any other plot which uses `plot_kde`, so I don't think it'll cause a problem. We can always add legend the same way in other plots (when required in future) as we did in `energyplot`.
There is one problem of the title position, which I can't seem to find solution for. We need some way to write the title in same line as legend or is it fine this way?

percygautam:
@ahartikainen I have changed some font sizes. Interactive legends are working fine. This is how they are looking now:
Screenshot 2020-02-01 at 11 39 40 AM
Screenshot 2020-02-01 at 11 40 25 AM
Screenshot 2020-02-01 at 11 41 14 AM

arviz/plots/backends/bokeh/kdeplot.py
```diff
@@ -226,10 +229,15 @@ def plot_kde(
...
                 **pcolormesh_kwargs
             )
+            glyphs.append(image)
             ax.x_range.range_padding = ax.y_range.range_padding = 0
 
     if backend_show(show):
         bkp.show(ax, toolbar_location="above")
+
+    if legend and label is not None:
+        return glyphs
```
OriolAbril: This is what is making the documentation build fail. Maybe bokeh plots can return `ax, (extra, elements)` or something like this? Maybe a return_glyphs argument only for internal use? @ahartikainen

ahartikainen: Yeah, maybe add some argument to bokeh function to return this for internal use. So it would default to false and can not be seen in the main function.

btw did you mean

`if legend and label is not None`

or

`if (legend is not None) and (label is not)` None

percygautam: I have done as @OriolAbril suggested. I just defined `return_glyphs` as `None` and returned it along the axes. If the user doesn't specify the legend, it will plot the graph without legend.

arviz/plots/backends/bokeh/densityplot.py
```diff
@@ -66,18 +66,17 @@ def plot_density(
     if data_labels is None:
         data_labels = {}
 
+    legend_items = {}
     for m_idx, plotters in enumerate(to_plot):
-        for ax_idx, (var_name, selection, values) in enumerate(plotters):
+        for _, (var_name, selection, values) in enumerate(plotters):
```
OriolAbril: the enumerate an be removed completely if the index is not needed.

arviz/plots/backends/bokeh/densityplot.py
```diff
@@ -165,35 +185,46 @@ def _d_helper(
...
         est = calculate_point_estimate(point_estimate, vec, bw)
-        ax.circle(est, 0, fill_color=color, line_color="black", size=markersize)
+        plotted.append(ax.circle(est, 0, fill_color=color, line_color="black", size=markersize))
 
     _title = Title()
     _title.text = vname
     ax.title = _title
+    ax.title.text_font_size = "13pt"
+
+    return plotted
```
OriolAbril: Why is plotted returned? Shouldn't the axes array be returned?

arviz/plots/backends/bokeh/kdeplot.py
```diff
@@ -226,11 +230,16 @@ def plot_kde(
...
             )
+            glyphs.append(image)
             ax.x_range.range_padding = ax.y_range.range_padding = 0
 
     if backend_show(show):
         bkp.show(ax, toolbar_location="above")
-    return ax
+
+    if legend and label is not None:
+        return_glyphs = glyphs
```
OriolAbril: Given that glyphs is always generated, it could always be returned instead of returning none depending on the arguments. If it depends on an argument I'd make it explicit

arviz/plots/backends/bokeh/densityplot.py
percygautam: `plotted` is the list of glyphs that are needed to specify legend.
We are first plotting the axes in main function, and then adding plots (`glyph renders`) on top of it, using list of glyphs from helper function. So, the helper function needs only to get the list of glyphs, not axes.

arviz/plots/backends/bokeh/densityplot.py
percygautam: Yes, makes sense.

arviz/plots/backends/bokeh/densityplot.py
OriolAbril: Ooh, I see now, sorry

percygautam:
Still getting sphinx error for `bokeh_plot_kde`. Any workaround?

OriolAbril:
The examples assume the return value is `ax`, now it is not anymore. You have to update them to `ax, _`. Docs should also be updated in `plot/kdeplot.py`

percygautam:
Okay, thanks. I'll do the changes.

ahartikainen:
I still think our plotting code should return array of `axes` for the end user and `glyphs` should be for internal use.

This should be the case until we figure out something more advanced.

percygautam:
I have provided an internal argument that enables/disables the return value of plot_kde. Is this what you're looking for? Also what should be added to its docstring?

arviz/plots/backends/bokeh/kdeplot.py
```diff
@@ -226,10 +230,15 @@ def plot_kde(
                 palette=colors,
                 **pcolormesh_kwargs
             )
+            glyphs.append(image)
             ax.x_range.range_padding = ax.y_range.range_padding = 0
 
     if backend_show(show):
         bkp.show(ax, toolbar_location="above")
+
+    if legend and label is not None and return_glyph:
```
OriolAbril: the if should be only on return_glyph

arviz/plots/kdeplot.py
```diff
@@ -88,10 +89,12 @@ def plot_kde(
         check the plotting method of the backend.
     show : bool, optional
         Call backend show function.
+    return_glyph : bool, optional
+        Internal argument to return glyphs for bokeh
 
     Returns
     -------
-    axes : matplotlib axes or bokeh figures
+    axes : matplotlib axes or bokeh figures and glyphs
```
OriolAbril: (docstring comment)

```
Returns
-------
axes : matplotlib.Axes or bokeh.plotting.Figure
    Object containing the kde plot
glyphs : list, optional
    Bokeh glyphs present in plot.  Only provided if ``return_glyph`` is True.
```

Followed numpy docs like [`np.unique`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.unique.html)

arviz/plots/backends/bokeh/kdeplot.py
percygautam: Then we are getting the problem of unused arguments `label` and `legend`.

arviz/plots/backends/bokeh/densityplot.py
```diff
@@ -90,8 +89,18 @@ def plot_density(
...
                 shade,
                 axis_map[label],
-                data_label=data_label,
             )
+            if data_label is not None:
+                if axis_map[label] in legend_items.keys():
+                    legend_items[axis_map[label]].append((data_label, plotted))
+                else:
+                    legend_items[axis_map[label]] = []
+                    legend_items[axis_map[label]].append((data_label, plotted))
```
OriolAbril: Could legend items be a defaultdict?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Add `skipna` argument to `hpd` and `summary` (#1035)
 * Added `transform` argument to `plot_trace`, `plot_forest`, `plot_pair`, `plot_posterior`, `plot_rank`, `plot_parallel`,  `plot_violin`,`plot_density`, `plot_joint` (#1036)
 * Add `marker` functionality to `bokeh_plot_elpd` (#1040)
+* Add the functionality for `interactive legends` for bokeh plots of `densityplot`, `energyplot` and `essplot` (#1024)
```
OriolAbril: Maybe link to bokeh docs on interactive pegends instead of formatting it like code?

arviz/plots/backends/bokeh/densityplot.py
OriolAbril: https://docs.python.org/3/library/collections.html#collections.defaultdict

percygautam: Yes, it will make the code neat. Thanks, for the suggestion.

OriolAbril:
LGTM! And I love this feature, I'll probably merge tomorrow. It would be great to include this in next release.

Note: I have edited the description so it won't close the issue. Reason is I think it would be great to test these interactive legends and afterwards try to get them working in plot_trace (definitely in another PR).

percygautam:
Exactly, my thoughts. I was going to ask you about the legend implementation in `plot_trace`. These interactive legends will make the visualization clean. (And I have added the description earlier such that issue won't close. I thought I'd leave it up for future, if new plots were added.)

Issue #1023, opened 2020-01-23
plot_forest only plots first element


fonnesbeck:
I'm using ArviZ 0.6.1 on Debian Linux and Python 3.7, attempting to plot the trace of a vector of parameters derived from sampling a VI approximation. However, only the first element of the parameter vector is plotted:

![image](https://user-images.githubusercontent.com/81476/73009278-dd660300-3dd5-11ea-8d25-f69e296cbc90.png)

I can confirm that beta is indeed vector-valued:

![image](https://user-images.githubusercontent.com/81476/73009345-fa023b00-3dd5-11ea-8f10-d741b97bfd6b.png)

Oddly, I cannot replicate this behavior on other platforms (e.g. colab using 0.6.1).

fonnesbeck:
I can confirm that this also occurs on the current master.

ahartikainen:
How do you transform your samples to inferencedata? We usually assume `chain,draw,*shape` for the input (or 1D, which is then just `draw` (maybe))

fonnesbeck:
I'm just passing the result of `approx.sample()` to `plot_forest`, and that object is:

    <MultiTrace: 1 chains, 500 iterations, 8 variables>

fonnesbeck:
Doing an explicit cast via `az.from_pymc3(vi_trace)` does not fix the problem.

aloctavodia:
I can confirm this. It is also affecting other plots (but not summary). I can work on this, but not until tomorrow.

fonnesbeck:
OK, cool--no rush. Glad its not just me.

OriolAbril:
> Doing an explicit cast via `az.from_pymc3(vi_trace)` does not fix the problem.

Are all the variables in the `InferenceData`? What is the output of `az.from_pymc3(vi_trace).posterior`?

> It is also affecting other plots (but not summary). I can work on this, but not until tomorrow.

This seems to indicate the issue is in `plot_forest`, I just want to confirm `from_pymc3` is working fine.

fonnesbeck:
Passing the trace to summary generates a table of all values, so since `from_pymc3` is being called there as well, it shouldn't be an issue with that function.

aloctavodia:
Closing as fixed by #1027

PR #1027, opened 2020-01-24
Remove jitting of purge_duplicates()

justinbois:
The function `purge_duplicates()` in `arviz/plots/plots_utils.py` is currently jitted with Numba in `forceobj` mode. This results in incorrect dropping of duplicates. When this function is used in plot utilities for converting from xarrays to Numpy arrays, names/entries were then incorrectly dropped.

I updated the `purge_duplicates()` function to be both efficient (using sets, but still preserving order) and not jitted. See the comparison of results below.

The algorithm was gleaned from a [Stack Overflow discussion](https://stackoverflow.com/questions/480214/). I believe I properly attributed and reproduced the copyright in the code, based on instructions from Stack Overflow [here](https://stackoverflow.blog/2009/06/25/attribution-required/). Because we do not " remix, transform, or build upon the material" (as per the CC-BY-SA 4.0 license), I do not think we need to have to distribute under the same license as the original, so it should be fine as it licensing-wise.

```python
import arviz as az


def purge_duplicates(list_in):
    """Remove duplicates from list while preserving order.

    Parameters
    ----------
    list_in: Iterable

    Returns
    -------
    list
        List of first occurrences in order
    """
    # Algorithm taken from Stack Overflow,
    # https://stackoverflow.com/questions/480214. Content by Georgy 
    # Skorobogatov (https://stackoverflow.com/users/7851470/georgy) and
    # Markus Jarderot 
    # (https://stackoverflow.com/users/22364/markus-jarderot), licensed
    # under CC-BY-SA 4.0. 
    # https://creativecommons.org/licenses/by-sa/4.0/. 

    seen = set()
    seen_add = seen.add
    return [x for x in list_in if not (x in seen or seen_add(x))]


print('ArviZ version:', az.__version__)
print("Dropping duplicates out of [0, 1, 2, 3] with jitting:",
      az.plots.plot_utils.purge_duplicates([0, 1, 2, 3]))

print("Dropping duplicates out of [0, 1, 2, 3] without jitting:",
      purge_duplicates([0, 1, 4, 2, 3]))

print("Dropping duplicates out of [0, 1, 4, 2, 3, 2] with jitting:",
      az.plots.plot_utils.purge_duplicates([0, 1, 2, 3, 2]))

print("Dropping duplicates out of [0, 1, 4, 2, 3, 2] without jitting:",
      purge_duplicates([0, 1, 4, 2, 3, 2]))
```

And the output demonstrating the problem and the fix:
```
ArviZ version: 0.6.1
Dropping duplicates out of [0, 1, 2, 3] with jitting: [0]
Dropping duplicates out of [0, 1, 2, 3] without jitting: [0, 1, 4, 2, 3]
Dropping duplicates out of [0, 1, 4, 2, 3, 2] with jitting: [0]
Dropping duplicates out of [0, 1, 4, 2, 3, 2] without jitting: [0, 1, 4, 2, 3]
```


OriolAbril:
Good catch! Thanks! I think this will solve #1023 and expaim differences between local machine amd colab

OriolAbril:
@justinbois #990 has already been merged, so rebasing/merging latest changes in master should fix the failing tests. It would also be great if you could include this in the [changelog](https://github.com/arviz-devs/arviz/blob/master/CHANGELOG.md#maintenance-and-fixes). Other than this, LGTM, thanks!

ahartikainen:
No need to rebase, only need to fix lint tests.

aloctavodia:
Thanks @justinbois!

PR #1029, opened 2020-01-26
Update schema

OriolAbril:
## Description
Some updates to the schema specification following PRs #794 and #983. Also fixes several minor issues with docs in `from_emcee` and `plot_ppc`. 

For now only the [PyMC3 example](https://dfm.io/nbview/?url=https%3A%2F%2Fgithub.com%2Farviz-devs%2Farviz%2Fblob%2Fupdate_schema%2Fdoc%2Fschema%2FPyMC3_schema_example.ipynb) has been updated. I think it makes more sense to wait until predictions are supported in PyStan (which should not be too difficult) before updating it. 

## Checklist


ahartikainen:
I'm fine with this being "legal", but I can see that it will bite back at some point

    {"developer": names}
    ...
    coords={"developer": candidate_devs}

Also for pystan (and maybe the rest) this is not possible (without some hacks), because prediction is read in at the same time as the rest of the posterior.

But I like the idea of having external `from_pystan_predictions` (also for cmdstan+) function, so standalone `generated quantities` can be added later.

OriolAbril:
True, I missed that. We can change it to `developer_candidates`

ahartikainen:
I used `candidate developer` (with a space 😨 😁) in pystan example (see pr)

PR #1033, opened 2020-01-27
Added the new line for the API link documentation

percygautam:
## Description

Added the new line for API link documentation.

## Checklist

      PR format?

### DIFF
--- a/doc/sphinxext/gallery_generator.py
+++ b/doc/sphinxext/gallery_generator.py
@@ -39,8 +39,8 @@ MPL_RST_TEMPLATE = """
 
 .. image:: {img_file}
 
-**Python source code:** :download:`[download source: {fname}]<{fname}>`
-**API documentation:** {api_name} 
+|   **Python source code:** :download:`[download source: {fname}]<{fname}>`
+|   **API documentation:** {api_name} 
 
 .. literalinclude:: {fname}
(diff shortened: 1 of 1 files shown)


percygautam:
@OriolAbril Is this method for new line okay?

doc/sphinxext/gallery_generator.py
```diff
@@ -39,8 +39,8 @@ def execfile(filename, globals=None, locals=None):
 
 .. image:: {img_file}
 
-**Python source code:** :download:`[download source: {fname}]<{fname}>`
-**API documentation:** {api_name} 
+|   **Python source code:** :download:`[download source: {fname}]<{fname}>`
+|   **API documentation:** {api_name}
```
OriolAbril: I would use `\n` before `**API documentation**...`

percygautam: I tried to use this, but it doesn't seem to be supported in `rsT`.

OriolAbril: Oh, I would have thought that given that the file is written by the python script the newline would be directly present in the rst file, good to know. Then leaving a blank line between two links should do the trick

percygautam: Yes, that does the trick. :)

PR #1034, opened 2020-01-29
Fix yaml formatting in paper.md

mpadge:
I'm doing some analyses of all JOSS papers, and discovered that the `yaml` header of yours is not quite formatted properly - there was a stray white space following the end `---`, which this PR fixes.


aloctavodia:
thanks @mpadge!

PR #1037, opened 2020-01-29
Added example for concat method

### LINKED ISSUE #552: Add documentation example for combining InferenceData
## Tell us about it

Add documentation for combining `az.InferenceData`

https://github.com/arviz-devs/arviz/pull/547

percygautam:
## Description

Fixes #552 

## Checklist

      PR format?


arviz/data/inference_data.py
```diff
@@ -252,6 +252,46 @@ def concat(*args, dim=None, copy=True, inplace=False, reset_dim=True):
...
+
+        In [1]: dataB = az.convert_to_inference_data(dataset, group = "prior")
+           ...: dataB
+
+    We have created another ``InferenceData`` object with group 'prior'. Now we will concatenate
+    these two ``InferenceData`` objects:
+
+    .. ipython::
+
+        In [1]: az.concat(dataA, dataB)
```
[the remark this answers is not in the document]
percygautam: @OriolAbril  The inplace is `False` by default, it is returning the concatenated inference data object. I'll use `from_dict` method.

percygautam:
@OriolAbril  Do I add more examples regarding these? I don't understand the question.

OriolAbril:
In the first versions of the example, the inference data created had groups with several variables and each of these variables had multiple dimensions. Now it looks like the generated inference data has groups that contain only a scalar?

percygautam:
Okay, I understand now. So, you're asking for something like this, right? (I don't think coords is necessary)
```
data = {
         "a": (["chain", "draw", "a_dim"], np.random.normal(size=(4, 100, 3))),
         "b": (["chain", "draw"], np.random.normal(size=(4, 100))),
     }  
coords={
        "chain": (["chain"], np.arange(4)),
        "draw": (["draw"], np.arange(100)),
        "a_dim": (["a_dim"], ["x", "y", "z"])
    }
dataA = az.from_dict(posterior=data, coords=coords)
dataB = az.from_dict(prior=data, coords=coords)
az.concat(dataA,dataB)
```

OriolAbril:
Yes, something like this

percygautam:
I have used the same example as above. Should we also add examples for cases for concatenating over `chain` or `draw`?

arviz/data/inference_data.py
```diff
@@ -252,6 +252,43 @@ def concat(*args, dim=None, copy=True, inplace=False, reset_dim=True):
...
+           ...: data = {
+           ...:         "a": (["chain", "draw", "a_dim"], np.random.normal(size=(4, 100, 3))),
+           ...:         "b": (["chain", "draw"], np.random.normal(size=(4, 100))),
+           ...:         }
+           ...: coords = {
+           ...:         "chain": (["chain"], np.arange(4)),
+           ...:         "draw": (["draw"], np.arange(100)),
+           ...:         "a_dim": (["a_dim"], ["x", "y", "z"]),
+           ...:         }
+           ...: dataA = az.from_dict(posterior=data, coords=coords)
```
OriolAbril: I tried this code locally and it does not create an InferenceData object properly. ArviZ idiom should be followed:

```
data = { 
   "a": np.random.normal(size=(4, 100, 3)),
   "b": np.random.normal(size=(4, 100))
}
coords = {"a_dim": ["x", "y", "z"]}
dataA = az.from_dict(data, coords=coords, dims={"a": ["a_dim"]})
```

percygautam: Thanks, for the correction. It seems the xarray dataset (used in the previous case) map dims automatically, but we need to specify the mapping in case of `from_dict`. Is this the case?

OriolAbril: xarray's Dataset allows several notations, in the one used, the data and coordinates are specified as dicts of tuples, where the tuple contains both the data and the dimensions. ArviZ uses only one syntax, dicts of arrays for all 3 of them, the values of the dicts are directly understood as data, coordinate values or dimension list

percygautam: Okay, thanks for the clarification :)

arviz/data/inference_data.py
```diff
@@ -252,6 +252,41 @@ def concat(*args, dim=None, copy=True, inplace=False, reset_dim=True):
...
+
+        In [1]: import arviz as az
+           ...: import numpy as np
+           ...: data = {
+           ...:         "a": (["chain", "draw", "a_dim"], np.random.normal(size=(4, 100, 3))),
+           ...:         "b": (["chain", "draw"], np.random.normal(size=(4, 100))),
+           ...:         }
+           ...: coords = {
+           ...:         "a_dim": ["x", "y", "z"]
+           ...:         }
```
OriolAbril: This can be in the same line, it is not too long

arviz/data/inference_data.py
```diff
@@ -252,6 +252,41 @@ def concat(*args, dim=None, copy=True, inplace=False, reset_dim=True):
...
+    unique groups by default. We first create an InferenceData object:
+
+    .. ipython::
+
+        In [1]: import arviz as az
+           ...: import numpy as np
+           ...: data = {
+           ...:         "a": (["chain", "draw", "a_dim"], np.random.normal(size=(4, 100, 3))),
+           ...:         "b": (["chain", "draw"], np.random.normal(size=(4, 100))),
+           ...:         }
```
OriolAbril: We should follow black formatting, the `}` should be on the same level as the `d` in data and the two lines in between indented 4 spaces.

arviz/data/inference_data.py
```diff
@@ -252,6 +252,39 @@ def concat(*args, dim=None, copy=True, inplace=False, reset_dim=True):
...
+           ...: data = {
+           ...:     "a": (["chain", "draw", "a_dim"], np.random.normal(size=(4, 100, 3))),
+           ...:     "b": (["chain", "draw"], np.random.normal(size=(4, 100))),
+           ...: }
+           ...: coords = {"a_dim": ["x", "y", "z"]}
+           ...: dataA = az.from_dict(data, coords=coords, dims={"a": ["a_dim"]})
+           ...: dataA
+
+    We have created an ``InferenceData`` object with default group 'posterior'. Now, we will
+    create another InferenceData object:
```
OriolAbril: ``InferenceData`` in the second appearence

arviz/data/inference_data.py
```diff
@@ -252,6 +252,39 @@ def concat(*args, dim=None, copy=True, inplace=False, reset_dim=True):
     InferenceData
         A new InferenceData object by default.
         When `inplace==True` merge args to first arg and return `None`
+
+    Examples
+    --------
+    Use ``concat`` method to concatenate InferenceData objects. This will concatenates over
+    unique groups by default. We first create an InferenceData object:
```
OriolAbril: ``InferenceData`` in both cases

percygautam:
Hey @OriolAbril , Sorry for the delay. I was on the move. I'll add examples for chain and draw.

OriolAbril:
Don't worry, there is no rush!

Issue #950, opened 2019-12-08
Bokeh marker options


ahartikainen:
Each plot needs a marker option implemented for the bokeh backend.

http://docs.bokeh.org/en/1.3.2/docs/reference/models/markers.html

Ideally, we have a marker option somewhere and helper function that takes in {marker as a string, marker name, marker function, custom js for marker} and outputs correct function.

These markers should also be added to docstrings or at least give some link for supported markers.

percygautam:
@ahartikainen Should the new marker param takes only string as value(like `*` for asterisk and so on) or other values also like marker_name etc?
 I thinking of using switch call to different marker functions, as they all have same params, in `function_generator` helper function. Is my approach viable or is there any other method of doing it?

ahartikainen:
We are currently using methods from `bokeh.plotting.figure` to plot different markers.

I think we could change our way of plotting so we would use explicitly `bokeh.models.markers`

https://docs.bokeh.org/en/latest/docs/reference/models/markers.html

So I think switch (btw, in python just use dictionary with `.get` method adding a default value) could work. We would have a function that takes in the marker given for the plot, checks if it is a `str` and if it is, then it would translate it to a marker object, we could have some predefined markers similar to matplotlib / `bokeh.figure.scatter`.

So in the end user could define a custom marker and just pass that to our function. (This could be arbitrary shape e.g.)

We also need to think that the use of the default parameters are compatible with `bokeh.models.markers`

percygautam:
@ahartikainen @OriolAbril I have implemented the Marker option for `eloped_plot` (type of scatter plot). I am getting something like this:
![Screenshot 2020-01-29 at 2 54 46 AM](https://user-images.githubusercontent.com/20489158/73307164-b11e0c80-4243-11ea-808e-624c7a0b64d9.png)
![Screenshot 2020-01-29 at 2 55 28 AM](https://user-images.githubusercontent.com/20489158/73307221-cd21ae00-4243-11ea-9460-159090aaa3b6.png)

I am confused about some plots whether the marker option needs to be added to them or not (like kde plots, energy plots, dist plot, etc). Can you specify the plots on which markers to be added?

PR #1040, opened 2020-02-03
[WIP] Added marker param to bokeh

percygautam:
## Description

Related to issue #950 . This is how `elpd_plot` looks like for different marker options:
For `Cross`:
Screenshot 2020-02-03 at 4 48 59 PM
For `Diamond`:
Screenshot 2020-02-03 at 4 50 28 PM
For `Asterisk`:
Screenshot 2020-02-03 at 4 50 47 PM

## Checklist

      PR format?


percygautam:
I have added marker param to `elpd_plot`. I am not sure of my method, @ahartikainen @OriolAbril Can you confirm this? I'll add this param to other scatter plots after review.
Also, as we have to use markers list to validate the input param value, it'll be a hassle to define this list in every backend bokeh file. So, I'm thinking of defining this as global variable in some other common file an then importing it according to need.

arviz/plots/elpdplot.py
```diff
@@ -22,6 +22,7 @@ def plot_elpd(
     ax=None,
     ic=None,
     scale=None,
+    marker="auto",
```
OriolAbril: We should try to get some common API between matplotlib and bokeh. The marker in matplotlib can be modified via `plot_kwargs`. @ahartikainen what do you think will be better, pop marker from `plot_kwargs` or integrate the new `marker` with matplotlib backend?

arviz/rcparams.py
```diff
@@ -107,6 +108,29 @@ def _validate_boolean(value):
...
+                "InvertedTriangle",
+                "Square",
+                "SquareCross",
+                "SquareX",
+                "Triangle",
+                "X",
+            },
+            case_sensitive=True,
+        ),
+    ),
```
OriolAbril: I think it will be more readable to create a `_validate_bokeh_marker = _make_validate_choice...` around line 94 and then here use this new function so everything fits in a couple of lines max.

OriolAbril:
Regarding validation of bokeh markers, it is not ideal, but maybe for now the `_validate_bokeh_marker` in rcParams could be used?

percygautam:
@OriolAbril I am thinking of creating a function `_validate_bokeh_marker` which will concise the code in rcparam and also can be imported to any backend bokeh file to validate user-provided marker param. Will that work?

percygautam:
@OriolAbril Also I was checking out matplotlib markers [list](https://matplotlib.org/api/_as_gen/matplotlib.markers.MarkerStyle.html#matplotlib.markers.MarkerStyle), Do we also implement our bokeh markers like this?

ahartikainen:
I think they can be different, let's just give the options in docstring.

Or do we want to map back and forward these?

OriolAbril:
I would use the corresponding markers for each library, and mixing them won't really improve functionality, at most it will ease plotting but at a maintenance cost of keeping them all up-to-date.

My main issue is with `marker` as a function argument. I think there are two options:

* set marker via `plot_kwargs={"marker": marker_str}` (I am inclined towards this) -> modify proposed bokeh code
* set marker via `marker=marker_str` -> modify current matplotlib code

I don't think we should consider the option of using the first option for matplotlib and the second for bokeh (which is what currently happens in this PR)

ahartikainen:
Yeah, that sounds good. It should not be argument but part of that dict. 

Let's just add small example to docstring.

percygautam:
@OriolAbril @ahartikainen I have changed the code as discussed. Now marker is passed as dict argument to `plot_kwargs`. What type of example should be added docstring to illustrate this feature?

arviz/rcparams.py
```diff
@@ -107,8 +169,10 @@ def _validate_boolean(value):
     "plot.bokeh.figure.width": (500, _validate_positive_int),
     "plot.bokeh.figure.height": (500, _validate_positive_int),
     "plot.bokeh.show": (True, _validate_boolean),
+    "plot.bokeh.marker": ("Cross", _validate_marker("bokeh")),
     "plot.matplotlib.constrained_layout": (True, _validate_boolean),
     "plot.matplotlib.show": (False, _validate_boolean),
+    "plot.matplotlib.marker": ("+", _validate_marker("matplotlib")),
```
OriolAbril: I would not include matplotlib marker here. matplotlib already has its own rcparams and marker can be handled there. Also, matplotlib markers allow even more options than these strings. Sticking with bokeh should be enough

arviz/plots/backends/bokeh/elpdplot.py
```diff
@@ -146,12 +147,12 @@ def _plot_atomic_elpd(
     ylabels_shown,
     plot_kwargs,
 ):
-    ax_.cross(
-        np.asarray(xdata),
-        np.asarray(ydata),
-        line_color=plot_kwargs.get("color", "black"),
-        size=plot_kwargs.get("s"),
-    )
+    marker = _validate_marker("bokeh")(plot_kwargs.get("marker"))
```
OriolAbril: `plot_kwargs.get("marker", rcParams["plot.bokeh.marker"])` otherwise the value stored in rcParams is never used.

arviz/plots/elpdplot.py
```diff
@@ -144,7 +144,10 @@ def plot_elpd(
 
     if plot_kwargs is None:
         plot_kwargs = {}
-    plot_kwargs.setdefault("marker", "+")
+    if backend == "matplotlib":
+        plot_kwargs.setdefault("marker", "+")
```
OriolAbril: Adding on the other comment, if we want rcParams to control the marker in matplotlib elpd plot, we can just remove this line. That will make the marker fall back to its default in [matplotlib rcparams](https://github.com/matplotlib/matplotlib/blob/master/matplotlibrc.template#L603) and users will be able to modify it from there.

arviz/rcparams.py
percygautam: Okay, makes sense.

arviz/plots/elpdplot.py
percygautam: The matplotlib rcParam has `.` as default marker option. This will make the matplotlib example and bokeh example generated graph different. Do we also need to modify the example code?

OriolAbril: No need to modify the example if you don't want to, it will use a circle instead of a cross.

percygautam: As in all other examples, both 'bokeh' and 'matplotlib' have the same graphs. I think it'll be better to add `plot_kwargs = {"marker":"+"}` to matplotlib files. It'll also be helpful to illustrate the use of `marker` in the plots. What are your thoughts?

percygautam: I'll start to add these changes to other scatter plots.

OriolAbril: I am not sure there are many instances of scatter plots, I think we generally use `plot` with linestyle `None`. The docs in elpdplot should be fixed though, `plot_kwargs` are passed to `plt.scatter` in this case

percygautam: @OriolAbril  Apart from scatter plots , normal `plot`'s marker can also be changed. For `essplot`, we are passing kwargs `{'color': 'red', 'linestyle': 'none', 'linewidth': 1.5, 'markersize': 6.0, 'marker': '+', 'zorder': 3, 'label': 'quantile'}` to `ax.plot`. We can change marker in kwarg to change the marker

percygautam: There is similar case for `mcseplot` where we are explicitly defining the kwargs.

arviz/rcparams.py
```diff
@@ -93,6 +93,30 @@ def _validate_boolean(value):
     return value is True or value == "true"
 
 
+def _validate_marker(value):
```
OriolAbril: nit: `_validate_bokeh_marker`

arvizrc.template
```diff
@@ -18,6 +18,7 @@ plot.bokeh.figure.dpi        : 60          # dots (pixels) per inch
 plot.bokeh.figure.height     : 500         # num of pixels
 plot.bokeh.figure.width      : 500         # num of pixels
 plot.bokeh.show              : true        # call bokeh.plotting.show for figure or grid of figures. One of "true", "false"
+plot.bokeh.marker            : Cross       # specify the marker type used to plot
```
OriolAbril: nit: sort rcparams groups alphabetically

arviz/plots/elpdplot.py
percygautam: > I am not sure there are many instances of scatter plots, I think we generally use `plot` with linestyle `None`. The docs in elpdplot should be fixed though, `plot_kwargs` are passed to `plt.scatter` in this case

I have corrected the docstring.

ahartikainen:
Hi, could also add possibility to change size of the markers? Maybe another PR

arviz/plots/backends/bokeh/elpdplot.py
```diff
@@ -146,12 +147,14 @@ def _plot_atomic_elpd(
...
     plot_kwargs,
 ):
-    ax_.cross(
-        np.asarray(xdata),
-        np.asarray(ydata),
-        line_color=plot_kwargs.get("color", "black"),
-        size=plot_kwargs.get("s"),
+    marker = _validate_bokeh_marker(plot_kwargs.get("marker"))
+    marker_func = getattr(mk, marker)
+    sizes = np.ones(len(xdata)) * plot_kwargs.get("s")
```
ahartikainen: Maybe we could first try to read `"size"` and then `"s"`

    plot_kwargs.get("size", plot_kwargs.get("s"))

OriolAbril: I think this has to be handled before passing to the backend

ahartikainen: This is difference between mpl and bokeh

OriolAbril: Yes, I meant that if I remember correctly, the common code adds the "s" key to plot_kwargs

ahartikainen: It would still be the default

OriolAbril:
@percygautam there are some merge conflicts in rcparams, do not hesitate to ask for guidance if you need it. After fixing this and Ari's comment we can merge.

arviz/plots/backends/bokeh/elpdplot.py
percygautam: Hey @ahartikainen @OriolAbril , This problem will be fixed after #1041 is fixed, which I'm currently working on. Should I add temporary check(for now) in main plot file?

OriolAbril: I think it can wait

percygautam:
Should I add it to changelog?

Issue #796, opened 2019-08-16
Move pointwise log likelihood data to log_likelihood group


OriolAbril:
Following #794,  log likelihood data should be moved to the new `log_likelihood` group of the inference data objects. PyStan, PyMC3, emcee and dict have been updated in the PR, cmdstan, cmdstanpy, pyro and numpyro are still pending. For users familiar with the libraries it should be an achievable issue.

ahartikainen:
I can probably fix cmdstanpy / cmdstan after 794 is merged

VincentBt:
#1044 does it for numpyro and pyro

OriolAbril:
#1064 did this for cmdstan and cmdstanpy. Only tensorflow_probability is missing

OriolAbril:
Closing as `from_tfp` needs a thorough rewrite. I have made sure using `log_likelihood` group is explicitly stated in #812

PR #1044, opened 2020-02-04
Move pointwise log likelihood data to log_likelihood group (Pyro + NumPyro)

VincentBt:
Follows #796, for Pyro and NumPyro: I added "log_likelihood" to InferenceData, and added tests for multiple observed variables. 

Thanks for a Pyro/NumPyro expert to review this (@fehiepsi maybe?). Indeed, I am not entirely sure about the Pyro part: the test_multiple_observed_rv fails if y2 and y1 have different dimensions. 
For the rest, I copied io_pymc3.py, in particular for coord_name and dims.

Sorry in advance for the possible mistakes, it's my first commit.

### DIFF
--- a/arviz/tests/test_data_numpyro.py
+++ b/arviz/tests/test_data_numpyro.py
@@ -58,7 +58,6 @@ class TestDataNumPyro:
         import numpyro
         import numpyro.distributions as dist
         from numpyro.infer import MCMC, NUTS
-        
         y1 = np.random.randn(10)
         y2 = np.random.randn(100)
         def model_example_multiple_obs(y1=None, y2=None):
@@ -66,7 +65,7 @@ class TestDataNumPyro:
             numpyro.sample('y1', dist.Normal(x, 1), obs=y1)
             numpyro.sample('y2', dist.Normal(x, 1), obs=y2)
         nuts_kernel = NUTS(model_example_multiple_obs)
-        mcmc = MCMC(nuts_kernel, num_warmup=20, num_samples=20)
+        mcmc = MCMC(nuts_kernel, num_samples=10, num_warmup=2)
         mcmc.run(PRNGKey(0), y1=y1, y2=y2)
         inference_data = from_numpyro(mcmc)
         test_dict = {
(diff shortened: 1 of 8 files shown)


arviz/tests/test_data_numpyro.py
```diff
@@ -25,10 +26,10 @@ class Data:
     def get_inference_data(self, data, eight_schools_params):
         posterior_samples = data.obj.get_samples()
         model = data.obj.sampler.model
-        posterior_predictive = Predictive(model, posterior_samples).get_samples(
+        posterior_predictive = Predictive(model, posterior_samples).__call__(
```
fehiepsi: nit: `.__call__` is not required

arviz/tests/test_data_numpyro.py
```diff
@@ -25,10 +26,10 @@ class Data:
...
             PRNGKey(1), eight_schools_params["J"], eight_schools_params["sigma"]
         )
-        prior = Predictive(model, num_samples=500).get_samples(
+        prior = Predictive(model, num_samples=500).__call__(
```
fehiepsi: ditto: `.__call__` is not required

arviz/tests/test_data_numpyro.py
```diff
@@ -3,6 +3,7 @@
 import pytest
 from jax.random import PRNGKey
 from numpyro.infer import Predictive
+import jax.numpy
```
fehiepsi: nit: this import is not required I guess?

OriolAbril:
Can you update also Azure configuration file (in the same way as Travis) and add this to the changelog?

Everything else looks good, after these last changes we can merge

VincentBt:
> In case the try except must be a block as it is now, the `pass` should probably be a `return None`

You're right, I changed the code as you suggested.

Ready to merge.

PS: is it possible to run the Travis and Azure checks on my computer for the specific files that changed? That would have decreased the number of unnecessary commits (not talking about having to wait for all the checks to finish on this page before getting a failure)

OriolAbril:
Travis and Azure mainly run the tests with pytest and the linters. If you have the required dependencies installed, you can run `pytest arviz/tests/test_file.py` or `pytest arviz/tests -k "key1 and key2 or key3"` to tests only a subset of the files/test functions. In addition, running `.scripts/lint.sh` should mimic the lint job. 

If you want to keep special versions in your environment you can always use docker to run tests locally being sure the dependencies installed will be the same as the ones in Azure and Travis first job. There is a section on this in the [contributing guide](https://github.com/arviz-devs/arviz/blob/master/CONTRIBUTING.md#developing-in-docker). The `--test` flag is a shortcut to run all tests, to run only some of them, enter a bash session in docker and run the specific pytest command. 

Finally, the documentation is also generated automatically and sphinx (or sphinx inside docker) can be used to build it locally and check everything works. In many cases it may not be necessary.

OriolAbril:
Thanks for the PR!

Issue #939, opened 2019-12-04
Use of pymc3 _model pointer will cause future problems


rpgoldman:
**Describe the bug**
In `data/io_pymc3/` the `PyMC3Converter` finds the PyMC3 model -- which it needs to interpret the trace, through a pointer hidden in the `pm.MultiTrace` object.

This use will not be compatible with out-of-sample prediction, which in PyMC3 requires using a thinned trace in a *different* model (in general, one with different predictors and different dimensions).

See https://github.com/arviz-devs/arviz/blob/3aba6c9d1831199631a8e48ecc82ed482bdf00e8/arviz/data/io_pymc3.py#L45-L59

We have already added an explicit `model` argument to the translator.  The next step would be to deprecate using the existing back door and move towards making the PyMC3 `Model` object a mandatory argument.

Issue #415, opened 2018-11-17
Standard for documentation types


ColCarroll:
## Tell us about it

Documentation uses `numpydoc` and is linted using `pydocstyle`. However, the format of the type is not checked.  Specifically, [densityplot.py](https://github.com/arviz-devs/arviz/blob/212c5977aba29be8f0ed25cf21e05ffac916dc7b/arviz/plots/densityplot.py#L40) now uses proper type hinting, while everywhere else uses a less rigorous style emulating other pydata projects (`int, optional`, instead of `Optional[int]`, for example)

I am most in favor of consistency across the code base, which would suggest changing densityplot.py. I would also prefer type hinting everywhere to our current standard.

**Specifically**: I think we should manually urge new contributions to match current documentation practices, but I would also accept a pull request that converts a reasonable portion of the codebase to the new style.

## Thoughts on implementation

My feelings on this are weak enough that I *do not* think it should be in CONTRIBUTING.md, since every requirement there feels like a hurdle to engagement. If we ended up in a place with type hinting everywhere, then I would be in favor of adding it.

canyon289:
A couple points
1. Putting something that looks like type hints in the docstring isn't actually type hints. Type hints (mostly) go in the argument list for a function. From everything I can read numpy docstrings and type hinting can coexist and don't clobber each other. I have a weak prior on this though and am continuing to read

2. I'm strongly in favor of adopting keeping numpy style docstrings. Some of our major dependencies  call for them explicitly and I don't see a good reason to diverge from our larger community

https://matplotlib.org/devel/documenting_mpl.html#formatting-conventions
http://xarray.pydata.org/en/stable/contributing.html#about-the-xarray-documentation
https://docs.python.org/3/library/typing.html

3. I do agree that docstring style does pose a barrier. Perhaps as a compromise we can accept pull requests where the docstring style doesn't match but make an issue to go back and fix it

OriolAbril:
I vote for numpy style docstrings and leaving the issue open to encourage anybody who might feel like it to use type hinting in the function arguments. It would have some advantages, mainly when using IDEs as far as I know.

OriolAbril:
Should probably be added to #798 so it is clear what should be followed.

percygautam:
@OriolAbril So, we have to extend the contribution guide about new links suggested by @canyon289 . This can also be added to the new PR template being added. Also, are you suggesting to change all docstrings to `mypy` format?

OriolAbril:
[...]

PR #1045, opened 2020-02-05
extend model arg usage in io_pymc3 to fix plot_ppc with prior

OriolAbril:
## Description
Try to improve handling of prior groups and observed data in `io_pymc3`. Related to #1002. It introduces an important change in behaviour, it proposes to exclude the `observed_data` group whenever predictions are present. @corriebar @rpgoldman 

I also extended tests on `io_pymc` and improved the helper function `check_multiple_attrs`.

And finally, ~how do you feel about deprecating the use of `from_pymc3` without model nor trace (or even without model). I was thinking on adding a warning.~ For now not using a model nor trace will raise a pending deprecation warning as a start. @canyon289 @aloctavodia @ColCarroll 

Related to #939. 

## Checklist

      PR format?

### DIFF
--- a/arviz/data/io_pymc3.py
+++ b/arviz/data/io_pymc3.py
@@ -1,5 +1,6 @@
 """PyMC3-specific conversion code."""
 import logging
+import warnings
 from typing import Dict, List, Any, Optional, TYPE_CHECKING
 from types import ModuleType
 
@@ -85,6 +86,14 @@ class PyMC3Converter:  # pylint: disable=too-many-instance-attributes
             self.ndraws = len(trace)
         else:
             self.nchains = self.ndraws = 0
+
+        if model is None:
+            warnings.warn(
+                "Using `from_pymc3` without the model will be deprecated in a future release. "
+                "Not using the model will return less accurate and less useful results. "
+                "Make sure you use the model argument or call from_pymc3 within a model context.",
+                PendingDeprecationWarning
+            )
 
         self.prior = prior
         self.posterior_predictive = posterior_predictive
--- a/arviz/data/io_pymc3.py
+++ b/arviz/data/io_pymc3.py
@@ -7,7 +7,7 @@ import numpy as np
 import xarray as xr
 from .. import utils
 from .inference_data import InferenceData, concat
-from .base import requires, dict_to_dataset, generate_dims_coords, make_attrs
+from .base import requires, dict_to_dataset, generate_dims_coords, make_attrs, CoordSpec, DimSpec
 
 if TYPE_CHECKING:
     import pymc3 as pm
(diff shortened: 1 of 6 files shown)


arviz/data/io_pymc3.py
```diff
@@ -121,8 +121,7 @@ def arbitrary_element(dct: Dict[Any, np.ndarray]) -> np.ndarray:
     def find_observations(self) -> Optional[Dict[str, Var]]:
         """If there are observations available, return them as a dictionary."""
         has_observations = False
-        if self.trace is not None:
```
rpgoldman: I don't really like leaving out the check that tells the caller why they are not getting observations out of their trace when they expect it.  Perhaps replace my error with a `UserWarning`?
It seems like Arviz tries to get whatever information it can out of a trace, and ignores whatever it can't figure out.  This makes it relatively robust, but means that the user doesn't know that they could get *more* information into the `InferenceData` by providing a model.

rpgoldman:
Per my comment in the code -- I think it would be a good idea to require supply of the model.  I think this is particularly important because the user *could* get an inappropriate model out of the trace.
This may seem far-fetched, but for PyMC3 one actually has to move a trace to a new model in some cases in order to do out of sample predictions.

arviz/tests/helpers.py
```diff
@@ -165,8 +165,10 @@ def check_multiple_attrs(test_dict, parent):
 
     Args
     ----
-    test_dict: dict
-        Its structure should be `{dataset1_name: [var1, var2], dataset2_name: [var]}`
+    test_dict: dict of {str : list of str}
```
rpgoldman: What's the motivation for requiring that a variable be absent?

arviz/data/io_pymc3.py
OriolAbril: I completely agree, I was just waiting in order to replace it with an informative warning or a deprecation warning.

arviz/tests/helpers.py
OriolAbril: In the case of a group, I was thinking in cases such as `log_likelihood` in sample stats, I see more uses to the group case though (e.g. `test_multiple_observed_rv_without_observations` in test_data_pymc where we want to test there is no observed data because there should not be)

rpgoldman: Possibly change the `test_dict` type annotation to type hint format?  `Dict[str, List[str]]`.

For that matter, you could add type hints, but of course that's not required.

OriolAbril: I will try to add proper type hinting this evening (following #415) and leave the docstring in pydata style. Some output examples are also in order I think. The `assert error False == ["observed_data", ("sample_stats", "~log_likelihood")]` messages should be useful for figuring what is the problem (in this case, observed data group is expected but not found and log likelihood should not be in sample stats but it is)

OriolAbril:
I think this is ready to merge

OriolAbril:
Only the `model` argument warning-deprecation is left to do, after that we can merge

OriolAbril:
I decided to start with the deprecation on no model nor trace (it will be triggered in cases such as only prior or only posterior predictive/predictions). We can make it more strict next release by requiring a model even when the trace is present and eventually remove the `_straces` trick. In my opinion, PyMC3  users should not mind having to call `from_pymc3` within a model context or passing the model as kwarg, it already is the workflow for sample, sample_prior_predictive...

I think it is ready to merge

rpgoldman:
I agree. If the tests pass again, we should merge. Looks very good!

OriolAbril:
This allows to make prior predictive checks right away with PyMC3!

```
with pm.Model():
    # define model
    idata = az.from_pymc3(prior) # has to be either from a model context or
                                 # passing model=model to from_pymc3

az.plot_ppc(idata, group="prior")
```

I am not really sure about the way to advertise this though, is there a PyMC3 tutorial on prior predictive sampling?  @canyon289 @aloctavodia

AlexAndorra:
This is great -- thanks guys!!
To my knowledge, there is no tutorial dedicated to prior pred checks on PyMC's website. I found:

- [The putting notebook](https://docs.pymc.io/notebooks/putting_workflow.html#Prior-Predictive-Checks), where Colin does some prior checks -- but maybe too complicated a use-case for an introduction.

- I updated [the radon NB](https://docs.pymc.io/notebooks/multilevel_modeling.html) a few weeks ago, and among other things added prior checks. The new version is not up yet (but it's on master if you wanna check it out), but this could be a good avenue to advertise this new use.

- [One NB](https://docs.pymc.io/notebooks/posterior_predictive.html) dedicated to _posterior_ predictive checks. We could extend (and update) it -- this would make a useful educational ressource to point to.

In general, I noticed on the Discourse that people find it hard to understand the concept of prior/posterior pred checks -- or just don't know about them. So I think an intro tutorial would be really useful.

PR #1046, opened 2020-02-05
Remove bokeh popups during doc generation

### LINKED ISSUE #927: Gallery examples don't pop up the window with the plot
For example, https://arviz-devs.github.io/arviz/examples/plot_pair.html just exits silently.

percygautam:
## Description

Fixes #927 . Remove bokeh popups during doc generation and corrected code for matplotlib.  Doc generation is tested locally.

## Checklist

      PR format?


doc/sphinxext/gallery_generator.py
```diff
@@ -24,7 +24,7 @@
 
 from bokeh.io import export_png
 from bokeh.layouts import gridplot
-
+from arviz.rcparams import rc_context
```
percygautam: I can't seem to import rcparam file and am getting the error `ValueError: attempted relative import beyond top-level package`. Is there any workaround to this?

doc/sphinxext/gallery_generator.py
```diff
@@ -398,6 +400,9 @@ def main(app):
                 img_file=ex.pngfilename,
                 api_name=ex.apiname,
             )
+            if backend == "matplotlib":
+                with open(op.join(target_dir, ex.pyfilename), "a") as f:
+                    f.write("plt.show()")
```
percygautam: I used this workaround to remove matplotlib popups. Is this okay?

doc/sphinxext/gallery_generator.py
OriolAbril: Do either of `import arviz` and then use `arviz.rc_context` or `from arviz import rc_context` work?

doc/sphinxext/gallery_generator.py
OriolAbril: I think it would be better to use something like [this](https://github.com/arviz-devs/arviz/blob/dfd353d69a9b9eed4a871c5f3d2d37255da996ed/doc/sphinxext/gallery_generator.py#L307) we tried this with bokeh initially. The generated website will be the same, but people having cloned the repo will be able to run `python example.py` and get a plot as plt.show() will be in the file

doc/sphinxext/gallery_generator.py
percygautam: Sorry for the miscommunication, actually `from ....arviz/rcparams import rc_context` was showing the error. `from arviz.rcparams import rc_context` is working fine.

doc/sphinxext/gallery_generator.py
percygautam: I have done the changes.

PR #1047, opened 2020-02-05
Add quantile feature to forestplot ridgeplot

### LINKED ISSUE #756: plot_forest extension to allow plot_kde with quantiles
Hi,

Thanks for your great work on this. I am happy to help in the future. I think a great addition to the plot_forest function using ridge plot would be to allow plotting KDEs with quantiles indicated. 

With any simple trace, this code would get at it but is not polished:

```
import arviz as az
import seaborn as sns

fig, axs = plt.subplots(nrows=len(trace.varnames), sharex=True, sharey=True, figsize=(5,9))
for ax, n in zip(axs,trace.varnames) :

    az.plot_kde(trace.get_values(n), quantiles=[0.25, 0.5, 0.75], ax=ax, contour=False)
    ax.set_ylabel(n, rotation=0)
    ax.set_yticks([])
 
sns.despine(left=True, right=True, bottom=False, top=True)

```

percygautam:
## Description

Fixes #756 . Add new feature quantiles to ridgeplot of forestplot.
## Checklist

      PR format?


percygautam:
Currently implemented for `matplotlib`. It looks something like:
Screenshot 2020-02-06 at 1 06 16 AM

There is a problem of having step size too small in plots. `ax.fill_between` will inverse fill between `x[i]` and `x[i+1]`, this cause the quantile line to be very thin. (almost disappear in more dense step sized plots like `defs Italy`). How can we clearly show these quantiles?

aloctavodia:
As a workaround you can stretch the figure along the x-axis, something like `figsize=(14, 4)`. Another solution is to re-think how we display quantiles for kdes. Maybe instead of not filling the area we could use two alternating shades (using two different alpha values) ¿?

percygautam:
@aloctavodia I tinkered a little with figsize generation, and below size seems okayish:
```
    if figsize is None:
        if ridgeplot_quantiles is None:
            figsize = (min(12, sum(width_ratios) * 2), plot_handler.fig_height())
        else:
            figsize = (min(12, sum(width_ratios) * 5), plot_handler.fig_height() * 1.2)
```
Graph looks like this :
Screenshot 2020-02-11 at 2 34 41 AM

About alternate shades for quantiles, Should I add that also in this? Should alternate shades also be added to normal kdeplot with quantiles?

aloctavodia:
Not sure, but I guess a good solution is to work on this first and after merge work on a separated PR for the shade feature. For the next PR we may even think on a general function that is called by the kdeplot and the ridgeplot.

aloctavodia:
I do not think the function should change the figsize automatically, the user should be in control of the figsize

percygautam:
Okay, Let's get this one done first. I have changed the figsize for ridgeplot_quantiles, if it is not explicitly defined by the user. It's creating a figure of `(12, 5.2)`. Is this correct or should we specify some different default figsize for quantiles plot?

percygautam:
I have added  the quantile feature to bokeh `ridgeplot`, and changed the default figsize to clearly show the quantiles (for [0.2, 0.8]):
Screenshot 2020-02-12 at 4 42 36 PM

Also, The figsize of normal ridgeplot (without quantiles) is not normalized, making the bokeh figure to look cluttered. Should I change that also?

arviz/plots/backends/bokeh/forestplot.py
```diff
@@ -58,7 +59,10 @@ def plot_forest(
     )
 
     if figsize is None:
-        figsize = (min(12, sum(width_ratios) * 2), plot_handler.fig_height())
```
aloctavodia: ohh, sorry I misread your previous example. This is OK

arviz/plots/backends/matplotlib/forestplot.py
```diff
@@ -484,7 +505,7 @@ def treeplot(self, qlist, credible_interval):
 
     def ridgeplot(self, mult, ridgeplot_kind):
         """Get data for each ridgeplot for the variable."""
-        xvals, yvals, pdfs, colors = [], [], [], []
+        xvals, yvals, pdfs, pdfs_q, colors = [], [], [], [], []
```
ahartikainen: this could be 

    xvals = yvals = pdfs = pdfs_q = colors = []

but it is not that important, so no need to change it

PR #1049, opened 2020-02-06
use get_bins function to compute bins for density and posterior plot

agustinaarroyuelo:
Hello everyone. I have made some changes in density and posterior plot. Bins were not computed using the `get_bins` function. Also `plot_density` did not take the `ax` argument.

Before
density_before
posterior_before

After
density_after
posterior_after


aloctavodia:
LGTM. This should be added to the changelog.

PR #1050, opened 2020-02-06
[WIP] Add plot examples to plot_violin

amukh18:
## Description

Adds plot examples (with and without data transformation) to `plot_violin`

## Checklist

      PR format?
- [ ] Is the code style correct (follows pylint and black guidelines)?


arviz/plots/violinplot.py
```diff
@@ -84,8 +87,31 @@ def plot_violin(
...
+        >>> az.plot_violin(data)
+
+    Show a default violin plot, but with a transformation applied to the data
+
+    .. plot::
+        :context: close-figs
+
+        >>> import arviz as az
+        >>> data = az.load_arviz_data('centered-eight', transform=np.log)
+        >>> az.plot_violin(data)
```
OriolAbril: No need to reimport nor reload the data, variables from the previous examples are kept. Also, the `transform` argument is not present in `load_arviz_data`, it has to be used in `plot_violin` while being careful of applying the transform (log) only to positive values. In this case I would do `az.plot_violin(data, var_names="tau", transfrom=np.log)`

amukh18: I have fixed the mistakes. However, when I originally started working on this pull request, I saw examples in docstrings of other functions reimport and reload data. Shall I modify the docstrings of the other functions?

PR #1063, opened 2020-02-10
Explain sample_stats naming convention

### LINKED ISSUE #1053: Explain naming convention in sample_stats
The section on sample stats group in [InferenceData scheme](https://arviz-devs.github.io/arviz/schema/schema.html#sample-stats) is missing a description of the naming convention and what does every variable represents. 

The bullet points should be extended with variables still missing and all explanations added/improved. After this is done, it may be interesting to have something like [`SUPPORTED_GROUPS`](https://github.com/arviz-devs/arviz/blob/master/arviz/data/inference_data.py#L14-L26) for sample stats variables.

nitishp25:
## Description

fixes #1053

## Checklist

- [x] Does the PR follow [official](https://github.com/arviz-devs/arviz/blob/master/CONTRIBUTING.md#pull-request-checklist) PR format?

### DIFF
--- a/arviz/data/io_pystan.py
+++ b/arviz/data/io_pystan.py
@@ -655,7 +655,6 @@ def get_sample_stats(fit, warmup=False):
         name = re.sub("__$", "", key)
         name = "diverging" if name == "divergent" else name
         data[name] = values
-    data.update({"max_treedepth": False})
 
     data_warmup = OrderedDict()
     if warmup:
@@ -792,7 +791,6 @@ def get_sample_stats_stan3(fit, variables=None, ignore=None):
         name = re.sub("__$", "", key)
         name = "diverging" if name == "divergent" else name
         data[name] = values
-    data.update({"max_treedepth": False})
 
     return data
 
--- a/arviz/data/io_pystan.py
+++ b/arviz/data/io_pystan.py
@@ -655,6 +655,7 @@ def get_sample_stats(fit, warmup=False):
         name = re.sub("__$", "", key)
         name = "diverging" if name == "divergent" else name
         data[name] = values
+    data.update({"max_treedepth": False})
 
     data_warmup = OrderedDict()
     if warmup:
@@ -791,6 +792,7 @@ def get_sample_stats_stan3(fit, variables=None, ignore=None):
         name = re.sub("__$", "", key)
         name = "diverging" if name == "divergent" else name
         data[name] = values
+    data.update({"max_treedepth": False})
 
     return data
 
--- a/doc/source/schema/schema.md
+++ b/doc/source/schema/schema.md
@@ -54,8 +54,10 @@ The name convention used for `sample_stats` variables is the following:
 * `step_size`: The current integration step size.
 * `step_size_nom`: The nominal integration step size. The `step_size` may differ from this, for example if the step size is jittered. Should only be present if `step_size` is also present and it varies between samples (i.e. step size is jittered).
 * `tree_depth`: The number of tree doublings in the balanced binary tree.
-* `n_steps`: The number of leapfrog steps computed.
-* `diverging`: (boolean) Indicates the presence of leapfrog transitions with large energy deviation from starting and subsequent termination of the trajectory.
+* `n_steps`: The number of leapfrog steps computed. It is related to `tree_depth` with `n_steps <=
+  2^tree_dept`.
+* `diverging`: (boolean) Indicates the presence of leapfrog transitions with large energy deviation
+  from starting and subsequent termination of the trajectory. "large" is defined as `max_energy_error` going over a threshold.
 * `energy`: The value of the Hamiltonian energy for the accepted proposal (up to an
 additive constant).
 * `energy_error`: The difference in the Hamiltonian energy between the initial point and
--- a/doc/source/schema/schema.md
+++ b/doc/source/schema/schema.md
@@ -42,14 +42,14 @@ Samples from the posterior distribution p(theta|y).
 ### `sample_stats`
 Information and diagnostics for each `posterior` sample, provided by the inference
 backend. It may vary depending on the algorithm used by the backend (i.e. an affine
-invariant sampler has no energy associated). Therefore none of this parameters
+invariant sampler has no energy associated). Therefore none of these parameters
 should be assumed to be present in the `sample_stats` group. The convention
 below serves to ensure that _if_ a variable is present with one of these names
 it will correspond to the definition included here.
 
 The name convention used for `sample_stats` variables is the following:
 
-* `lp`: The log posterior density (up to an additive constant).
+* `lp`: The joint log posterior density for the model (up to an additive constant).
 * `acceptance_rate`: The average acceptance probabilities of all possible samples in the proposed tree.
 * `step_size`: The current integration step size.
 * `step_size_nom`: The nominal integration step size. The `step_size` may differ from this, for example if the step size is jittered. Should only be present if `step_size` is also present and it varies between samples (i.e. step size is jittered).
--- a/doc/source/schema/schema.md
+++ b/doc/source/schema/schema.md
@@ -40,20 +40,28 @@ Moreover, each group contains the following attributes:
 Samples from the posterior distribution p(theta|y).
 
 ### `sample_stats`
-Information and diagnostics for each `posterior` sample, provided by the inference backend. It may vary depending on the algorithm used by the backend (i.e. an affine invariant sampler has no energy associated). The name convention used for `sample_stats` variables is the following:
-* `lp`: (unnormalized) log probability for sample
-* `step_size`: the step size used for this sample
-* `step_size_bar`: the current best known step-size
-* `tune`: boolean variable indicating if the sampler is tuning or sampling
-* `depth`: HMC-NUTS only, the depth of the tree that was used to generate this sample
-* `tree_size`: HMC-NUTS only, the number of leafs of the sampling tree, when the sample was accepted
-* `n_leapfrog`: HMC-NUTS only, number of steps selected in each iteration to traverse the posterior
-* `mean_tree_accept`: HMC-NUTS only, the mean acceptance probability for the tree that generated this sample
-* `diverging`: HMC-NUTS only, boolean variable indicating divergent transitions
-* `energy`: HMC-NUTS only, the energy at the point in phase-space where the sample was accepted
-* `energy_error`: HMC-NUTS only, the difference in energy between the start and the end of the trajectory
-* `max_energy_error`: HMC-NUTS only, the maximum difference in energy along the whole trajectory
-* `accept/accept_stat`: acceptance probability for each sampler, when multiple samplers are used
+Information and diagnostics for each `posterior` sample, provided by the inference
+backend. It may vary depending on the algorithm used by the backend (i.e. an affine
+invariant sampler has no energy associated). Therefore none of this parameters
+should be assumed to be present in the `sample_stats` group. The convention
+below serves to ensure that _if_ a variable is present with one of these names
+it will correspond to the definition included here.
+
+The name convention used for `sample_stats` variables is the following:
+
+* `lp`: The log posterior density (up to an additive constant).
+* `acceptance_rate`: The average acceptance probabilities of all possible samples in the proposed tree.
+* `step_size`: The current integration step size.
+* `step_size_nom`: The nominal integration step size. The `step_size` may differ from this, for example if the step size is jittered. Should only be present if `step_size` is also present and it varies between samples (i.e. step size is jittered).
+* `tree_depth`: The number of tree doublings in the balanced binary tree.
+* `n_steps`: The number of leapfrog steps computed.
+* `diverging`: (boolean) Indicates the presence of leapfrog transitions with large energy deviation from starting and subsequent termination of the trajectory.
+* `energy`: The value of the Hamiltonian energy for the accepted proposal (up to an
+additive constant).
+* `energy_error`: The difference in the Hamiltonian energy between the initial point and
+the accepted proposal.
+* `max_energy_error`: The maximum absolute difference in Hamiltonian energy between the initial point and all possible samples in the proposed tree.
+
 
 ### `log_likelihood`
 Pointwise log likelihood data. Samples should match with `posterior` ones and its variables
--- a/doc/source/schema/schema.md
+++ b/doc/source/schema/schema.md
@@ -45,10 +45,10 @@ Information and diagnostics for each `posterior` sample, provided by the inferen
 * `step_size`: the step size used for this sample
 * `step_size_bar`: the current best known step-size
 * `tune`: boolean variable indicating if the sampler is tuning or sampling
-* `depth`: the depth of the tree that was used to generate this sample
-* `tree_size`: the number of leafs of the sampling tree, when the sample was accepted
-* `n_leapfrog`: number of steps selected in each iteration to traverse the posterior
-* `mean_tree_accept`: the mean acceptance probability for the tree that generated this sample
+* `depth`: HMC-NUTS only, the depth of the tree that was used to generate this sample
+* `tree_size`: HMC-NUTS only, the number of leafs of the sampling tree, when the sample was accepted
+* `n_leapfrog`: HMC-NUTS only, number of steps selected in each iteration to traverse the posterior
+* `mean_tree_accept`: HMC-NUTS only, the mean acceptance probability for the tree that generated this sample
 * `diverging`: HMC-NUTS only, boolean variable indicating divergent transitions
 * `energy`: HMC-NUTS only, the energy at the point in phase-space where the sample was accepted
 * `energy_error`: HMC-NUTS only, the difference in energy between the start and the end of the trajectory
--- a/doc/source/schema/schema.md
+++ b/doc/source/schema/schema.md
@@ -42,16 +42,18 @@ Samples from the posterior distribution p(theta|y).
 ### `sample_stats`
 Information and diagnostics for each `posterior` sample, provided by the inference backend. It may vary depending on the algorithm used by the backend (i.e. an affine invariant sampler has no energy associated). The name convention used for `sample_stats` variables is the following:
 * `lp`: (unnormalized) log probability for sample
-* `step_size`
-* `step_size_bar`
+* `step_size`: the step size used for this sample
+* `step_size_bar`: the current best known step-size
 * `tune`: boolean variable indicating if the sampler is tuning or sampling
-* `depth`:
-* `tree_size`:
-* `mean_tree_accept`:
+* `depth`: the depth of the tree that was used to generate this sample
+* `tree_size`: the number of leafs of the sampling tree, when the sample was accepted
+* `n_leapfrog`: number of steps selected in each iteration to traverse the posterior
+* `mean_tree_accept`: the mean acceptance probability for the tree that generated this sample
 * `diverging`: HMC-NUTS only, boolean variable indicating divergent transitions
-* `energy`: HMC-NUTS only
-* `energy_error`
-* `max_energy_error`
+* `energy`: HMC-NUTS only, the energy at the point in phase-space where the sample was accepted
+* `energy_error`: HMC-NUTS only, the difference in energy between the start and the end of the trajectory
+* `max_energy_error`: HMC-NUTS only, the maximum difference in energy along the whole trajectory
+* `accept/accept_stat`: acceptance probability for each sampler, when multiple samplers are used
 
 ### `log_likelihood`
 Pointwise log likelihood data. Samples should match with `posterior` ones and its variables
--- a/doc/source/schema/schema.md
+++ b/doc/source/schema/schema.md
@@ -63,6 +63,7 @@ additive constant).
 * `energy_error`: The difference in the Hamiltonian energy between the initial point and
 the accepted proposal.
 * `max_energy_error`: The maximum absolute difference in Hamiltonian energy between the initial point and all possible samples in the proposed tree.
+* `int_time`: The total integration time (static HMC sampler)
 
 
 ### `log_likelihood`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
   for markdown/notebook parsing in docs ([1406](https://github.com/arviz-devs/arviz/pull/1406))
 * Incorporated `input_core_dims` in `hdi` and `plot_hdi` docstrings ([1410](https://github.com/arviz-devs/arviz/pull/1410))
 * Add documentation pages about experimental `SamplingWrapper`s usage ([1373](https://github.com/arviz-devs/arviz/pull/1373))
+* Add `sample_stats` naming convention to the InferenceData schema ([1063](https://github.com/arviz-devs/arviz/pull/1063))
 
 ### Experimental
 * Modified `SamplingWrapper` base API ([1373](https://github.com/arviz-devs/arviz/pull/1373))


nitishp25:
@OriolAbril I have updated the descriptions and added a few missing variables. Are there any more variables to be added?

doc/schema/schema.md
```diff
@@ -68,16 +68,18 @@ Samples from the posterior distribution p(theta|y).
...
+* `step_size`: the step size used for this sample
+* `step_size_bar`: the current best known step-size
 * `tune`: boolean variable indicating if the sampler is tuning or sampling
-* `depth`:
-* `tree_size`:
-* `mean_tree_accept`:
+* `depth`: the depth of the tree that was used to generate this sample
+* `tree_size`: the number of leafs of the sampling tree, when the sample was accepted
+* `n_leapfrog`: number of steps selected in each iteration to traverse the posterior
+* `mean_tree_accept`: the mean acceptance probability for the tree that generated this sample
```
OriolAbril: I think this are also HMC/NUTS only

doc/schema/schema.md
```diff
@@ -68,16 +68,18 @@ Samples from the posterior distribution p(theta|y).
...
 * `diverging`: HMC-NUTS only, boolean variable indicating divergent transitions
-* `energy`: HMC-NUTS only
-* `energy_error`
-* `max_energy_error`
+* `energy`: HMC-NUTS only, the energy at the point in phase-space where the sample was accepted
+* `energy_error`: HMC-NUTS only, the difference in energy between the start and the end of the trajectory
+* `max_energy_error`: HMC-NUTS only, the maximum difference in energy along the whole trajectory
+* `accept/accept_stat`: acceptance probability for each sampler, when multiple samplers are used
```
OriolAbril: We should stick to a single name, is this because from_pymc uses one name and from_pystan another?

nitishp25: Yes, it'll depend on the library used

ahartikainen: I think currently only real `sample_stats` parameter is `divergent`. Rest of them are still optional. I would not start renaming optional parameters until we use them for something.

OriolAbril: Yep, I also wanted to use this opportunity to add `max_tree_depth` (boolean) or something similar in order to be able to reproduce plots from https://arxiv.org/abs/1903.08008 (e.g. figures 3 or 4)

nitishp25: Should ```depth``` and ```treedepth``` be renamed to ```max_tree_depth``` then?

OriolAbril: No, this would be a new variable, it would have to be boolean like `diverging`. One of the possible arguments to pass to NUTS is the maximum tree depth; many samples reaching this maximum depth can be an indicator of sampling or modeling issues.

arviz/data/io_pystan.py
```diff
@@ -571,6 +571,7 @@ def get_sample_stats(fit):
         name = re.sub("__$", "", key)
         name = "diverging" if name == "divergent" else name
         data[name] = values
+    data.update({"max_treedepth": False})
```
ahartikainen: I don't think this is the correct thing to do.

I would not touch `io_` functions in schema PR.

nitishp25: Ohh, should I make a separate PR for this? Please also suggest the correct way to add this variable.

OriolAbril: It should be a separate PR after we have discussed what variables are we interested in adding to `sample_stats`, there is no rush.

nitishp25: Alright! Are there any more changes to be made in this naming convention?

OriolAbril:
I was thinking about it and I think we should either rename it (preferably) or not have it in sample stats. My main concerns are: as I understand it, one of the goals of ArviZ is also to ease the comparison between different inference libraries, we do not use it for now, but our users may use it and accessing a value in an inference data should not depend on the inference library; the second concern is that we may have one use for it at some point (I can't think of nothing else than including the acceptance prob warning in our summary right now but who knows) the naming differences may block or delay the new feature. This second concern is why I prefer renaming over removing it.

nitishp25:
I see. I will rename it once you confirm @OriolAbril @ahartikainen

sethaxen:
I also support renaming.

ahartikainen:
For Stan the outputs for sample_stats are

| original name | current name | unified name | meaning | source |
| ---------------|----------------|-----------------|-----------|---------|
| lp__ | lp |  | the log posterior density (up to a constant) <sup>1</sup> | [src1](https://mc-stan.org/docs/2_22/reference-manual/notation-for-samples-chains-and-draws.html)
| accept_stat__ | accept_stat | | the average acceptance probabilities of all possible samples in the proposed tree. | [src2](https://mc-stan.org/bayesplot/reference/MCMC-nuts.html#quick-definitions) |
| stepsize__  | stepsize | | the step size used by NUTS in its Hamiltonian simulation. | [src2](https://mc-stan.org/bayesplot/reference/MCMC-nuts.html#quick-definitions) |
| treedepth__ | treedepth | | the depth of tree used by NUTS, which is the log (base 2) of the number of leapfrog steps taken during the Hamiltonian simulation. | [src2](https://mc-stan.org/bayesplot/reference/MCMC-nuts.html#quick-definitions) |
| n_leapfrog__ | n_leapfrog | | the actual number of leapfrog steps computed | [src3](https://mc-stan.org/docs/2_22/reference-manual/hmc-algorithm-parameters.html) | 
| divergent__ | diverging | | the number of leapfrog transitions with diverging error. Because NUTS terminates at the first divergence this will be either 0 or 1 for each iteration. | [src2](https://mc-stan.org/bayesplot/reference/MCMC-nuts.html#quick-definitions) |
| energy__ | energy | | the value of the Hamiltonian (up to an additive constant) at each iteration. | [src2](https://mc-stan.org/bayesplot/reference/MCMC-nuts.html#quick-definitions) |

<sup>1. The lp__ value also represents the potential energy in the Hamiltonian system and is rate bounded by
the randomly supplied kinetic energy each iteration, which follows a Chi-square distribution in the number
of parameters.</sup>

ahartikainen:
Then there is also other settings

see for example for CmdStan output

# stan_version_major = 2
# stan_version_minor = 18
# stan_version_patch = 0
# model = eight_schools_nc_model
# method = sample (Default)
#   sample
#     num_samples = 100
#     num_warmup = 1000 (Default)
#     save_warmup = 0 (Default)
#     thin = 1 (Default)
#     adapt
#       engaged = 1 (Default)
#       gamma = 0.050000000000000003 (Default)
#       delta = 0.80000000000000004 (Default)
#       kappa = 0.75 (Default)
#       t0 = 10 (Default)
#       init_buffer = 75 (Default)
#       term_buffer = 50 (Default)
#       window = 25 (Default)
#     algorithm = hmc (Default)
#       hmc
#         engine = nuts (Default)
#           nuts
#             max_depth = 10 (Default)
#         metric = diag_e (Default)
#         metric_file =  (Default)
#         stepsize = 1 (Default)
#         stepsize_jitter = 0 (Default)
# id = 0 (Default)
# data
#   file = eight_schools.data.R
# init = 2 (Default)
# random
#   seed = 779839997
# output
#   file = eight_schools_output1.csv
#   diagnostic_file =  (Default)
#   refresh = 100 (Default)

These are also probably needed, and also for other libs too.

sethaxen:
> Then there is also other settings
> 
> see for example for CmdStan output

These would probably go under attributes though, right?

ahartikainen:
Yes, I would put them under attributes.

Maybe under `sampler_settings` or something similar.

sethaxen:
> Yes, I would put them under attributes.
> 
> Maybe under `sampler_settings` or something similar.

That sounds good. Perhaps that should be a separate issue.

For Turing.jl, we have the following sample stats for HMC:
- `acceptance_rate`: MH stats, i.e. sum of MH accept prob for all leapfrog steps ([src](https://github.com/TuringLang/AdvancedHMC.jl/blob/c13d1b111fde94a9c4ac34ffd88cf31fbaac7428/src/trajectory.jl#L412))
- `hamiltonian_energy`: value of the hamiltonian energy for the accepted proposal to within an additive constant
- `hamiltonian_energy_error`: difference in the hamiltonian energy between the initial point and the proposed point,
- `is_adapt`: boolean, whether the current sample is part of adaptation
- `max_hamiltonian_energy_error`: energy in tree with largest absolute difference from initial energy ([src](https://github.com/TuringLang/AdvancedHMC.jl/blob/c13d1b111fde94a9c4ac34ffd88cf31fbaac7428/src/trajectory.jl#L414))
- `n_steps`: total # of leap frog steps, i.e. phase points in a trajectory ([src](https://github.com/TuringLang/AdvancedHMC.jl/blob/c13d1b111fde94a9c4ac34ffd88cf31fbaac7428/src/trajectory.jl#L413))
- `numerical_error`: termination due to large energy deviation from starting (possibly numerical errors) ([src](https://github.com/TuringLang/AdvancedHMC.jl/blob/c13d1b111fde94a9c4ac34ffd88cf31fbaac7428/src/trajectory.jl#L376))
- `lp`/`log_density`: log probability to within an additive constant
- `tree_depth`: the number of tree doublings in the balanced binary tree
- `step_size `: current integration step size ([src](https://github.com/TuringLang/AdvancedHMC.jl/blob/514b33d170f95d17194433bfdc00e65e035b5bb9/src/integrator.jl#L28))
- `nom_step_size`: Get the nominal integration step size. The current integration step size may differ from this, for example if the step size is jittered. Nominal step size is usually used in adaptation. ([src](https://github.com/TuringLang/AdvancedHMC.jl/blob/514b33d170f95d17194433bfdc00e65e035b5bb9/src/integrator.jl#L17))

For SMC:
- `le`: The log evidence retrieved from the particle
- `weight`: The weight of the particle the sample was retrieved from.

Others that are supposedly parameters but I've never seen used:
- `elapsed`
- `eval_num`
- `lf_eps`

OriolAbril:
I'll try to summarize the information gathered grouping equivalent sampler stats from different libraries. Descriptions and sources should still be checked in the original comment. Feel free to edit (I think members should have permission).

## HMC 

| Stan | Turing.jl | PyMC3 | Pyro | NumPyro | Unified name | 
| ---------------|----------------|-----------------|-----------|---------|-------|
| lp__ | lp/log_density | model_logp | - | potential_energy | lp |
| accept_stat__ | acceptance_rate | mean_tree_accept | acceptance rate | accept_prob | |
| stepsize__  | step_size | step_size | - | adapt_state.step_size | |
| - | nom_step_size | - | - | | | 
| - | - | step_size_bar | - | | |
| treedepth__ | tree_depth | depth | - | | |
| n_leapfrog__ | n_steps | tree_size | - | num_steps | |
| divergent__ | numerical_error | diverging | divergences | | diverging |
| energy__ | hamiltonian_energy | energy | - | | energy |
| - | hamiltonian_energy_error | energy_error | - | | |
| - | max_hamiltonian_energy_error | max_energy_error | - | | |
| - | is_adapt | tune | - | | |

@fehiepsi could you please check Pyro and NumPyro names? I think I am on the right track but not completely sure.

I don't know enough about tfp as to include anything here. Is there anybody we could tag that comes to mind?

## SMC
any thoughts @aloctavodia ?

| Turing.jl | PyMC3 | Unified name |
|----|----|----|
| le | | |
| weight | | |

## MH?

Notes:

I think the only sampler stats currently used (in plotting or in stats) are `diverging` and `energy`.

PyMC3 reference: [src1](https://docs.pymc.io/notebooks/sampler-stats.html), 
Pyro reference: [src](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/mcmc/hmc.py#L349)
NumPyro reference: [src](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/mcmc.py#L40)

fehiepsi:
Thank @OriolAbril, they are correct names.

ahartikainen:
What tfp use?

junpenglao:
TFP does not have internal naming convention, as they are function output (tensor or array) and user are free to name it whatever they want - I was manually mapping it eg: https://colab.research.google.com/github/tensorflow/probability/blob/master/tensorflow_probability/examples/jupyter_notebooks/Modeling_with_JointDistribution.ipynb#scrollTo=4qQdOPk90f7t

ahartikainen:
Stan hmc has these (I need verify)

    accept_stat__ stepsize__ int_time__

aloctavodia:
Currently PyMC3's SMC does not return any statistics, but I should fix that.

OriolAbril:
I updated my previous comment with the table summary to try to restart the discussion. I tried to use the names that felt had more consensus (e.g. more than one library had similar or equal names), please comment, I actually don't know the reasons behind any of the naming conventions chosen by each library. 

The one I am having trouble with `step_size` related parameters. After warmup (whose sample stats should be stored in warmup groups, see #1126), `step_size` should have converged to a given value, therefore we are basically storing repeated values unless the step size is jittered (which seems to only be available in Turing). It feels unnecessary to store `step_size` in an `(nchains, ndraws)` array, but given the jitter possibility it may be the simplest way to accommodate all the libraries into a single naming convention (with `nom_step_size` generally missing). 

Regarding `step_size_bar`, I thought it could be removed from `sample_stats`: PyMC3 docs say that after tuning `step_size` is set to `step_size_bar` but it does not seem to be the case. All stored ArviZ inferencedata objects have different values for `step_size` and `step_size_bar` (each of them is constant for all draws but they are different between them).

nitishp25:
Btw, the renaming work is to be done in this PR itself right? Or leave this one only for the descriptions of the existing `sample_stats`?

canyon289:

OriolAbril:
This one depends on core contributors having some consensus on the names to use. Maybe a lab meeting could help in finishing this?

canyon289:
Sounds good. Ill add it as topic for next lab meeting

doc/schema/schema.md
```diff
@@ -68,16 +68,18 @@ Samples from the posterior distribution p(theta|y).
 ### `sample_stats`
 Information and diagnostics for each `posterior` sample, provided by the inference backend. It may vary depending on the algorithm used by the backend (i.e. an affine invariant sampler has no energy associated). The name convention used for `sample_stats` variables is the following:
 * `lp`: (unnormalized) log probability for sample
-* `step_size`
-* `step_size_bar`
+* `step_size`: the step size used for this sample
```
sethaxen, line 71: Can you clarify whether this is the jittered or unjittered step size (where jittering is done)?

doc/schema/schema.md
```diff
@@ -68,16 +68,18 @@ Samples from the posterior distribution p(theta|y).
...
+* `step_size_bar`: the current best known step-size
```
sethaxen, line 72: Best by what criterion?

doc/schema/schema.md
```diff
@@ -68,16 +68,18 @@ Samples from the posterior distribution p(theta|y).
...
 * `tune`: boolean variable indicating if the sampler is tuning or sampling
-* `depth`:
-* `tree_size`:
-* `mean_tree_accept`:
+* `depth`: HMC-NUTS only, the depth of the tree that was used to generate this sample
+* `tree_size`: HMC-NUTS only, the number of leafs of the sampling tree, when the sample was accepted
+* `n_leapfrog`: HMC-NUTS only, number of steps selected in each iteration to traverse the posterior
```
sethaxen, line 76: In Turing at least `n_leapfrog` will be collected even if plain old HMC is used instead of NUTS. I think Stan is the same way? Why must this be NUTS-only?

sethaxen, line 76: Could you elaborate on the relationship between these? I know in the past when I ran this done, the value of one bounded the others (e.g. I believe `2^depth` bounds `tree_size` and `n_leapfrog`, but how does `tree_size` relate to `n_leapfrog`?) and it would be good while defining these to explain those constraints.

doc/schema/schema.md
```diff
@@ -68,16 +68,18 @@ Samples from the posterior distribution p(theta|y).
...
+* `mean_tree_accept`: HMC-NUTS only, the mean acceptance probability for the tree that generated this sample
 * `diverging`: HMC-NUTS only, boolean variable indicating divergent transitions
-* `energy`: HMC-NUTS only
-* `energy_error`
-* `max_energy_error`
+* `energy`: HMC-NUTS only, the energy at the point in phase-space where the sample was accepted
```
sethaxen, line 79: Is it where the sample was accepted or proposed? e.g. if a transition was diverging, the initial point will be returned, but which energy is usually returned?

doc/schema/schema.md
```diff
@@ -68,16 +68,18 @@ Samples from the posterior distribution p(theta|y).
...
+* `energy_error`: HMC-NUTS only, the difference in energy between the start and the end of the trajectory
```
sethaxen, line 80: Could you clarify what "end" means here?

doc/schema/schema.md
```diff
@@ -68,16 +68,18 @@ Samples from the posterior distribution p(theta|y).
...
+* `max_energy_error`: HMC-NUTS only, the maximum difference in energy along the whole trajectory
```
sethaxen, line 81: Is it the maximum difference in energy or the maximum difference between the energy of the initial point and all energies along the trajectory?

doc/schema/schema.md
```diff
@@ -68,16 +68,18 @@ Samples from the posterior distribution p(theta|y).
...
+* `step_size`: the step size used for this sample
+* `step_size_bar`: the current best known step-size
 * `tune`: boolean variable indicating if the sampler is tuning or sampling
-* `depth`:
-* `tree_size`:
-* `mean_tree_accept`:
+* `depth`: HMC-NUTS only, the depth of the tree that was used to generate this sample
+* `tree_size`: HMC-NUTS only, the number of leafs of the sampling tree, when the sample was accepted
+* `n_leapfrog`: HMC-NUTS only, number of steps selected in each iteration to traverse the posterior
+* `mean_tree_accept`: HMC-NUTS only, the mean acceptance probability for the tree that generated this sample
```
sethaxen, line 77: I believe this is being replaced by `acceptance_rate`, but so is `accept_stat` below?

doc/schema/schema.md
sethaxen: e.g. for Turing this is _jittered_ step size.

doc/schema/schema.md
ahartikainen: I think in Stan initial point is not returned when transition is diverging. Point along the path between the initial and diverging point is selected.

OriolAbril:
We have agreed on the names, see table in [previous comment](https://github.com/arviz-devs/arviz/pull/1063#issuecomment-589771200), but we still need to update the PR to match said table and add the definitions. There are some definitions in previous comments when describing Stan and Turing sample stats.

ahartikainen:
Explicit name is better, then let's add description somewhere what is what.

E.g. attrs field could contain something?

OriolAbril:
I think having that on the schema plus maybe advertising it a little more once its more definitive is enough, no need to add in attrs

doc/schema/schema.md
```diff
@@ -66,18 +66,28 @@ Moreover, each group contains the following attributes:
...
+it will correspond to the definition included here.
+
+The name convention used for `sample_stats` variables is the following:
+
+* `lp`: The log posterior density (up to an additive constant).
+* `acceptance_rate`: The average acceptance probabilities of all possible samples in the proposed tree.
+* `step_size`: The current integration step size.
+* `step_size_nom`: The nominal integration step size. The `step_size` may differ from this, for example if the step size is jittered. Should only be present if `step_size` is also present and it varies between samples (i.e. step size is jittered).
+* `tree_depth`: The number of tree doublings in the balanced binary tree.
+* `n_steps`: The number of leapfrog steps computed.
```
OriolAbril, line 83: I am not not completely sure about the equivalencies in the table above. Not sure if the relation should be `n_steps = 2^tree_depth` or `n_steps <= 2^tree_dept`. 

So far I think that in Stan the relation is `=`, in PyMC3 it's `<=`, no idea in Turing, in NumPyro only `n_steps` exists and `tree_depth` is calculated manually in `io_numpyro` so the relation is `=` and none of the two exist in Pyro.
@

doc/schema/schema.md
```diff
@@ -66,18 +66,28 @@ Moreover, each group contains the following attributes:
...
+* `diverging`: (boolean) Indicates the presence of leapfrog transitions with large energy deviation from starting and subsequent termination of the trajectory.
+* `energy`: The value of the Hamiltonian energy for the accepted proposal (up to an
+additive constant).
+* `energy_error`: The difference in the Hamiltonian energy between the initial point and
+the accepted proposal.
+* `max_energy_error`: The maximum absolute difference in Hamiltonian energy between the initial point and all possible samples in the proposed tree.
```
OriolAbril, line 89: @sethaxen I think the definition will be clear now but wanted to be extra sure about it and also about the fact that Turing stores the difference and not the energy with the max difference

doc/schema/schema.md
```diff
@@ -66,18 +66,28 @@ Moreover, each group contains the following attributes:
...
+
+* `lp`: The log posterior density (up to an additive constant).
+* `acceptance_rate`: The average acceptance probabilities of all possible samples in the proposed tree.
+* `step_size`: The current integration step size.
+* `step_size_nom`: The nominal integration step size. The `step_size` may differ from this, for example if the step size is jittered. Should only be present if `step_size` is also present and it varies between samples (i.e. step size is jittered).
+* `tree_depth`: The number of tree doublings in the balanced binary tree.
+* `n_steps`: The number of leapfrog steps computed.
+* `diverging`: (boolean) Indicates the presence of leapfrog transitions with large energy deviation from starting and subsequent termination of the trajectory.
+* `energy`: The value of the Hamiltonian energy for the accepted proposal (up to an
+additive constant).
```
OriolAbril, line 86: I think so far this is correct and is true for all libraries, however I have no idea about what should it say (if anything) about what is `energy` when there is a diverging transition. Not sure if this propagates to `energy_error` or if defining what does `accepted proposal` means when diverging is enough.

OriolAbril (commented):
I would like to merge this soon, it has been stale long enough, so if the discussion about definitions takes more than 1-2 weeks I'll merge and we can fix the definitions later.

nitishp25:
Don't we also have to rename these variables in the idata converters of these libraries so that the created `sample_stats` group contains the updated names?

OriolAbril:
> Don't we also have to rename these variables in the idata converters of these libraries so that the created sample_stats group contains the updated names?

Yes, but this will come later in follow-up PRs

nitishp25:
Ohh, so will you be working on this PR now?

doc/schema/schema.md
sethaxen: Should be straightforward to empirically verify. I'll do so for Turing.

doc/schema/schema.md
sethaxen: I had thought that if a transition was divergent then the entire transition was rejected, i.e. the returned sample was always the starting point. In an earlier comment I think @ahartikainen said that's not what Stan at least does. I've been meaning to look into this.

doc/schema/schema.md
sethaxen: So Turing seems to have `n_steps <= 2^(tree_depth + 1)`.

doc/schema/schema.md
sethaxen: Yes, this is correct. In fact, `diverging` and `max_energy_error` should be closely related. In almost all cases, `max_energy_error` exceeding a threshold defines a `diverging` transition (almost all because negative error of the magnitude of the threshold wouldn't trigger `diverging` but is probably much rarer).

doc/schema/schema.md
OriolAbril: Empirically confirmed that PyMC3 has `n_steps < 2^(tree_depth)` and Stan `n_steps < 2^(tree_depth + 1)`. 

Moreover, in both cases `n_steps` is always a power of 2 minus one: 3, 7, 15...

doc/schema/schema.md
sethaxen: Yeah @ahartikainen was totally right about this. When a divergent transition is encountered, the tree doubling is rejected, but the already-accepted tree is not.

OriolAbril (approved):

doc/source/schema/schema.md
```diff
@@ -40,18 +40,30 @@ Moreover, each group contains the following attributes:
...
-* `depth`:
-* `tree_size`:
-* `mean_tree_accept`:
-* `diverging`: HMC-NUTS only, boolean variable indicating divergent transitions
-* `energy`: HMC-NUTS only
-* `energy_error`
-* `max_energy_error`
+Information and diagnostics for each `posterior` sample, provided by the inference
+backend. It may vary depending on the algorithm used by the backend (i.e. an affine
+invariant sampler has no energy associated). Therefore none of this parameters
```
ahartikainen, line 45:
```suggestion
invariant sampler has no energy associated). Therefore none of these parameters
```

doc/source/schema/schema.md
```diff
@@ -40,18 +40,30 @@ Moreover, each group contains the following attributes:
...
+should be assumed to be present in the `sample_stats` group. The convention
+below serves to ensure that _if_ a variable is present with one of these names
+it will correspond to the definition included here.
+
+The name convention used for `sample_stats` variables is the following:
+
+* `lp`: The log posterior density (up to an additive constant).
```
ahartikainen, line 52:
```suggestion
* `lp`: The joint log posterior density for the model (up to an additive constant).
```

I'm not sure if this will make it more clear

ahartikainen (approved):
Do we want to mention `sample_stats_prior` which is analogous to `sample_stats` but the 1to1 correspondence is with the `prior` and not with the `posterior`

ahartikainen:
Oh, did we miss the `int_time__` --> https://mc-stan.org/docs/2_25/cmdstan-guide/mcmc-intro.html

    int_time__ - total integration time (static HMC sampler)

Also lp__ is mentioned as `the total log probability density (up to an additive constant) at each sample`

OriolAbril:
> Oh, did we miss the `int_time__` 

Looks like it. `int_time` looks like a good name, unless someone wants to propose an alternative.

almostmeenal:
> I'll try to summarize the information gathered grouping equivalent sampler stats from different libraries. Descriptions and sources should still be checked in the original comment. Feel free to edit (I think members should have permission).
> 
> ## HMC
> Stan	Turing.jl	PyMC3	Pyro	NumPyro	Unified name
> lp__	lp/log_density	model_logp	-	potential_energy	lp
> accept_stat__	acceptance_rate	mean_tree_accept	acceptance rate	accept_prob	acceptance_rate
> stepsize__	step_size	step_size	-	adapt_state.step_size	step_size
> -	nom_step_size	-	-		
> -	-	step_size_bar	-		
> treedepth__	tree_depth	depth	-		tree_depth
> n_leapfrog__	n_steps	tree_size	-	num_steps	n_steps
> divergent__	numerical_error	diverging	divergences	diverging	diverging
> energy__	hamiltonian_energy	energy	-	energy	energy
> -	hamiltonian_energy_error	energy_error	-		energy_error
> -	max_hamiltonian_energy_error	max_energy_error	-		max_energy_error
> -	~is_adapt~	~tune~	-		_removed_ (see #1126)
> @fehiepsi could you please check Pyro and NumPyro names? I think I am on the right track but not completely sure.
> 
> I don't know enough about tfp as to include anything here. Is there anybody we could tag that comes to mind?
> 
> ## SMC
> any thoughts @aloctavodia ?
> 
> Turing.jl	PyMC3	Unified name
> le		
> weight		
> ## MH?
> Notes:
> 
> I think the only sampler stats currently used (in plotting or in stats) are `diverging` and `energy`.
> 
> PyMC3 reference: [src1](https://docs.pymc.io/notebooks/sampler-stats.html), (MH related, not really sure they are relevant: [src2](https://github.com/pymc-devs/pymc3/blob/master/pymc3/step_methods/metropolis.py#L197), [src3](https://github.com/pymc-devs/pymc3/blob/master/pymc3/step_methods/metropolis.py#L305), [src4](https://github.com/pymc-devs/pymc3/blob/master/pymc3/step_methods/metropolis.py#L631))
> Pyro reference: [src](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/mcmc/hmc.py#L349)
> NumPyro reference: [src](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/mcmc.py#L40)

Maybe I missed something, but I saw

`perf_counter_diff `, `process_time_diff `and `perf_counter_start `in `trace.sample_stats`

OriolAbril:
These were added after this comment was written, let's continue this in https://github.com/pymc-devs/pymc-examples/pull/95 better

PR #1070, opened 2020-02-16
Traceplot legend

### LINKED ISSUE #837: plot_trace with `combined=True` to have alternating colors and labels
## Tell us about it

I'm very new at this, so not sure how much this problem is relevant for other users.
When using `plot_trace` with `combined=True`:
```python
import matplotlib as plt
import arviz as az
ax = az.plot_trace(data, var_names=('theta_t', 'theta'), combined=True, compact=True, figsize = (20, 8))
```
the lines come up in the same color and without labels, so it's hard to differenciate between them
Screenshot from 2019-09-30 17-06-18

It would be great to have colors by default in the lines. And maybe an option to show labels
Something like:
Screenshot from 2019-09-30 17-05-55

## Thoughts on implementation
Alternating colors - I don't see any reason not to have them.
For the labels - I understand that there could be thousands of lines, so maybe have a limit, to show something like the first 5 labels, or maybe add an argument to the method with `num_labels` (and then show "..." to indicate there are more not being shown)

I've been looking at https://arviz-devs.github.io/arviz/_modules/arviz/plots/traceplot.html for a while but it's way above my league. I just don't understand what's going on.

Thanks for reading!

OriolAbril:
## Description

Fixes #837. Allows customization of properties to distinguish lines in the same subplot with all combinations of combine and compact. Introduces arguments `compact_prop` and `chain_prop` with format `(property_name, (value1, value2...))` to set color, linestyle, linewidth...

Matplotlib version is kind of working, only labels are missing, plot format is OK. Bokeh is still more preliminar. It would also be interesting to implement interactive legends in bokeh for traceplot, but I currently do not plan to do this here.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,12 @@
   and `essplot` (#1024)
 * New defaults for cross validation: `loo` (old: waic) and `log` -scale (old: `deviance` -scale) (#1067)
 * **Experimental Feature**: Added `arviz.wrappers` module to allow ArviZ to
-  refit the models if necessary
-* **Experimental Feature**: Added `reloo` function to ArviZ
+  refit the models if necessary (#771)
+* **Experimental Feature**: Added `reloo` function to ArviZ (#771)
 * ArviZ version to InferenceData attributes. (#1086)
-* Add `log_likelihood` argument to `from_pymc3`
+* Add `log_likelihood` argument to `from_pymc3` (#1082)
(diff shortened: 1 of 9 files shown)


OriolAbril:
Current behaviour:

1. combined=False, compact=False (default)

image

2. combined=True, compact=False

image

3. combined=False, compact=True

image

4. combined=True, compact=True

image

How does this look? cc @canyon289 @julian-pani

PR #1072, opened 2020-02-16
Remove hpd redundant copy

bantmen:
## Description
`hpd` makes a redundant copy for the unimodal use case since the input array is not being modified. 

Wrote some code just to double check:
```
a = np.random.random((5000,))
b = a.copy()
hpd(b, 0.9)
assert np.array_equal(a, b)
```

## Checklist


aloctavodia:
thanks @bantmen!

PR #1073, opened 2020-02-17
[WIP]Add de-aliaser function for plots

### LINKED ISSUE #1041: matplotlib kwarg de-aliasing helper function
## Tell us about it

There are many kwarg arguments in matplotlib that accept an alias, which can get quite messy when ArviZ sets default values for them. I propose to pass the kwarg dicts through this helper function and then always work internally with the complete name.

## Thoughts on implementation

Use a matplotlib internal function if it exists? Otherwise we should create one (or several, see note).

Note: some aliases are not universal like the case of color and c in a scatter plot.

References that may (or not) be useful: [Line2D properties](https://matplotlib.org/api/_as_gen/matplotlib.lines.Line2D.html#matplotlib.lines.Line2D) and [Collection properties](https://matplotlib.org/api/collections_api.html#matplotlib.collections.Collection)

percygautam:
## Description

Fixes #1041 . Added function to dealiase the plot kwargs passed to plots by user. Working example can be found in [cookbook](https://github.com/percygautam/arviz-examples/blob/master/de-Aliasing%20helper%20function.ipynb).
## Checklist


percygautam:
Okay, I'd do the changes. We can use a single function for dealiasing.

percygautam:
> I would take advantage of the fact that alias_mapping argument in cbook.normalize_kwargs can be a subclass of Artist objects.

@OriolAbril This functionality is only present in unreleased latest version `3.2.0` and not in version used by arviz `3.1.2`. 
https://github.com/matplotlib/matplotlib/blob/3b2a89825e5e5a98b8b9478c8f103cc22b9b706f/lib/matplotlib/cbook/__init__.py#L1768-L1772

As a workaround I'd do something like this:
```
def dealiser(args): 
    return cbook.normalize_kwargs(args, getattr(mpl.collections.PathCollection,"_alias_map", {}))
```
We can use this workaround until arviz shifts to new version of matplotlib.

OriolAbril:
The workaround sounds good, in this case it is probably better to define a handful of dealiasers (or a dealiaser with some kind argument) to allow dealiasing for both plot and scatter and also hexbin of hist if they were to return a different artist than plot and scatter

percygautam:
I have applied `dealiser` to all the cases I can find. Please check if I missed something.
Also, what should we do in the case of `compareplot`?(we are defining arviz specific kwargs)

percygautam:
I have checked the files and confirmed that only default prop names are used. 
Also, for case we set default by using alias, then kwargs dict will contain both alias prop(set by arviz) and dealiased prop (set by user), for such case full names are given precedence.( ref: https://github.com/matplotlib/matplotlib/blob/3b2a89825e5e5a98b8b9478c8f103cc22b9b706f/lib/matplotlib/cbook/__init__.py#L1722). So, only user-defined property will be used in plots.

Anyway, currently it won't cause any error and we can correct that alias aren't used in future .

arviz/plots/essplot.py
```diff
@@ -270,6 +273,8 @@ def plot_ess(
     kwargs.setdefault("label", kind)
     if hline_kwargs is None:
         hline_kwargs = {}
+    else:
+        hline_kwargs = matplotlib_kwarg_dealiaser(hline_kwargs, "plot")
     hline_kwargs.setdefault("linewidth", hline_kwargs.pop("lw", _linewidth))
     hline_kwargs.setdefault("linestyle", hline_kwargs.pop("ls", "--"))
```
OriolAbril: These can be simplified now, no need to change now though

arviz/plots/mcseplot.py
```diff
@@ -140,6 +141,8 @@ def plot_mcse(
     kwargs.setdefault("zorder", 3)
     if extra_kwargs is None:
         extra_kwargs = {}
+    else:
+        extra_kwargs = matplotlib_kwarg_dealiaser(extra_kwargs, "plot")
     extra_kwargs.setdefault("linestyle", extra_kwargs.pop("ls", "-"))
     extra_kwargs.setdefault("linewidth", extra_kwargs.pop("lw", _linewidth / 2))
```
OriolAbril: Same here, I'm basically marking what I see to have it somewhere

arviz/plots/mcseplot.py
```diff
@@ -149,6 +152,8 @@ def plot_mcse(
         sd_mcse = mcse(data, var_names=var_names, method="sd")
         if text_kwargs is None:
             text_kwargs = {}
+        else:
+            text_kwargs = matplotlib_kwarg_dealiaser(text_kwargs, "text")
         text_x = text_kwargs.pop("x", 1)
         text_kwargs.setdefault("fontsize", text_kwargs.pop("size", xt_labelsize * 0.7))
```
OriolAbril: idem

arviz/plots/pairplot.py
```diff
@@ -131,13 +131,20 @@ def plot_pair(
...
         plot_kwargs = {}
-
-    if kind == "scatter":
-        plot_kwargs.setdefault("marker", ".")
-        plot_kwargs.setdefault("lw", 0)
+    else:
+        if kind == "scatter":
+            plot_kwargs = matplotlib_kwarg_dealiaser(plot_kwargs, "scatter")
+            plot_kwargs.setdefault("marker", ".")
+            plot_kwargs.setdefault("lw", 0)
```
OriolAbril: I'd use linewidth here. For now it works, but I have tried it and it will raise an error in the next release:

```
plt.plot([1,2], [2,0], ls="--", linestyle=":")                                                                                                                                                      
/home/oriol/venvs/arviz-dev/bin/ipython:1: MatplotlibDeprecationWarning: Saw kwargs ['ls', 'linestyle'] which are all aliases for 'linestyle'.  Kept value from 'linestyle'.  Passing multiple aliases for the same property will raise a TypeError in 3.3.
```

percygautam: Okay, I'll do the changes.

percygautam:
Should I add more test cases or are these enough?

percygautam:
The dealiasing function is not working for python 3.5. Any workaround?

OriolAbril:
> The dealiasing function is not working for python 3.5. Any workaround?

matplotlib 3.1 droped support for python 3.5, so CI uses 3.0.3 which probably has different alias. Checking the keys and for which kinds it fails should allow to remove the combination from tests

OriolAbril:
@percygautam I just had an idea. Currently most plots do something similar to:

```
if kwargs is None:
    kwargs = {}
else:
    kwargs = matplotlib_kwarg_dealiaser(kwargs)
```

We could simplify this by making `matplotlb_kwarg_dealiaser` also accept None in which case it would return `{}` straight away and not call matplotlib's kwarg normalizer. All test suite was reorganized, so a rebase/merge from master is needed.

percygautam:
I've done the changes and left `pairplot` as before.

percygautam:
@OriolAbril There are two cases for kwargs, whether it is for matplotlib or it's for bokeh. If an alias(say 'lw') or full name( say 'linewidth') is passed to the normalized function, we'll get the dealiased name or same full name. Say a kwarg is passed that is not the param of matplotlib, it the normalized function will return it the same. 
The error handling for wrong params is done at the backend.
I have modified test to show the desired change. (not valid param get passed on as it is)

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@
   and `essplot` (#1024)
 * New defaults for cross validation: `loo` (old: waic) and `log` -scale (old: `deviance` -scale) (#1067)
 * **Experimental Feature**: Added `arviz.wrappers` module to allow ArviZ to
+  refit the models if necessary
+* **Experimental Feature**: Added `reloo` function to ArviZ
+* Added new helper function `matplotlib_kwarg_dealiaser` (#1073)
   refit the models if necessary (#771)
 * **Experimental Feature**: Added `reloo` function to ArviZ (#771)
```
OriolAbril: There have been some issues with merging

arviz/plots/mcseplot.py
```diff
@@ -134,28 +135,28 @@ def plot_mcse(
     (figsize, ax_labelsize, titlesize, xt_labelsize, _linewidth, _markersize) = _scale_fig_size(
         figsize, textsize, rows, cols
     )
-    kwargs.setdefault("linestyle", kwargs.pop("ls", "none"))
-    kwargs.setdefault("linewidth", kwargs.pop("lw", _linewidth))
-    kwargs.setdefault("markersize", kwargs.pop("ms", _markersize))
+    kwargs.setdefault("linestyle", "none")
```
OriolAbril: `kwargs` should be dealiased before setting the defaults.

arviz/plots/pairplot.py
```diff
@@ -85,10 +85,8 @@ def plot_pair(
     Examples
     --------
     KDE Pair Plot
-
     .. plot::
         :context: close-figs
-
```
OriolAbril: I think these black lines are important for docs to render properly, could you check all inline example images are shown?

arviz/plots/mcseplot.py
percygautam: I must have missed this one.

OriolAbril:
I'll merge if tests pass

arviz/plots/pairplot.py
percygautam: Yes, there was some problem with blank lines. I checked all examples are working fine now.

OriolAbril: Thanks!

PR #1074, opened 2020-02-17
Improved tooltip information for bokeh pair_plot

### LINKED ISSUE #952: Bokeh hover tooltip
We need to implement better default hover settings.

https://docs.bokeh.org/en/1.4.0/docs/reference/models/tools.html#bokeh.models.tools.HoverTool

amukh18:
## Checklist

Added improved tooltip parameters  to  bokeh backend files of plots
To fix #952


arviz/plots/backends/bokeh/pairplot.py
```diff
@@ -66,7 +70,7 @@ def plot_pair(
 
         if ax is None:
             ax = bkp.figure(
-                width=int(figsize[0] * dpi), height=int(figsize[1] * dpi), **backend_kwargs
+                width=int(figsize[0] * dpi), height=int(figsize[1] * dpi), **backend_kwargs, tooltips=TOOLTIPS
```
ahartikainen: This does the job, but maybe we could add tooltip into backend_kwargs with `.setdefault`, if user wants to change the default behaviour.

amukh18:
I'm not sure if I use the setdefault function with backend_kwargs correctly

arviz/plots/backends/bokeh/pairplot.py
```diff
@@ -42,6 +42,11 @@ def plot_pair(
         ),
         **backend_kwargs,
     }
+    TOOLTIPS = [
+        (flat_var_names[0], "@{}".format(flat_var_names[0])),
+        (flat_var_names[1], "@{}".format(flat_var_names[1])),
+    ]
+    backend_kwargs.setdefault("tooltips", TOOLTIPS)
```
ahartikainen: This looks good.

ahartikainen:
You need to run black, but otherwise I think this is ready.

We can start to add other metadata after we have done rewriting for some of the plotting core code.

amukh18:
How about the tooltip parameters for the other plots? Is it part of the metadata?

ahartikainen:
Let's do them plot by plot

amukh18:
Ok I'll  edit my PR comment to say that it is only for plot_pair

amukh18:
And how do we begin rewriting the core code?

ahartikainen:
Core plotting stuff is done later. Let's first get these simplified tooltips working for plots with inner cds.

amukh18:
Ok I'll add the necessary parameters for the other plots using cds.

ahartikainen:
I don't think that last PR is valid/correctly set?

amukh18:
I am sorry but I do not understand. I thought the plotters and var_names parameters were similar to flat_var names

ahartikainen:
have you tested that output is correct?

amukh18:
I have run the notebooks in the docs as suggested by the contributing docs with only InferenceDataCookbook returning some unrelated errors. Could you please show me how I should proceed?

amukh18:
Ah I get it. I forgot to run pytest.

ahartikainen:
Have you tested what is actually changed by this pr? Like visually?

amukh18:
I am looking at the bokeh plots as they appear in my browser while pytest runs.

ahartikainen:
You could just manually plot and see the results

```
import arviz as az

# in jupyter notebook use this
az.backends.output_notebook()
# to save a specific file (I think both of these can be called at the same time)
# az.backends.output_file("filename.html")

az.plot_pair(data, var_names=["mu", "tau"], backend="bokeh") 
```
and then test some other plot

```
az.plot_forest(data, var_names=["mu", "tau"], backend="bokeh") 
```

amukh18:
But doesn't this use the release that came with pip? I think I would need to add some local paths to the notebook for it to import my local copy of the module. Sorry, but I am a little inexperienced at this.

OriolAbril:
You can use `pip install .` from your local repo (it won't work from anywhere) to install the local version and remove the other arviz version installed.

amukh18:
@OriolAbril This worked! Thank you very much!

ahartikainen:
I usually do

    pip install -e .

So it installs it inplace.

amukh18:
@ahartikainen In order to better understand the code snippet you suggested for adding to the `plot_pair` bokeh file, I tried to trace the running of all the functions including the ones in plot_utils.py such as `xarray_var_iter`, `xarray_sel_iter`,` make_label`, `selection_to_string` and `_zip_dims` by running the posterior group of the centered-eight data through all of these in a notebook. The final tooltip output I got was  `[('mu', '@mu'), ('theta\nChoate', '@theta\nChoate')]`. Could you please tell me what the \n is for?

ahartikainen:
\n is a newline character added when the labels are done

amukh18:
@ahartikainen Ok thank you.

ahartikainen:
Maybe clean this PR (leave pair_plot) and let's think how we can fix the other plots later.

amukh18:
Sorry about the delay. I have been looking for a way to extract the variable names for the mcse plot, but I haven't found a way yet.

ahartikainen:
It is not just variable names, we also beed CDS and those are not implemented yet.

amukh18:
Hasn't CDS already been used in mcse plots? Oh and sorry, I meant to say labels instead of variable names in my previous comment. And have I correctly extracted the labels for the trace plot?

ahartikainen:
Only for rug, not the main plot.

amukh18:
Yes, I noticed that right now. So can I continue adding the tooltip parameters for forest plot?

ahartikainen:
Forest plot is density-vased plot, so we can skip it for now.

Let me check for trace. There we can update the tooltip for raw traces.

arviz/plots/backends/bokeh/traceplot.py
```diff
@@ -62,9 +62,17 @@ def plot_trace(
         **backend_kwargs,
     }
     dpi = backend_kwargs.pop("dpi")
+    labels = [
+        make_label(var_name, selection) for idx, (var_name, selection, value) in enumerate(plotters)
+    ]
```
ahartikainen: You need to have the same names than those that go into the CDS. And that tooltip should change based on the axis.

And In traceplot the x is `index` (this could be draw or ndraw * chain + draw) and y is the value of the parameter.

amukh18: Could you please tell me what ndraw is?

amukh18:
Also, in the issue, you said that you would like to have a way to add tooltips after the figure is created. I am not sure if I correctly understand, but do you mean that there should be an option to somehow "toggle" the tooltips in the plot when the plot appears in the html file?

arviz/plots/backends/bokeh/traceplot.py
ahartikainen: number of draws

ahartikainen:
It would be interesting to know, can tooltips be changed after the figure has been created?

arviz/plots/backends/bokeh/traceplot.py
amukh18: By the same "names" do you mean the python variable names like cds_name instead of var_name?

amukh18: Ah never mind I think I understand now.

amukh18: Are the names like "theta_t_ARVIZ_CDS_SELECTION_school_Choate" ?

amukh18:
Do you mean something using MultiSelect from bokeh.models.widgets?

arviz/plots/backends/bokeh/traceplot.py
ahartikainen: yes.

amukh18:
I finally got how to change the tooltip names according to the axes but on running the following test code snippet, I got some tooltip values represented by '???':
`import arviz as az
coords = {'school': ['Choate', 'Lawrenceville']}
az.plot_trace(data, var_names=('theta_t', 'theta'), coords=coords, backend="bokeh")`

amukh18:
I think I have made a mistake. In the tooltips, were the x value and y value supposed to be printed?

amukh18:
I think I have made a mistake. In the tooltips, were the x value and y value supposed to be printed separately?

amukh18:
I think I have made a mistake. In the tooltips, were the x value and y value supposed to be printed separately?

amukh18:
@ahartikainen any opinions about changes I should make to my commit?

ahartikainen:
I need to test the behavior

ahartikainen:
I fixed the `pair_plot` code, but I still need to go through the other files

amukh18:
Thank you. I am familiarising myself with the changes you made. Is there any part of the code I should be working on right now?

ahartikainen:
I think you could reset the traceplot changes and after that let's merge this in

    git checkout arviz/plots/backend/bokeh/traceplot.py

Remember to pull at the start. `git pull`

amukh18:
Don't the other plots also require changes?

ahartikainen:
For now, no. There are multiple things needed to be fixed before we can have good hover info.

PR #1076, opened 2020-02-17
Removed bugs from traceplot lines

### LINKED ISSUE #929: plot_trace lines is unclear and it may yield unexpected results
**Describe the bug**
The argument `lines` for the function `plot_trace` can give unexpected results. Moreover, the documentation is a bit nebulous.
 
**To Reproduce**
A toy example is defined
```python
import pymc3 as pm
import arviz as az
import numpy as np

# fake data
mu_real = 0
sigma_real = 1
n_samples = 150

Y = np.random.normal(loc=mu_real, scale=sigma_real, size=n_samples)

with pm.Model() as model:
    mu = pm.Normal('mu', mu=0, sigma=10)  
    sigma = pm.HalfNormal('sigma', sigma=10)
    likelihood = pm.Normal('likelihood', mu=mu, sigma=sigma, observed=Y)
    trace = pm.sample()
```
As per [documentation](https://arviz-devs.github.io/arviz/generated/arviz.plot_trace.html#arviz.plot_trace), the argument `lines` accepts a tuple in the form `(var_name, {‘coord’: selection}, [line, positions])`. So, the command 
```python
az.plot_trace(trace, lines=(('mu', {}, mu_real),))
```
yields correctly 
image

I can also pass a list of tuples or a list of tuples and lists and it will work fine:
```
az.plot_trace(trace, lines=[('mu', {}, mu_real)]) # list of tuples
az.plot_trace(trace, lines=[['mu', {}, mu_real]]) # list of lists
az.plot_trace(trace, lines=[['mu', {}, mu_real], ('sigma', {}, sigma_real)]) # list of lists and tuples
```
however, I cannot pass a simple tuple because I will get a `KeyError: 0`
```python
az.plot_trace(trace, lines=(['mu', {}, mu_real]))
az.plot_trace(trace, lines=(('mu', {}, mu_real)))
```
Also, I can pass a variable or coordinate name that do not exist and Arviz will not complain---but not lines will be plotted (here I would expect a warning)
```python
az.plot_trace(trace, lines=[('hey', {}, mu_real)])
az.plot_trace(trace, lines=[('mu', {'hey'}, mu_real)])
```
image
The weird behavior happens when I pass a string:
```python
az.plot_trace(trace, lines=[('mu', {}, 'hey')])
```
image

**Expected behavior**
The [documentation](https://arviz-devs.github.io/arviz/generated/arviz.plot_trace.html#arviz.plot_trace) could be improved and the function could check the inputs. In addition to what described above, the placeholder `[line, positions]` in `(var_name, {‘coord’: selection}, [line, positions])` should be something like `[line_positions]` otherwise one may think (like myself :) ) that two values should be inserted (one for `line` and one for `positions`).

**Additional context**
I am using Win10, fresh conda environment with PyMC3 and Arviz from master.

Possibly related https://github.com/pymc-devs/pymc3/issues/3495, https://github.com/pymc-devs/pymc3/issues/3497

percygautam:
## Description

Fixes #929 . Removed bugs from traceplotlines
## Checklist

### DIFF
--- a/arviz/plots/backends/matplotlib/traceplot.py
+++ b/arviz/plots/backends/matplotlib/traceplot.py
@@ -133,7 +133,7 @@ def plot_trace(
         for line in lines:
             if line[0] not in all_var_names:
                 invalid_var_names.add(line[0])
-        if len(invalid_var_names):
+        if invalid_var_names:
             warnings.warn(
                 "A valid var_name should be provided, found {} expected from {}".format(
                     invalid_var_names, all_var_names
--- a/arviz/plots/backends/matplotlib/traceplot.py
+++ b/arviz/plots/backends/matplotlib/traceplot.py
@@ -127,7 +127,8 @@ def plot_trace(
 
     # Check the input for lines
     if lines is not None:
-        all_var_names = set([plotter[0] for plotter in plotters])
+        all_var_names = set(plotter[0] for plotter in plotters)
+
         invalid_var_names = set()
         for line in lines:
             if line[0] not in all_var_names:
--- a/arviz/plots/backends/matplotlib/traceplot.py
+++ b/arviz/plots/backends/matplotlib/traceplot.py
@@ -1,9 +1,9 @@
 """Matplotlib traceplot."""
 
+import warnings
 import matplotlib.pyplot as plt
 from matplotlib.lines import Line2D
 import numpy as np
-import warnings
 
 from . import backend_kwarg_defaults, backend_show
 from ...distplot import plot_dist
@@ -49,8 +49,8 @@ def plot_trace(
     rug : bool
         If True adds a rugplot. Defaults to False. Ignored for 2D KDE. Only affects continuous
         variables.
-    lines : tuple
-        Tuple of (var_name, {'coord': selection}, [line, positions]) to be overplotted as
+    lines : tuple or list
+        list of tuple of (var_name, {'coord': selection}, [line_positions]) to be overplotted as
         vertical lines on the density and horizontal lines on the trace.
     combined : bool
         Flag for combining multiple chains into a single line. If False (default), chains will be
@@ -127,8 +127,10 @@ def plot_trace(
 
     # Check the input for lines
     if lines is not None:
-        if lines[0] not in [plotters[i][0] for i in range(len(plotters))]:
-            warnings.warn("A valid var_name should be provided")
+        all_var_names = set([plotters[i][0] for i in range(len(plotters))])
+        for line in lines:
+            if line[0] not in all_var_names:
+                warnings.warn("A valid var_name should be provided, found {} expected one of {}".format(line[0], all_var_names))
 
     for idx, (var_name, selection, value) in enumerate(plotters):
         value = np.atleast_2d(value)
@@ -225,7 +227,7 @@ def plot_trace(
                 line_values = [vlines]
             else:
                 line_values = np.atleast_1d(vlines).ravel()
-                if not all(np.issubdtype(type(val), np.number) for val in line_values):
+                if not np.issubdtype(line_values.dtype, np.number):
                     raise ValueError(
                         "line-positions should be numeric, found {}".format(line_values)
                     )
--- a/arviz/plots/backends/matplotlib/traceplot.py
+++ b/arviz/plots/backends/matplotlib/traceplot.py
@@ -3,6 +3,7 @@
 import matplotlib.pyplot as plt
 from matplotlib.lines import Line2D
 import numpy as np
+import warnings
 
 from . import backend_kwarg_defaults, backend_show
 from ...distplot import plot_dist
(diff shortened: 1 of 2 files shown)


arviz/plots/backends/matplotlib/traceplot.py
```diff
@@ -219,6 +225,10 @@ def plot_trace(
                 line_values = [vlines]
             else:
                 line_values = np.atleast_1d(vlines).ravel()
+                if not all(np.issubdtype(type(val), np.number) for val in line_values):
```
OriolAbril: line_values will be a 1d numpy array here, `np.issubdtype(line_values, np.number)` should work

arviz/plots/backends/matplotlib/traceplot.py
```diff
@@ -124,6 +125,11 @@ def plot_trace(
 
     _, axes = plt.subplots(len(plotters), 2, squeeze=False, figsize=figsize, **backend_kwargs)
 
+    # Check the input for lines
+    if lines is not None:
+        if lines[0] not in [plotters[i][0] for i in range(len(plotters))]:
+            warnings.warn("A valid var_name should be provided")
```
OriolAbril: I don't think this will work at a quick glance (have not tried it yet though). Also now is probably a good time to fix docs, `lines` is a `list of tuple of (str, dict, float or array_like)`, so `lines[0]` will be a tuple. You should first get all var_names in a set (from plotters) and then check all var_names in lines tuples exist. It would be great to include _all_ incorrect variable names in warning message too.

arviz/plots/backends/matplotlib/traceplot.py
percygautam: `np.issubdtype` works only on individual datatypes, following code is throwing the error:
```
line_values = np.atleast_1d("hi").ravel()
if np.issubdtype(line_values, np.number):
    print("Failed")
```
Error: `TypeError: data type not understood`. 
We need to check for the individual elements.

arviz/plots/backends/matplotlib/traceplot.py
percygautam: Yes, I wrote the code for keeping only a particular case in mind. This is failing for many other cases. I'll do the changes.

arviz/plots/backends/matplotlib/traceplot.py
OriolAbril: `np.issubdtype(line_values.dtype, np.number)` np.issubdtype expects a dtype, not an array, forgot about that

percygautam: Yes, it'll work for same datatypes array `np.issubdtype(line_values.dtype, np.number)`. Can  it be case, if line_position is array of different datatypes say `[1, 3, "hi"]`!

OriolAbril: numpy arrays have one dtype, and cast all inputs to this dtype. Therefore, as numbers can always be cast to str but not the other way around, the dtype would be string/unicode... something like this

percygautam: Okay, thanks for the clarification :)

arviz/plots/backends/matplotlib/traceplot.py
```diff
@@ -127,8 +127,10 @@ def plot_trace(
 
     # Check the input for lines
     if lines is not None:
-        if lines[0] not in [plotters[i][0] for i in range(len(plotters))]:
-            warnings.warn("A valid var_name should be provided")
+        all_var_names = set([plotters[i][0] for i in range(len(plotters))])
+        for line in lines:
+            if line[0] not in all_var_names:
+                warnings.warn("A valid var_name should be provided, found {} expected one of {}".format(line[0], all_var_names))
```
OriolAbril: This will print one warning per missing variable, we could try to tell about all of them at once:

```
missing_vars = set([line for line in lines if line not in all_var_names])
warnings.warn("{} var_names were not found in data to be plotted ({}) and will be omitted".format(missing_vars, all_var_names))
```

Please improve and adapt on this if you like the idea

arviz/plots/backends/matplotlib/traceplot.py
```diff
@@ -124,6 +125,21 @@ def plot_trace(
...
 
+    # Check the input for lines
+    if lines is not None:
+        all_var_names = set(plotter[0] for plotter in plotters)
+
+        invalid_var_names = set()
+        for line in lines:
+            if line[0] not in all_var_names:
+                invalid_var_names.add(line[0])
+        if len(invalid_var_names):
```
OriolAbril: `if invalid_var_names:`

arviz/tests/test_plots_matplotlib.py
```diff
@@ -156,6 +156,21 @@ def test_plot_trace_max_subplots_warning(models):
     assert axes.shape
 
 
+@pytest.mark.parametrize("kwargs", [{"var_names": ["mu", "tau"], "lines": [("hey", {}, [1])]}])
+def test_plot_trace_invalid_varname_warning(models, kwargs):
+    with pytest.warns(UserWarning):
```
OriolAbril: can you add a `match` argument too? maybe `match="valid var.+should be provided"`

arviz/tests/test_plots_matplotlib.py
```diff
@@ -156,6 +156,21 @@ def test_plot_trace_max_subplots_warning(models):
...
+        axes = plot_trace(models.model_1, **kwargs)
+    assert axes.shape
+
+
+@pytest.mark.parametrize(
+    "bad_kwargs", [{"var_names": ["mu", "tau"], "lines": [("mu", {}, ["hey"])]}]
+)
+def test_plot_trace_bad_lines_value(models, bad_kwargs):
+    with pytest.raises(ValueError):
```
OriolAbril: same here

PR #1079, opened 2020-02-18
[WIP] Modify pairplot to include jointplot and cornerplot like features

agustinaarroyuelo:
The motivation behind this PR is that plots that are currently made with `plot_joint` can be made with `plot_pair`. Also, the `point_estimate` key word argument and `scatter_kde` kind will allow pair plots similar to fairly used [cornerplots](https://corner.readthedocs.io/en/latest/pages/quickstart.html).  
Here are some examples:
pairplot+jointplot_4
pairplot+jointplot3

I will really appreciate suggestions, especially on function testing.

## Checklist

- [x] Follows official PR format
- [x] Changes are listed in changelog


ahartikainen:
Looks good. Not needed for this PR, but it could be interesting to have this property for bokeh, but for any point that is selected.

arviz/plots/backends/matplotlib/pairplot.py
```diff
@@ -39,13 +44,48 @@ def plot_pair(
...
+            ax_hist_y.set_ylim(ax.get_ylim())
+
+            # Personalize axes
+            ax_hist_x.tick_params(labelleft=False, labelbottom=False)
+            ax_hist_y.tick_params(labelleft=False, labelbottom=False)
+        else:
+            if ax is None:
+                fig, ax = plt.subplots(figsize=figsize, **backend_kwargs)
 
         if kind == "scatter":
```
OriolAbril: how about reusing the code for scatter and kde? Something like:

```
if "scatter" in kind:
...
if "kde" in kind: # not elif
...
if "hexbin" in kind:
...
```

For now the valid kinds do not allow for more alternatives, but this is easily extended to a hexbin plot covering the scatter plot so only the point further from the mean are seen (in a corner.py style fashion)

arviz/plots/backends/matplotlib/pairplot.py
```diff
@@ -58,12 +98,23 @@ def plot_pair(
...
 
+        else:
+            ax.plot(infdata_group[0], infdata_group[1], zorder=-1, **plot_kwargs)
+            plot_kde(
+                infdata_group[0],
+                infdata_group[1],
+                ax=ax,
+                contourf_kwargs={"alpha": 0},
+                contour_kwargs={"colors": "k"},
+                fill_last=False,
```
OriolAbril: Building on previous comment, I'd use setdefaults and provided fill_last (maybe default fill_last to None so we can use True or False depending on plot still allowing user to customize this). Something like:

```
plot_kwargs.setdefault("contour_kwargs", {})
plot_kwargs["contour_kwargs"].setdefault("alpha", 0)
...
plot_kde(
    ...,
    **plot_kwargs
)
```

If reusing code it may also be a good idea to move the setdefault part somewhere else.

arviz/plots/backends/matplotlib/pairplot.py
```diff
@@ -75,6 +126,18 @@ def plot_pair(
...
+            pe_x = calculate_point_estimate(point_estimate, x)
+            pe_y = calculate_point_estimate(point_estimate, y)
+            if diagonal:
+                ax_hist_x.axvline(pe_x, **point_estimate_kwargs)
+                ax_hist_y.axhline(pe_y, **point_estimate_kwargs)
+
+            ax.axvline(pe_x, **point_estimate_kwargs)
+            ax.axhline(pe_y, **point_estimate_kwargs)
+
+            ax.scatter(pe_x, pe_y, marker="s", s=figsize[0] + 50, **point_estimate_kwargs, zorder=4)
```
OriolAbril: I think scatter and axvline have some kwarg incompatibilities, I am not sure they should share kwargs. Maybe there could be a warning in docs (along the lines of explaining where are kwargs eventually passed to)

agustinaarroyuelo:
Thanks Oriol for your comments! I am going through them :+1:

arviz/plots/plot_utils.py
```diff
@@ -685,7 +685,7 @@ def get_plotting_function(plot_name, plot_module, backend):
     return plotting_method
 
 
-def calculate_point_estimate(point_estimate, values, bw):
+def calculate_point_estimate(point_estimate, values, bw=4.5):
```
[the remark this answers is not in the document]
aloctavodia: I was asking myself the same question :-) I think it makes sense. Maybe Agustina wants to add it in a future PR, not pushing, just suggesting :-)

arviz/plots/backends/matplotlib/pairplot.py
agustinaarroyuelo: I am thinking that if multiple kinds can take place in the same plot, `plot_kwargs` will eventually present kwarg incompatibilities. Maybe we should have `scatter_kwargs`, `kde_kwargs` and `hexbin_kwargs`. Or just `plot_kwargs` and find a way to distribute key words.

aloctavodia: Having separated `kwargs` is much more flexible for the user and also easier to maintain than having to properly parse a `plot_kwargs` and distribute its content .

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## v0.x.x Unreleased
 
 ### New features
+* Added `diagonal`, `point_estimate`, `point_estimate_kwargs`, `point_estimate_marker_kwargs`, `scatter_kwargs`, `kde_kwargs` and `hexbin_kwargs` to `plot_pair` #1079
```
OriolAbril: I would focus more on the feature, even more so when it will probably be used constantly by most users. They can always check the api to see exactly what arguments have changed

arviz/data/io_pymc3.py
```diff
@@ -288,6 +288,8 @@ def constant_data_to_xarray(self):
         for var in self.model.deterministics:
             ancestors = self.theano.tensor.gof.graph.ancestors(var.owner.inputs)
             # no dependency on a random variable
+
+            # pylint: disable=E1101
```
OriolAbril: I think pylint in [CI builds](https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=1859&view=logs&j=8b1f4e88-90b9-5fe3-64d9-9b2a3a22d1e4&t=8729fc61-5b5c-512c-a278-229e5112b707&l=20) did not complain about this, it should not be necessary. If it is triggered locally, maybe updating pymc3 and pylint will solve the issue.

arviz/plots/backends/bokeh/pairplot.py
```diff
@@ -42,156 +49,123 @@ def plot_pair(
...
+                    width, height = get_width_and_height(jointplot, rotate)
+                    ax_ = bkp.figure(width=width, height=height, **backend_kwargs,)
+                    row_ax.append(ax_)
+            ax.append(row_ax)
+        ax = np.array(ax)
+    else:
+        assert ax.shape == (numvars - var, numvars - var)
 
+    tmp_flat_var_names = None
+    if len(flat_var_names) == len(list(set(flat_var_names))):
```
OriolAbril: I think `len(set(flat_var_names))` will do the trick even without converting to list.

arviz/plots/backends/bokeh/pairplot.py
```diff
@@ -42,156 +49,123 @@ def plot_pair(
...
-            source_dict = dict(zip(flat_var_names, [list(post) for post in infdata_group]))
-        else:
-            tmp_flat_var_names = ["{}__{}".format(name, str(uuid4())) for name in flat_var_names]
-            source_dict = dict(zip(tmp_flat_var_names, [list(post) for post in infdata_group]))
-        if divergences:
-            divergenve_name = "divergences_{}".format(str(uuid4()))
-            source_dict[divergenve_name] = (
-                np.array(diverging_mask).astype(bool).astype(int).astype(str)
-            )
+    # pylint: disable=R1702
```
OriolAbril: can you use `# pylint: disable=too-many-nested-blocks`? this way we have an idea of what kind of check we are skipping.

arviz/plots/backends/bokeh/pairplot.py
```diff
@@ -42,156 +49,123 @@ def plot_pair(
...
                     if divergences:
                         ax[j, i].circle(var1, var2, source=source, view=source_nondiv)
                     else:
                         ax[j, i].circle(var1, var2, source=source)
 
-                elif kind == "kde":
+                if "kde" in kind:
                     var1_kde = infdata_group[i]
-                    var2_kde = infdata_group[j + 1]
+                    var2_kde = infdata_group[j + var]
```
OriolAbril: we can define var1_ary and var2_ary outside the ifs and use them when needed, kde, hexbin, point_estimates...

ahartikainen:
I recommend that you rebase your branch.

First update your master against arviz master.

Then go to your branch 

    git rebase -i master

And then you can remove unneeded commits and keep only your.

`rebase -i` should open a text editor (vim for example) and then you can edit commits as needed.

agustinaarroyuelo:
Thank you Ari for your help. 
I did what you suggested and removed the commits that were previous to rebasing. This commits were duplicated.  When I tried to push this to my branch, the tip was behind its remote counterpart. Git suggested `git pull ...`, but pulling from my branch would have reverted the changes made by `git rebase -i master`. Therefore I forced the push, using `--force-with-lease` just in case. Let me know if it is ok.

ahartikainen:
`git push -f` is needed, I forgot to add that.

arviz/plots/pairplot.py
```diff
@@ -68,13 +76,29 @@ def plot_pair(
         Matplotlib axes or bokeh figures.
     divergences_kwargs : dicts, optional
         Additional keywords passed to ax.scatter for divergences
-    plot_kwargs : dicts, optional
-        Additional keywords passed to ax.plot, az.plot_kde or ax.hexbin
+    scatter_kwargs:
```
aloctavodia: Given the changes parts of the docstring are outdated. For example 

    kind : str
        Type of plot to display (scatter, kde or hexbin)

is not an "or" but and/or. And probably we should explicitly indicate the overlaying feature.

Instead of
`Plot a scatter or hexbin matrix of the sampled parameters.`

we should have something like:
 
Scatter, kde and/or hexbin matrix with (optional) marginals on the diagonal

aloctavodia: Also the examples need to be updated. It's OK for me if we do that in another PR (but we should open and issue so we do not forget).

OriolAbril:
Changelog should be updated to list changes in unreleased version, not 0.7.0

agustinaarroyuelo:
Thanks everyone for your comments and help!

Issue #1077, opened 2020-02-18
Memory issues with log likelihood in io_pymc3


OriolAbril:
In the presence of a large number of observations, the log likelihood array cannot be created due to memory limitations. This is partly due to the fact that the code to store log likelihood values in io_pymc3 relies on growing lists inside loops which is not memory efficient and it could even reach the point of the log likelihood not fitting in memory. 

I propose a 2 part solution, which afterwards could then be improved by storing in-file the results with or without dask... 

The first part (which would also solve other issues related to io_pymc3) is to add an argument to exclude the log likelihood data. This argument could also be extended to pyro and numpyro, other converters already allow being explicit on this. 

The second part is to preallocate the arrays which will reduce the memory usage and also slightly speed up conversion. Fix could be based on https://github.com/pymc-devs/pymc3/pull/3556 which fixed a similar issue with posterior predictive

nitishp25:
Will this argument contain a list of variables for which we calculate log_likelihood?

OriolAbril:
I am not sure about this, I think errors when present will affect all variables altogether, but it could make sense as a memory saving measure.

nitishp25:
Yes I mean keeping it None as default will avoid generating ```log_likelihood``` but the second part of memory pre-allocation will have some effect in preventing errors right? (when the user will input the variables for ```log_likelihood```) If not then we can use dask arrays instead of numpy?

OriolAbril:
I am not sure I follow, sorry, maybe I could have been more clear from the start. There are actually two issues in one here:

The first is with memory usage. From this side, the code does work, but it is not very efficient with memory usage. This could eventually end up raising a MemoryError, but in general what happens is that the computer freezes (one example [here](https://discourse.pymc.io/t/plotting-a-trace-potentially-causes-my-python-kernel-to-crash/4142)). Thinking about this however does pose a question: What if there are so many observations that the arrays do not fit in memory? Then no memory usage optimization can solve the problem, hence the proposal to allow users to avoid log likelihood storage (always defaulting to storing it though). On the long run, storing on file or using dask would allow us to work with data not fitting in memory, however, plots and algorithms do not work with dask yet, so it doesn't really make sense to look into this as of now. Here storing one instead of 2 variables could be relevant in terms of memory usage.

[...]

Issue #395, opened 2018-11-08
Arviz incompatible with PyMC3 trace subsetting


mfansler:
After upgrading to Arviz 0.2.1 from 0.1.0, I found that many of my PyMC3 trace objects are no longer compatible with Arviz.  In particular, failure happens in models where I subsetted the variables being traced during sampling. It seems that the conversion from a PyMC3 trace object to Arviz's InferenceData object ultimately expects some internal variables to be traced, whereas PyMC3's trace subsetting excludes all but those listed.

## Example
```
import numpy as np
import pymc3 as pm
import arviz as az

observed = np.random.binomial(500, p=0.4, size=30)

with pm.Model() as model:
    mu = pm.DiscreteUniform('mu', lower=100, upper=1000)
    p = pm.Beta('p', alpha=1, beta=1)
    N = pm.Binomial('N', n=mu, p=p, observed=observed)

    trace = pm.sample(trace=[mu, p])

az.plot_trace(trace)
```
### Result
> TypeError: Missing required input: p_logodds__

One might hope that restricting the trace plot to only a variable of interest might get beyond it, e.g.

    az.plot_trace(trace, var_names=['mu'])

but this fails just as above.  The stack trace notably includes `from_pymc3`, and calling this directly on the trace object will also fail.

### Commentary
Certainly, in this toy model one has no problem tracing everything, but in large complex models, tracing all variables can be wasteful and unnecessary.  Perhaps an alternative is for PyMC3 to still keep those internal variables in the trace, even when a subset is requested.  Moreover, since `pm.sample` expects a list of actual variables and not just strings (e.g., `[mu, p]` above) for the `trace` argument, I don't see any way to tell the sampler to continue saving those internal variables, e.g. `'p_logodds__'` that are still relevant to the ones being included.

## Full Error Result

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-26-f1462efd3db0> in <module>()
----> 1 az.plot_trace(trace)

~/anaconda/envs/pymc3/lib/python3.6/site-packages/arviz/plots/traceplot.py in plot_trace(data, var_names, coords, figsize, textsize, lines, combined, kde_kwargs, hist_kwargs, trace_kwargs)
     72         >>> az.plot_trace(data, var_names=('theta_t', 'theta'), coords=coords, lines=lines)
     73     """
---> 74     data = convert_to_dataset(data, group='posterior')
     75 
     76     if coords is None:

[...]

PR #1082, opened 2020-02-20
[WIP] Reduce memory usage in log_likelihood io_pymc3

nitishp25:
## Description

Related to issue #1077
## Checklist


arviz/utils.py
```diff
@@ -488,3 +488,56 @@ def flatten_inference_data_to_dict(
 
                         data_dict[var_name_dim] = var_values[loc]
     return data_dict
+
+
+class _sampledict:
```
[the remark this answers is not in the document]
nitishp25: Yes, I'll use the PyMC3 one.

arviz/data/io_pymc3.py
```diff
@@ -153,18 +155,25 @@ def _extract_log_likelihood(self):
...
+        if isinstance(self.log_likelihood, list):
+            cached = [
+                (var, var.logp_elemwise)
+                for var in self.model.observed_RVs
+                if var in self.log_likelihood
+            ]
+        else:
+            cached = [(var, var.logp_elemwise) for var in self.model.observed_RVs]
+        log_likelihood_dict = utils._sampledict(  # pylint: disable=protected-access
+            len(self.trace.chains)
```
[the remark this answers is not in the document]
nitishp25: Actually, what this does is pass the length for the first dim for the array for each variable (key). Since, each key is looped for total chains, I guess the shape of array for the key would be ```len(self.trace.chains) + val.shape```? Do you want to initialize the whole dict in one go?

arviz/data/io_pymc3.py
```diff
@@ -153,18 +155,25 @@ def _extract_log_likelihood(self):
         """Compute log likelihood of each observation."""
         # If we have predictions, then we have a thinned trace which does not
         # support extracting a log likelihood.
-        cached = [(var, var.logp_elemwise) for var in self.model.observed_RVs]
-        log_likelihood_dict = {}
+        if isinstance(self.log_likelihood, list):
```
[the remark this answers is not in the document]
nitishp25: Makes sense, thanks!

arviz/data/io_pymc3.py
```diff
@@ -153,18 +155,25 @@ def _extract_log_likelihood(self):
...
+            cached = [
+                (var, var.logp_elemwise)
+                for var in self.model.observed_RVs
+                if var in self.log_likelihood
```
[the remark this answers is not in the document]
nitishp25: Yes you're right!

arviz/data/io_pymc3.py
```diff
@@ -361,6 +370,7 @@ def from_pymc3(
     *,
     prior: Optional[Dict[str, Any]] = None,
     posterior_predictive: Optional[Dict[str, Any]] = None,
+    log_likelihood: Optional = True,
```
OriolAbril: `log_likelihood: Union[bool, Iterable[str]] = True`

nitishp25:
These are the benchmarks I ran using memory profiler on [this code](https://github.com/nitishp25/arviz-bench/blob/master/logp_benchmarks.py).

The first 4 graphs show the memory being used with time for samples of different chains.

The last graph summarizes the maximum memory used for different chains and the memory difference between the *Preallocated code* and *Non-preallocated code*.

Thanks to @OriolAbril for all the help!

 
  

## Summary
summary1

OriolAbril:
This looks great, thanks! It would also be great if you could review @rpgoldman given that it uses `defaultTrace` (only if you have time!). 

2 things left to do:

Could you add some tests to check `log_likelikhood` argument behaves as suposed to? There is no need to check the memory usage in a test. maybe a pytest.mark.parametrize added to `test_multiple_observed_rv` to check that default returns both observed variables, false returns no log_likelihood group and using a str returns only the selected variable. [`check_multiple_attrs`](https://github.com/arviz-devs/arviz/blob/dc20591d3932041fcaecbbd0c93943f9f1cfb786/arviz/tests/helpers.py#L161) allows to ensure both cases, either an attr being present or not, the help (and other cases where it is used) should be clear.

On second thought, using pymc3's `defaultTrace` would force users to have a very recent pymc3 version, maybe it is not ideal

nitishp25:
> Could you add some tests to check `log_likelikhood` argument behaves as suposed to? There is no need to check the memory usage in a test. maybe a pytest.mark.parametrize added to `test_multiple_observed_rv` to check that default returns both observed variables, false returns no log_likelihood group and using a str returns only the selected variable. [`check_multiple_attrs`](https://github.com/arviz-devs/arviz/blob/dc20591d3932041fcaecbbd0c93943f9f1cfb786/arviz/tests/helpers.py#L161) allows to ensure both cases, either an attr being present or not, the help (and other cases where it is used) should be clear.

Yes I'm working on testing these cases only. Thanks!

> On second thought, using pymc3's `defaultTrace` would force users to have a very recent pymc3 version, maybe it is not ideal

Then the previous one with the code defined in utils would be more preferable?

rpgoldman:
I'll try to check in the next day or so.  Busy today!

rpgoldman:
Minor nit: it's `_DefaultTrace` that is used, not `defaultTrace` -- took me a few minutes to figure this out.

arviz/tests/test_data_pymc.py
```diff
@@ -206,23 +206,30 @@ def test_missing_data_model(self):
...
+            inference_data = from_pymc3(trace=trace, log_likelihood=log_likelihood)
         test_dict = {
             "posterior": ["x"],
             "observed_data": ["y1", "y2"],
             "log_likelihood": ["y1", "y2"],
             "sample_stats": ["diverging", "lp"],
         }
         fails = check_multiple_attrs(test_dict, inference_data)
-        assert not fails
+        if log_likelihood is True:
```
OriolAbril: I'd be better to modify the test_dict, something like:

```
if not log_likelihood:
    test_dict.pop("log_likelihood")
    test_dict["~log_likelihood"] = []
if isinstance(log_likelihood, list):
    test_dict["log_likelihood"] = ["y1", "~y2"]
```

OriolAbril:
Also, make sure to get your branch up to date with master!

arviz/tests/external_tests/test_data_pymc.py
```diff
@@ -215,12 +216,19 @@ def test_multiple_observed_rv(self):
             pm.Normal("y2", x, 1, observed=y2_data)
             trace = pm.sample(100, chains=2)
             inference_data = from_pymc3(trace=trace)
+            inference_data = from_pymc3(trace=trace, log_likelihood=log_likelihood)
```
OriolAbril: It looks like the conversion is done twice.

arviz/data/io_pymc3.py
```diff
@@ -193,7 +202,7 @@ def sample_stats_to_xarray(self):
     @requires("model")
     def log_likelihood_to_xarray(self):
         """Extract log likelihood and log_p data from PyMC3 trace."""
-        if self.predictions:
+        if self.predictions or not self.log_likelihood:
             return None
         data = self._extract_log_likelihood()
```
OriolAbril: Extra idea. We could put a

    try:
         data = self._extract...
     except TypeError:
         warnings.warn("could not compute log likelihood. log_likelihood group will be omitted. Check your model object or set log_likelihood=False")
        return None

Which if I am correct should fix several issues such as https://github.com/arviz-devs/arviz/issues/395 (it has a minimal example to reproduce so it should be easy to check the issue is fixed) or even https://github.com/pymc-devs/pymc3/issues/3728

nitishp25: Yes it does fix both the issues!

nitishp25:
@rpgoldman Can you explain what the **2** peaks near the middle of the graphs above represent? 

They seem to occur at the start of allocation of ```posterior_predictive```. And each of those peaks contain more peaks equal to the no. of chains of the samples.
Similar peaks are observed in allocation of ```log_likelihood``` but they are relatively smaller (use lesser memory, also only equal to no. of chains, unlike 2 * no. of chains for ```posterior_predictive```).
If you want I can give you the visuals to refer.

rpgoldman:
> @rpgoldman Can you explain what the **2** peaks near the middle of the graphs above represent?

So you are asking about the two spikes that come at the start of the process, as opposed to the two near the end. Correct?

Do we know if they are happening in the building of `chain_likelihoods`,  the allocation of the `_TraceDict`, `log_likelilhood_dict`, or somewhere else?  It seems like that is the most important question to answer.  Once we know that, we will know if it is something inside my `_TraceDict` code (likely the `insert`method), or if it's the list creations in the loop over the chains.

Can you derive plots like the above where the x axis plots position in the code, instead of clock time?

OriolAbril:
Back in the day when I did some experiments on this, I found the pattern to be: first slope corresponds to model creation and trace allocation, plateau is posterior sampling, the gentle slope is posterior predictive sampling and steep slope is conversion to inference data (which in terms of memory is basically retrieving and allocating log_likelihood data).

This would leave the two spikes at the beginning of posterior predictive sampling or at the end of posterior sampling. While writing this I realized it is not difficult to identify which of the two is happening, so I ran a quick example similar to the one above with just posterior sampling, no posterior predictive sampling. I see these same peaks towards the end of sampling.

rpgoldman:
I changed posterior predictive sampling to pre-allocate memory because it used to be building enormous python lists and then translating them into numpy arrays.  Now we pre-allocate the arrays and then just fill them.  Posterior predictive sampling is still much slower than I would like (I have an open PR for this), but pre-allocation substantially reduces memory usage.  That could be what is causing this issue.

I'm a bit confused about why this question is here on the ArviZ development repo, instead of on the `pymc3-devs1 one.

OriolAbril:
I have done a quick check between pymc3 versions, only posterior sampling, no posterior predictive sampling nor conversion to inference data. This is the result for 6 chains:

pymc_version_6chains

It is here because we saw this behaviour while checking that PR did indeed reduce memory usage in `log_likelihood_to_xarray` and we didn't have much idea of what was happening. Now that the issue seems more clear, it is probably better to move to pymc issue or discourse.

If there are no issues with `log_likelihood_to_xarray` I'll merge

rpgoldman:
> I have done a quick check between pymc3 versions, only posterior sampling, no posterior predictive sampling nor conversion to inference data. 

That's a relief!  It can't be my fault, then, because I didn't change anything in the normal sampling, only posterior predictive. 😌


> If there are no issues with `log_likelihood_to_xarray` I'll merge

Please do!

arviz/data/io_pymc3.py
```diff
@@ -153,18 +155,25 @@ def _extract_log_likelihood(self):
...
+            cached = [(var, var.logp_elemwise) for var in self.model.observed_RVs]
+        else:
+            cached = [
+                (var, var.logp_elemwise)
+                for var in self.model.observed_RVs
+                if var.name in self.log_likelihood
+            ]
+        log_likelihood_dict = self.pymc3.sampling._DefaultTrace(  # pylint: disable=protected-access
+            len(self.trace.chains)
+        )
```
OriolAbril: we should wrap this in a try except block, so that when pymc3 version is an old one, the error risen explains that pymc3 should be upgraded or arviz downgraded, otherwise users won't know what hit them.

There is no need to edit requirements file though, as pymc3 development is shown there.

arviz/data/io_pymc3.py
```diff
@@ -153,18 +155,30 @@ def _extract_log_likelihood(self):
...
+                for var in self.model.observed_RVs
+                if var.name in self.log_likelihood
+            ]
+        try:
+            log_likelihood_dict = self.pymc3.sampling._DefaultTrace(  # pylint: disable=protected-access
+                len(self.trace.chains)
+            )
+        except AttributeError:
+            raise AttributeError(
+                "Either upgrade PyMC3 to latest version or downgrade ArviZ for log_likelihood."
```
OriolAbril: "Installed version of ArviZ requires PyMC3>=3.8. Please upgrade with `pip install pymc3>=3.8` "
    "or `conda install -c conda-forge pymc3>=3.8`."

PR #1084, opened 2020-02-21
To transform colors in plotting functions to hex

### LINKED ISSUE #956: Bokeh, transform all colors to hex
Input all colors to bokeh functions as hex.

We can use matplotlib helper functions to transform special syntax

    color="C0"

etc

amukh18:
## Checklist

To transform colors in plotting functions to hex. 
To fix #956

### DIFF
--- a/arviz/plots/khatplot.py
+++ b/arviz/plots/khatplot.py
@@ -1,6 +1,6 @@
 """Pareto tail indices plot."""
 import matplotlib.pyplot as plt
-from matplotlib.colors import to_rgba_array
+from matplotlib.colors import to_rgba_array,to_hex
 import matplotlib.cm as cm
 import numpy as np
 from xarray import DataArray
@@ -137,13 +137,14 @@ def plot_khat(
     hlines_kwargs.setdefault("linestyle", [":", "-.", "--", "-"])
     hlines_kwargs.setdefault("alpha", 0.7)
     hlines_kwargs.setdefault("zorder", -1)
-    hlines_kwargs.setdefault("color", "C1")
+    hlines_kwargs.setdefault("color", to_hex("C1"))
 
     if coords is None:
         coords = {}
 
     if color is None:
         color = "C0"
+    color = to_hex(color)
 
     if isinstance(khats, np.ndarray):
         khats = khats.flatten()
--- a/arviz/plots/khatplot.py
+++ b/arviz/plots/khatplot.py
@@ -1,6 +1,6 @@
 """Pareto tail indices plot."""
 import matplotlib.pyplot as plt
-from matplotlib.colors import to_rgba_array, to_hex
+from matplotlib.colors import to_rgba_array,to_hex
 import matplotlib.cm as cm
 import numpy as np
 from xarray import DataArray
@@ -8,6 +8,7 @@ from xarray import DataArray
 from .plot_utils import (
     _scale_fig_size,
     color_from_dim,
+    vectorized_to_hex,
     format_coords_as_labels,
     get_plotting_function,
     matplotlib_kwarg_dealiaser,
@@ -144,7 +145,6 @@ def plot_khat(
 
     if color is None:
         color = "C0"
-    color = to_hex(color)
 
     if isinstance(khats, np.ndarray):
         khats = khats.flatten()
@@ -201,6 +201,7 @@ def plot_khat(
     khats = khats if isinstance(khats, np.ndarray) else khats.values.flatten()
     alphas = 0.5 + 0.2 * (khats > 0.5) + 0.3 * (khats > 1)
     rgba_c[:, 3] = alphas
+    rgba_c=vectorized_to_hex(rgba_c)
 
     plot_khat_kwargs = dict(
         hover_label=hover_label,
--- a/arviz/plots/khatplot.py
+++ b/arviz/plots/khatplot.py
@@ -1,6 +1,6 @@
 """Pareto tail indices plot."""
 import matplotlib.pyplot as plt
-from matplotlib.colors import to_rgba_array,to_hex
+from matplotlib.colors import to_rgba_array, to_hex
 import matplotlib.cm as cm
 import numpy as np
 from xarray import DataArray
--- a/arviz/plots/khatplot.py
+++ b/arviz/plots/khatplot.py
@@ -139,8 +139,8 @@ def plot_khat(
     hlines_kwargs.setdefault("linestyle", [":", "-.", "--", "-"])
     hlines_kwargs.setdefault("alpha", 0.7)
     hlines_kwargs.setdefault("zorder", -1)
-    hlines_kwargs.setdefault("color","C1")
-    hlines_kwargs["color"]=vectorized_to_hex(hlines_kwargs["color"])
+    hlines_kwargs.setdefault("color", "C1")
+    hlines_kwargs["color"] = vectorized_to_hex(hlines_kwargs["color"])
 
     if coords is None:
         coords = {}
--- a/arviz/plots/khatplot.py
+++ b/arviz/plots/khatplot.py
@@ -139,7 +139,8 @@ def plot_khat(
     hlines_kwargs.setdefault("linestyle", [":", "-.", "--", "-"])
     hlines_kwargs.setdefault("alpha", 0.7)
     hlines_kwargs.setdefault("zorder", -1)
-    hlines_kwargs.setdefault("color", vectorized_to_hex("C1"))
+    hlines_kwargs.setdefault("color","C1")
+    hlines_kwargs["color"]=vectorized_to_hex(hlines_kwargs["color"])
 
     if coords is None:
         coords = {}
--- a/arviz/plots/khatplot.py
+++ b/arviz/plots/khatplot.py
@@ -1,6 +1,6 @@
 """Pareto tail indices plot."""
 import matplotlib.pyplot as plt
-from matplotlib.colors import to_rgba_array,to_hex
+from matplotlib.colors import to_rgba_array, to_hex
 import matplotlib.cm as cm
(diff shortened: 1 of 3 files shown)


arviz/plots/khatplot.py
```diff
@@ -136,13 +136,14 @@ def plot_khat(
...
     hlines_kwargs.setdefault("zorder", -1)
-    hlines_kwargs.setdefault("color", "C1")
+    hlines_kwargs.setdefault("color", to_hex("C1"))
 
     if coords is None:
         coords = {}
 
     if color is None:
         color = "C0"
+    color = to_hex(color)
```
OriolAbril: Ooohh, I see.  Plot khat accepts plenty of options, from a color string to a dimension name. From the [docs](https://arviz-devs.github.io/arviz/generated/arviz.plot_khat.html#arviz.plot_khat).

> color : str or array_like, optional
>
>    Colors of the scatter plot, if color is a str all dots will have the same color, if it is the size of the observations, each dot will have the specified color, otherwise, it will be interpreted as a list of the dims to be used for the color code

the conversion to hex should be done somewhere later in the code (after dim names have been converted to colors) and also take into account that colors can be arrays, we should probably define a function in plot_utils to take into account vectorized cases.

I didn't think about this before.

amukh18: Thanks for the clarification. I'll make the changes accordingly.

amukh18: I have been trying to make the changes for quite some time now but I am not quite getting them right. Just wanted to double-check: the Matplotlib color argument for to_rgba and to_hex is essentially a list right?

OriolAbril:
Talking from memory, these are all the cases that have to be taken into account. I'd recommend start considering one or a couple of them and progressively add more cases. It is a task that requires to be familiar with both matplotlib's color handling and the extra particularities added by ArviZ (I have tried to list them here, but you should double check).

The first division is between different kinds of plots:
* plt.plot and most matplotlib functions can take the color argument in several ways:
  * named color string
  * hex color string (hex colors start with "#", so even something like "#010412" cannot be converted to a float)
  * 1D array of rgb (length 3) or rgba (length 4) values
* plt.scatter: https://matplotlib.org/3.1.3/api/_as_gen/matplotlib.pyplot.scatter.html
  * named color string
  * hex color string (hex colors start with "#", so even something like "#010412" cannot be converted to a float)
  * 1D array of strings (either named colors or hex)
  * 1D array of numbers, values are mapped to a color using a colormap
  * 2D array of numbers

The cases using plot and similar functions do not need to handle vectorized cases, and `to_hex` works straight out of the box for them. Cases using scatter need to handle also arrays with multiple colors, and apply `to_hex` not to the whole array but to each color. Moreover, 1D arrays of numbers are not valid colors, it is the colormap what converts a scalar number into a color, this has to be taken into account too (https://github.com/arviz-devs/arviz/blob/master/arviz/plots/khatplot.py#L179-L195 tries to handle all the possibilities, therefore the conversion to hex must go after this).

Extra thing to consider. Some ArviZ plots (like khat) also accept a dimension name as a coordinate, in which case the value of the coordinate is used to define a colormap (there is an example in the api documentation https://arviz-devs.github.io/arviz/generated/arviz.plot_khat.html#arviz.plot_khat). Dimension names generally cannot be converted to colors (and never should be converted right away, only via colormap so this has to be taken into account too). Luckily, https://github.com/arviz-devs/arviz/blob/master/arviz/plots/khatplot.py#L179-L195 also handles dimension names as colors, so calling _vectorized to_hex_ after these lines should solve the dimension name handling.

amukh18:
So you mean I should create a separate function called `vectorized to_hex` to handle these vectorised cases as you had said in a previous comment? This would help the API example with color="observed_county" work I guess.

amukh18:
And thank you for your detailed explanation!

amukh18:
Could you please tell me what these three lines in khatplot.py do?
`khats = khats if isinstance(khats, np.ndarray) else khats.values.flatten()
    alphas = 0.5 + 0.2 * (khats > 0.5) + 0.3 * (khats > 1)
    rgba_c[:, 3] = alphas`
I think I got the hex conversion to work but I was getting an error in the last line of the above three lines.

amukh18:
Also, one more thing. I think the hex conversion should be done after Line 199 of khatplot.py so that the alpha values can be assigned to the RGBA values.

amukh18:
I was going to implement the `vectorized_to_hex ` function in plot_utils.py. Is there any special ordering of functions in plot_utils.py which I may need to follow so as to place `vectorized_to_hex` below one and above another existing function?

OriolAbril:
> I was going to implement the `vectorized_to_hex ` function in plot_utils.py. Is there any special ordering of functions in plot_utils.py which I may need to follow so as to place `vectorized_to_hex` below one and above another existing function?

No specific order

arviz/plots/plot_utils.py
```diff
@@ -602,6 +603,23 @@ def color_from_dim(dataarray, dim_name):
...
+    rgba_c: vector of RGBA values
+
+    Returns
+    -------
+    rgba_hex : vector of hex values
+    """
+    rgba_hex = []
+    for rgba in range(len(rgba_c)):
+        rgba_hex.append(to_hex(rgba_c[rgba]))
+    return rgba_hex
```
OriolAbril: I think this could be a list comprehension, it's better to avoid lists growing inside of loops if possible.

The keep_alpha (not sure the name is exactly this one, I hope the idea is clear) should also be explicit and passed to to hex, we may want to keep or not the alpha value in the resulting hex color.

amukh18: For that, should I add a keep_alpha parameter in the main plot function too?

OriolAbril: No, the default is to set different alpha values to the colors themselves, so that the relevant points (large khat) are opaque whereas low khat is quite transparent. This can already be changed using plot_kwargs, there is no need for a new keyword argument.

amukh18:
I was looking at `plot_parallel` and ran one of the examples once with matplotlib as the backend and once with bokeh as the backend. Both plots returned different colors. I saw that in the main plot file, the color for the divergent and non-divergent data had been removed from `parallel_kwargs` for bokeh backend and the colors had been predefined in the bokeh backend file, leading to different colors in the two plots Should I modify this?

OriolAbril:
I would first focus on `vectorized_to_hex` and trying to make it accept all the possibilities I described above (except list of numbers and dim names). This will allow us to use the function in a more general way. 

One thing that comes to mind to try is to do the following:

```
def vectorized_to_hex(c, keep_alpha=False):
    try:
        hex_color = to_hex(c, keep_alpha)
    except ...:
        hex_color = [to_hex(color, keep_alpha) for color in c]
    return hex_color
```

Our goal would be to be able to do:

```
vectorized_to_hex("black")  # works
vectorized_to_hex(["red", "black", "C0"])  # works
...
```

arviz/plots/khatplot.py
```diff
@@ -136,7 +137,7 @@ def plot_khat(
     hlines_kwargs.setdefault("linestyle", [":", "-.", "--", "-"])
     hlines_kwargs.setdefault("alpha", 0.7)
     hlines_kwargs.setdefault("zorder", -1)
-    hlines_kwargs.setdefault("color", "C1")
+    hlines_kwargs.setdefault("color", to_hex("C1"))
```
OriolAbril: After we have managed to generalize vectorized to hex, we could modify this to:

```
hlines_kwargs.setdefault("color", "C1")
hlines_kwargs["color"] = vectorized_to_hex(hlines_kwargs["color"])  # color can be anything accepted by hlines
```

which would bring us closer to using the same colors in bokeh and in matplotlib which is the motivation of converting the colors to hex, matplotlib already works fine.

arviz/plots/plot_utils.py
```diff
@@ -602,6 +603,28 @@ def color_from_dim(dataarray, dim_name):
...
+        to select if alpha values should be kept in the final hex values.
+
+    Returns
+    -------
+    rgba_hex : vector of hex values
+    """
+    try:
+        hex_color=to_hex(c, keep_alpha=False)
+
+    except:
```
OriolAbril: The error type we want to catch should be explicit. Try calling `to_hex(["k", "r"])` yourself and see what is the error risen.

arviz/plots/khatplot.py
```diff
@@ -136,7 +138,7 @@ def plot_khat(
     hlines_kwargs.setdefault("linestyle", [":", "-.", "--", "-"])
     hlines_kwargs.setdefault("alpha", 0.7)
     hlines_kwargs.setdefault("zorder", -1)
-    hlines_kwargs.setdefault("color", "C1")
+    hlines_kwargs.setdefault("color", vectorized_to_hex("C1"))
```
OriolAbril: This will only transform `C1` to hex without transforming `hlines_kwargs["color"]` to hex if the users inputs a value for it. Try printing `hlines_kwargs` right below here and compare this approach to the other proposal if you do not pass any hlines_kwargs and if you do pass something like `["k", "k", "k", "r"]`

amukh18:
I have tried adding a test for vectorized_to_hex. I will add the code to handle colormaps by tomorrow.

amukh18:
Could you please tell me how pytest test functions are generally tested? I'm not sure if I'm doing it the right way.

OriolAbril:
Pytest is called from the command line by doing `pytest path/to/tests`, in ArviZ's case this is `pytest arviz/tests` to run all tests or `pytest arviz/tests/base_tests` to skip data relared tests which require pymc3, pystan, pyro, tensorflow probability...

When called like this, pytest will look into all files starting with `test_` and execute all functions also starting with `test_` in tgese files

PR #1088, opened 2020-02-24
[WIP] Benchmark main ArviZ code

### LINKED ISSUE #782: A better way to benchmark the speed tests
The benchmark tests which are in use now can be improved much more. Importing arviz into the benchmark tests throws the following error: 

Screenshot from 2019-08-01 21-45-25

The workaround right now is to include only the jitted methods in `benchmarks.py`. It is indicative of the extent of the speedup but isn't very accurate.

OriolAbril:
## Description
Benchmarks currently have a copied version of ArviZ's functions in their code instead of using ArviZ itself. Once finished it will fix  #782. 

Moving the benchmarks one folder below allows to import ArviZ. Then, to ease comparison, it would be great to use asv parametrization options, so that both numba and non-numba version are run in the same environment. 

## Checklist

- [ ] _Code style  correct (follows pylint and black guidelines)_ -> should benchmarks comply with lint too?


ahartikainen:
What different things we want to test?

All diagnostics / stats functions?

Also, should we implement LineProfiler for all the functions?

Also maybe a viz about our codebase?

https://pycallgraph.readthedocs.io/en/master/
https://stackoverflow.com/questions/582336/how-can-you-profile-a-python-script

And then add one file that does some processing for the results?

OriolAbril:
I would start (small) with asv benchmarks timing some common functions. Also, if they have numba option, compare between numba and no numba. This will warn us when there is a suspicious change in the code. 

One example is `_fast_kde` which currently has no numba speedup whatsoever since it was moved to `plot_utils`. In the moving changed the import from `stats_utils.histogram` to `np.histogram`. (trying to fix this causes the cyclic import, we should move fast_kdes to their own file or maybe to stats_utils)

Once we are happy with these benchmarks we can try to go deeper, check memory usage, line profiling...

nitishp25:
Can't we import `histogram`, `stats_variance_1d` and `stats_variance_2d` functions as well from `az.stats.stats_utils` and make the benchmarks using Numba `params` like you have done for the rest of the functions?

OriolAbril:
I think there are some issues due to cyclic imports. We have to check functions in stats utils and plot utils to make sure that plot utils imports stats utils and stats utils _does not_ import plot utils.

OriolAbril:
@nitishp25 also feel free to check out this branch locally to build on top of this and submit the work in a different PR

OriolAbril:
I'll close in favour of #1142

PR #1089, opened 2020-02-26
Add layout properties for Bokeh backend.

ahartikainen:
## Description

Use `bokeh.layouts.layout` by default, when `show is True`. This enables user to define `rcParams` for `sizing_mode` and also setting the `toolbar_location`.

## Checklist


ahartikainen:
We probably could add also rcParam for the plot output?

E.g. `bokeh.layout.shape = default` would use the default order given our plot. But user could set it for .. `square` which would unravel axes (remove `None`) and then reshape so the plots fill a square.

Maybe other options could be possible based on maximum columns and maximum rows?

layout also eats rugged list of lists, so we could have 3 / 2 / 3 structure etc?

Or maybe this could be a function that user can use manually (`show=False`)?

ahartikainen:
cc @sethaxen @AlexAndorra

ahartikainen:
```
import arviz as az
import bokeh.plotting as bkp
from bokeh.layouts import layout
import numpy as np

def make_square(arr):
    """Flat list to square"""
    n = np.ceil(len(arr)**0.5)
    new_arr = arr + [None for _ in range(n**2 - len(arr))]
    return np.array(new_arr).reshape(n,n).tolist()

az.backends.output_file("pair_plot_centered_eight.html")

axes = az.plot_pair(data, backend="bokeh", divergences=True, show=False)
ravelled_axes = [ax for ax in axes.ravel() if ax is not None]
padded_square = make_square(ravelled_axes)
bkp.show(layout(padded_square, sizing_mode="stretch_both"))
```

ahartikainen:
I think this has the main functionality. We need to add examples at some point to docs.

@OriolAbril @canyon289 and others, any better names / functionality for the row / column hack?

Should we use ncolumn / nrow structure --> 3column 2column 5row 6row etc for rcParam? How should we handle those cases?

row or column in  param --> regex --> if not found --> normal test?

fyi

    3column --> three columns --> shape == (-1, 3)

OriolAbril:
LGTM

PR #1090, opened 2020-02-26
[WIP] Add new groups to io_pyro

nitishp25:
## Description

Add the following groups to `io_pyro`:
- [x] `predictions`
- [x] `constant_data`
- [x] `predictions_constant_data`

Tests:
- [x] `predictions`
- [x] `constant_data`
- [x] `predictions_constant_data`

## Checklist

### DIFF
--- a/arviz/data/io_pyro.py
+++ b/arviz/data/io_pyro.py
@@ -4,7 +4,7 @@ import numpy as np
 from packaging import version
 import xarray as xr
 
-from .inference_data import InferenceData
+from .inference_data import InferenceData, concat
 from .base import requires, dict_to_dataset, generate_dims_coords, make_attrs
 from .. import utils
 
@@ -21,7 +21,15 @@ class PyroConverter:
     ndraws = None  # type: int
 
     def __init__(
-        self, *, posterior=None, prior=None, posterior_predictive=None, coords=None, dims=None
+        self,
+        *,
+        posterior=None,
+        prior=None,
+        posterior_predictive=None,
+        predictions=None,
+        coords=None,
+        dims=None,
+        pred_dims=None
     ):
         """Convert Pyro data into an InferenceData object.
 
@@ -33,10 +41,14 @@ class PyroConverter:
             Prior samples from a Pyro model
         posterior_predictive : dict
             Posterior predictive samples for the posterior
+        predictions: dict
+            Out of sample predictions
         coords : dict[str] -> list[str]
             Map of dimensions to coordinates
         dims : dict[str] -> list[str]
             Map variable names to their coordinates
+        pred_dims: dict
+            Dims for predictions data. Map variable names to their coordinates.
         """
         self.posterior = posterior
         if posterior is not None:
@@ -45,9 +57,14 @@ class PyroConverter:
             self.nchains = self.ndraws = 0
         self.prior = prior
         self.posterior_predictive = posterior_predictive
+        self.predictions = predictions
         self.coords = coords
         self.dims = dims
+        self.pred_dims = pred_dims
         import pyro
+
+        if self.predictions is not None and self.pred_dims is None:
+            raise ValueError("Prediction dims are needed for predictions group.")
 
         self.pyro = pyro
         if posterior is not None:
@@ -107,11 +124,10 @@ class PyroConverter:
                 return None
         return dict_to_dataset(data, library=self.pyro, coords=self.coords, dims=self.dims)
 
-    @requires("posterior_predictive")
-    def posterior_predictive_to_xarray(self):
-        """Convert posterior_predictive samples to xarray."""
+    def translate_posterior_predictive_dict_to_xarray(self, dct, dims):
+        """Convert posterior_predictive or prediction samples to xarray."""
         data = {}
-        for k, ary in self.posterior_predictive.items():
+        for k, ary in dct.items():
             ary = ary.detach().cpu().numpy()
             shape = ary.shape
             if shape[0] == self.nchains and shape[1] == self.ndraws:
@@ -124,7 +140,19 @@ class PyroConverter:
                     "posterior predictive shape not compatible with number of chains and draws."
                     "This can mean that some draws or even whole chains are not represented."
                 )
-        return dict_to_dataset(data, library=self.pyro, coords=self.coords, dims=self.dims)
+        return dict_to_dataset(data, library=self.pyro, coords=self.coords, dims=dims)
+
+    @requires("posterior_predictive")
+    def posterior_predictive_to_xarray(self):
+        """Convert posterior_predictive samples to xarray."""
+        return self.translate_posterior_predictive_dict_to_xarray(
+            self.posterior_predictive, self.dims
+        )
+
+    @requires("predictions")
+    def predictions_to_xarray(self):
+        """Convert predictions (out of sample predictions) to xarray."""
+        return self.translate_posterior_predictive_dict_to_xarray(self.predictions, self.pred_dims)
 
     def priors_to_xarray(self):
         """Convert prior samples (and if possible prior predictive too) to xarray."""
@@ -183,13 +211,25 @@ class PyroConverter:
                 "sample_stats": self.sample_stats_to_xarray(),
                 "log_likelihood": self.log_likelihood_to_xarray(),
                 "posterior_predictive": self.posterior_predictive_to_xarray(),
+                "predictions": self.predictions_to_xarray(),
                 **self.priors_to_xarray(),
                 "observed_data": self.observed_data_to_xarray(),
             }
         )
 
 
-def from_pyro(posterior=None, *, prior=None, posterior_predictive=None, coords=None, dims=None):
+def from_pyro(
+    posterior=None,
+    *,
+    prior=None,
+    posterior_predictive=None,
+    predictions=None,
+    coords=None,
+    dims=None,
+    pred_dims=None,
+    idata_origin=None,
+    inplace=False
+):
     """Convert Pyro data into an InferenceData object.
 
     Parameters
@@ -200,15 +240,45 @@ def from_pyro(posterior=None, *, prior=None, posterior_predictive=None, coords=N
         Prior samples from a Pyro model
     posterior_predictive : dict
         Posterior predictive samples for the posterior
+    predictions: dict
+        Out of sample predictions
     coords : dict[str] -> list[str]
         Map of dimensions to coordinates
     dims : dict[str] -> list[str]
         Map variable names to their coordinates
+    pred_dims: dict
+        Dims for predictions data. Map variable names to their coordinates.
+    idata_origin: InferenceData, optional
+        If supplied, then modify this inference data in place, adding `predictions` group.
+    inplace: boolean, optional
+        If idata_origin is supplied and inplace is True, merge the `predictions` into idata_origin,
+        rather than returning a fresh InferenceData object.
     """
-    return PyroConverter(
+    data = PyroConverter(
         posterior=posterior,
         prior=prior,
         posterior_predictive=posterior_predictive,
+        predictions=predictions,
         coords=coords,
         dims=dims,
+        pred_dims=pred_dims,
     ).to_inference_data()
+
+    if inplace and not idata_origin:
+        raise ValueError(
+            (
+                "Do not pass True for inplace unless passing"
+                "an existing InferenceData as idata_orig"
+            )
+        )
+    if idata_origin is None:
+        return data
+    else:
+        if inplace:
+            concat([idata_origin, data], dim=None, inplace=True)
+            return idata_origin
+        else:
+            # if we are not returning in place, then merge the old groups into the new inference
+            # data and return that.
+            concat([data, idata_origin], dim=None, copy=True, inplace=True)
+            return data
--- a/arviz/data/io_pyro.py
+++ b/arviz/data/io_pyro.py
@@ -4,7 +4,7 @@ import numpy as np
 from packaging import version
 import xarray as xr
 
-from .inference_data import InferenceData, concat
+from .inference_data import InferenceData
 from .base import requires, dict_to_dataset, generate_dims_coords, make_attrs
 from .. import utils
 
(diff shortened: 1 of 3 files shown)


nitishp25:
`from_pyro` works with only predictions though it cannot retain (draws, chains) in this case. Works fine otherwise.

arviz/data/io_pyro.py
```diff
@@ -183,13 +211,25 @@ def to_inference_data(self):
...
+    posterior=None,
+    *,
+    prior=None,
+    posterior_predictive=None,
+    predictions=None,
+    coords=None,
+    dims=None,
+    pred_dims=None,
+    idata_origin=None,
+    inplace=False
```
OriolAbril: I would not add this to `from_pyro`, they could go to a `from_pyro_predictions` (to mimic pymc3 pattern) but I am not sure it is worth it to include such a function it yet, it would basically be only a `concat` call, so users can call `az.concat` directly

arviz/data/io_pyro.py
```diff
@@ -45,10 +57,15 @@ def __init__(
...
         self.prior = prior
         self.posterior_predictive = posterior_predictive
+        self.predictions = predictions
         self.coords = coords
         self.dims = dims
+        self.pred_dims = pred_dims
         import pyro
 
+        if self.predictions is not None and self.pred_dims is None:
+            raise ValueError("Prediction dims are needed for predictions group.")
```
OriolAbril: dims are not needed, if not present they will get some default names by ArviZ internal functions.

OriolAbril:
> Should I add arguments for pred_chains and pred_draws for this case?

Maybe an nchains argument? It should be enough with only one, as the number of samples is already a dimension

arviz/data/io_pyro.py
nitishp25: But the thing is, the `predictions` variables have the same name as the `posterior_predictive` variables, so it uses the `posterior_predictive` coords which has different value and causes error.

OriolAbril: I am not sure I follow, predictions should not have access to `dims` at any point, only to `pred_dims`. Then the possible cases are:
* pred dims is None -> default dim and coord names
* pred dims is a dict -> use pred dims

arviz/data/io_pyro.py
```diff
@@ -45,10 +66,18 @@ def __init__(
             self.nchains = self.ndraws = 0
```
OriolAbril: I have just seen the `if posterior is not None` and chain, draw definition is repeated below. This one can be removed

arviz/data/io_pyro.py
```diff
@@ -45,10 +66,18 @@ def __init__(
...
         self.prior = prior
         self.posterior_predictive = posterior_predictive
+        self.predictions = predictions
+        self.constant_data = constant_data
+        self.predictions_constant_data = predictions_constant_data
         self.coords = coords
         self.dims = dims
+        self.pred_dims = pred_dims
+        self.num_chains = num_chains
```
OriolAbril: I would handle this in the `else: self.nchains = self.ndraws = 0` below, to set self.nchains to num_chains and then get the number of draws from the predictions, posterior predictive or prior (I think io_pymc does something similar).

arviz/data/io_pyro.py
```diff
@@ -107,24 +136,38 @@ def log_likelihood_to_xarray(self):
...
+    @requires("posterior_predictive")
+    def posterior_predictive_to_xarray(self):
+        """Convert posterior_predictive samples to xarray."""
+        return self.translate_posterior_predictive_dict_to_xarray(
+            self.posterior_predictive, self.dims
+        )
+
+    @requires("predictions")
+    def predictions_to_xarray(self):
+        """Convert predictions (out of sample predictions) to xarray."""
```
OriolAbril: I would either remove the out of sample predictions or change it to out of sample posterior predictive

arviz/data/io_pyro.py
nitishp25: Yes, both predictions and predictions_constant_data use `pred_dims` instead of `dims`. I have raised that error just in case the user passes the predictions but does not pass `pred_dims`.

All other groups are using the default `dims`, only the `predictions` and `predictions_constant_data` groups need `pred_dims` because their variable names are same as `posterior_predictive` and `constant_data` respectively but their data has actually different dimensions. So they cannot use the default `dims` in any case. You can see an example [here](https://github.com/nitishp25/arviz-playground/blob/master/arviz-pyro-example.ipynb).

Are we on the same page now? Other groups will always use the default `dims` but predictions and predictions_constant_data must always use `pred_dims` right?

OriolAbril: Yes, my point is that dims can be None, which makes ArviZ use some defaults, pred_dims should have the option of being none and generating default values, not using dims, generating the defaults that correspond to the dataset. I am not sure what I am missing, I thought default dims were generated on a dataset basis, not on an inference data basis

nitishp25: Ohh yes dims are generated on dataset basis.

> * pred dims is None -> default dim and coord names

Sorry I got lost here, thought that you were talking about `self.dims` :)

nitishp25:
Can I now add the tests?

arviz/data/io_pyro.py
```diff
@@ -33,22 +44,36 @@ def __init__(
...
+        predictions_constant_data: dict
+            Constant data used for out-of-sample predictions.
         coords : dict[str] -> list[str]
             Map of dimensions to coordinates
         dims : dict[str] -> list[str]
             Map variable names to their coordinates
+        pred_dims: dict
+            Dims for predictions data. Map variable names to their coordinates.
+        num_chains: int
+            Number of chains used for sampling. Only needed when posterior is not provided.
```
OriolAbril: Maybe "ignored if posterior is present instead"

arviz/data/io_pyro.py
OriolAbril: Also, as it is only used in init, it does not need to be saved in self

arviz/data/io_pyro.py
```diff
@@ -57,8 +82,25 @@ def __init__(
...
+                # pylint: disable=line-too-long
+                raise ValueError(
+                    """When constructing InferenceData must have at least
+                                    one of trace, prior, posterior_predictive or predictions."""
+                )
+            aelem = arbitrary_element(get_from)
+            self.ndraws = aelem.shape[0] // self.num_chains
         else:
-            self.nchains = self.ndraws = 0
+            raise ValueError("`num_chains` is needed if trace is not given.")
```
OriolAbril: I would not raise an error, for example with prior, sampling is generally not performed with multiple chains, from_pyro with only prior or predictions and no num chains should work (even if merging afterwards does not or does not work properly)

nitishp25: Then should I set chains and draws to 0?

OriolAbril: In the interest of not sending warnings when they are not needed, maybe it would be best to set `num_chains=1` in the function definitions. I think it will do the same as the else you are proposing given that num_chains is ignored if posterior is present.

OriolAbril:
> Can I now add the tests?

Definitely, I think you have already used `check_multiple_attrs`, so I'll skip the introduction, as always, ask if you have any doubt.

nitishp25:
@OriolAbril, do you think these tests are enough? Any tests for `pred_dims` or `num_chains` or any changes to be made in `test_inference_data_no_posterior`?

nitishp25:
> Another idea, please say if you think it would be useful or not. I have seen that constant data (either for model or for predictions) is generally already in a dict, however it is a dict of pytorch tensors, do you think it would be useful to allow a dict of tensors as constant data argument? It could be handled with a try except or checking `ìf hasattr("detach", value)`?

I think constant data already works with a dict of tensors. Is there any incompatibility?

OriolAbril:
> I think constant data already works with a dict of tensors. Is there any incompatibility?

Not sure, there should probably be a test for this (or just say it is not supported), like numpy, xarray hardly ever raises an error when creating arrays/datasets, however, the result can be unexpected in many cases:

```
np.array(((1, 2), (1, 4)))                                                                                                         
# array([[1, 2],
#        [1, 4]])

np.array(((1, 2), (1, 4, 3)))                                                                                                      
# array([(1, 2), (1, 4, 3)], dtype=object)

np.array(((1,2), (1, (4, 3))))                                                                                                   
# array([[1, 2],
#        [1, (4, 3)]], dtype=object)
```

We should make sure to add in docstring to convert to array or add a test checking that the generated dataset has the right shape, not length one and dtype object or something similar.

arviz/tests/external_tests/test_data_pyro.py
```diff
@@ -72,6 +79,38 @@ def test_inference_data_no_posterior(self, data, eight_schools_params):
...
+        predictions = Predictive(model, posterior_samples)(
+            predictions_data["J"], torch.from_numpy(predictions_data["sigma"]).float()
+        )
+        constant_data = {"J": 8, "sigma": eight_schools_params["sigma"]}
+        predictions_constant_data = {"J": 8, "sigma": predictions_data["sigma"]}
+        # only prior
+        inference_data = from_pyro(prior=prior)
+        test_dict = {"prior": ["mu", "tau", "eta"]}
+        fails = check_multiple_attrs(test_dict, inference_data)
+        assert not fails
```
OriolAbril: I think using

    assert not fails, "only prior: {}".format(fails)

will yield more informative error messages, not sure about formatting working though. I'll try to check this

arviz/tests/external_tests/test_data_pyro.py
```diff
@@ -123,3 +162,71 @@ def model_example_multiple_obs(y1=None, y2=None):
...
+    def test_inference_data_num_chains(self, data):
+        posterior_samples = data.obj.get_samples()
+        model = data.obj.kernel.model
+        predictions_data = {"J": 8, "sigma": np.array([5.0, 7.0, 12.0, 4.0, 6.0, 10.0, 3.0, 9.0])}
+        predictions = Predictive(model, posterior_samples)(
+            predictions_data["J"], torch.from_numpy(predictions_data["sigma"]).float()
+        )
+
+        inference_data = from_pyro(predictions=predictions, num_chains=2)
+        nchains = inference_data.predictions["obs"].shape[0]
```
OriolAbril: `inference_data.predictions.dims["chain"]` should return the length of chain dim

nitishp25:
> Not sure, there should probably be a test for this (or just say it is not supported), like numpy, xarray hardly ever raises an error when creating arrays/datasets, however, the result can be unexpected in many cases:
> 
> ```
> np.array(((1, 2), (1, 4)))                                                                                                         
> # array([[1, 2],
> #        [1, 4]])
> 
> np.array(((1, 2), (1, 4, 3)))                                                                                                      
> # array([(1, 2), (1, 4, 3)], dtype=object)
> 
> np.array(((1,2), (1, (4, 3))))                                                                                                   
> # array([[1, 2],
> #        [1, (4, 3)]], dtype=object)
> ```
> 
> We should make sure to add in docstring to convert to array or add a test checking that the generated dataset has the right shape, not length one and dtype object or something similar.

Sorry, I don't understand. The incorrect dims case would only be there when pyro samples data from the model without chains (like in case of `prior`) right? And if the user converts the above 3 cases to tensors then only the first one would work and other 2 wouldn't be converted to a tensor in the first place. Do you mean that the user could pass such examples without converting to a tensor or that these are the values after the passed tensors have been detached to arrays? I didn't get your point

OriolAbril:
I have played a little with tensors and it looks like xarray converts them automatically to arrays. 

My concern came from the fact that when prompted with the object:

```
tensor([ 4.5000,  6.0000,  7.0000, 12.0000, 18.0000], dtype=torch.float64)
``` 

it can either be converted to an array of dtype float and length 5 or to an array of dtype object and length 1 whose first position contains the whole tensor. Everything seems to work properly, even test error messages. Thanks!

arviz/tests/external_tests/test_data_pyro.py
```diff
@@ -123,3 +172,65 @@ def model_example_multiple_obs(y1=None, y2=None):
...
+            "prior": ["mu", "tau", "eta"],
+            "prior_predictive": ["obs"],
+        }
+        fails = check_multiple_attrs(test_dict, inference_data)
+        assert not fails
+
+        dims = inference_data.posterior_predictive["obs"].shape[2:]
+        pred_dims = inference_data.predictions["obs"].shape[2:]
+        assert dims == (8,)
+        assert pred_dims == (8,)
```
OriolAbril: `inference_data.posterior_predictive.dims["school"]` and `inference_data.predictions.dims["school_pred"]`, then the assert will be `dims == 8`

OriolAbril: This can be moved to the `test_inference_data` test above too, both start with the same code.

arviz/tests/external_tests/test_data_pyro.py
```diff
@@ -123,3 +172,65 @@ def model_example_multiple_obs(y1=None, y2=None):
...
+
+    def test_inference_data_num_chains(self, predictions_data):
+        predictions = predictions_data
+        inference_data = from_pyro(predictions=predictions, num_chains=2)
```
OriolAbril: `num_chains=chains` (use `chains` fixture imported from helpers to make sure chains is actually the number of chains in posterior)

arviz/tests/external_tests/test_data_pyro.py
```diff
@@ -70,6 +78,11 @@ def test_inference_data(self, data, eight_schools_params, predictions_data):
         fails = check_multiple_attrs(test_dict, inference_data)
         assert not fails
 
+        # test dims
+        dims = inference_data.posterior_predictive.dims["school"]
+        pred_dims = inference_data.predictions.dims["school_pred"]
+        assert dims == 8, pred_dims == 8
```
OriolAbril: these should still be one in each line, otherwise the `pred_dims == 8` will only be executed if assert dims fails.

nitishp25:
I'll add the groups to numpyro soon

OriolAbril:
Thanks!

PR #1092, opened 2020-02-27
Normalize dataframe dtypes

ahartikainen:
## Description

Normalize pandas dataframe columns.

Main points are:
- `int32` --> `int64` 
- `float32` --> `float64`

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 * Stabilize covariance matrix for `plot_kde_2d` (#1075)
 * Removed extra dim in `prior` data in `from_pyro` (#1071)
 * Moved CI and docs (build & deploy) to Azure Pipelines and started using codecov (#1080)
+* Normalize pandas dataframe dtypes (#1092)
 
 ### Deprecation
 * `from_pymc3` now requires PyMC3>=3.8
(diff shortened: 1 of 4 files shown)


ahartikainen:
@sethaxen I think this should work for Julia?

I don't think we have any mixed columns of int / float so I did not add any checks for that? (if we have any nan dtype should be float)

sethaxen:
Thanks! So far I don't see any change in the Julia function. Now that the code in https://github.com/arviz-devs/ArviZ.jl/pull/35 is refactored, I can do a little more digging into what I'm getting from ArviZ and try to figure out why it's not converting correctly.

Linking the relevant issues for future reference: https://github.com/arviz-devs/ArviZ.jl/issues/34

ahartikainen:
Did you try against this PR?

sethaxen:
Yes, and I confirmed by inserting an error in the `normalize_` function and seeing it raised. But when I removed the error, I saw no change in behavior.

ahartikainen:
So maybe we should close this PR and see if there are some other way to fix your problem?

sethaxen:
I think so for now, yes. I suspect there's an issue in the PyCall side. Thanks for trying this out!

PR #1101, opened 2020-03-03
Revise docstring for plot_density().

rpgoldman:
Previously, this specified the `ax` argument as taking an (optional)
axes.  This is incorrect: it takes a 2D array of axes or bokeh figures.

Perhaps we should explicitly point out that a 2D array is required.  The code requires *at least* 2D, but I'm not sure what would happen if one passed a higher-dimensional array.

### DIFF
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -47,8 +47,9 @@ def plot_autocorr(
     textsize: float
         Text size scaling factor for labels, titles and lines. If None it will be autoscaled based
         on figsize.
-    ax: axes, optional
-        Matplotlib axes or bokeh figures.
+    ax: numpy array of matplotlib axes or bokeh figures, optional
+        A 2D array of locations into which to plot the densities. If not supplied, Arviz will create
+        its own array of plot areas (and return it).
     backend: str, optional
         Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
     backend_config: dict, optional
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -47,7 +47,7 @@ def plot_autocorr(
     textsize: float
         Text size scaling factor for labels, titles and lines. If None it will be autoscaled based
         on figsize.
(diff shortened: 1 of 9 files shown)


rpgoldman:
> Yeah this is correct. Maybe comment that the order needs to be similar as the created axes array?

TBH, I am not sure I understand how `default_grid()` works well enough to explain the ordering.  It's clear that the size of the array must correspond with the number of items to be plotted, but beyond that 🤷🏻‍♂️. 

How does the programmer predict how many items there will be?  Presumably the number of variables times the size of the coordinates for each variable.

Can we say something about the order in which the plots will be emitted, so that the programmer can decide the shape of the array?  The order of variables in `var_names`, and then the order of dimensions in the xarray, I expect.

Shouldn't this take coordinates as an argument to select subsets, or do we expect the caller to subset the `InferenceData` themselves?

I will do another pass over this, and then we can merge.

rpgoldman:
@ahartikainen I modified several other docstrings where there are grids of axes.  I was not sure what to do about pairplot, though, since I don't understand what sort of compound of axes is required by this function, since it doesn't use `default_grid()`, and uses `plt.subplots()` with backend keyword args in a way I couldn't figure out.

Anyway, if you are ok with my corrections, please merge this.  Any more explanation would require someone else to do it: this is all I could do with my current level of understanding.

OriolAbril:
Is it a must for ax to be numpy array? Or a list would be converted to array?

rpgoldman:
I'll try to check.  I never bothered to worry about this because subplots gives me a numpy array back, so I have never tried to pass in a python list, and I'm not sure why one would.

This is a thing I don't really like about numpy: it's got such an expansive notion of what is "array-like" that it's hard to know what will or won't work without experimentation.  But that's my background speaking: I'm a computer scientist and not a statistician or experimental scientist.

rpgoldman:
@OriolAbril I checked and the `ax` arguments get pushed through `np.atleast_2D`, which accepts array-like arguments, so I have revised the docstrings per your suggestion.

If that's good enough, let's squash merge this.

Issue #855, opened 2019-10-31
Unexpected behavior when using `pm.hpd` with multidimensional (ndim>2)posterior arrays


narendramukherjee:
**Describe the bug**
I am using `hpd` on multi-dimensional posteriors with larger than 2 dimensions. One example of where this can come up is if I use a 2D array to represent the coefficients for the interaction between levels of 2 different predictors. The output trace of this 2D array of coefficients from PyMC3 will now be 3D, with the 0th dimension representing the MCMC samples.

I find that with a 3D trace/array, `hpd` treats dimension 1 as the MCMC dimension (instead of dimension 0, the actual MCMC dimension) 

**To Reproduce**
```
#12000 MCMC samples from a 10x3 dimensional distribution
a = np.random.normal(size=(12000, 10, 3))
pm.hpd(a).shape
```

**Output: (3, 12000, 2)** 
**Should instead be: (10, 3, 2)**

This happens because the assumption in `hpd` is that `ndim==2` if `ndim>1`: https://github.com/arviz-devs/arviz/blob/master/arviz/stats/stats.py#L346

`for row in ary.T` works properly for arrays that have 2 dimensions, but for anything larger than that, it infers the wrong MCMC dimension.

**Suggestion:** Will it make sense to have an `axis` parameter to specify the axis of the MCMC samples, much like say `np.mean` etc?

ahartikainen:
Possible solution:

First dimension = chains
Second dimension = draws
3rd --> = parameter dimensions.

You need to add one dimension to you data.

Or use pymc3 trace object with `vars` info?

AlexAndorra:
Ah yes [I had to do that](https://github.com/aloctavodia/BAP/blob/master/extras/multinomial_ppcs.ipynb) the other day. My solution was to slice the array you give to `az.hpd`: `az.hpd(a[:, :, 0]).shape` should give you (10, 2) -- I'm using `az` instead of `pm` here as PyMC is based on ArviZ for diagnostics.

So, if the `3`in the shape corresponds to categories for instance, you do `az.hpd(a[:, :, i])` for each of them and you should get a (10, 3, 2) by concatenation and transposition 🎉 

That being said, implementing an `axis` paremeter could indeed be interesting -- if easily doable -- but I'm no expert here.

narendramukherjee:
[...]

PR #1107, opened 2020-03-04
extend make_ufunc and improve wrap_xarray_ufunc defaults

OriolAbril:
## Description
Currently `az.hpd` uses a recursive pattern to compute hpd values for 1d or 2d arrays. It would be great to create an `_hpd` function that is then used by `az.hpd` via `wrap_xarray_ufunc` (like it is done with ess, rhat...). 

I have realized that `make_ufunc` does not support the function returning extra output dimensions, which would be needed for hpd to work and return a single array. This extends `make_ufunc` and improves slightly the defaults for `wrap_xarray_ufunc`. 

There is one drawback/divergence from the ufunc default arguments. `xr.apply_ufunc` performs a separate call of the ufunc for each data variable (in the case of datasets) but the kwargs passed to the ufunc are always the same for all data variables. This behaviour makes the `out` argument not usable when the shapes of each data variable differ, the same out array is passed to all calls. To sort this out I thought about adding an `out_shape` argument which will be present to all ufuncs generated with `make_ufunc`. This argument indicates the shape of the extra dims (i.e. in hpd, `out_shape=(2,)`). 

Therefore, this PR allows to do things like:

```python
import arviz as az
data = az.load_arviz_data("centered_eight").posterior
az.wrap_xarray_ufunc(
    az.hpd, 
    data, 
    output_core_dims =[['hpd']], 
    func_kwargs={"out_shape": (2,)}
)
# <xarray.Dataset>
# Dimensions:  (hpd: 2, school: 8)
# Coordinates:
#  * school   (school) object 'Choate' 'Deerfield' ... "St. Paul's" 'Mt. Hermon'
# Dimensions without coordinates: hpd
# Data variables:
#     mu       (hpd) float64 -2.118 10.4
#     theta    (school, hpd) float64 -3.707 17.34 -4.039 ... 16.92 -5.665 15.27
#     tau      (hpd) float64 0.5692 9.386

az.wrap_xarray_ufunc(
    az.hpd, data, 
    input_core_dims=[["draw"]], 
    output_core_dims =[['hpd']], 
    func_kwargs={"out_shape": (2,)}
)
# <xarray.Dataset>
# Dimensions:  (chain: 4, hpd: 2, school: 8)
# Coordinates:
#   * chain    (chain) int64 0 1 2 3
#   * school   (school) object 'Choate' 'Deerfield' ... "St. Paul's" 'Mt. Hermon'
# Dimensions without coordinates: hpd
# Data variables:
#     mu       (chain, hpd) float64 -1.996 9.312 -2.358 ... 11.52 -0.7842 9.985
#     theta    (chain, school, hpd) float64 -4.077 17.44 -3.104 ... -3.708 14.21
#     tau      (chain, hpd) float64 0.6768 8.881 1.013 9.1 ... 10.64 0.5001 8.994
```

It should be feasible to convert current `az.hpd` to `_hpd` removing the recursion and adding a flatten and create a new `az.hpd` that like `ess` accepts either arrays, datasets or inference data objects, coordinate values for hpd dim could also be added... The two examples above showcase that it could even have a `combined` argument to allow users to make the calculation on a per chain basis without using wrap_xarray_ufunc directly.

The PR has no tests yet, maybe we could add them in another PR? Or someone wants to hijack the PR and add the tests here? I'm not sure I'll have much time to thoroughly test this in the near future.

Related to #855. cc @percygautam @ahartikainen @aloctavodia 

## Checklist

- [ ] New features are properly documented (with an example if appropriate)? **DOES NOT!**
- [ ] Includes new or updated tests to cover the new feature **DOES NOT!**

### DIFF
--- a/arviz/stats/stats_utils.py
+++ b/arviz/stats/stats_utils.py
@@ -209,7 +209,7 @@ def wrap_xarray_ufunc(
         func_kwargs = {}
 
     kwargs.setdefault(
-        "input_core_dims", tuple(("chain", "draw") for _ in range(len(func_args) + 1))
+        "input_core_dims", tuple(("chain", "draw") for _ in range(len(func_args) + len(datasets)))
     )
     ufunc_kwargs.setdefault("n_dims", len(kwargs["input_core_dims"][-1]))
     kwargs.setdefault("output_core_dims", tuple([] for _ in range(ufunc_kwargs.get("n_output", 1))))
--- a/arviz/stats/stats_utils.py
+++ b/arviz/stats/stats_utils.py
@@ -114,27 +114,35 @@ def make_ufunc(
     elif check_shape is None:
         check_shape = False
 
-    def _ufunc(*args, out=None, **kwargs):
+    def _ufunc(*args, out=None, out_shape=None, **kwargs):
         """General ufunc for single-output function."""
         arys = args[:n_input]
+        n_dims_out = None
         if out is None:
-            out = np.empty(arys[-1].shape[:-n_dims])
+            if out_shape is None:
+                out = np.empty(arys[-1].shape[:-n_dims])
+            else:
+                out = np.empty((*arys[-1].shape[:-n_dims], *out_shape))
+                n_dims_out = -len(out_shape)
         elif check_shape:
             if out.shape != arys[-1].shape[:-n_dims]:
                 msg = "Shape incorrect for `out`: {}.".format(out.shape)
                 msg += " Correct shape is {}".format(arys[-1].shape[:-n_dims])
                 raise TypeError(msg)
-        for idx in np.ndindex(out.shape):
+        for idx in np.ndindex(out.shape[:n_dims_out]):
             arys_idx = [ary[idx].ravel() if ravel else ary[idx] for ary in arys]
             out[idx] = np.asarray(func(*arys_idx, *args[n_input:], **kwargs))[index]
         return out
 
-    def _multi_ufunc(*args, out=None, **kwargs):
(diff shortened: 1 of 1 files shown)


percygautam:
@OriolAbril I'd like to take on the test work. I am currently working on #855, using the `out_shape`, the issue can be solved:
```
import xarray as xr
import arviz as az
import numpy as np
a = np.random.normal(size=(12000, 10, 3))
a = xr.DataArray(a).to_dataset(name= 'a')
az.wrap_xarray_ufunc(az.hpd, a, input_core_dims=[["dim_0"]], output_core_dims =[['hpd']], func_kwargs={"out_shape": (2,)})

# <xarray.Dataset>
# Dimensions:  (dim_1: 10, dim_2: 3, hpd: 2)
# Dimensions without coordinates: dim_1, dim_2, hpd
# Data variables:
#     a        (dim_1, dim_2, hpd) float64 -1.839 1.886 -1.82 ... -1.916 1.835
```

arviz/stats/stats_utils.py
```diff
@@ -114,27 +114,35 @@ def make_ufunc(
     elif check_shape is None:
         check_shape = False
 
-    def _ufunc(*args, out=None, **kwargs):
+    def _ufunc(*args, out=None, out_shape=None, **kwargs):
         """General ufunc for single-output function."""
         arys = args[:n_input]
```
canyon289: Where does n_input come from?

canyon289: nevermind, I see this is a function within a function

OriolAbril:
+1 on the notebook!

PR #1109, opened 2020-03-05
Fix model warning.

rpgoldman:
The check for whether the PyMC3 model has been found looked at the model variable, instead of the model property.

This fixes a false positive deprecation warning.


rpgoldman:
@OriolAbril Please have a look at this.  I think it's good to merge, but  it would be good for you to review it.

Issue #1105, opened 2020-03-03
New example pairplots


agustinaarroyuelo:
Example plots for `plot_pair` function are outdated due to changes done in #1079.

agustinaarroyuelo:
I will start working on this

PR #1110, opened 2020-03-06
[WIP] add new examples for plot_pair

agustinaarroyuelo:
## Description
I added new examples for plot_pair. I tried to showcase in them the possibility of obtaining cornerplots, as mentioned in #1079. I also fixed some bugs. This PR solves #1105 . 

I tried to build the documentation locally, but I was having trouble with docker. I was later explained that this isn't a requirement. 

## Checklist

### DIFF
--- a/examples/matplotlib/mpl_plot_pair_point_estimate.py
+++ b/examples/matplotlib/mpl_plot_pair_point_estimate.py
@@ -1,5 +1,5 @@
 """
-Point Estimate Pairplor
+Point Estimate Pairplot
 ===============
 
 _thumb: .2, .5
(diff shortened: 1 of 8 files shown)


examples/matplotlib/mpl_plot_pair_point_estimate.py
```diff
@@ -0,0 +1,24 @@
+"""
+Point Estimate Pairplor
```
hectormz: Hi @agustinaarroyuelo I was checking out your new examples and noticed your docstrings for the example `.py` files say Pairplo**r** instead of what I assume should be Pairplo**t**.

And the files don't end in newlines, which PEP8 doesn't like

agustinaarroyuelo: Hi, thanks for letting me know!

OriolAbril:
Maybe we should also update the plot_joint example to use `plot_pair` and get a similar looking plot? Given that we are deprecating `plot_joint` we should update docs to not encourage its usage anymore.

agustinaarroyuelo:
Should the example include a deprecation message?

Issue #1112, opened 2020-03-08
Skip tests relying on absent optional/external libaries


hectormz:
**Describe the issue**
`Arviz` is meant to use/interact with external libraries that are optional, found in `requirements-external.txt`, `requirements-optional.txt`, with tests found in `/tests/external-tests/`. I'd propose making these tests optional when the other libraries aren't installed, so someone not touching those areas could run tests locally without installing all the packages.

**Suggested change**
In [`pyjanitor`](https://github.com/ericmjl/pyjanitor/blob/dev/tests/spark/conftest.py) we support the use of some optional, external libraries and skip tests if the libraries can't be loaded.

I propose modifying the test files to something like this:
```python
try:
    import emcee
except ImportError:
    emcee= None

@pytest.mark.skipif(emcee is None, reason="emcee tests only required for CI")
def some_emcee_test():
```

CI should still run as is, but it'd be a bit easier for new people to get involved and test without the external libraries. 

One issue with this is at least once instance with this (for `pyro` at least) is when a `skipif` decorator is also checking for `pyro.version` which will cause an error when it is `None` etc.

ahartikainen:
We could probably do that.

I currently run only the testfile. It might be a bit inconvenient.

    pytest arviz/.../pystan_tests_file.py

OriolAbril:
I also like the idea, it'd be great if you can make a PR, in the meantime, another alternative is to use the `k` flag, something like `pytest arviz/tests/ -k "not bokeh and not pyro" ` would skip all tests with bokeh or pyro in name.

hectormz:
@OriolAbril  I wasn't aware of that skip, is that built into `pytest`? nice!

I was looking into the `pytest.mark.skipif` documentation and found there is an option to skip tests based on missing a dependency: [`pytest.importorskip`](https://docs.pytest.org/en/latest/skipping.html?highlight=skipping#skipping-on-a-missing-import-dependency)

Assuming that our tests for external libraries are in separate files, this might work well. In `test_data_pyro.py`, the imports
```python
import torch
import pyro
from pyro.infer import Predictive
```

can be replaced with
```python
torch = pytest.importorskip("torch")
pyro = pytest.importorskip("pyro")
Predictive = pytest.importorskip("pyro.infer.Predictive")
```

[...]

PR #1113, opened 2020-03-09
Skip tests for optional/extra dependencies when not installed

hectormz:
## Description

When running all tests locally `ImportError`s are thrown if optional/extra dependencies are not installed. By skipping these tests when not installed, users can still run the entire test suite if desired. The entire test suite should still be run via CI.

This behavior is largely achieved by changing
```python
import xx
```
to
```python
xx = pytest.importorskip(xx)
```
which will try to import module `xx`, and will skip all tests in file if the module is not installed.

Individual tests may also be skipped using the decorator:
```python
@pytest.mark.skipif(importlib.util.find_spec("xx") is None, reason="xx tests only required for CI")

```

The last remaining issue is `numba`, which is used in 20+ tests, used in a variety of files. One option is to make `numba` an entry in `requirements-dev.txt` instead of `requirements-optional.txt`, or add the skip decorators to each of the 20+ tests.

This PR addresses  Issue #1112 .

## Checklist

- [X] Changes are listed in [changelog](https://github.com/arviz-devs/arviz/blob/master/CHANGELOG.md#v0xx-unreleased)

### DIFF
--- a/arviz/tests/external_tests/test_data_pystan.py
+++ b/arviz/tests/external_tests/test_data_pystan.py
@@ -18,12 +18,12 @@ from ..helpers import (  # pylint: disable=unused-import
     pystan_version,
 )
 
-# Check if either stan or pystan is installed
+# Check if either pystan or pystan3 is installed
 pystan_installed = (importlib.util.find_spec("pystan") is not None) or (
     importlib.util.find_spec("stan") is not None
 )
 pytestmark = pytest.mark.skipif(
-    ~pystan_installed, reason="test requires stan/pystan which is not installed"
+    ~pystan_installed, reason="test requires pystan/pystan3 which is not installed"
 )
 
 
--- a/arviz/tests/external_tests/test_data_pystan.py
+++ b/arviz/tests/external_tests/test_data_pystan.py
@@ -21,7 +21,9 @@ from ..helpers import (  # pylint: disable=unused-import
 pystan_installed = (importlib.util.find_spec("pystan") is not None) or (
     importlib.util.find_spec("stan") is not None
 )
-pytestmark = pytest.mark.skipif(~pystan_installed, reason="pystan/stan tests only required for CI")
+pytestmark = pytest.mark.skipif(
+    ~pystan_installed, reason="test requires stan/pystan which is not installed"
+)
 
 
 class TestDataPyStan:
--- a/arviz/tests/external_tests/test_data_pystan.py
+++ b/arviz/tests/external_tests/test_data_pystan.py
@@ -24,7 +24,7 @@ pystan_installed = (importlib.util.find_spec("pystan") is not None) or (
     importlib.util.find_spec("stan") is not None
 )
 pytestmark = pytest.mark.skipif(
-    not pystan_installed & (not running_on_ci()),
+    not (pystan_installed | running_on_ci()),
     reason="test requires pystan/pystan3 which is not installed",
 )
 
--- a/arviz/tests/external_tests/test_data_pystan.py
+++ b/arviz/tests/external_tests/test_data_pystan.py
@@ -1,9 +1,12 @@
 # pylint: disable=no-member, invalid-name, redefined-outer-name
+import importlib
 from collections import OrderedDict
+
 import numpy as np
 import pytest
 
 from arviz import from_pystan
+
 from ...data.io_pystan import get_draws, get_draws_stan3  # pylint: disable=unused-import
 from ..helpers import (  # pylint: disable=unused-import
     chains,
@@ -13,6 +16,12 @@ from ..helpers import (  # pylint: disable=unused-import
     load_cached_models,
     pystan_version,
 )
+
+# Check if either stan or pystan is installed
+pystan_installed = (importlib.util.find_spec("pystan") is not None) or (
+    importlib.util.find_spec("stan") is not None
+)
+pytestmark = pytest.mark.skipif(~pystan_installed, reason="pystan/stan tests only required for CI")
 
 
 class TestDataPyStan:
@@ -217,7 +226,8 @@ class TestDataPyStan:
     @pytest.mark.skipif(pystan_version() != 2, reason="PyStan 2.x required")
     def test_index_order(self, data, eight_schools_params):
         """Test 0-indexed data."""
-        import pystan  # pylint: disable=import-error
+        # Skip test if pystan not installed
+        pystan = pytest.importorskip("pystan")  # pylint: disable=import-error
 
         fit = data.model.sampling(data=eight_schools_params)
         if pystan.__version__ >= "2.18":
--- a/arviz/tests/external_tests/test_data_pystan.py
+++ b/arviz/tests/external_tests/test_data_pystan.py
@@ -13,6 +13,7 @@ from ..helpers import (  # pylint: disable=unused-import
     check_multiple_attrs,
     draws,
     eight_schools_params,
+    importorskip,
     load_cached_models,
     pystan_version,
 )
@@ -229,7 +230,7 @@ class TestDataPyStan:
     def test_index_order(self, data, eight_schools_params):
         """Test 0-indexed data."""
         # Skip test if pystan not installed
-        pystan = pytest.importorskip("pystan")  # pylint: disable=import-error
+        pystan = importorskip("pystan")  # pylint: disable=import-error
 
         fit = data.model.sampling(data=eight_schools_params)
         if pystan.__version__ >= "2.18":
--- a/arviz/tests/external_tests/test_data_pystan.py
+++ b/arviz/tests/external_tests/test_data_pystan.py
@@ -16,6 +16,7 @@ from ..helpers import (  # pylint: disable=unused-import
     importorskip,
     load_cached_models,
     pystan_version,
+    running_on_ci,
 )
 
 # Check if either pystan or pystan3 is installed
@@ -23,7 +24,8 @@ pystan_installed = (importlib.util.find_spec("pystan") is not None) or (
     importlib.util.find_spec("stan") is not None
 )
 pytestmark = pytest.mark.skipif(
-    ~pystan_installed, reason="test requires pystan/pystan3 which is not installed"
+    ~pystan_installed & ~running_on_ci(),
+    reason="test requires pystan/pystan3 which is not installed",
 )
 
 
--- a/arviz/tests/external_tests/test_data_pystan.py
+++ b/arviz/tests/external_tests/test_data_pystan.py
@@ -24,7 +24,7 @@ pystan_installed = (importlib.util.find_spec("pystan") is not None) or (
     importlib.util.find_spec("stan") is not None
 )
 pytestmark = pytest.mark.skipif(
-    ~pystan_installed & ~running_on_ci(),
+    not pystan_installed & (not running_on_ci()),
     reason="test requires pystan/pystan3 which is not installed",
 )
 
--- a/arviz/tests/helpers.py
+++ b/arviz/tests/helpers.py
@@ -16,8 +16,6 @@ from packaging.version import Version
 from ..data import InferenceData, from_dict
 
 _log = logging.getLogger(__name__)
-# ARVIZ_CI_MACHINE is True if tests run on CI, where ARVIZ_CI_MACHINE env variable exists
-ARVIZ_CI_MACHINE = os.environ.get("ARVIZ_CI_MACHINE") is not None
 
 
 @pytest.fixture(scope="module")
@@ -608,6 +606,8 @@ def importorskip(
     Example::
         docutils = pytest.importorskip("docutils")
     """
+    # ARVIZ_CI_MACHINE is True if tests run on CI, where ARVIZ_CI_MACHINE env variable exists
+    ARVIZ_CI_MACHINE = os.environ.get("ARVIZ_CI_MACHINE") is not None
     if ARVIZ_CI_MACHINE:
         __import__(modname)
         mod = sys.modules[modname]
--- a/arviz/tests/helpers.py
+++ b/arviz/tests/helpers.py
@@ -623,8 +623,7 @@ def importorskip(
             # of existing directories with the same name we're trying to
             # import but without a __init__.py file
             warnings.simplefilter("ignore")
-            __import__(modname)
-        mod = sys.modules[modname]
+            mod = importlib.import_module(modname)
         if minversion is None:
             return mod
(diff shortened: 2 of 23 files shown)


hectormz:
Hi @OriolAbril thanks!

 I'll fix the `pylint` issues, and move the imports down.

I'll update the skipif messages to: "test requires xx which is not installed"

I'll look into moving all the `numba` files to their own file(s). If it looks a little more complicated organizing them in a clear way, I might move that to a new Issue/PR. Here are the failing `numba` tests currently:

```
FAILED tests/base_tests/test_diagnostics.py::test_numba_bfmi - ValueError: Numba is not installed
FAILED tests/base_tests/test_diagnostics.py::test_numba_rhat[rank] - ValueError: Numba is not installed
FAILED tests/base_tests/test_diagnostics.py::test_numba_rhat[split] - ValueError: Numba is not installed
FAILED tests/base_tests/test_diagnostics.py::test_numba_rhat[folded] - ValueError: Numba is not installed
FAILED tests/base_tests/test_diagnostics.py::test_numba_rhat[z_scale] - ValueError: Numba is not installed
FAILED tests/base_tests/test_diagnostics.py::test_numba_rhat[identity] - ValueError: Numba is not installed
FAILED tests/base_tests/test_diagnostics.py::test_numba_mcse[mean] - ValueError: Numba is not installed
FAILED tests/base_tests/test_diagnostics.py::test_numba_mcse[sd] - ValueError: Numba is not installed
FAILED tests/base_tests/test_diagnostics.py::test_numba_mcse[quantile] - ValueError: Numba is not installed
FAILED tests/base_tests/test_diagnostics.py::test_ks_summary_numba - ValueError: Numba is not installed
FAILED tests/base_tests/test_diagnostics.py::test_geweke_numba - ValueError: Numba is not installed
FAILED tests/base_tests/test_diagnostics.py::test_mcse_error_numba[True-1] - ValueError: Numba is not installed
FAILED tests/base_tests/test_diagnostics.py::test_mcse_error_numba[True-20] - ValueError: Numba is not installed
FAILED tests/base_tests/test_diagnostics.py::test_mcse_error_numba[False-1] - ValueError: Numba is not installed
FAILED tests/base_tests/test_diagnostics.py::test_mcse_error_numba[False-20] - ValueError: Numba is not installed
FAILED tests/base_tests/test_stats.py::test_summary_include_circ[True] - ValueError: Numba is not installed
FAILED tests/base_tests/test_stats.py::test_summary_include_circ[False] - ValueError: Numba is not installed
FAILED tests/base_tests/test_stats.py::test_numba_stats - ValueError: Numba is not installed
FAILED tests/base_tests/test_utils.py::test_utils_fixture - ModuleNotFoundError: No module named 'numba'
FAILED tests/base_tests/test_utils.py::test_conditional_jit_numba_decorator_keyword - ValueError: too many values to unpack (expected 2)
FAILED tests/base_tests/test_utils.py::test_numba_utils - ValueError: Numba is not installed
FAILED tests/base_tests/test_utils.py::test_numba_var[0-0] - ValueError: Numba is not installed
FAILED tests/base_tests/test_utils.py::test_numba_var[0-1] - ValueError: Numba is not installed
FAILED tests/base_tests/test_utils.py::test_numba_var[1-0] - ValueError: Numba is not installed
FAILED tests/base_tests/test_utils.py::test_numba_var[1-1] - ValueError: Numba is not installed
```

The easiest/laziest solution might be to create new files with `numba` tests:

`test_utils_numba.py`
`test_stats_numba.py`
`test_diagnostis_numba.py`

This would preserve the clarity of the original goal that `numba` is helping achieve.

hectormz:
Hi @OriolAbril I ended up moving the `numba` tests as I proposed. I tested locally with and without `numba` to verify passed or skipped tests

arviz/tests/helpers.py
```diff
@@ -571,9 +571,14 @@ def pystan_version():
     """
     try:
         import pystan  # pylint: disable=import-error
+        version = int(pystan.__version__[0])
     except ImportError:
-        import stan as pystan  # pylint: disable=import-error
-    return int(pystan.__version__[0])
+        try:
+            import stan as pystan  # pylint: disable=import-error
```
canyon289: Curious, why is this extra block of logic needed?

canyon289:
This pr is great. It does mean that the maintainers need to be extra vigilant with CI and ensure that no tests are skipped before merging. Do you know if theres a way with pytest to fail ir any tests are skipped?

arviz/tests/helpers.py
hectormz: @canyon289 In the event that neither `pystan` or `stan` are installed, there would be no imported `pystan`, which would then have no version attribute. I believe that this function is called via `test_data_pystan.py`, and will throw an error just while collecting the tests. Even though this file would eventually be skipped because neither libraries are installed via:
```python
pytestmark = pytest.mark.skipif(
    ~pystan_installed, reason="test requires stan/pystan which is not installed"
)
```

This was my way to handle potentially two exceptions, and also indicate that there is no form of `stan` installed by returning `None`. (I have neither `stan`/`pystan` package, so this was a solution that I could get to work for me)

hectormz:
Valid point @canyon289 ! There may be an option that is builtin to `pytest` (that I haven't found yet), but I came across [`pytest-error-for-skips`](https://pypi.org/project/pytest-error-for-skips/) :
>Pytest plugin to treat skipped tests as test failures.

>This is nice if you want to ensure that your CI tests really run all tests and don't skip tests because of missing dependencies.

>Simply execute your tests via `pytest --error-for-skips ...` and all skipped tests become test failures.

But I must imagine that there might be something completely integrated into `pytest`

hectormz:
[2016 `pytest` issue on the matter (#1364)](https://github.com/pytest-dev/pytest/issues/1364)

ahartikainen:
Could we have env variable which is also tested with the import func?

canyon289:
> Could we have env variable which is also tested with the import func?

Curious what this would do

OriolAbril:
I have to point out that some tests are skipped in CI, some features require the latest version of a given library which means that test can be executed in "external_latest" but not in "external_especial"

hectormz:
I'm guessing @ahartikainen is proposing to create an env variable on the CI machine like `ARVIZ_CI_MACHINE`. Instead of the `importorskip` calls at the top, we would do something like:

```python
if os.environ.get('ARVIZ_CI_MACHINE')=='True':
    ARVIZ_CI_MACHINE = True
else:
    ARVIZ_CI_MACHINE = False

try:
    import pystan
    pystan_installed = True
except ImportError:
    pystan_installed = False

pytestmark = pytest.mark.skipif(
    ~pystan_installed & ~ARVIZ_CI_MACHINE, reason="test requires stan/pystan which is not installed"
)
```

hectormz:
@OriolAbril so it would be a good idea to separate acceptable and unacceptable CI skips?

ahartikainen:
Maybe we could monkeypatch pytest.importskip function?

OriolAbril:
I have seen this example in pytest docs about importorskip usage: http://doc.pytest.org/en/latest/example/parametrize.html#indirect-parametrization-of-optional-implementations-imports, maybe we could base the monkeypatch on something similar? 

If CI import and return module, if not in CI use importorskip?

hectormz:
I've been looking at monkeypatching `pytest.importskip` (and `_pytest.outcomes.importorskip`) and haven't been able to do so. I'm not sure `pytest` can monkeypatch itself. I've done this in the context of `conftest.py` for all tests, or individual tests. I started with just trying to delete `importorskip` entirely:

```python
@pytest.fixture(autouse=True)
def no_importorskip(monkeypatch):
    """Remove pytest.importorskip for all tests."""
    monkeypatch.delattr("pytest.importorskip")
```

I _think_ if we want to go down this route, we can implement our own `importorskip()` (take original code and add ability to check for env variable, and stick it in `arviz.utils` etc).

ahartikainen:
yeah, that sounds like the correct option

hectormz:
@ahartikainen let me know of this implementation. I also added tests for the custom `importorskip` to test local and CI experience

arviz/tests/external_tests/test_data_pystan.py
```diff
@@ -1,19 +1,31 @@
...
     load_cached_models,
     pystan_version,
 )
 
+# Check if either stan or pystan is installed
+pystan_installed = (importlib.util.find_spec("pystan") is not None) or (
+    importlib.util.find_spec("stan") is not None
+)
+pytestmark = pytest.mark.skipif(
+    ~pystan_installed, reason="test requires stan/pystan which is not installed"
```
ahartikainen: Maybe this could be

`... requires pystan/pystan3 which is not ...`

arviz/tests/helpers.py
ahartikainen: Yeah, this sounds fine, maybe this could be `import stan` --> `stan.__version__`

arviz/tests/helpers.py
```diff
@@ -571,11 +573,63 @@ def pystan_version():
...
+        import warnings
+
+        compile(modname, "", "eval")  # to catch syntaxerrors
+
+        with warnings.catch_warnings():
+            # make sure to ignore ImportWarnings that might happen because
+            # of existing directories with the same name we're trying to
+            # import but without a __init__.py file
+            warnings.simplefilter("ignore")
+            __import__(modname)
```
ahartikainen: is there any difference between these

    __import__(modname)
    mod = sys.modules[modname]

vs

    mod = importlib.import_module(modname)

arviz/tests/external_tests/test_data_pystan.py
hectormz: Sure I'll make that change

arviz/tests/helpers.py
hectormz: I think they should be mostly similar for our uses. I used the former as I copied the existing code in `pytest.importorksip`.

But I can use `importlib.import_module`. It might clear up the failures in CI to import `jax.random.PRNGKey` and `pyro.infer.Predictive`.


I can't seem to easily install `jaxlib` on Windows, so I can't test this out locally.

hectormz:
In addition to the points raised by @ahartikainen , I'm looking to see why these CI imports failed:
```
ModuleNotFoundError: No module named 'jax.random.PRNGKey'; 'jax.random' is not a package

ModuleNotFoundError: No module named 'pyro.infer.Predictive'
```

arviz/tests/helpers.py
hectormz: Hi @ahartikainen is that just regarding line 580? Still try to import `pystan`, and if not present try to import `stan`, but no need to `import stan as pystan`?

hectormz:
@ahartikainen I don't think `test_precompile_models()` in `helpers.py` is being discovered and run with `pytest arviz\tests` (I searched through it in verbose log). It might be better to move it to the `test_helpers.py` file I created.

ahartikainen:
> @ahartikainen I don't think `test_precompile_models()` in `helpers.py` is being discovered and run with `pytest arviz\tests` (I searched through it in verbose log). It might be better to move it to the `test_helpers.py` file I created.

tbh, it is not a test, but a way to precompile models for CI, so at some point, we can start gathering performance data for our tests and they should be reproducible, in most cases.

I cheated with `test_` so pytest would work with it.

So users should not run it (without manually calling it)

ahartikainen:
> In addition to the points raised by @ahartikainen , I'm looking to see why these CI imports failed:
> 
> ```
> ModuleNotFoundError: No module named 'jax.random.PRNGKey'; 'jax.random' is not a package
> 
> ModuleNotFoundError: No module named 'pyro.infer.Predictive'
> ```

Yeah, better to stay with __import__, I think it does some extra magic.

hectormz:
> tbh, it is not a test, but a way to precompile models for CI, so at some point, we can start gathering performance data for our tests and they should be reproducible, in most cases.
> 
> I cheated with `test_` so pytest would work with it.
> 
> So users should not run it (without manually calling it)

okay got it @ahartikainen 


> edit.
> I think we probably need to change the import for the classes
> 
> ```
> class = getattr(import_module('my_module'), 'my_class')
> ```
> 
> https://stackoverflow.com/questions/41678073/import-class-from-module-dynamically

I'll give that shot!

ahartikainen:
Is this now skipping pystan tests?

hectormz:
Seems that way. I was going to investigate why. To sort this out, I setup a virtual machine to mimic the CI, but pystan tests ran locally.

ahartikainen:
Is it trying to import once and then just fails?

arviz/tests/external_tests/test_data_pystan.py
```diff
@@ -1,17 +1,31 @@
...
     pystan_version,
+    running_on_ci,
+)
+
+# Check if either pystan or pystan3 is installed
+pystan_installed = (importlib.util.find_spec("pystan") is not None) or (
+    importlib.util.find_spec("stan") is not None
+)
+pytestmark = pytest.mark.skipif(
+    not pystan_installed & (not running_on_ci()),
```
ahartikainen: not pystan_installed & (not running_on_ci())

not True & (not True) == True

(not True) & (not True) == False

ahartikainen: Or you can use

    not ( pystan_installed |  running_on_ci() )

hectormz: I think that logic is cleaner. But shouldn't the current logic work because it equates to False, so it shouldn't get skipped

ahartikainen: It is currently true (missing parenthesis)

    not True & (not True) == True

hectormz: Ah yes okay. I updated that, will see if it does the trick. And I see that there are some built-in emcee skips looking for emcee3, so that should still be okay

hectormz:
Thanks @ahartikainen for getting this through!

canyon289:
Looked it over. It looks like it works, there some sections I can't read but I don't want to hold this up :)

canyon289:
Also this is great work. Looks well implemented. thank you so much for persevering and getting it through

hectormz:
@canyon289 there are sections you physically can't access in order to inspect, or some sections that don't make sense? 

If there's something I could clarify or improve to avoid future headache, I'd be happy to update it now that it's working.

Happy to contribute 🙂

canyon289:
@hectormz It's likely less about your code and more about me not understanding the depth of pytest at this time. I didn't mean that as a criticism of your work!

hectormz:
No offense taken!

PR #1114, opened 2020-03-09
[WIP] Codecov troubleshooting

OriolAbril:
## Description
From time to time codecov does not recieve/update properly checks and coverage.

## Checklist


scripts/codecov_uploader.py
```diff
@@ -0,0 +1,12 @@
+from os import environ
+from time import sleep
+from codecov import main as codecov
+
+for _ in range(10):
+    try:
+        codecov(["--token", environ["CODECOV_TOKEN"], "--name", environ["NAME"], "--required"])
```
ahartikainen: is this the correct format? I think they should go in as arguments

scripts/codecov_uploader.py
```diff
@@ -0,0 +1,12 @@
...
+    except Exception:
+        sleep(30)
```
ahartikainen: maybe add print what exception is thrown

ahartikainen:
It should now work, right?

Do we want to raise error if codecov fails?

We would need to add this after the for loop

    if [ $? -ne 0 ]
        exit 1

or change 

    $output = python -m codecov ...
    if [ $output -eq 0 ]
        break
    ...
    if [ $output -ne 0 ]
        exit 1

OriolAbril:
It does work! Thanks a lot! 

There is one job whose [upload failed the 5 times](https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=2105&view=logs&j=e4994452-efbb-51bd-c9fc-f1d030f5bbfb&t=b874d77c-8497-5843-d5d5-e253609482a4&l=32), which is why codecov has not uploaded. I think that the timeout in Azure is kind of long, so I'll try a larger wait time.

ahartikainen:
It would be interesting to know why those fail

PR #1117, opened 2020-03-14
Revamping HPD

### LINKED ISSUE #855: Unexpected behavior when using `pm.hpd` with multidimensional (ndim>2)posterior arrays

percygautam:
## Description

fixes #855  Make hpd work with multidimensional arrays.
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,16 @@
 ## v0.x.x Unreleased
 
 ### New features
-* Integrate jointplot into pairplot, add point-estimate and overlay of plot kinds #1079
+* Integrate jointplot into pairplot, add point-estimate and overlay of plot kinds (#1079)
 ### Maintenance and fixes
-
+* Fixed behaviour of `credible_interval=None` in `plot_posterior` (#1115)
+* Fixed hist kind of `plot_dist` with multidimensional input (#1115)
+* Fixed `TypeError` in `transform` argument of `plot_density` and `plot_forest` when `InferenceData is a list or tuple (#1121)`
 ### Deprecation
 
 ### Documentation
+* Image thumbs generation updated to be Bokeh 2 compatible (#1116)
+* Add new examples for `plot_pair` (#1110)
 
 
 ## v0.7.0 (2020 Mar 2)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Integrate jointplot into pairplot, add point-estimate and overlay of plot kinds (#1079)
 * Allow xarray.Dataarray input for plots.(#1120)
 * Revamped the `hpd` function to make it work with mutidimensional arrays, InferenceData and xarray objects (#1117)
+* Skip test for optional/extra dependencies when not installed (#1113)
 ### Maintenance and fixes
 * Fixed behaviour of `credible_interval=None` in `plot_posterior` (#1115)
 * Fixed hist kind of `plot_dist` with multidimensional input (#1115)
@@ -213,4 +214,3 @@
 ## v0.3.0 (2018 Dec 14)
 
 * First Beta Release
-
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,11 @@
 * Add `num_chains` and `pred_dims` arguments to io_pyro #1090
 * Integrate jointplot into pairplot, add point-estimate and overlay of plot kinds (#1079)
 * Allow xarray.Dataarray input for plots.(#1120)
-* Revamped the `hpd` function to make it work with muti-dimensional arrays (#1117)
+* Revamped the `hpd` function to make it work with mutidimensional arrays, InferenceData and xarray objects (#1117)
 ### Maintenance and fixes
 * Fixed behaviour of `credible_interval=None` in `plot_posterior` (#1115)
 * Fixed hist kind of `plot_dist` with multidimensional input (#1115)
-* Fixed `TypeError` in `transform` argument of `plot_density` and `plot_forest` when `InferenceData is a list or tuple` (#1121)
+* Fixed `TypeError` in `transform` argument of `plot_density` and `plot_forest` when `InferenceData` is a list or tuple (#1121)
 ### Deprecation
 
 ### Documentation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,11 @@
 * Add `num_chains` and `pred_dims` arguments to io_pyro #1090
 * Integrate jointplot into pairplot, add point-estimate and overlay of plot kinds (#1079)
 * Allow xarray.Dataarray input for plots.(#1120)
+* Revamped the `hpd` function to make it work with muti-dimensional arrays (#1117)
 ### Maintenance and fixes
 * Fixed behaviour of `credible_interval=None` in `plot_posterior` (#1115)
 * Fixed hist kind of `plot_dist` with multidimensional input (#1115)
-* Fixed `TypeError` in `transform` argument of `plot_density` and `plot_forest` when `InferenceData is a list or tuple (#1121)`
+* Fixed `TypeError` in `transform` argument of `plot_density` and `plot_forest` when `InferenceData is a list or tuple` (#1121)
 ### Deprecation
 
 ### Documentation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,11 @@
 ## v0.x.x Unreleased
 
 ### New features
+* Integrate jointplot into pairplot, add point-estimate and overlay of plot kinds #1079
+* Add out-of-sample groups (`predictions` and `predictions_constant_data`) and `constant_data` group to pyro translation #1090
+* Add `num_chains` and `pred_dims` arguments to io_pyro #1090
 * Integrate jointplot into pairplot, add point-estimate and overlay of plot kinds (#1079)
+* Allow xarray.Dataarray input for plots.(#1120)
 ### Maintenance and fixes
 * Fixed behaviour of `credible_interval=None` in `plot_posterior` (#1115)
 * Fixed hist kind of `plot_dist` with multidimensional input (#1115)
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -324,11 +324,11 @@ def hpd(
     Parameters
     ----------
     ary : obj
-        onject containing posterior samples.
+        object containing posterior samples.
         Any object that can be converted to an az.InferenceData object.
         Refer to documentation of az.convert_to_dataset for details.
     group : str, optional
-         Specifies which InferenceData group should be used to calculate hpd.  Defaults to 'posterior'
+         Specifies which InferenceData group should be used to calculate hpd. Defaults to 'posterior'
     var_names : list
         Names of variables to include in the hpd report
     credible_interval : float, optional
@@ -342,10 +342,14 @@ def hpd(
         modes are well separated.
     skipna : bool
         If true ignores nan values when computing the hpd interval. Defaults to false.
+    kwargs : dict, optional
+        Additional keywords passed to ax.scatter
+        For example, to calculate hpd over "chain" dimension, Pass "input_core_dims" to "chain" in
+        the dictionary.
 
     Returns
     -------
-    np.ndarray
+    np.ndarray or xarray.Dataset, depending upon input
         lower(s) and upper(s) values of the interval(s).
 
     Examples
@@ -381,14 +385,13 @@ def hpd(
             return _hpd(ary, credible_interval, circular, skipna)
         ary = convert_to_dataset(ary)
         kwargs.setdefault("input_core_dims", [["chain"]])
-        return _wrap_xarray_ufunc(_hpd, ary, func_kwargs=func_kwargs, **kwargs).to_array().values[0]
+        return _wrap_xarray_ufunc(_hpd, ary, func_kwargs=func_kwargs, **kwargs).x.values
 
     ary = convert_to_dataset(ary, group=group)
     var_names = _var_names(var_names, ary)
 
     ary = ary if var_names is None else ary[var_names]
 
-    kwargs.setdefault("input_core_dims", [["chain", "draw"]])
     return _wrap_xarray_ufunc(_hpd, ary, func_kwargs=func_kwargs, **kwargs)
 
 
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -410,7 +410,7 @@ def hpd(
     kwargs.setdefault("output_core_dims", [["hpd", "mode"] if multimodal else ["hpd"]])
     if not multimodal:
         func_kwargs["circular"] = circular
-    if multimodal:
+    else:
         func_kwargs["max_modes"] = max_modes
 
     func = _hpd_multimodal if multimodal else _hpd
@@ -419,8 +419,8 @@ def hpd(
         if len(ary.shape) == 1:
             func_kwargs.pop("out_shape")
             hpd_data = func(ary, **func_kwargs)
-            out = hpd_data[~np.isnan(hpd_data)]
-            return out.reshape(out.shape[0] // 2, 2) if multimodal else hpd_data
+            return hpd_data[~np.isnan(hpd_data).all(axis=1), :] if multimodal else hpd_data
+
         ary = convert_to_dataset(ary)
         kwargs.setdefault("input_core_dims", [["chain"]])
         res = _wrap_xarray_ufunc(func, ary, func_kwargs=func_kwargs, **kwargs)
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -44,7 +44,7 @@ __all__ = [
 
 
 def compare(
-    dataset_dict, ic=None, method="BB-pseudo-BMA", b_samples=1000, alpha=1, seed=None, scale=None
+        dataset_dict, ic=None, method="BB-pseudo-BMA", b_samples=1000, alpha=1, seed=None, scale=None
 ):
     r"""Compare models based on PSIS-LOO `loo` or WAIC `waic` cross-validation.
 
@@ -306,15 +306,16 @@ def _ic_matrix(ics, ic_i):
 
 
 def hpd(
-    ary,
-    *,
-    group="posterior",
-    var_names=None,
-    credible_interval=None,
-    circular=False,
-    multimodal=False,
-    skipna=False,
-    **kwargs
+        ary,
+        *,
+        group="posterior",
+        var_names=None,
+        sel=None,
+        credible_interval=None,
+        circular=False,
+        multimodal=False,
+        skipna=False,
+        **kwargs
 ):
     """
     Calculate highest posterior density (HPD) of array for given credible_interval.
@@ -330,8 +331,10 @@ def hpd(
     group : str, optional
          Specifies which InferenceData group should be used to calculate hpd.
          Defaults to 'posterior'
-    var_names : list
+    var_names : list, optional
         Names of variables to include in the hpd report
+    sel: dict, optional
+        To calculate hpd over selection on all groups.
     credible_interval : float, optional
         Credible interval to compute. Defaults to 0.94.
     circular : bool, optional
@@ -384,6 +387,12 @@ def hpd(
 
         In [1]: az.hpd(data, **{"input_core_dims": [["chain"]]})
 
+    We can also calculate the hpd over a particular selection over all groups:
+
+    .. ipython::
+
+        In [1]: az.hpd(data, sel={"chain":[0, 1, 3]}, **{"input_core_dims": [["draw"]]})
+
     """
     if credible_interval is None:
         credible_interval = rcParams["stats.credible_interval"]
@@ -410,6 +419,8 @@ def hpd(
         return _wrap_xarray_ufunc(_hpd, ary, func_kwargs=func_kwargs, **kwargs).x.values
 
     ary = convert_to_dataset(ary, group=group)
+    if sel is not None:
+        ary = ary.sel(**sel)
     var_names = _var_names(var_names, ary)
 
     ary = ary[var_names] if var_names else ary
@@ -579,7 +590,7 @@ def loo(data, pointwise=False, reff=None, scale=None):
             ess_p = ess(posterior, method="mean")
             # this mean is over all data variables
             reff = (
-                np.hstack([ess_p[v].values.flatten() for v in ess_p.data_vars]).mean() / n_samples
+                    np.hstack([ess_p[v].values.flatten() for v in ess_p.data_vars]).mean() / n_samples
             )
 
     log_weights, pareto_shape = psislw(-log_likelihood, reff)
@@ -865,20 +876,20 @@ def r2_score(y_true, y_pred):
 
 
 def summary(
-    data,
-    var_names: Optional[List[str]] = None,
-    fmt: str = "wide",
-    kind: str = "all",
-    round_to=None,
-    include_circ=None,
-    stat_funcs=None,
-    extend=True,
-    credible_interval=None,
-    order="C",
-    index_origin=None,
-    skipna=False,
-    coords: Optional[CoordSpec] = None,
-    dims: Optional[DimSpec] = None,
+        data,
+        var_names: Optional[List[str]] = None,
+        fmt: str = "wide",
+        kind: str = "all",
+        round_to=None,
+        include_circ=None,
+        stat_funcs=None,
+        extend=True,
+        credible_interval=None,
+        order="C",
+        index_origin=None,
+        skipna=False,
+        coords: Optional[CoordSpec] = None,
+        dims: Optional[DimSpec] = None,
 ) -> Union[pd.DataFrame, xr.Dataset]:
     """Create a data frame with summary statistics.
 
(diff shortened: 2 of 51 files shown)


percygautam:
I have considered the multimodal case and modified hpd tests. I can't seem to find how to use this modified hpd in `summary` function.

arviz/stats/stats.py
```diff
@@ -305,7 +305,7 @@ def _ic_matrix(ics, ic_i):
     return rows, cols, ic_i_val
 
 
-def hpd(ary, credible_interval=None, circular=False, multimodal=False, skipna=False):
+def hpd(ary, *, credible_interval=None, circular=False, multimodal=False, skipna=False, **kwargs):
```
OriolAbril: we can also add a group argument (like the one in plotting functions) in case someone wanted to calculate hpd on posterior predictive, predictions...

also a `var_names` could be added.

arviz/stats/stats.py
```diff
@@ -349,84 +351,92 @@ def hpd(ary, credible_interval=None, circular=False, multimodal=False, skipna=Fa
...
-            dx = range_x / len(density)
-            bins = np.linspace(lower, upper, len(density))
-        else:
-            bins = get_bins(ary)
-            _, density, _ = histogram(ary, bins=bins)
-            dx = np.diff(bins)[0]
+        return _hpd_multimodal(ary, credible_interval, skipna)
+
+    if isinstance(ary, np.ndarray):
+        ary = convert_to_dataset(ary)
```
OriolAbril: If input is array, output should be array too (like wrap_xarray_ufunc does).

We have to decide what should happen though, do we want `hpd` to be backwards compatible when input is ndarray? Or should we assume ArviZ chain, draw, *shape convention? @aloctavodia

arviz/stats/stats.py
```diff
@@ -349,84 +351,92 @@ def hpd(ary, credible_interval=None, circular=False, multimodal=False, skipna=Fa
...
+
+    kwargs.setdefault("input_core_dims", [["chain", "draw"]])
+    kwargs.setdefault("output_core_dims", [["hpd"]])
+    func_kwargs = {
+        "credible_interval": credible_interval,
+        "circular": circular,
+        "skipna": skipna,
+        "out_shape": (2,),
+    }
+    return _wrap_xarray_ufunc(_hpd, ary, func_kwargs=func_kwargs, **kwargs)
```
OriolAbril: does wrap xarray ufunc work on inference data objects? Is the input supposed to be a dataset? I think now that we are improving and rewriting, it should behave similarly to `ess`/`rhat`, accept both inference data/dataset or ndarray.

arviz/stats/stats.py
```diff
@@ -349,84 +351,92 @@ def hpd(ary, credible_interval=None, circular=False, multimodal=False, skipna=Fa
...
-        hdi_min = ary[min_idx]
-        hdi_max = ary[min_idx + interval_idx_inc]
 
-        if circular:
-            hdi_min = hdi_min + mean
-            hdi_max = hdi_max + mean
-            hdi_min = np.arctan2(np.sin(hdi_min), np.cos(hdi_min))
-            hdi_max = np.arctan2(np.sin(hdi_max), np.cos(hdi_max))
+def _hpd_multimodal(ary, credible_interval, skipna):
+    """Compute hpd if the distribution is multimodal"""
```
OriolAbril: Add the flatten here too.

arviz/stats/stats.py
aloctavodia: backwards compatibility when input is ndarray sounds reasonable.

percygautam: Okay, I'll do it.

arviz/stats/stats.py
percygautam: > does wrap xarray ufunc work on inference data objects? 

No, currently I don't think it works. We need to modify the `wrap_xarray_ufunc`.

arviz/stats/stats.py
percygautam: group argument is valid for InferenceData object input, right? Or is there some other case.

percygautam:
I have made the changes. It is working fine for ndarray and datasets. Currently, I am first converting ndarray to dataset then using `wrap_xarray_ufunc`. 
I tried using directly on ndarray but the hpd is calculated is always calculated on the last dimension. Converting it to ndarray gives me control over which dimension to calculate hpd for.
Any workaround for this?

arviz/stats/stats.py
```diff
@@ -349,86 +365,104 @@ def hpd(ary, credible_interval=None, circular=False, multimodal=False, skipna=Fa
...
 
-        idx = np.argsort(-density)
-        intervals = bins[idx][density[idx].cumsum() <= credible_interval]
-        intervals.sort()
+    if isinstance(ary, np.ndarray):
+        if len(ary.shape) == 1:
+            return _hpd(ary, credible_interval, circular, skipna)
+        ary = convert_to_dataset(ary)
+        kwargs.setdefault("input_core_dims", [["chain"]])
+        return _wrap_xarray_ufunc(_hpd, ary, func_kwargs=func_kwargs, **kwargs).to_array().values[0]
```
OriolAbril: I think this should be `...**kwargs).x.values` (as it comes from an array, the name will be `x`, this should avoid the extra `variable` dimension added by `to_array`)

arviz/stats/stats.py
```diff
@@ -349,86 +365,104 @@ def hpd(ary, credible_interval=None, circular=False, multimodal=False, skipna=Fa
...
+    ary = convert_to_dataset(ary, group=group)
+    var_names = _var_names(var_names, ary)
 
-        hpd_intervals = []
-        for interval in intervals_splitted:
-            if interval.size == 0:
-                hpd_intervals.append((bins[0], bins[0]))
-            else:
-                hpd_intervals.append((interval[0], interval[-1]))
+    ary = ary if var_names is None else ary[var_names]
```
OriolAbril: This should not be necessary after having called `_var_names`

arviz/stats/stats.py
```diff
@@ -349,86 +365,104 @@ def hpd(ary, credible_interval=None, circular=False, multimodal=False, skipna=Fa
...
 
-        hpd_intervals = np.array(hpd_intervals)
+    kwargs.setdefault("input_core_dims", [["chain", "draw"]])
```
OriolAbril: This is not needed either, these are the default in `wrap_xarray_ufunc`

arviz/stats/stats.py
OriolAbril: We may get to apply ufunc on idata objects, but clearly not right now, and probably not directly in hpd/ess functions.

arviz/tests/base_tests/test_stats.py
```diff
@@ -47,6 +47,12 @@ def test_hpd():
     assert_array_almost_equal(interval, [-1.88, 1.88], 2)
 
 
+def test_hpd_multidimension():
```
OriolAbril: There are some more cases to test:

* test ndarray input:
   * 1d array -> done in `test_hpd`
   * 2d array -> missing, this is important to make sure that hpd with `wrap_apply_ufunc` is backwards compatible
   * 3d+ array -> done in this test
* test idata input
   * missing, the `centered_eight` fixture could be used for this. This should test returned shapes, returned types (xarray dataset) and that the `group` and `var_names` arguments work

arviz/stats/stats.py
```diff
@@ -305,16 +305,32 @@ def _ic_matrix(ics, ic_i):
...
+def hpd(
+    ary,
+    *,
+    group="posterior",
+    var_names=None,
+    credible_interval=None,
+    circular=False,
+    multimodal=False,
+    skipna=False,
+    **kwargs
```
OriolAbril: should be documented too

arviz/tests/base_tests/test_stats.py
OriolAbril: It would also be great to add a test with idata case using `kwargs` to get the hpd of every chain

arviz/stats/stats.py
percygautam: I don't understand. We have to get data_vars of ary from var_names, right? `_var_names` will just check for the input var_names in data_vars and return list of var_names. Am I correct?

arviz/tests/base_tests/test_stats.py
percygautam: I have added all the specified tests, except for `group` one. Can you help me with this one?

arviz/stats/stats.py
percygautam: I have documented kwargs. I have specified the return type to be depending on the input.

arviz/stats/stats.py
OriolAbril: exactly, var_names cannot be none after calling `_var_names`: https://github.com/arviz-devs/arviz/blob/master/arviz/utils.py#L13

OriolAbril: Oh, okey, the selection is necessary, the if is None is not, I could have expressed myself better

arviz/stats/stats.py
```diff
@@ -326,10 +342,14 @@ def hpd(ary, credible_interval=None, circular=False, multimodal=False, skipna=Fa
         modes are well separated.
     skipna : bool
         If true ignores nan values when computing the hpd interval. Defaults to false.
+    kwargs : dict, optional
+        Additional keywords passed to ax.scatter
```
OriolAbril: passed to `wrap_xarray_ufunc`, maybe add a link to wrap_ufunc docs too?

 The explanation below would be better as an example instead of here in argument description I think.

arviz/tests/base_tests/test_stats.py
```diff
@@ -47,6 +47,35 @@ def test_hpd():
...
+    assert result.shape == (5, 2,)
+
+
+def test_hpd_multidimension():
+    normal_sample = np.random.randn(12000, 10, 3)
+    result = hpd(normal_sample)
+    assert result.shape == (10, 3, 2,)
+
+
+def test_hpd_idata():
```
OriolAbril: These tests should use `data` instead of `normal_sample`

OriolAbril: you can also use the fixture instead of calling load_arviz_data: https://github.com/arviz-devs/arviz/blob/master/arviz/tests/base_tests/test_stats.py#L107

arviz/tests/base_tests/test_stats.py
```diff
@@ -47,6 +47,35 @@ def test_hpd():
...
+
+    result = hpd(normal_sample, **{"input_core_dims": [["chain"]]})
+    assert isinstance(result, Dataset)
+    assert result.dims == {"draw": 500, "hpd": 2, "school": 8}
+
+    result = hpd(normal_sample, var_names=["mu", "theta"])
+    assert isinstance(result, Dataset)
+    assert result.dims == {"hpd": 2, "school": 8}
+    assert list(result.data_vars.keys()) == ["mu", "theta"]
+
```
OriolAbril: to test `group` argument you could use something like:

```
def test_hpd_idata_group(centered_eight):
    result_posterior = hpd(centered_eight, group="posterior", var_names="mu")
    result_prior = hpd(centered_eight, group="prior", var_names="mu")
    assert result_prior.dims == {"hpd": 2}
    assert result_posterior.mu.values[0] > result_prior.mu.values[0]
    assert result_posterior.mu.values[1] < result_prior.mu.values[1]
```

Not really sure about the values and indexes, please check. The posterior distribution should be more constrained than the prior, therefore the posterior hpd should lie inside the prior hpd (which is what this tries to check)

arviz/stats/stats.py
percygautam: Ah, I copied it from somewhere. Sorry. I will add a few examples.

arviz/tests/base_tests/test_stats.py
percygautam: Okay, I understand the logic. 
But, I am getting `mu.values` for posterior to be [-2.11840595 10.4032071 ]  and for prior to be [-9.93081274  8.76928874]. (not completely inside prior)
> assert result_posterior.mu.values[1] < result_prior.mu.values[1]

So, this is failing.?

OriolAbril: The region in the prior is much larger, but centered around 0 whereas the region in the posterior is smaller but not centered around zero, therefore, even though it is smaller than the prior hpd it does not lie inside prior hpd region, you can change it to some other check that actually works. I was also thinking that `sel` should also be added, otherwise users can select theta variable but nothing else.

percygautam: Okay I understand. Can I check the range of prior and posterior? (By doing `result_posterior.mu.values[1] - result_posterior.mu.values[0]`)

Are you suggesting to add new `sel` param?

OriolAbril: You can check the range yes. And yes, I am suggesting to add a sel argument

arviz/stats/stats.py
```diff
@@ -305,16 +305,36 @@ def _ic_matrix(ics, ic_i):
...
+        *,
+        group="posterior",
+        var_names=None,
+        sel=None,
+        credible_interval=None,
+        circular=False,
+        multimodal=False,
+        skipna=False,
+        **kwargs
+):
```
OriolAbril: I would try to minimize the changes to the call signature, these changes are not backwards compatible. I think they should be, otherwise trying to make the behaviour on arrays backwards compatible does not make much sense

percygautam: Yes, it makes sense. Should I pass them through `kwargs`?

OriolAbril: no need to, just make the new arguments be optional positional arguments after the current ones

percygautam:
Do we start the work of multimodal here or wait for everyone's opinion? Maybe create a new issue for multimodal discussion.

arviz/tests/base_tests/test_stats.py
```diff
@@ -47,6 +47,52 @@ def test_hpd():
...
+    assert result.shape == (10, 3, 2,)
+
+
+def test_hpd_idata(centered_eight):
+    data = centered_eight.posterior
+    result = hpd(data)
+    assert isinstance(result, Dataset)
+    assert result.dims == {"school": 8, "hpd": 2}
+
+    result = hpd(data, **{"input_core_dims": [["chain"]]})
```
OriolAbril: You can use `input_core_dims=[["chain"]]`, no need to create a dict and then `**`

arviz/tests/base_tests/test_stats.py
```diff
@@ -47,6 +47,52 @@ def test_hpd():
...
+    result_prior = hpd(centered_eight, group="prior", var_names="mu")
+    assert result_prior.dims == {"hpd": 2}
+    range_posterior = result_posterior.mu.values[1] - result_posterior.mu.values[0]
+    range_prior = result_prior.mu.values[1] - result_prior.mu.values[0]
+    assert range_posterior < range_prior
+
+
+def test_hpd_sel(centered_eight):
+    data = centered_eight.posterior
+    result = hpd(data, sel={"chain":[0, 1, 3]}, **{"input_core_dims": [["draw"]]})
```
OriolAbril: idem

OriolAbril:
I would work on multimodal in this same PR, otherwise, behaviour of ArviZ development version between merging this and the other will be quite confusing.

percygautam:
> Regarding multimodal case, I was thinking of a possible workaround, let's see what everyone thinks about this. The idea would be to have _hpd_multimodal return a result with shape (2, 10) (the 10 is completely arbitrary and could be modified, even be an argument). Of these 10 pairs of values, the first would be hpd intervals and the last ones will probably not be needed, and would be then set to nan (in _hpd_multimodal). Eventually hpd in the case of multimodal would drop nan values and return the hpd dataset with the proper shape (also, different variables may have different number of modes and this approach should solve this too).

Should I start implementing this or wait?

OriolAbril:
Let's go ahead with multimodal :muscle:

percygautam:
I have tried to implement the multimodal case for a single input. Currently, I am filling only the first dimension with NaNs. 
How do we decide the size (10) of the first dimension in this case?

arviz/stats/stats.py
```diff
@@ -342,91 +364,139 @@ def hpd(ary, credible_interval=None, circular=False, multimodal=False, skipna=Fa
...
-            hdi_max = np.arctan2(np.sin(hdi_max), np.cos(hdi_max))
+    intervals_splitted = np.split(intervals, np.where(np.diff(intervals) >= dx * 1.1)[0] + 1)
 
-        hpd_intervals = np.array([hdi_min, hdi_max])
+    hpd_intervals = []
+    for interval in intervals_splitted:
+        if interval.size == 0:
+            hpd_intervals.append((bins[0], bins[0]))
+        else:
+            hpd_intervals.append((interval[0], interval[-1]))
```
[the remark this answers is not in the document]
OriolAbril: On second thought, this could be checked before the loop checking the length of `intervals_splitted`, and if necessary, truncating `intervals_splitted` and issuing the warning

arviz/stats/stats.py
```diff
@@ -342,91 +364,139 @@ def hpd(ary, credible_interval=None, circular=False, multimodal=False, skipna=Fa
...
+        kwargs.setdefault("input_core_dims", [["chain"]])
+        return _wrap_xarray_ufunc(func, ary, func_kwargs=func_kwargs, **kwargs).x.values
+
+    func_kwargs["out_shape"] = (10, 2,) if multimodal else (2,)
+    ary = convert_to_dataset(ary, group=group)
+    if sel is not None:
+        ary = ary.sel(**sel)
+    var_names = _var_names(var_names, ary)
+
+    ary = ary[var_names] if var_names else ary
```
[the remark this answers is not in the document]
percygautam: Don't the `_var_names` returns `None` if user doesn't pass `var_names` argument (i.e. if `var_names` is  None)?
https://github.com/arviz-devs/arviz/blob/33d9d9fb55e2ae2b7a938f8e1673015a01ae507f/arviz/utils.py#L21-L25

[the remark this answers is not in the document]
OriolAbril: ooh, true, it's the converters to array the ones that handle the None automatically, good catch

arviz/stats/stats.py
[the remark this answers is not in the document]
percygautam: What could be the default value for `max_modes`? I am currently using 10.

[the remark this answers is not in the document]
aloctavodia: I guess if we think about this in detail and/or run a few tests we should be able to figure out a reasonable default number for this, based on the accuracy/reliability of the method as the number of modes increase. In the meantime 10 sounds good and reminds me of this story 

In 1975, at the age of 99, Leonor Acevedo Borges, the writer’s mother, died. At the wake, a woman expresses her condolences to Borges and comments:
    Poor little Leonetta! To die so close to his 100th birthday! If she had waited a little longer…To what Borges answers:- I see madam that you are devoted to the metric system.

arviz/stats/stats.py
```diff
@@ -342,91 +367,149 @@ def hpd(ary, credible_interval=None, circular=False, multimodal=False, skipna=Fa
...
-        if ary.dtype.kind == "f":
-            density, lower, upper = _fast_kde(ary)
-            range_x = upper - lower
-            dx = range_x / len(density)
-            bins = np.linspace(lower, upper, len(density))
-        else:
-            bins = get_bins(ary)
-            _, density, _ = histogram(ary, bins=bins)
-            dx = np.diff(bins)[0]
+        func_kwargs["max_modes"] = max_modes
```
OriolAbril: could be an if/else

arviz/stats/stats.py
```diff
@@ -342,91 +367,149 @@ def hpd(ary, credible_interval=None, circular=False, multimodal=False, skipna=Fa
...
+
+    func = _hpd_multimodal if multimodal else _hpd
+
+    if isinstance(ary, np.ndarray):
+        if len(ary.shape) == 1:
+            func_kwargs.pop("out_shape")
+            hpd_data = func(ary, **func_kwargs)
+            out = hpd_data[~np.isnan(hpd_data)]
+            return out.reshape(out.shape[0] // 2, 2) if multimodal else hpd_data
```
OriolAbril: can be `...reshape(-1,2)...`, however, I think that something like 

    return hpd_data[~np.isnan(hpd_data).all(axis=0)] if multimodal else hpd_data  # maybe ...axis=0), :]...

would work to remove the extra rows.

arviz/stats/stats.py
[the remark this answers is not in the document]
percygautam: Okay. I think I am also one of the devoted ones. 😄

arviz/stats/stats.py
percygautam: Thanks, it works fine. (axis need to be 1)

percygautam:
I think ndarray input with ndims > 2 is not supported.

OriolAbril:
You can add a comment to ignore the pylint warning, it probably misses the pop

CHANGELOG.md
```diff
@@ -8,10 +8,11 @@
 * Add `num_chains` and `pred_dims` arguments to io_pyro #1090
 * Integrate jointplot into pairplot, add point-estimate and overlay of plot kinds (#1079)
 * Allow xarray.Dataarray input for plots.(#1120)
+* Revamped the `hpd` function to make it work with muti-dimensional arrays (#1117)
 ### Maintenance and fixes
 * Fixed behaviour of `credible_interval=None` in `plot_posterior` (#1115)
 * Fixed hist kind of `plot_dist` with multidimensional input (#1115)
-* Fixed `TypeError` in `transform` argument of `plot_density` and `plot_forest` when `InferenceData is a list or tuple (#1121)`
+* Fixed `TypeError` in `transform` argument of `plot_density` and `plot_forest` when `InferenceData is a list or tuple` (#1121)
```
OriolAbril: ` should close after InferenceData

CHANGELOG.md
```diff
@@ -8,10 +8,11 @@
 * Add `num_chains` and `pred_dims` arguments to io_pyro #1090
 * Integrate jointplot into pairplot, add point-estimate and overlay of plot kinds (#1079)
 * Allow xarray.Dataarray input for plots.(#1120)
+* Revamped the `hpd` function to make it work with muti-dimensional arrays (#1117)
```
OriolAbril: not only multidimensional arrays, also inferencedata and xarray objects

arviz/stats/stats.py
```diff
@@ -342,93 +367,151 @@ def hpd(ary, credible_interval=None, circular=False, multimodal=False, skipna=Fa
...
-            dx = range_x / len(density)
-            bins = np.linspace(lower, upper, len(density))
-        else:
-            bins = get_bins(ary)
-            _, density, _ = histogram(ary, bins=bins)
-            dx = np.diff(bins)[0]
+    func = _hpd_multimodal if multimodal else _hpd
 
-        density *= dx
+    if isinstance(ary, np.ndarray):
```
OriolAbril: This should be only if the array is 1d or 2d:

```
isarray = isinstance(ary, np.ndarray)
if isarray and ary.ndim <= 2:
```

If the array has 3 or more dimensions, it should assume ArviZ dim order: `(chain, draw, *shape)`. hpd should still return a numpy array though:

```
...
hpd_data = _wrap_xarray_ufunc(func, ary, func_kwargs=func_kwargs, **kwargs)
hpd_data = hpd_data.dropna("mode", how="all") if multimodal else hpd_data
return hpd_data.x.values if isarray else hpd_data
```

arviz/tests/base_tests/test_stats.py
```diff
@@ -47,6 +47,52 @@ def test_hpd():
...
+def test_hpd_2darray():
+    normal_sample = np.random.randn(12000, 5)
+    result = hpd(normal_sample)
+    assert result.shape == (5, 2,)
+
+
+def test_hpd_multidimension():
+    normal_sample = np.random.randn(12000, 10, 3)
+    result = hpd(normal_sample)
+    assert result.shape == (10, 3, 2,)
```
OriolAbril: This line will have to be updated to check that the result shape is the desired `(3, 2)`

percygautam: Earlier, we were calculating `hpd` over one dimension only, for ndarrays. So, for backward compatibility I have set default to be calculated only over 'chain' for ndarrays. So, the result still would be (10, 3, 2,).

OriolAbril: The issue is that calculating hpd only over chain is a very bad default, we'll keep the behaviour (for now) in 2d array case to keep backwards compatibility, but 3d arrays are not supported, so we do not have the backwards compatibility constraint.

percygautam: Okay, I have done the changes.

almostmeenal:
i ran some cell with pm.stats.hpd in an example notebook and ig its removed or something cause i get the error - no attribute 'hpd'. I tried az.hpd too, same thing. im probably missing something, is the function renamed or stuff.

ahartikainen:
Did you try az.hdi? Please open a new issue if that doesn't work.

almostmeenal:
@ahartikainen thankyou that worked!

PR #1119, opened 2020-03-14
missed some export_png instances

OriolAbril:
## Description

One day I'll try to see if Azure can build docs but not upload in PRs.


ahartikainen:
Maybe we can build docs and upload to artifacts?

PR #1121, opened 2020-03-15
Fix bug in `transform` arg when InferenceData is List or Tuple

AlexAndorra:
## Description
`plot_density` and `plot_forest` both accept a list or tuple of InferenceData, but the `transform` argument was not iterating through the list or tuple when it should. This was causing something like `az.plot_forest(idata var_names=["a"], transform=scipy.special.expit)` to raise a `TypeError: ufunc 'expit' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''`.

I think the modifications I made will fix this issue -- please tell me if you have any comment.
Thanks and cheers 🖖

### DIFF
--- a/arviz/plots/densityplot.py
+++ b/arviz/plots/densityplot.py
@@ -151,7 +151,6 @@ def plot_density(
 
         >>> az.plot_density([centered, non_centered], var_names=["mu"], bw=.9)
     """
-
     if not isinstance(data, (list, tuple)):
         datasets = [convert_to_dataset(data, group=group)]
     else:
--- a/arviz/plots/densityplot.py
+++ b/arviz/plots/densityplot.py
@@ -151,11 +151,14 @@ def plot_density(
 
         >>> az.plot_density([centered, non_centered], var_names=["mu"], bw=.9)
     """
-    if transform is not None:
-        data = transform(data)
+
     if not isinstance(data, (list, tuple)):
+        if transform is not None:
+            data = transform(data)
         datasets = [convert_to_dataset(data, group=group)]
     else:
+        if transform is not None:
+            data = [transform(datum) for datum in data]
         datasets = [convert_to_dataset(datum, group=group) for datum in data]
 
     var_names = _var_names(var_names, datasets)
--- a/arviz/plots/densityplot.py
+++ b/arviz/plots/densityplot.py
@@ -153,13 +153,12 @@ def plot_density(
     """
 
     if not isinstance(data, (list, tuple)):
-        if transform is not None:
-            data = transform(data)
         datasets = [convert_to_dataset(data, group=group)]
     else:
-        if transform is not None:
-            data = [transform(datum) for datum in data]
         datasets = [convert_to_dataset(datum, group=group) for datum in data]
+
+    if transform is not None:
+        datasets = [transform(dataset) for dataset in datasets]
 
     var_names = _var_names(var_names, datasets)
     n_data = len(datasets)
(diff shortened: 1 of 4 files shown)


arviz/plots/densityplot.py
```diff
@@ -151,11 +151,14 @@ def plot_density(
 
         >>> az.plot_density([centered, non_centered], var_names=["mu"], bw=.9)
     """
-    if transform is not None:
-        data = transform(data)
+
     if not isinstance(data, (list, tuple)):
+        if transform is not None:
```
canyon289: Im being so pick here but could you move transform up to line 154 and delete one of the duplicate lines?

arviz/plots/densityplot.py
```diff
@@ -151,11 +151,14 @@ def plot_density(
...
+            data = transform(data)
         datasets = [convert_to_dataset(data, group=group)]
     else:
+        if transform is not None:
+            data = [transform(datum) for datum in data]
```
OriolAbril: I think this won't work because here data is still inference data object, it should go somewhere below, once datasets has already been defined. This will also simplify the code because datasets is always a list, even if it has only one element.

arviz/plots/densityplot.py
AlexAndorra: But I don't think there is a duplicate, is there?
The data transformation is different depending on whether or not `data` is a list, so I'm not sure I can do the check and transform the data *before* checking `data`'s type 🤔

arviz/plots/densityplot.py
AlexAndorra: Ooh yeah, good catch! Plus, this will look more like the implementation in `plot_forest`

AlexAndorra:
Updated the fix with your comments guys, but I don't know why the checks failed 🤔

OriolAbril:
What has failed is the lint of documentation style: https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=2134&view=logs&j=e6a7683b-6131-58a8-ef68-5f3a9120796c&t=a013c8b0-ff63-556c-b2e6-c616f9dda162&l=14

You have to check inside the job that failed and once there check which one of the steps is failing, in this case it is pydocstyle, could be pytest or pylint.

AlexAndorra:
Ok thanks, I'll do that when I have some time 👌

OriolAbril:
@AlexAndorra could you also add a test to prevent this from happening again?

For plot forest, changing this [line](https://github.com/arviz-devs/arviz/blob/master/arviz/tests/base_tests/test_plots_matplotlib.py#L208) to:

    ({"var_names": "mu", "transform": lambda x: x+1}, 1),

should do the trick (`+1` is allowed for arrays but not for lists), and something similar should be possible for plot_density.

arviz/plots/forestplot.py
```diff
@@ -145,8 +145,9 @@ def plot_forest(
     """
     if not isinstance(data, (list, tuple)):
         data = [data]
+
     if transform is not None:
-        data = transform(data)
+        data = [transform(datum) for datum in data]
```
OriolAbril: This has to be moved below to 

```
datasets = [convert_to_dataset(datum) for datum in reversed(data)]
if transform is not None:
    datasets = [transform(dataset) for datum in datasets]
datasets = get_coords(datasets, list(reversed(coords)) if isinstance(coords, (list, tuple)) else coords)

```

data is still of type inference data, tests are catching that the issue was not yet solved, so tests are working perfectly!

AlexAndorra: Ow, ok! Very clear, thanks Oriol!

AlexAndorra:
Indeed -- I'll do it this afternoon

PR #1136, opened 2020-04-03
Black lint in CI

OriolAbril:
## Description

I think black is not working properly yet, only bad formatting in examples seems to fail builds.

## Checklist


ahartikainen:
I think we need to test if black returns 0 or not and then call exit manually

OriolAbril:
I think checking both folders at the same time should do the trick, I was going to test that now

PR #1142, opened 2020-04-07
[WIP] Improve benchmarks

### LINKED ISSUE #782: A better way to benchmark the speed tests

nitishp25:
## Description
Fixes #782 and builds on top of the work done in #1088.

I've moved the following functions to resolve cyclic imports that occured when `histogram` was imported for `_fast_kde`:
 - `_fast_kde`,  `_fast_kde_2d`, `get_bins` and `_sturges_formula` --> **numeric_utils**
 - `get_coords` --> **utils**

I've also modified the benchmarks and will cover more functions soon.

cc @OriolAbril, @ahartikainen

## Checklist


nitishp25:
Should I continue with adding more benchmarks?

aloctavodia:
`get_bins` and `_sturges_formula` are mainly helper functions to compute plots, not to compute statistics (inside `stats`). That's why they are inside `plot_utils`. Any reason to move them to `stats_utils`?

OriolAbril:
It has been moved to solve a cyclic import due to the following dependencies:

* `_fast_kde` should import `_histogram` from `stats_utils`, it currently does not because of this cyclic import which means that `_fast_kde` does not use numba anymore ->**` imports stats_utils`**
* `calculate_point_estimate` needs `_fast_kde` -> **` imports fast_kde`** (assuming it is moved to its own file) -> **` imports stats_utils`**
* `hpd` (multimodal only) needs `get_bins` -> **` imports plot_utils`**

I proposed to move `_sturges_formula` too (even though it is not necessary) to have it in the same file as `get_bins`

nitishp25:
It looks complicated but it's done to make the whole `stats` module independent of `plots` so that `plots` can import functions from `stats` module easily

asv_benchmarks/benchmarks/benchmarks.py
```diff
@@ -0,0 +1,107 @@
...
+import scipy.signal as ss
+import warnings
+import arviz as az
+from arviz.stats.stats_utils import _circular_standard_deviation, histogram, stats_variance_2d
+from arviz.kde_utils import _fast_kde, _fast_kde_2d
+
+
+class Hist:
+    params = (True, False)
+    param_names = "Numba"
```
OriolAbril, line 17: This have to be a single element tuple, the string is interpreted as an iterable and only the first letter is used: https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=2274&view=logs&j=273a7ac9-cc40-5d2f-3776-d61d22a0ef9c&t=8fe29edb-6726-5e00-142e-7c337980bc77&l=60

See `Atleast_Nd` for an example

asv_benchmarks/benchmarks/benchmarks.py
```diff
@@ -0,0 +1,107 @@
...
+        if numba_flag:
+            az.Numba.enable_numba()
+        else:
+            az.Numba.disable_numba()
+
+    def time_histogram(self, numba_flag):
+        histogram(self.data, bins=100)
+
+
+class Variance:
```
OriolAbril, line 30: variance benchmark is failing, could you look into it? https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=2274&view=logs&j=273a7ac9-cc40-5d2f-3776-d61d22a0ef9c&t=8fe29edb-6726-5e00-142e-7c337980bc77&l=99

nitishp25: I was trying to debug this so I ran the benchmarks with `--show-stderr` flag but it only showed
```
asv: benchmark timed out (timeout 60.0s)
```
And when I ran the benchmarks with both `--quick --show-stderr` flags, it worked but the function took 31s to run with `Numba=False` and when I used the function in a notebook with the same data and numba disabled it ran in milliseconds. I think it's some issue with the benchmarks configuration, I'll continue looking.

aloctavodia:
What about having a `numerics.py` (or some other name), and put there `histogram`, `_fast_kde`, `_fast_kde_2d`, `get_bins` and `_sturges_formula`?

asv_benchmarks/benchmarks/benchmarks.py
nitishp25: My bad, the data was too large so the function was actually really slow without Numba speed-up  (taking ~30s and not milliseconds after I rechecked) and it caused the benchmark to fail :(

I'll reduce the data.

asv_benchmarks/benchmarks/benchmarks.py
```diff
@@ -0,0 +1,103 @@
...
+    param_names = ("Numba",)
+
+    def setup(self, numba_flag):
+        self.data = np.random.rand(10000, 1000)
+        if numba_flag:
+            az.Numba.enable_numba()
+        else:
+            az.Numba.disable_numba()
+
+    def time_variance(self, numba_flag):
```
OriolAbril, line 37: we should probably name this one time_variance_2d

asv_benchmarks/benchmarks/benchmarks.py
```diff
@@ -0,0 +1,103 @@
...
+        if numba_flag:
+            az.Numba.enable_numba()
+        else:
+            az.Numba.disable_numba()
+
+    def time_fast_kde_normal(self, numba_flag, n):
+        _fast_kde(self.x)
+
+class Fast_KDE_2d:
+    params = [(True, False), (10**5, 10**6)]
```
OriolAbril, line 71: I have realized that with these shapes, there is no speedup when using numba, I tried to look if the same issue as with fast_kde_1d happened where the speedup was lost in a future PR, and it looks like the speed-up is still there, but is only seen for larger sizes. 

Could we try something like this and see how do dimensions affect performance?

```
class Fast_KDE_2d:
    params = [
        (True, False), 
        ((100, 10**4), (10**4, 100), (1000, 1000))
    ]
    param_names = ("Numba", "shape")

    def setup(self, numba_flag, shape):
        self.x = np.random.randn(*shape)
        self.y = np.random.randn(n//10, 10)
        if numba_flag:
        ...
```

nitishp25: Ooh, I'll implement this.

Also, in #1088 you mentioned that more functions need to be covered in benchmarks and later on memory benchmarks and line profiling also to be added, so should I start adding all this now in this PR or sometime later?

OriolAbril: Maybe it's better to wait for another PR

nitishp25: > I have realized that with these shapes, there is no speedup when using numba, I tried to look if the same issue as with fast_kde_1d happened where the speedup was lost in a future PR, and it looks like the speed-up is still there, but is only seen for larger sizes.
> 
> Could we try something like this and see how do dimensions affect performance?
> 
> ```
> class Fast_KDE_2d:
>     params = [
>         (True, False), 
>         ((100, 10**4), (10**4, 100), (1000, 1000))
>     ]
>     param_names = ("Numba", "shape")
> 
>     def setup(self, numba_flag, shape):
>         self.x = np.random.randn(*shape)
>         self.y = np.random.randn(n//10, 10)
>         if numba_flag:
>         ...
> ```

It gives the following error:
```
ValueError: all the input array dimensions for the concatenation axis must match exactly, but along dimension 1, the array at index 0 has size 1000000 and the array at index 1 has size 100
```
Do `x` and `y` have to be of the same size always?

Also, I did `self.y = np.random.randn(shape[0]//10, 10)` instead of `self.y = np.random.randn(n//10, 10)`. This is what you meant?

OriolAbril: I forgot to update y, it should be created like x, they must be the same shape. Also, there shoud be no n instances anymore

OriolAbril:
I was thinking that benchmarks should also be formatted with black. I think 3 lines have to be changed

* `black arviz/ examples/ asv_benchmarks/` in contributing guide (in PR checklist)
* [`python -m black --check arviz examples asv_benchmarks`](https://github.com/arviz-devs/arviz/blob/master/.azure-pipelines/azure-pipelines-external.yml#L122)
* similarly in https://github.com/arviz-devs/arviz/blob/3352a8adfc42bd5ba2aa90efbde28df18f1eb74d/scripts/lint.sh#L12

asv_benchmarks/benchmarks/benchmarks.py
nitishp25: Yup, I had removed `n`. The only problem was the different sizes of `x` and `y`. You'll see after the tests finish that there's still no speed-up :(

scripts/lint.sh
```diff
@@ -9,7 +9,11 @@ python -m pydocstyle --convention=numpy ${SRC_DIR}/arviz/
 echo "Success!"
 
 echo "Checking code style with black..."
+<<<<<<< HEAD
 python -m black -l 100 --check ${SRC_DIR}/arviz/ ${SRC_DIR}/examples/
+=======
+python -m black -l 100 --check ${SRC_DIR}/arviz/ ${SRC_DIR}/examples/ ${SRC_DIR}/asv_benchmarks/
+>>>>>>> update fast_kde_2d benchmark and changelog
```
OriolAbril, line 16: there has been some merge issue

PR #1148, opened 2020-04-11
Prior-Posterior plot

### LINKED ISSUE #749: Prior-Posterior Plot
Worked on an example with @raulpl to display how the prior changes after running inference. Here is an example with code!

```python
rugby = az.load_arviz_data('rugby')

prior_def = rugby.prior.defs.sel(team='Italy')
post_def = rugby.posterior.defs.sel(team='Italy')


fig = plt.figure(figsize=(10, 7))
ax1 = plt.subplot2grid((2, 2), (0, 0), fig=fig)
ax2 = plt.subplot2grid((2, 2), (0, 1), fig=fig)
ax3 = plt.subplot2grid((2, 2), (1, 0), colspan=2, fig=fig)


az.plot_kde(prior_def, ax=ax1)
ax1.set_title("Prior")

az.plot_kde(post_def, ax=ax2)
ax2.set_title("Posterior")

az.plot_kde(prior_def, ax=ax3)
az.plot_kde(post_def.values, ax=ax3)
ax3.set_title("Both");
```
image

percygautam:
## Description

This PR implements a new prior/posterior plot (fixes #749 ). 
## Checklist


percygautam:
Currently, matplotlib backend for prior/posterior is implemented. The plot from following code looks like this: 

```
import arvizz as az
data = az.load_arviz_data("rugby")
az.plot_pp(data, vars=["defs"], coords={"team" : ["Italy", "France"]})
plt.show()
```
pp_plot

This is the initial implementation of the issue. Further work includes styling changes, adding observed data support, making bokeh backend (maybe more). All comments are welcome about the new features/modifications in the plot.

arviz/plots/backends/matplotlib/ppplot.py
```diff
@@ -0,0 +1,102 @@
...
+            rows * cols, rows, cols, figsize=figsize, backend_kwargs=backend_kwargs
+        )
+        axes = []
+        for i in range(cols, rows * cols, 2 * cols):
+            axes += list(ax[i - cols : i])
+            gs = ax[i].get_gridspec()
+            for j in range(i, i + cols):
+                ax[j].remove()
+            axbig = fig.add_subplot(gs[i // cols, :])
+            axes.append(axbig)
```
OriolAbril, line 44: It feels cleaner to not use `_create_axes_grid` and use completely custom code with gridspec. idea in pseudocode:

```
fig = plt.figure(constrained_layout=rcParams[...])
gs = fig.add_gridspec(ncols=ngroups, nrows=nvars*2)
for i in range(nvars):
    for j in range(ngroups):
        ax_singles = [fig.add_subplot(gs[2*i, j])]
    ax_combined = fig.add_subplot(gs[2*i+1, :])
# combine axes into array, create array in a whole different way
# another option would be creating an array of shape (nvars, ngroup+1)
# so ax[i, :-1] = ax_singles and ax[i, -1] = ax_combined
```

The structure of the axes array has to be defined and explicited to create here, check shape and add to docstring. I think using a flattened array will be confusing.

arviz/plots/backends/matplotlib/ppplot.py
```diff
@@ -0,0 +1,102 @@
...
+        axes = np.ravel(ax)
+        if len(axes) != length_plotters:
+            raise ValueError(
+                "Found {} variables to plot but {} axes instances. They must be equal.".format(
+                    length_plotters, len(axes)
+                )
+            )
+
+    matplotlib.rc("xtick", labelsize=7)
+    matplotlib.rc("ytick", labelsize=7)
```
OriolAbril, line 56: I think `scale_fig_size` should be used here to get textsize, labelsize and linewidth: https://github.com/arviz-devs/arviz/blob/master/arviz/plots/plot_utils.py#L31

percygautam: It's already set in kdeplot. So, I am just removing them.

arviz/plots/ppplot.py
```diff
@@ -0,0 +1,135 @@
...
+    .. plot::
+        :context: close-figs
+
+        >>> import arvizz as az
+        >>> data = az.load_arviz_data('radon')
+        >>> az.plot_pp(data, vars=["defs"], coords={"team" : ["Italy"]})
+
+    """
+
+    groups = ["prior", "posterior"]
```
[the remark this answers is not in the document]
percygautam: Currently, I am doing it for prior/posterior only. Once, it's complete, I'll start adding other groups too. (if need be)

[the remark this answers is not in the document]
OriolAbril: Perfect!

arviz/plots/backends/matplotlib/ppplot.py
```diff
@@ -0,0 +1,91 @@
...
+    backend_kwargs,
+    show,
+):
+    """Matplotlib pp plot."""
+
+    backend_kwargs = {**backend_kwarg_defaults(), **backend_kwargs}
+
+    if ax is None:
+        axes = np.empty((nvars, ngroups + 1), dtype=object)
+        fig = plt.figure(constrained_layout=backend_kwargs, figsize=figsize)
```
OriolAbril, line 32: I think this won't work, maybe `...figure(**backend_kwargs, figsize...)`?

arviz/plots/backends/matplotlib/ppplot.py
```diff
@@ -0,0 +1,91 @@
...
+        gs = fig.add_gridspec(ncols=ngroups, nrows=nvars * 2)
+        for i in range(nvars):
+            axes_single = []
+            for j in range(ngroups):
+                axes_single.append(fig.add_subplot(gs[2 * i, j]))
+            axes[i, :] = np.concatenate([axes_single, [fig.add_subplot(gs[2 * i + 1, :])]])
```
OriolAbril, line 38: You can store them in the array right away:

```
for j in range(ngroups):
    axes[i, j] = fig.add_subplot(gs[2 * i, j])
axes[i, -1] = fig.add_subplot(gs[2 * i + 1, :])
```

arviz/plots/backends/matplotlib/ppplot.py
```diff
@@ -0,0 +1,91 @@
...
+        for idx2, (var, coord, data,) in enumerate(plotter):
+            _pp_helper(
+                ax=axes[idx2, idx],
+                data=data,
+                group=group,
+                var_name=var,
+                coord_name=coord,
+                legend=legend,
+                fill_kwargs=fill_kwargs,
+                plot_kwargs=plot_kwargs,
```
OriolAbril, line 62: fill_kwargs and plot_kwargs should be removed. If necessary the user can do

 `plot_pp(..., posterior_kwargs={"rug": True, "fill_kwargs": {"alpha": 0.7}})`

Thus, passing `posterior_kwargs` to `plot_kde` as `**kwargs` can directly handle plot_kwargs, fill_kwargs, even rug_kwargs if desired.

percygautam: Okay, I understand.

arviz/plots/ppplot.py
```diff
@@ -0,0 +1,135 @@
...
+
+        group = getattr(data, group)
+        coord = {}
+        for key in coords.keys():
+            coord[key] = np.where(np.in1d(group[key], coords[key]))[0]
+
+        plotters = [
+            tup for tup in xarray_var_iter(group.isel(coord), var_names=vars, combined=True,)
+        ]
+        pp_plotters.append(plotters)
```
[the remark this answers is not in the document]
percygautam: Yes, that reduces it.

arviz/plots/ppplot.py
```diff
@@ -0,0 +1,135 @@
...
+        autoscaled based on figsize.
+    coords : dict
+        Dictionary mapping dimensions to selected coordinates to be plotted.
+        Dimensions without a mapping specified will include all coordinates for
+        that dimension. Defaults to including all coordinates for all
+        dimensions if None.
+    legend : bool
+        Add legend to figure. By default True.
+    ax: axes, optional
+        Matplotlib axes or bokeh figures.
```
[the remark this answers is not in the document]
percygautam: What should I add here? Should I specify shape? Or `a numpy 2d array of matplotlib axes` is enough?

percygautam:
What should be the default behaviour, if coords and/or var_names are not provided?

arviz/plots/ppplot.py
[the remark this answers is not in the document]
OriolAbril: I would explain that the returned object will have shape `(nvars, 3/4)` where the last column is the combined plot and the first columns are the single plots.

OriolAbril:
> What should be the default behaviour, if coords and/or var_names are not provided?

I would try to mimic other plotting functions. If no var names are present, plot all the names in posterior group and assume vars in prior have the same name, if they don't, users should use the `data_pair` argument (like in `plot_ppc`) which is currently missing. Regarding coords, I think the same behaviour can be applied.

arviz/plots/ppplot.py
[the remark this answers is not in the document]
percygautam: `nvars` is not defined firsthand. Can we still add that in documentation?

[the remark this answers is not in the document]
percygautam: `nvars` is not defined firsthand. Can we still add that in documentation?

percygautam:
I have tried to apply it to `vars`. Can you confirm it's the right way? Then I'll apply it to `coords`.

arviz/plots/ppplot.py
```diff
@@ -0,0 +1,169 @@
...
+    datasets = [getattr(data, group) for group in groups]
+
+    if var_names is None:
+        var_names = list(datasets[0].data_vars)
+    var_names = _var_names(var_names, datasets[0])
+    vars = [var_names]
+    for i in range(1, len(datasets)):
+        var_name =[data_pairs.get(var, var) for var in var_names]
+        var_name = _var_names(var_name,  datasets[i])
+        vars.append(var_name)
```
OriolAbril, line 123: Code inspired in `plot_density` and `plot_forest` while trying to save the distances: here is the [first relevant piece](https://github.com/arviz-devs/arviz/blob/master/arviz/plots/densityplot.py#L159-L162) and here is the [second one](https://github.com/arviz-devs/arviz/blob/master/arviz/plots/forestplot.py#L155).

The trickiest part is var_names handling, while making sure that vars are kept paired (or kept in trios if it were to be extended to prior predictive/observed data/posterior predictive). The code on var names looks good, my only comment is in prevision of adding the 3 plot case, in which case, data_pairs would not work. Maybe it would be better to use a list/list of lists? something like:
* if var_names = None, use as var names all data_vars in first dataset
* if var_name is str, make `var_names = [var_names]
* if var_name is list and var_names[0] is str, var_names = [var_names for _ in datasets]
* otherwise go on and see if it works, var_names should be a list of lists. Note that I used ifs, not elif

Once `datasets` is created, you should be able to use the same code as in `plot_density` and `plot_forest`. First transform, then select coords. Note that `get_coords` should be able to handle a list of datasets and that therefore `coords` can be either a dict or a list of dicts (unlike the description in `plot_forest` docstring specifies), it would look similar to the code below.

arviz/plots/ppplot.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+    pp_plotters = [
+        list(xarray_var_iter(get_coords(data, coords), var_names=var_names, combined=True))
+        for data , var in zip(datasets, vars)
+    ]
```
OriolAbril, line 129: Make sure `coords` is a list of same length as datasets (like in forestplot) and zip also on coords `... for data, var, coords_i in zip(datasets, vars, coords)`

arviz/plots/backends/matplotlib/ppplot.py
```diff
@@ -0,0 +1,79 @@
...
+    return axes
+
+
+def _pp_helper(ax, data, group, var_name, coord_name, legend, **kwargs):
+    plot_kde(
+        data,
+        label="{} {} {}".format(group, var_name, coord_name) if legend else None,
+        ax=ax,
+        **kwargs,
+    )
```
OriolAbril, line 79: I think this function can be omited and call plot_kde on the main code. Also, using fstrings to create label string will simplify the code and improve readability. I would actually use `_make_label` and then prepend the group to the generated string (thus, if we add configuration options to make_label via rcParams or something similar, changes will be used here too).

arviz/plots/ppplot.py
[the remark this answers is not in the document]
aloctavodia: what about setting the shape to (nvars, 3)?

arviz/plots/ppplot.py
percygautam: I can't seem to understand, why is this necessary after the above changes?

arviz/plots/ppplot.py
```diff
@@ -108,24 +97,31 @@ def plot_pp(
...
-    vars = [var_names]
-    for i in range(1, len(datasets)):
-        var_name =[data_pairs.get(var, var) for var in var_names]
-        var_name = _var_names(var_name,  datasets[i])
+
+    if isinstance(var_names, str):
+        var_names = [var_names]
+
+    if isinstance(var_names, list) or isinstance(var_names, tuple):
+        var_names = [list(var_names) for _ in datasets]
```
OriolAbril, line 109: `var_names` should allow string, list of strings and list of lists but end up being a list of lists always:

```
if isinstance(var_names, str):
   var_names = [var_names]
# now it is a list
if isinstance(var_names[0], str):
    var_names = [var_names for _ in datasets]
# now it is a list of lists the same length of datasets with the var names that correspond to the dataset
```

arviz/plots/ppplot.py
```diff
@@ -108,24 +97,31 @@ def plot_pp(
...
+
+    vars = []
+    for i in range(len(datasets)):
+        var_name = _var_names(var_names[i],  datasets[i])
         vars.append(var_name)
```
OriolAbril: this should not be needed anymore after the changes above

arviz/plots/ppplot.py
OriolAbril: that would be to use different coordinates in the different datasets

arviz/plots/ppplot.py
percygautam: Okay, I understand

arviz/plots/ppplot.py
percygautam: Currently, `coords` are selected using:
```
datasets = get_coords(
        datasets, list(reversed(coords)) if isinstance(coords, (list, tuple)) else coords
    )
```
After this, I don't think we need to apply the above changes. I also tested `az.plot_pp(data, var_names=["defs"], coords=[{"team" : [ "France"]}, {"team" : [ "Italy"]}])` and its working fine. Am I mistaking something?

arviz/plots/ppplot.py
percygautam: I was doing it earlier to check if the `var_names` are present in all groups or not? In that case, the `_var_names` will throw the error.

OriolAbril: I missed that, it can still be added once var_names is a list of lists:

    var_names = [_var_names(vars, dataset) for vars, dataset in zip(var_names, datasets)]

or something similar

percygautam: Okay :)

arviz/plots/ppplot.py
OriolAbril: I am not sure we expect the same behaviour :thinking:. Imagine you have an idata with posterior group containing only a variable `defs` with coords `defs_dim_0: [0, 1, 2]` and prior group containing only a variable `defs_prior` with coords `defs_prior_dim_0: [0, 1, 2]` (it is common to have different names between groups even though they may be the same var, and these are the default dim names, could be `team` and `team_prior`, the point is that coord name could be different too. Would is be possible to compare posterior and prior only for coord==0?

OriolAbril: True, it should work now!

arviz/plots/ppplot.py
```diff
@@ -0,0 +1,157 @@
...
+        var_names = [var_names]
+
+    if isinstance(var_names[0], str):
+        var_names = [var_names for _ in datasets]
+
+    if transform is not None:
+        datasets = [transform(dataset) for dataset in datasets]
+
+    datasets = get_coords(
+        datasets, list(reversed(coords)) if isinstance(coords, (list, tuple)) else coords
```
OriolAbril, line 112: I _think_ it should not be reversed

percygautam: Ohh, yes you're right.

arviz/plots/ppplot.py
```diff
@@ -0,0 +1,157 @@
...
+    ngroups = len(groups)
+
+    (figsize, ax_labelsize, _, xt_labelsize, linewidth, markersize) = _scale_fig_size(
+        figsize, textsize, 2 * nvars, ngroups
+    )
+
+    posterior_plot_kwargs = posterior_kwargs.pop("plot_kwargs", dict())
+    posterior_plot_kwargs.setdefault("color", "red")
+    posterior_plot_kwargs.setdefault("linewidth", linewidth)
+    posterior_kwargs["plot_kwargs"] = posterior_plot_kwargs
```
OriolAbril, line 129: I think there is no need to pop, `posterior_plot_kwargs["plot_kwargs"].setdefault(...)` should work

percygautam:
What should be the next steps? Should I start adding other groups or start testing first?

arviz/plots/ppplot.py
```diff
@@ -105,11 +106,13 @@ def plot_pp(
...
         var_names = [var_names for _ in datasets]
 
+    var_names = [_var_names(vars, dataset) for vars, dataset in zip(var_names, datasets)]
+
     if transform is not None:
         datasets = [transform(dataset) for dataset in datasets]
 
     datasets = get_coords(
-        datasets, list(reversed(coords)) if isinstance(coords, (list, tuple)) else coords
+        datasets, list(coords) if isinstance(coords, (list, tuple)) else coords
```
OriolAbril, line 115: I think bow the conversion to list and if else can also be skipped

percygautam:
I have started with the tests and added the basic test. What other tests can we add?
Also, regarding the new name of the plot as you mentioned earlier, we should ask the community about it. (I don't have any ideas)

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -1159,3 +1160,20 @@ def test_plot_mcse_no_divergences(models):
     idata.sample_stats = idata.sample_stats.rename({"diverging": "diverging_missing"})
     with pytest.raises(ValueError, match="not contain diverging"):
         plot_mcse(idata, rug=True)
+
+@pytest.mark.parametrize(
+    "kwargs",
+    [
+        {},
+        {"var_names": ["theta"]},
+        {"var_names": ["mu"], "coords": {"theta_dim_0": [0, 1]}},
```
OriolAbril, line 1179: var_names should be theta, otherwise it will either be ignored or raise an error

percygautam: I don't understand as all these vars are present in fake data generated in `helpers.py` and the test is working fine.

OriolAbril: My point is that `coords` is not doing anything, so it is not actually testing it works. You can see what happens is a similar situation with the eight schools model:

```
az.plot_pp(idata, var_names="mu", coords={"school": [0,1]})
# and compare with 
az.plot_pp(idata, var_names="mu", coords={"school": [3,7]})
```

percygautam: Okay I understand.

percygautam: I think we can add other tests after adding more groups or can you suggest some tests to add?

percygautam:
I am planning to add other groups to the plot, starting with `observed_data`. I have a doubt regarding the acquisition of data. In the example given in #749 we are doing something like:
```
    li = []
    for i in pp_ob_plotters:
        li.append(i[2])
    az.plot_kde(np.array(li).flatten(), plot_kwargs={'color':'r'},ax=ax3)
```
Are we looking for the same in the plot? How will we select `var_names` for these other groups?

OriolAbril:
I think most of the plotting is already coded to accept any number of groups, the only thing that needs some extra work being rcParams. Have you checked what happens if you modify the groups to be 3?

percygautam:
The `var_names` are not the same in all the groups. e.g. `posterior_predictive` and `observed_data` have vars: `home_points` and `away_points`.

OriolAbril:
There are two things to take into account here, var_names being different should be proporly handled by the code (var names can be a list of lists), thus prior and posterior can be plotted together and so can prior predictive, observed data and posterior predictive.

The other factor us the _variables_ themselves being different. It does not make sense to compare prior and prior predictive, they refer to conceptually different and not comparable quantities

percygautam:
Okay, I understand now and the code is working fine for other groups. I have done a minor change regarding groups. (can be changed after metagroups PR).

> the only thing that needs some extra work being rcParams.

What work are you referring too?

OriolAbril:
Only the two combinations shown in the issue make sense to plot. I think other group combinations should NOT be supported, it does not really make sense to use metagroups here (not in plot_ppc for that matter) as they basically use a very restrictive and fixed set of group combinations. 

I would therefore add an argument to choose between either prior-posterior (maybe `kind="latent"`) or prior predictive-observed data-posterior predictive (`kind="observed"`?). The work missing then is adding the `observed_data_kwargs` so that prior_kwargs is used for prior/prior predictive... and if observed_data is not used, its kwargs are just ignored.

As a note on metagroups and why I proposed the idea, functions of interest that can be applied to idata objects can very different, from functions to be applied only to one group or to these same subsets (undoing a scaling in observed data that also has to be applied to predictive groups, including predictions if present), to functions to be applied to all groups (loading datasets to memory) and to practically any combination of groups (even though variables in posterior and observed data are conceptually different, they can share dimensions and coordinates that users may want to rename, reshape, stack...)

percygautam:
okay, got it 👍

arviz/plots/backends/matplotlib/ppcplot.py
```diff
@@ -1,401 +1,68 @@
-"""Matplotib Posterior predictive plot."""
-from matplotlib import animation
+"""Matplotib Prior Posterior plot."""
```
ahartikainen: Matplotlib

arviz/plots/backends/matplotlib/ppcplot.py
```diff
@@ -1,4 +1,4 @@
-"""Matplotib Posterior predictive plot."""
+"""Matplolib Posterior predictive plot."""
```
OriolAbril, line 1: now the final t of plot is missing

arviz/plots/ppplot.py
```diff
@@ -0,0 +1,181 @@
...
+
+    nvars = len(pp_plotters[0])
+    ngroups = len(groups)
+
+    (figsize, ax_labelsize, _, xt_labelsize, linewidth, markersize) = _scale_fig_size(
+        figsize, textsize, 2 * nvars, ngroups
+    )
+
+    posterior_kwargs.setdefault("plot_kwargs", dict())
+    posterior_kwargs["plot_kwargs"].setdefault("color", "red")
```
OriolAbril, line 148: I would use "C0" and then covert to hex (I think the hex PR is still pending but ready to merge), however "C0" is a matplotlib defined color, if we ever want to not depend on mpl and allow having only mpl or bokeh to plot this would make things harder. cc @ahartikainen @canyon289

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -1169,3 +1170,34 @@ def test_plot_mcse_no_divergences(models):
...
+        {"var_names": ["theta"], "coords": {"theta_dim_0": [0, 1]}},
+        {"var_names": ["eta"], "posterior_kwargs": {"rug": True, "rug_kwargs": {"color": "r"}}},
+        {"var_names": ["mu"], "prior_kwargs": {"fill_kwargs": {"alpha": 0.5}}},
+        {
+            "var_names": ["tau"],
+            "prior_kwargs": {"plot_kwargs": {"color": "r"}},
+            "posterior_kwargs": {"plot_kwargs": {"color": "b"}},
+        },
+    ],
+)
```
OriolAbril, line 1189: add one case using kind

arviz/plots/ppplot.py
OriolAbril: then use c1 and c2 below to have colors depend on style set

arviz/plots/backends/matplotlib/ppcplot.py
percygautam: LOL, sorry. I'll correct it.

arviz/plots/ppplot.py
```diff
@@ -0,0 +1,181 @@
...
+        Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
+    backend_kwargs: bool, optional
+        These are kwargs specific to the backend being used. For additional documentation
+        check the plotting method of the backend.
+    show : bool, optional
+        Call backend show function.
+
+    Returns
+    -------
+    axes : matplotlib axes or bokeh figures
```
rpgoldman, line 76: What's the return type more specifically? Is it a numpy array of axes or figures?  It can't be a single `matplotlib.axes.Axes`, right?

arviz/plots/ppplot.py
```diff
@@ -0,0 +1,181 @@
...
+    ax=None,
+    prior_kwargs=None,
+    posterior_kwargs=None,
+    observed_kwargs=None,
+    backend=None,
+    backend_kwargs=None,
+    show=None,
+):
+    """
+    Plot for posterior/Prior.
```
rpgoldman, line 30: Maybe "Plots that compare fitted and unfitted distributions." as the one-liner?

Then maybe a more extended description of exactly what the comparisons will be?

E.g.

"The resulting plots will show the compared distributions (as KDEs) both on separate axes (particularly useful when one of them is substantially tighter than another), and plotted together, so three plots per distribution.

The groups plotted are dictated by the value of the ``kind`` argument."

arviz/plots/ppplot.py
rpgoldman: Ah, I see -- the return value is actually described in the description of the `ax` argument.  Suggest you move (or copy) that material down to "Returns", and change the parameter description to read "The ``ax`` argument should have shape ``(nvars, 3)``, where the last column is for the combined before/after plots and columns 0 and 1 are for the before and after plots, respectively."

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -1185,6 +1185,7 @@ def test_plot_mcse_no_divergences(models):
             "prior_kwargs": {"plot_kwargs": {"color": "r"}},
             "posterior_kwargs": {"plot_kwargs": {"color": "b"}},
         },
+        {"var_names": ["theta"], "kind": "latent"},
```
OriolAbril, line 1188: latent is the default, better to use `"observed"` to make sure code does not depend on number of groups and check so every time the code is modified

arviz/plots/ppplot.py
```diff
@@ -0,0 +1,182 @@
...
+
+    nvars = len(pp_plotters[0])
+    ngroups = len(groups)
+
+    (figsize, ax_labelsize, _, xt_labelsize, linewidth, markersize) = _scale_fig_size(
+        figsize, textsize, 2 * nvars, ngroups
+    )
+
+    posterior_kwargs.setdefault("plot_kwargs", dict())
+    posterior_kwargs["plot_kwargs"].setdefault("color", "C0")
```
OriolAbril, line 149: I know this is currently matplotlib only, but now that conversion_to_hex has been merged, this should be updated to:

```
posterior_kwargs["plot_kwargs"].setdefault("color", "C0")
posterior_kwargs["plot_kwargs"]["color"] = vectorized_to_hex(posterior_kwargs["plot_kwargs"]["color"])
```

to ease the implementation of the bokeh version using the same colors as mpl one

OriolAbril: In fact, maybe:

```
posterior_kwargs["plot_kwargs"]["color"] = vectorized_to_hex(posterior_kwargs["plot_kwargs"].get("color", "C0"))
```

is more pythonic? I _think_ it does the same. Also, shouldn't `posterior_kwargs["plot_kwargs"]` be dealiased before setting the defaults?

arviz/plots/ppplot.py
```diff
@@ -0,0 +1,182 @@
...
+    posterior_kwargs["plot_kwargs"].setdefault("linewidth", linewidth)
+
+    prior_kwargs.setdefault("plot_kwargs", dict())
+    prior_kwargs["plot_kwargs"].setdefault("color", "C1")
```
OriolAbril, line 153: same as above

arviz/plots/ppplot.py
```diff
@@ -0,0 +1,182 @@
...
+    prior_kwargs["plot_kwargs"].setdefault("linewidth", linewidth)
+
+    observed_kwargs.setdefault("plot_kwargs", dict())
+    observed_kwargs["plot_kwargs"].setdefault("color", "C2")
```
OriolAbril, line 157: same as above

arviz/plots/ppplot.py
percygautam: Okay, will do. I also thought earlier about dealiasing, but procrastinated it for the end. Will add now.

aloctavodia:
> I think this is ready to merge, only minor comments. I would change the name though as `plot_pp` will probably be confused with `plot_ppc` often. I don't really know to what I'd change it to though

what do you think of `plot_compare_densities`? And what about making the kind "latent" more explicit and use "prior_posterior"?

arviz/plots/ppplot.py
percygautam: > I think it does the same. Also, shouldn't posterior_kwargs["plot_kwargs"] be dealiased before setting the defaults?

The dealiasing has been handled in plot_kde itself. I don't think we need to dealiase here.

arviz/plots/backends/matplotlib/ppplot.py
```diff
@@ -1,4 +1,4 @@
-"""Matplolib Prior Posterior plot."""
+"""Matplotlib Density Comparision plot."""
```
ahartikainen, line 1: comparison

rpgoldman:
> > I think this is ready to merge, only minor comments. I would change the name though as `plot_pp` will probably be confused with `plot_ppc` often. I don't really know to what I'd change it to though
> 
> what do you think of `plot_compare_densities`? And what about making the kind "latent" more explicit and use "prior_posterior"?

I'm not sure that prior/posterior is more explicit than latent, because the former specifies what distributions are being compared and the latter specifies what variables are being compared.

Whatever we choose, I think it should be parallel.  So if we choose prior_posterior to replace latent, we should choose something like `observed_prior_predictive` and `observed_posterior_predictive` for the other `kind`, which is a lot of typing!

OriolAbril:
How about `plot_dist_comparison` as name?

arviz/plots/distcomparisonplot.py
```diff
@@ -0,0 +1,195 @@
...
+    var_names = [_var_names(vars, dataset) for vars, dataset in zip(var_names, datasets)]
+
+    if transform is not None:
+        datasets = [transform(dataset) for dataset in datasets]
+
+    datasets = get_coords(datasets, coords)
+    dc_plotters = [
+        list(xarray_var_iter(data, var_names=var, combined=True))
+        for data, var in zip(datasets, var_names)
+    ]
```
OriolAbril, line 142: cut `dc_plotters` list so that no more than `rcParams["plot.max_subplots"]` are generated (also ensure at least one var is plotted, if max_subplots were 2, plot one variable instead of none, plotting variables partially is not worth the effort).

arviz/plots/ppplot.py
OriolAbril: If the user passes `c` as part of `plot_kwargs` and we then do the `setdefault` with `color`, what does dealiasing do? Keeps `c` (should be as it is the user introduced one), keeps `color` or crashes?

canyon289:

percygautam:
I think most of the work for matplotlib backend is done? We can do the bokeh backend in a follow-up PR. Meanwhile, we can merge this?

canyon289:
@percygautam Tests are failing so those need to be fixed. Also its ok to do boken in a followup but  would add code in Bokeh that raises a warning that we haven't implemented if someone tries to use that backend, rather

arviz/plots/distcomparisonplot.py
```diff
@@ -0,0 +1,197 @@
...
+    ax=None,
+    prior_kwargs=None,
+    posterior_kwargs=None,
+    observed_kwargs=None,
+    backend=None,
+    backend_kwargs=None,
+    show=None,
+):
+    """Plot that compare fitted and unfitted distributions.
+    The resulting plots will show the compared distributions both on
```
OriolAbril, line 31:
```suggestion
    """Plot to compare fitted and unfitted distributions.

    The resulting plots will show the compared distributions both on
```

arviz/plots/distcomparisonplot.py
```diff
@@ -0,0 +1,197 @@
...
+    .. plot::
+        :context: close-figs
+
+        >>> import arviz as az
+        >>> data = az.load_arviz_data('radon')
+        >>> az.plot_dist_comparison(data, var_names=["defs"], coords={"team" : ["Italy"]})
+
+    """
+
+    all_groups = ["prior", "posterior"]
```
OriolAbril, line 95:
```suggestion
    """
    all_groups = ["prior", "posterior"]
```

arviz/plots/distcomparisonplot.py
```diff
@@ -0,0 +1,203 @@
...
+    )
+
+    if backend is None:
+        backend = rcParams["plot.backend"]
+    backend = backend.lower()
+    if backend == "bokeh":
+        warnings.warn(
+            "The bokeh backend is still under development. Using matplotlib bakend.", UserWarning
+        )
+    backend = "matplotlib"
```
canyon289, line 198: This should be moved to the backends, the high level plots shouldnt refer to backend at all. Also would prefer if plot doesnt load at all if bokeh is used and instead just raises warning

arviz/plots/backends/bokeh/distcomparisonplot.py
```diff
@@ -0,0 +1,23 @@
...
+    posterior_kwargs,
+    observed_kwargs,
+    backend_kwargs,
+    show,
+):
+    """Bokeh Density Comparison plot."""
+    warnings.warn(
+        "The bokeh backend is still under development. Use matplotlib bakend.", UserWarning
+    )
+    raise
```
OriolAbril, line 23:
```suggestion
    raise NotImplementedError("The bokeh backend is still under development. Use matplotlib bakend.")
```

Not sure if this will be black compliant or loo long

arviz/tests/base_tests/test_plots_bokeh.py
```diff
@@ -956,3 +957,8 @@ def test_plot_posterior_point_estimates(models, point_estimate):
 def test_plot_rank(models, kwargs):
     axes = plot_rank(models.model_1, backend="bokeh", show=False, **kwargs)
     assert axes.shape
+
+
+def test_plot_dist_comparison_warn(models):
+    with pytest.warns(UserWarning, match="The bokeh backend.+Using matplotlib bakend."):
```
OriolAbril, line 963:
```suggestion
    with pytest.raises(NotImplementedError, match="The bokeh backend.+Using matplotlib bakend."):
```

percygautam:
I think it's ready to merge!

PR #1149, opened 2020-04-12
Fix issues with ESS docstring.

rpgoldman:
There was something wrong in the embedded latex code that I had to fix.
Also there were some missing requirements from requirements-dev.

### DIFF
--- a/arviz/stats/diagnostics.py
+++ b/arviz/stats/diagnostics.py
@@ -75,7 +75,7 @@ def ess(data, *, var_names=None, method="bulk", relative=False, prob=None):
         Refer to documentation of ``az.convert_to_dataset`` for details.
         For ndarray: shape = (chain, draw).
         For n-dimensional ndarray transform first to dataset with ``az.convert_to_dataset``.
-    var_names : list of strings
+    var_names : string or list of strings
         Names of variables to include in the return value Dataset.
     method : str, optional, default "bulk"
         Select ess method. Valid methods are:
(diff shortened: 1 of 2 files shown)


requirements-dev.txt
```diff
@@ -4,3 +4,7 @@ pylint
 pytest
 pytest-cov
 black ; python_version >= '3.6'
+# needed to build docs:
+m2r
```
ahartikainen, line 8: We have `requirements-docs.txt`

rpgoldman:
@ahartikainen Thanks for explaining that. I can revert that change and re-push, or you could just drop it yourself if you want.

arviz/stats/diagnostics.py
```diff
@@ -66,18 +66,18 @@ def bfmi(data):
...
-        Refer to documentation of az.convert_to_dataset for details.
+        Any object that can be converted to an ``az.InferenceData`` object.
+        Refer to documentation of ``az.convert_to_dataset`` for details.
         For ndarray: shape = (chain, draw).
-        For n-dimensional ndarray transform first to dataset with az.convert_to_dataset.
-    var_names : list
-        Names of variables to include in the effective_sample_size_mean report
-    method : str
+        For n-dimensional ndarray transform first to dataset with ``az.convert_to_dataset``.
+    var_names : list of strings
```
aloctavodia, line 78: strings are allowed in the case of a single variable.

PR #1150, opened 2020-04-14
Add warning for log scale default in compare/loo/waic functions

AlexAndorra:
## Description
This PR adds a reminder about log scale default and how it should be interpreted (higher is better) for `az.waic` and `az.loo`. This merely mirrors what we already say in the doc.

For `az.compare` I added a `UserWarning` following the change to log scale in 0.7.0. This adresses [this issue](https://github.com/pymc-devs/pymc3/issues/3874), and the warning will be temporary, as discussed in the issue.

In passing, I trailed all the white spaces before colons in the docstrings of `stats.py`. This messes up the formatting on the website.

I'm here for any change, and thanks in advance for the review 🖖

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -142,9 +142,10 @@ def compare(
         scale_value = 1
         ascending = False
         warnings.warn(
-            "\nThe scale is now log by default. Use 'scale' argument if you "
-            "rely on a specific value.\n A higher log-score (or a lower "
-            "deviance) indicates a model with better predictive accuracy."
+            "\nThe scale is now log by default. Use 'scale' argument or "
+            "'stats.ic_scale' rcParam if you rely on a specific value.\nA higher "
+            "log-score (or a lower deviance) indicates a model with better predictive "
+            "accuracy."
         )
     else:
         if scale == "negative_log":
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -54,12 +54,12 @@ def compare(
 
     Parameters
     ----------
-    dataset_dict : dict[str] -> InferenceData
+    dataset_dict: dict[str] -> InferenceData
         A dictionary of model names and InferenceData objects
-    ic : str
+    ic: str
         Information Criterion (PSIS-LOO `loo` or WAIC `waic`) used to compare models. Defaults to
         ``rcParams["stats.information_criterion"]``.
-    method : str
+    method: str
         Method used to estimate the weights for each model. Available options are:
 
         - 'stacking' : stacking of predictive distributions.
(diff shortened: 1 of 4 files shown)


arviz/stats/stats_utils.py
```diff
@@ -427,6 +427,10 @@ def get_log_likelihood(idata, var_name=None):
 
 BASE_FMT = """Computed from {{n_samples}} by {{n_points}} log-likelihood matrix
 
+The scale is now log by default. Use 'scale' argument if you rely on a specific value.
+A higher log-score (or a lower deviance) indicates a model with better predictive 
+accuracy.
```
OriolAbril, line 432: I would put this on the bottom of the message (and if possible even below pointwise summary). I would then be able to copy/screenshot the upper part of the printed text and omit the warning

AlexAndorra: Good idea, will update

AlexAndorra:
Yeah, but weirdly it also take the argument's type when double-clicking it. When you remove the white space, it just selects the parameter, as it should -- don't know why 🤷 

Thanks for the review Oriol, I updated in light of your comments.

OriolAbril:
> Yeah, but weirdly it also take the argument's type when double-clicking it. When you remove the white space, it just selects the parameter, as it should -- don't know why shrug

Ohh, true, I hadn't noticed, and it looks like it happens with any documentation with space before the colon :thinking:

AlexAndorra:
Yeah that's weird 😅 

Do we still need `assert loo_data == loo_pointwise[: len(loo_data)]` in `test_loo_print`? It's this line that messes up the tests, because it's no longer the case.

ahartikainen:
I think we don't need it. What is it now?

AlexAndorra:
Now it includes the warning I introduced in this PR:
```
Computed from 2000 by 8 log-likelihood matrix
         Estimate       SE
elpd_loo   -30.81     1.43
p_loo        0.95        -
The scale is now log by default. Use 'scale' argument if you rely on a specific value. 
A higher log-score (or a lower deviance) indicates a model with better predictive 
accuracy.
```
when default,
```
Computed from 2000 by 8 log-likelihood matrix
         Estimate       SE
elpd_loo   -30.81     1.43
p_loo        0.95        -
------
Pareto k diagnostic values:
                         Count   Pct.
(-Inf, 0.5]   (good)        6   75.0%
 (0.5, 0.7]   (ok)          2   25.0%
   (0.7, 1]   (bad)         0    0.0%
   (1, Inf)   (very bad)    0    0.0%
The scale is now log by default. Use 'scale' argument if you rely on a specific value. 
A higher log-score (or a lower deviance) indicates a model with better predictive 
accuracy.
```
when pointwise.
Which explains why the test is failing.

OriolAbril:
It can be removed, black is also complaining, make sure to run black right before commiting

AlexAndorra:
Done! Thanks guys 👌

arviz/stats/stats.py
```diff
@@ -138,6 +141,11 @@ def compare(
     if scale == "log":
         scale_value = 1
         ascending = False
+        warnings.warn(
+            "\nThe scale is now log by default. Use 'scale' argument if you "
+            "rely on a specific value.\n A higher log-score (or a lower "
```
OriolAbril, line 146: space after \n can be removed

arviz/stats/stats_utils.py
```diff
@@ -439,6 +439,10 @@ def get_log_likelihood(idata, var_name=None):
    (0.7, 1]   (bad)      {{4:{0}d}} {{8:6.1f}}%
    (1, Inf)   (very bad) {{5:{0}d}} {{9:6.1f}}%
 """
+SCALE_WARNING_FORMAT = """
+The scale is now log by default. Use 'scale' argument if you rely on a specific value.
```
OriolAbril, line 443: could we use `Use 'scale' argument or 'stats.ic_scale' rcParam if...`? Same goes for warning in compare

AlexAndorra:
Thanks Oriol! It's now updated -- and I had already added it to the changelog 👌

PR #1154, opened 2020-04-19
Add `filter` feature for `var_names`

AlexAndorra:
## Description
- Currently, users can select variables entering `az.summary` and other plotting/diagnostics functions with the `var_names` kwarg. 
This selection must however be exact, meaning that if you have variables with similar names in your trace, you have to specify them all: `az.summary(idata, var_names=["alpha", "beta1", "beta2", "beta3"])`, which can be cumbersome for big models.

- This PR introduces the `filter_like` kwarg to allow users to select _all_ variables containing the given word. For instance, `az.summary(idata, var_names=["alpha", "beta"], filter_like=True)` would give the same result as above. This also works to _exclude_ variables with similar names: `az.summary(idata, var_names=["~beta"], filter_like=True)`

- `filter_like` defaults to `False` to ensure backwards-compatibility.

Before adding these changes to the changelog, I had pending questions:
1. Should we add `filter_like` to the other functionalities that use `var_names`, and not only to `summary`?
2. Should we include new tests for this feature?
- Bonus points: the new code may be too convoluted at times, so please tell me if you see a simpler, more pythonic way to do things.

I'm of course available for any changes, and thanks a lot in advances for your reviews 🖖

### DIFF
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -14,7 +14,7 @@ from ..utils import _var_names
 def plot_autocorr(
     data,
     var_names=None,
-    filter=None,
+    filter_vars=None,
     max_lag=None,
     combined=False,
     figsize=None,
@@ -38,7 +38,7 @@ def plot_autocorr(
         Variables to be plotted, if None all variable are plotted. Prefix the
         variables by `~` when you want to exclude them from the plot. Vector-value
         stochastics are handled automatically.
-    filter: {None, "like", "regex"}, optional, default=None
+    filter_vars: {None, "like", "regex"}, optional, default=None
         If `None` (default), interpret var_names as the real variables names. If "like",
         interpret var_names as substrings of the real variables names. If "regex",
         interpret var_names as regular expressions on the real variables names. A la
@@ -95,7 +95,7 @@ def plot_autocorr(
     .. plot::
         :context: close-figs
 
-        >>> az.plot_autocorr(data, var_names=['~thet'], filter="like", combined=True)
+        >>> az.plot_autocorr(data, var_names=['~thet'], filter_vars="like", combined=True)
 
 
     Specify maximum lag (x axis bound)
@@ -106,7 +106,7 @@ def plot_autocorr(
         >>> az.plot_autocorr(data, var_names=['mu', 'tau'], max_lag=200, combined=True)
     """
     data = convert_to_dataset(data, group="posterior")
-    var_names = _var_names(var_names, data, filter)
+    var_names = _var_names(var_names, data, filter_vars)
 
     # Default max lag to 100 or max length of chain
     if max_lag is None:
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -14,6 +14,7 @@ from ..utils import _var_names
 def plot_autocorr(
     data,
     var_names=None,
+    filter=None,
     max_lag=None,
     combined=False,
     figsize=None,
@@ -30,18 +31,24 @@ def plot_autocorr(
 
     Parameters
     ----------
-    data : obj
+    data: obj
         Any object that can be converted to an az.InferenceData object
         Refer to documentation of az.convert_to_dataset for details
-    var_names : list of variable names, optional
-        Variables to be plotted, if None all variable are plotted.
-        Vector-value stochastics are handled automatically.
-    max_lag : int, optional
+    var_names: list of variable names, optional
+        Variables to be plotted, if None all variable are plotted. Prefix the
+        variables by `~` when you want to exclude them from the plot. Vector-value
+        stochastics are handled automatically.
+    filter: Union[None, "like", "regex"], optional, default=None
+        If `None` (default), interpret var_names as the real variables names. If "like",
+        interpret var_names as substrings of the real variables names. If "regex",
+        interpret var_names as regular expressions on the real variables names. À la
+        `pandas.filter`.
+    max_lag: int, optional
         Maximum lag to calculate autocorrelation. Defaults to 100 or num draws, whichever is smaller
-    combined : bool
+    combined: bool
         Flag for combining multiple chains into a single chain. If False (default), chains will be
         plotted separately.
-    figsize : tuple
+    figsize: tuple
         Figure size. If None it will be defined automatically.
         Note this is not used if ax is supplied.
     textsize: float
@@ -57,12 +64,12 @@ def plot_autocorr(
     backend_kwargs: dict, optional
         These are kwargs specific to the backend being used. For additional documentation
         check the plotting method of the backend.
-    show : bool, optional
+    show: bool, optional
         Call backend show function.
 
     Returns
     -------
-    axes : matplotlib axes or bokeh figures
+    axes: matplotlib axes or bokeh figures
 
     Examples
     --------
@@ -88,7 +95,7 @@ def plot_autocorr(
     .. plot::
         :context: close-figs
 
-        >>> az.plot_autocorr(data, var_names=['mu', 'tau'], combined=True)
+        >>> az.plot_autocorr(data, var_names=['~theta'], filter="like", combined=True)
 
 
     Specify maximum lag (x axis bound)
@@ -99,7 +106,7 @@ def plot_autocorr(
         >>> az.plot_autocorr(data, var_names=['mu', 'tau'], max_lag=200, combined=True)
     """
     data = convert_to_dataset(data, group="posterior")
-    var_names = _var_names(var_names, data)
+    var_names = _var_names(var_names, data, filter)
 
     # Default max lag to 100 or max length of chain
     if max_lag is None:
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -38,10 +38,10 @@ def plot_autocorr(
         Variables to be plotted, if None all variable are plotted. Prefix the
         variables by `~` when you want to exclude them from the plot. Vector-value
         stochastics are handled automatically.
-    filter: Union[None, "like", "regex"], optional, default=None
+    filter: {None, "like", "regex"}, optional, default=None
         If `None` (default), interpret var_names as the real variables names. If "like",
         interpret var_names as substrings of the real variables names. If "regex",
-        interpret var_names as regular expressions on the real variables names. À la
+        interpret var_names as regular expressions on the real variables names. A la
         `pandas.filter`.
     max_lag: int, optional
         Maximum lag to calculate autocorrelation. Defaults to 100 or num draws, whichever is smaller
@@ -90,12 +90,12 @@ def plot_autocorr(
         >>> az.plot_autocorr(data, var_names=['mu', 'tau'] )
 
 
-    Combine chains collapsing by variable
+    Combine chains by variable and select variables by excluding some with partial naming
 
     .. plot::
         :context: close-figs
 
-        >>> az.plot_autocorr(data, var_names=['~theta'], filter="like", combined=True)
+        >>> az.plot_autocorr(data, var_names=['~thet'], filter="like", combined=True)
 
 
     Specify maximum lag (x axis bound)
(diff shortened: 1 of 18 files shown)


arviz/stats/stats.py
```diff
@@ -949,6 +950,9 @@ def summary(
         Refer to documentation of az.convert_to_dataset for details
     var_names: list
         Names of variables to include in summary
+    filter_like: bool, default=False
+        Whether to interpret var_names as substrings of the real variables names. À la
+        `pandas.filter("like")`.
```
OriolAbril, line 955: Does pandas.filter accept regular expressions?

arviz/utils.py
```diff
@@ -50,14 +51,26 @@ def _var_names(var_names, data):
...
+            var[1:] for var in var_names if var.startswith("~") and var not in all_vars
+        ]
         if excluded_vars:
-            var_names = [i for i in all_vars if i not in excluded_vars]
+            if filter_like:
+                for var in excluded_vars:
+                    if var not in all_vars:
+                        real_vars = [real_var for real_var in all_vars if var in real_var]
+                        excluded_vars.extend(real_vars)
+                        excluded_vars.remove(var)
```
OriolAbril, line 63: what would happen if one of the `real_vars` was the same as `var`? i.e. `theta` in eight schools -> exlcude `theta` and `theta_t`

arviz/stats/stats.py
AlexAndorra: Yes, if you use `regex` as kwarg instead of `like`

arviz/utils.py
AlexAndorra: Mmmh I don't think it's possible because `real_var` is in `all_vars`, but `var` is not in `all_vars`. Is that your question?

arviz/stats/stats.py
OriolAbril: Ohhh, I see now

arviz/utils.py
AlexAndorra: Are we on the same page here @OriolAbril ?

AlexAndorra:
Thanks for the review @OriolAbril ! 

- I implemented the regex feature, extended the `filter` kwarg to all functions taking `var_names` as arg, and added tests for this new feature in `test_utils.py`.

- I also updated the examples in the docstrings when useful and deleted the white space between args' names and the colon in docstrings (as discussed in another PR).

- Here is how this works now: imagine that `all_vars=["alpha", "beta1", "beta2", "p1", "p2", "phi"]`. Users will now be able to:
    - Filter by partial naming: `az.summary(idata, var_names=["alpha", "beta"], filter="like") --> ["alpha", "beta1", "beta2"]`
    - `az.summary(idata, var_names=["~beta"], filter="like") --> ["alpha", "p1", "p2", "phi"]`
    - Filter by regular expressions: `az.summary(idata, var_names=["^bet"], filter="regex") --> ["beta1", "beta2"]`
    - `az.summary(idata, var_names=["~p[0-9]+"], filter="regex") --> ["alpha", "beta1", "beta2", "phi"]`

Tell me if you see other necessary changes, otherwise I hope this is ready to merge 🖖

arviz/plots/autocorrplot.py
```diff
@@ -30,18 +31,24 @@ def plot_autocorr(
...
-        Vector-value stochastics are handled automatically.
-    max_lag : int, optional
+    var_names: list of variable names, optional
+        Variables to be plotted, if None all variable are plotted. Prefix the
+        variables by `~` when you want to exclude them from the plot. Vector-value
+        stochastics are handled automatically.
+    filter: Union[None, "like", "regex"], optional, default=None
+        If `None` (default), interpret var_names as the real variables names. If "like",
+        interpret var_names as substrings of the real variables names. If "regex",
+        interpret var_names as regular expressions on the real variables names. À la
```
ahartikainen, line 44: These special letters are nice until they hit some lib that handles only ascii etc (some versions of pip?; windows sometimes)

arviz/plots/autocorrplot.py
```diff
@@ -30,18 +31,24 @@ def plot_autocorr(
...
         Refer to documentation of az.convert_to_dataset for details
-    var_names : list of variable names, optional
-        Variables to be plotted, if None all variable are plotted.
-        Vector-value stochastics are handled automatically.
-    max_lag : int, optional
+    var_names: list of variable names, optional
+        Variables to be plotted, if None all variable are plotted. Prefix the
+        variables by `~` when you want to exclude them from the plot. Vector-value
+        stochastics are handled automatically.
+    filter: Union[None, "like", "regex"], optional, default=None
```
OriolAbril, line 41: it should use docstring formatting, not type hints (type hints go in the function definition). In this case it would be:

    filter: {None, "like", "regex"}, optional

See references in https://github.com/arviz-devs/arviz/blob/master/CONTRIBUTING.md#docstring-formatting-and-type-hints

arviz/utils.py
```diff
@@ -50,14 +54,34 @@ def _var_names(var_names, data):
...
 
+        excluded_vars = [
+            var[1:] for var in var_names if var.startswith("~") and var not in all_vars
+        ]
         if excluded_vars:
-            var_names = [i for i in all_vars if i not in excluded_vars]
-
-        existent_vars = np.isin(var_names, all_vars)
-        if not np.all(existent_vars):
+            if (filter == "like") or (filter == "regex"):
```
ahartikainen, line 61: filter in ("like, "regex")

We should also call `str(filter).lower()` before this line

arviz/plots/autocorrplot.py
```diff
@@ -88,7 +95,7 @@ def plot_autocorr(
     .. plot::
         :context: close-figs
 
-        >>> az.plot_autocorr(data, var_names=['mu', 'tau'], combined=True)
+        >>> az.plot_autocorr(data, var_names=['~theta'], filter="like", combined=True)
```
aloctavodia, line 98: The tittle should be different like     "Combine chains collapsing by variable and negate variable names".
setting `filter="like"`is not necessary to negate a variable, right?

arviz/utils.py
```diff
@@ -50,14 +54,34 @@ def _var_names(var_names, data):
...
+                    if pattern not in all_vars:
+                        if filter == "like":
+                            real_vars = [real_var for real_var in all_vars if pattern in real_var]
+                        else:
+                            # i.e filter == "regex"
+                            real_vars = [
+                                real_var for real_var in all_vars if re.search(pattern, real_var)
+                            ]
+                        excluded_vars.extend(real_vars)
+                        excluded_vars.remove(pattern)
```
ahartikainen, line 72: This will skip iterations. If needed, copy first the loop variable

arviz/plots/essplot.py
```diff
@@ -125,7 +132,7 @@ def plot_ess(
         :context: close-figs
 
         >>> az.plot_ess(
-        ...     idata, kind="quantile", var_names=["mu", "theta"], coords=coords
+        ...     idata, kind="quantile", var_names=['~theta'], filter="like", coords=coords
```
aloctavodia, line 135: Idem previous comment

arviz/utils.py
OriolAbril, line 72: Back to these two lines. I will try to better define my question, which could then be added as a test (I think). Imagine we have `ds.data_vars = ["theta", "theta_t"]` and we call a given function with `az.xyz(..., var_names="~theta", filter="like")`. We then have that `excluded_vars = ["theta"]` and in the pattern loop: `real_vars=["theta", "theta_t"]` which means that right after extending, `excluded_vars = ["theta", "theta", "theta_t"]` and then the `.remove` deletes the first occurrence of theta. Eventually, as `.remove` only deletes the first occurrence it does work, but it feels very convoluted. I would remove `pattern` at the beginning of the iteration and leave the extend at the end, this would make the operation feel more like _substituting the pattern for the list of matches_ and would not require people reading the code to know that `.remove` deletes only the first occurrence to be able to understand what is happening.

Just realized about the `if pattern not in all_vars` which would make all of the above not happen if `pattern == "theta"` right? Sorry I have not been able to try the code and my _brain python_ tends to miss important details.

arviz/utils.py
```diff
@@ -50,14 +54,34 @@ def _var_names(var_names, data):
...
+            var[1:] for var in var_names if var.startswith("~") and var not in all_vars
+        ]
         if excluded_vars:
-            var_names = [i for i in all_vars if i not in excluded_vars]
-
-        existent_vars = np.isin(var_names, all_vars)
-        if not np.all(existent_vars):
+            if (filter == "like") or (filter == "regex"):
+                for pattern in excluded_vars:
+                    if pattern not in all_vars:
```
OriolAbril, line 63: I _think_ this should not be here (see below too) because it should be valid to use the name itself even if we are using filter, the important check is the check above of `var not in all_vars` when creating `excluded_vars`

arviz/tests/base_tests/test_utils.py
```diff
@@ -64,6 +57,36 @@ def test_var_names_key_error(data):
...
+    data = from_dict(
+        posterior={
+            "alpha": np.random.randn(10),
+            "beta1": np.random.randn(10),
+            "beta2": np.random.randn(10),
+            "p1": np.random.randn(10),
+            "p2": np.random.randn(10),
+            "phi": np.random.randn(10),
+        }
+    ).posterior
```
OriolAbril, line 85: I think this could use `dict_to_dataset` to avoid doing the `.posterior` and get the dataset straight away.

arviz/tests/base_tests/test_utils.py
```diff
@@ -64,6 +57,36 @@ def test_var_names_key_error(data):
         _var_names(("theta", "tau", "bad_var_name"), data)
 
 
+@pytest.mark.parametrize(
+    "var_names_expected",
```
OriolAbril, line 61: given that this is `var_names`, `expected` and `filter`, I would name it `args`

arviz/plots/autocorrplot.py
AlexAndorra: Good catch! You're talking about "À", right?

arviz/utils.py
AlexAndorra: You sound just like pylint 😜

arviz/plots/autocorrplot.py
AlexAndorra: My bad, will update, thanks 👌

arviz/plots/autocorrplot.py
AlexAndorra: - You're right about the title, will do.
- Indeed, it `filter` is not necessary to negate, it was just to show an example of how to use it, to vary examples. I will use "thet" instead of the full variable name, to make `filter` necessary.

arviz/utils.py
AlexAndorra: I'm not sure I understand -- can you expand a little please?

arviz/plots/autocorrplot.py
ahartikainen: Yes

arviz/utils.py
ahartikainen: Actually this can blow-up too (infinite iteration). I will tell you when on computer.

(Never edit list while iterating over it)

arviz/tests/base_tests/test_utils.py
AlexAndorra: I opted for `var_args` -- `args` seemed too generic.

arviz/utils.py
AlexAndorra: Yes, you're right, in that case your example would not happen because `theta` is in `all_vars`. However, in this case, as the loop is not executed, only `theta` will be excluded and `theta_t` will still be there.
So the solution to you is to remove the `if pattern not in all_vars` condition, right?

On your second point, I'll indeed move the `.remove` -- or just get rid of it entirely (see Ari's comment above)

AlexAndorra:
- Addressed all comments
- Closed issue #1163 cc @canyon289

AlexAndorra:
Renamed the `filter` arg to `filter_vars`, both to keep pylint happy and to have a more explicit API. Should be ready to merge now 🎉

PR #1156, opened 2020-04-20
Fix documentation and deprecation warning in pair plot

rpgoldman:
## Description

The `kind` argument to pairplot was incorrectly described in the docstring.  Fixed to point out that it can be a list of strings.
Also there was a deprecation warning whose logic was wrong -- a user could be warned about a deprecated argument that they did not use.

## Checklist

Changes too trivial to be worth a Changelong entry.

### DIFF
--- a/arviz/plots/pairplot.py
+++ b/arviz/plots/pairplot.py
@@ -67,6 +67,7 @@ def plot_pair(
         in the x-direction and the y-direction.
     contour : bool, optional, deprecated, Defaults to True.
         If True plot the 2D KDE using contours, otherwise plot a smooth 2D KDE. Defaults to True.
+        **Note:** this default is implemented in the body of the code, not in argument processing.
     fill_last : bool
         If True fill the last contour of the 2D KDE plot. Defaults to True.
     divergences : Boolean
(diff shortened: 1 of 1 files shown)


arviz/plots/pairplot.py
```diff
@@ -56,15 +57,15 @@ def plot_pair(
...
+    kind : str or List[str]
         Type of plot to display (scatter, kde and/or hexbin)
     gridsize : int or (int, int), optional
         Only works for kind=hexbin.
         The number of hexagons in the x-direction. The corresponding number of hexagons in the
         y-direction is chosen such that the hexagons are approximately regular.
         Alternatively, gridsize can be a tuple with two elements specifying the number of hexagons
         in the x-direction and the y-direction.
-    contour : bool
+    contour : bool, optional, deprecated, Defaults to True.
```
AlexAndorra, line 68: You wrote it defaults to `True` here but set the default to `None` in the function definition above

rpgoldman: It actually _does_  default to `True`. But if I let the argument parser handle defaulting, we get the bogus warning. So I have to turn `None` into `True` myself, in code.

AlexAndorra: Yeah I saw that in your code -- just saying that, as is, the docstring contradicts the type hints for the function. Not a big deal I think though, this shouldn't delay merging.

rpgoldman: I just pushed a quick clarification in the docstring. It was easy to make, and it will avoid confusion.  Thanks.

OriolAbril:
LGTM, we can merge once black stops complaining

rpgoldman:
> LGTM, we can merge once black stops complaining

I don't have the privileges, so if my new commit fixes it, please do the merge.  Please squash if that isn't the default.

OriolAbril:
Thanks! Not only squash is the default, it also is the only option, other merges are disabled

Issue #1157, opened 2020-04-20
Best practices with plot_kde


rpgoldman:
## Short Description

I was trying to redo some of the plot_kde plots for the "rat tumor" example of Chapter 5 BDA3, building on the notebooks in [the PyMC3 resources repo](https://github.com/pymc-devs/resources/tree/master/BDA3).  I had a couple of problems, that may be more documentation issues than anything else:

1. It seems that one must label the x and y axes in a separate call -- there doesn't seem to be any way to specify the x and y axis labels in the call to `plot_kde`.  This seems normal enough to want to do that it should be supported.  Perhaps it's supported by one of the keywords arguments, but I couldn't figure out how and the examples in the docs weren't sufficient to figure it out.

2. I tried to use the `contour_kwds` argument to set the number of contours, but that didn't seem to work. Again, an example would have helped.

I had two less specific issues, that may be related:

1. In general I couldn't easily figure out how to use the extra keyword arguments, and I couldn't find examples.

2. The ArviZ plots I did ended up looking significantly uglier than the Seaborn plots that were in the old version of the notebook. I could use some help figuring out how to style the ArviZ plots.

## Code Example or link

From the diagram, here is the seaborn contour plot:

![image](https://user-images.githubusercontent.com/3274/79713372-32110100-8293-11ea-918f-3dacdf72fbc3.png)

and here is the best ArviZ equivalent I could come up with.  It's not very impressive:

![image](https://user-images.githubusercontent.com/3274/79713406-5371ed00-8293-11ea-9f63-58def748b9b6.png)

The seaborn code is as follows:

```
fig, ax = plt.subplots(figsize=(8, 7))
sns.kdeplot(np.log(alpha_trunc /  beta_trunc), np.log(alpha_trunc + beta_trunc),
            cmap=plt.cm.viridis, n_levels=11, ax=ax)
# ax.set_xlim(1.5, 2.1)
ax.set_ylim(1.7, 3.75)
ax.set_xlabel('log(alpha / beta)')
ax.set_ylabel('log(alpha + beta)');
```

and here's what I came up with as an ArviZ alternative:

```
fig, ax = plt.subplots(figsize=(8, 7))

az.plot_kde(values2=idata_10.posterior['log(alpha + beta)'], values=idata_10.posterior['log(alpha / beta)'], ax=ax,
                contour=True, 
                contour_kwargs={'levels': 10,}) # Setting different levels here does not have any effect.
    
[...]

PR #1158, opened 2020-04-20
add examples to customize 2D KDE

aloctavodia:
Following #1157 this add examples to the docstring of `plot_kde` and a new example to the gallery. 
This also sets `fill_last=False` as the default, because nobody likes the previous setting :-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Add out-of-sample groups (`predictions` and `predictions_constant_data`) and `constant_data` group to pyro and numpyro translation (#1090, #1125)
 * Add `num_chains` and `pred_dims` arguments to from_pyro and from_numpyro (#1090, #1125)
 * Integrate jointplot into pairplot, add point-estimate and overlay of plot kinds (#1079)
+
 ### Maintenance and fixes
 * Fixed behaviour of `credible_interval=None` in `plot_posterior` (#1115)
 * Fixed hist kind of `plot_dist` with multidimensional input (#1115)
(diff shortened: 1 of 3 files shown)


OriolAbril:
LGTM, and I completely agree with the `fill_last=False` default :tada:

rpgoldman:
See my feature request #1159 -- we should show usage of the `cmap` property to control the color scheme.  Understanding the effect of `cmap` turns out to be critical to getting attractive plots.

aloctavodia:
Two of the new examples show how to pass a `cmap` argument.

OriolAbril:
> Maybe this could be fixed by choosing a good default colormap for the users, so that at least they don't have to understand, instead of leaving it unpredictably to the state of the environment?

We should probably extend ArviZ native style like `arviz-darkgrid` and so one to be more explicit on the defaults (now they basically are designed to be called on top of matplotlib default and only list the parameters to override), however, for most arguments (at least in matplotlib, which is what I know better) we **want** the output to depend on matplotlib rcparams. As I see the situation, there are several possibilities, and the least bad is the current setting:

* We do not set the default in ArviZ to use matplotlib defaults. This has the advantage of letting users do `plt.rcparams["cmap"] = "plasma"` and have plasma be the default in native matplotlib, seaborn, ArviZ... (basically all other libraries on top of matplotlib that do not override the default). The main drawback is dependence on previous state as you said, but extending ArviZ styles should fix this.
* We use an ArviZ rcparam. It is similar to the previous situation, with the main difference that matplotlib defaults no longer affect ArviZ both for the good and for the bad, getting all plots to use the same cmap would require 2 commands (if we use both matplotlib and ArviZ). The main con I see is that if all libraries on top of matplotlib do that, defaults will escalate quickly.
* We set the default in the plotting functions. This hardcodes the default which would then be impossible to override at a general level.

 Would the first situation (which is basically the current one) work for your use case if styles were to guarantee reproducibility of results? I hope outlining the alternatives helps in understanding why plots behave like they do.

aloctavodia:
Using styles you can also do `plt.rcparams["cmap"] = "plasma"` if you want, styles overwrite the default rcParams and can be overwritten to. Setting values by default in the arviz style do not prevent they to be overwritten, so is not a general solution to the issue Robert is seeing, which is just the combination of two styles.

aloctavodia:
Thanks @rpgoldman for bringing up the issue.

PR #1164, opened 2020-04-22
New grayscale style

aloctavodia:
Adds new grayscale style. Originally intended for the BMCP book but probably useful for a more general use. It is based on the arviz-white style, but the color cycle has only five colors the first four are grays and the fifth is the ArviZ's "C0 blue", the cmap is matplotlib's gray cmap  

A few examples from the gallery (some of with their default colors changed)
0A


arviz/plots/styles/arviz-grayscale.mplstyle
```diff
@@ -0,0 +1,38 @@
...
+ytick.labelsize: 14
+legend.fontsize: 14
+
+axes.grid: False
+axes.facecolor: white
+axes.edgecolor: 0
+axes.linewidth: 1
+axes.spines.top: False
+axes.spines.right: False
+image.cmap: gray
```
ahartikainen, line 32: what cmap is this? The mpl default or the one in colorcet (etc+)

aloctavodia: The default from matplotlib, as I understand this one is perceptually uniform. But I could check again.

ahartikainen: It is not.

aloctavodia: :( OK, I will fix it

aloctavodia:
updated cmap figures






I was unsure if we should use the linear_grey_10_95_c0 or linear_grey_0_100_c0 scales. The first one does not include the extreme values, i.e the white and black. @ahartikainen What do you think?

canyon289:
This is nice!

canyon289:
> updated cmap figures
> 
> 
> 
> 
> 
> I was unsure if we should use the linear_grey_10_95_c0 or linear_grey_0_100_c0 scales. The first one does not include the extreme values, i.e the white and black. @ahartikainen What do you think?

If this is meant for printed book format should we ask publishers which plot will look better in print? My assumption for book is if readers decide to reproduce on computer the limitation of grayscale wont apply anymore and we tell them to use full colorscale, so this styles main use case is printed material

ahartikainen:
Also, for printing, we can force colours to be a bit lighter

aloctavodia:
This is more general than the book. Many journals use gray-scale figures or charge for color ones. So having an easy to use grey-scale could be useful for many users. Notice this adds the `cet_grey_r` and cet_grey_r` cmaps.

01_b

PR #1167, opened 2020-04-28
[WIP] Fix point estimate marker kwargs

agustinaarroyuelo:
## Description

This PR addresses issue #1166. I also reduced marginal plot size in joint plots and changed the argument name `diagonal` for `marginals`.
jointplot_kde

## Checklist

- [x] Follows [official](https://github.com/arviz-devs/arviz/blob/master/CONTRIBUTING.md#pull-request-checklist) PR format
- [x] Includes a sample plot to visually illustrate the changes (only for plot-related functions)
- [x] New features are properly documented (with an example if appropriate)?
- [ ] Changes are listed in [changelog](https://github.com/arviz-devs/arviz/blob/master/CHANGELOG.md#v0xx-unreleased)

PR #1168, opened 2020-04-29
Line endings

ahartikainen:
This is a test

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,4 @@
 /CHANGELOG.md merge=union
+
+# Set the default behavior, in case people don't have core.autocrlf set.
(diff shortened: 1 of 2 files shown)


ahartikainen:
pylint and astroid are failing... I don't know. Is it because we import `__version__` from `__init__` from some files?

AlexAndorra:
Yeah I noticed a weird error message in pylint yesterday in [Agustina's PR](https://github.com/arviz-devs/arviz/pull/1167). I think it comes from what @OriolAbril said: "pylint is broken due to a recent release, we should pin the previous version until solved".

ahartikainen:
I will merge this

Issue #998, opened 2020-01-14
Question: Massage InferenceData for multi-variable model comparison?


rpgoldman:
## Short Description

I have a PyMC3 model that is partitioned into 6 sub-models.  The observations are also partitioned into 6 subsets. This lets me apply different parameters based on the values of independent variables without complex indexing. I can't use mixing because there is no distribution over the independent variables.

I know model comparison only works for models with a single observed RV. But I was wondering: is there some way to "unpartition" the `InferenceData` so that Arviz can treat the six vectors of observations as one big vector, and compute model comparison metrics?

I can compare the submodels individually, but this does not properly take into account the hyperparameters that link them.

OriolAbril:
I think it is, but for now it will be convoluted and unnecessarily much more complicated than it needs to be. 

The first issue is getting the log_likelihood data from the PyMC3 model, for this you have to use the code in #794 (I can probably rebase and upload in some hours) or write/copy some function to get the log likelihood. 

The second issue is ArviZ not accepting multiple log likelihood instances, so the fastest way to get it done is probably to group all the _log likelihoods_ into a single `"log_likelihood"` variable, something similar to the following pseudocode (assuming code from #794):

```
idata.sampler_stats["log_likelihood"] = xr.concat(
    [idata.log_likelihoods[var_name] for var_name in var_names]
)
az.waic/loo(idata)
```

Note: I would start trying if it works with `xr.concat`, but maybe it is some [other function](xarray.pydata.org/en/stable/combining.html)

ahartikainen:
Do you want to sum them together or compare invidually.

If you can extract each cell (use `arviz.utils.flat_inference_data_to_dict`) and then either create a new `InferenceData` with `from_dict` (log_likelihood as a vector) or create a new `InferenceData` for each variable.

Then compare pair(s).

rpgoldman:
@ahartikainen I would like to compare the models as a whole (rather than comparing the submodels), so I think I want to do what @OriolAbril suggests: effectively concatenate all the observed random variables into one big random variable.  This is reasonable, because the observations are interchangeable by the model.

rpgoldman:
> I think it is, but for now it will be convoluted and unnecessarily much more complicated than it needs to be.
> 
> The first issue is getting the log_likelihood data from the PyMC3 model, for this you have to use the code in #794 (I can probably rebase and upload in some hours) or write/copy some function to get the log likelihood.
> 
[...]

Issue #987, opened 2020-01-06
WAIC/LOO for models with multiple observed variables


VincentBt:
(Related issue: #794 - cc @rpgoldman)

First, thank you for this great library!

I was told by @ahartikainen [here](https://github.com/arviz-devs/arviz/commit/a45e901d9f4076da40e63213eb9c28bdc671e10d#commitcomment-36660404) that multiple observations were not supported in arviz (at least for numpyro and pyro, but I guess for other libraries like PyMC3 as well).

My question is: to go from 1 observation to multiple observations <a href="https://www.codecogs.com/eqnedit.php?latex=\inline&space;$$y^1,&space;\dots,&space;y^m$$" target="_blank"><img src="https://latex.codecogs.com/gif.latex?\inline&space;$$y^1,&space;\dots,&space;y^m$$" title="$$y^1, \dots, y^m$$" /></a>, isn't it enough to sum all the log_likelihoods of observed variables to compute the waic and loo?

Indeed, the WAIC and LOO are computed based on <a href="https://www.codecogs.com/eqnedit.php?latex=\inline&space;$$p(y_i&space;|&space;\theta_s)$$" target="_blank"><img src="https://latex.codecogs.com/gif.latex?\inline&space;$$p(y_i&space;|&space;\theta_s)$$" title="$$p(y_i | \theta_s)$$" /></a> where <a href="https://www.codecogs.com/eqnedit.php?latex=\inline&space;$$y_i$$" target="_blank"><img src="https://latex.codecogs.com/gif.latex?\inline&space;$$y_i$$" title="$$y_i$$" /></a> are all the examples (observations) and <a href="https://www.codecogs.com/eqnedit.php?latex=\inline&space;$$\theta_s$$" target="_blank"><img src="https://latex.codecogs.com/gif.latex?\inline&space;$$\theta_s$$" title="$$\theta_s$$" /></a> are all the samples from the posterior of all parameters and hidden variables.
[...]

PR #1173, opened 2020-05-04
[proposal] support for IC with multiple variables

OriolAbril:
## Description
Given the number of possibilities and the even larger number of things that can go wrong, I propose to add only a `var_name` argument (or `name`/`log_likelihood` to avoid confusion with `var_names` that can be a list) and force users to manually add the combined log likelihood data on which they want to calculate loo/waic to `log_likelihood` group. 

We could add some `var_names` arg instead and a `combine` arg to sum or concat or maybe even try to cover groupby too, but this will always be more limited that creating the new variable and it will also be more obscure (i.e. two variables with the default coord names `var1_dim_0` and `var2_dim_0` will be broadcasted to a 2d array when added because their coord names are different, if the user has to manually create the dataarray of pointwise log likelihood, this can be seen before passing it to loo/waic). I find that both xarray and loo/waic are quite obscure and commonly misunderstood subjects to support this choice. The overhead to users I anticipate is checking examples to copy the combine code most similar to their situation which I actually see like an advantage as it will force them to think about which of the examples better fits their particular case.

closes (maybe) #998 and #987. By maybe I mean that everything will be _possible_ but we will still be missing detailed examples and guidance on how to do it. 

## Checklist

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -1328,7 +1328,7 @@ def waic(data, pointwise=None, var_name=None, scale=None):
         ``stats.ic_pointwise`` rcParam.
     var_name : str, optional
         The name of the variable in log_likelihood groups storing the pointwise log
-        likelihood data to use for loo computation.
+        likelihood data to use for waic computation.
     scale: str
         Output scale for WAIC. Available options are:
 
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -1322,7 +1322,7 @@ def waic(data, pointwise=None, var_name=None, scale=None):
     ----------
     data: obj
         Any object that can be converted to an az.InferenceData object. Refer to documentation of
-        az.convert_to_inference_data for details
+        ``az.convert_to_inference_data`` for details
     pointwise: bool
         If True the pointwise predictive accuracy will be returned. Defaults to ``stats.ic_pointwise``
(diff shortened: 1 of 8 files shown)


canyon289:
At a high level I think making things more explicit is good. That being said this is at a depth of diagnostics I dont quite understand. I'll check out this branch locally and give it a try

OriolAbril:
I think this is ready to merge, it would be great to include in the upcoming release

arviz/stats/stats.py
```diff
@@ -1278,9 +1283,13 @@ def waic(data, pointwise=False, scale=None):
...
         Any object that can be converted to an az.InferenceData object. Refer to documentation of
-        az.convert_to_inference_data for details
+        ``az.convert_to_inference_data`` for details
     pointwise: bool
-        if True the pointwise predictive accuracy will be returned. Defaults to False
+        If True the pointwise predictive accuracy will be returned. Defaults to
+        ``stats.ic_pointwise`` rcParam.
+    var_name : str, optional
+        The name of the variable in log_likelihood groups storing the pointwise log
+        likelihood data to use for loo computation.
```
AlexAndorra, line 1292: "... for ~~loo~~ waic computation ..."

canyon289:
Looks great. Merging

PR #1176, opened 2020-05-08
Replace credible_interval with hpd_interval

canyon289:
## Description
Per discussion in slack rename `credible_interval` to `hpd_interval`. Opening up for feedback before completing everything else in checklist/ Also change `auto` on posteriorplot to None to match api with other plotting functions

## Checklist


OriolAbril:
In posterior plot the default is set to auto because None is a valid value meaning do not plot hpd line. The behaviour would have to be modified in order to change the default

arviz/stats/stats.py
```diff
@@ -986,8 +986,8 @@ def summary(
     extend: boolean
         If True, use the statistics returned by ``stat_funcs`` in addition to, rather than in place
         of, the default statistics. This is only meaningful when ``stat_funcs`` is not None.
-    credible_interval: float, optional
-        Credible interval to plot. Defaults to 0.94. This is only meaningful when ``stat_funcs`` is
+    hpd_interval: float, optional
+        hpd interval to plot. Defaults to 0.94. This is only meaningful when ``stat_funcs`` is
```
AlexAndorra, line 990: Maybe you can fix here the issue you opened for `summary` docs by replacing "hpd interval to plot" by something like "hpd interval to compute and report"?

canyon289:
> In posterior plot the default is set to auto because None is a valid value meaning do not plot hpd line. The behaviour would have to be modified in order to change the default

I did this on purpose to match `forest_plot` api where `None` gets the default value for params. Maybe instead we change `forest_plot` default to `auto` and if `None` is specified we throw an exception?  Trying to unify the api here

https://arviz-devs.github.io/arviz/generated/arviz.plot_forest.html#arviz.plot_forest

canyon289:
> LGTM, thanks Ravin!
> Just had one comment, and Oriol's caveat above seems relevant.
> 
> Also, what's the deprecating plan here? Do we raise the warning for the 0.7.* series and deprecate in 0.8.0?

I think this is the right approach. Its how pandas does things

arviz/stats/stats.py
canyon289: Done Thanks!

arviz/utils.py
```diff
@@ -601,3 +601,28 @@ def get_coords(data, coords):
...
+    """Helper method to warn that credible interval will be deprecated"""
+
+    warnings.warn(
+        (
+            "Keyword argument credible_interval has been deprecated "
+            "Please replace with hpd_interval"
+        ),
+    )
+
+    if isinstance(credible_interval, str) and credible_interval == "auto":
```
canyon289, line 616: @OriolAbril Does this address your concerns?

canyon289: I'm also going to change this to an exception because it raises an exception further down and the warning is up top.

image

canyon289:
Leaving this open for 48 hours and if no comments will merge

arviz/plots/posteriorplot.py
```diff
@@ -62,8 +63,11 @@ def plot_posterior(
     textsize: float
         Text size scaling factor for labels, titles and lines. If None it will be autoscaled based
         on figsize.
-    credible_interval: float, optional
-        Credible intervals. Defaults to 0.94. Use None to hide the credible interval
+    hpd_interval: float, optional
+        Plots highest posterior density interval for chosen percentage of density.
+        Use 'hide' to hide the credible interval. Defaults to 0.94.
+    hpd_interval: float or str, optional
+        Credible intervals. Defaults to 0.94. Use 'hide' to hide the hpd interval
```
OriolAbril, line 70: docs got duplicated here

OriolAbril: There are also other docstring where description of hpd_interval says "Credible interval...".

arviz/rcparams.py
```diff
@@ -240,7 +240,7 @@ def validate_iterable(value):
     ),
     "plot.matplotlib.constrained_layout": (True, _validate_boolean),
     "plot.matplotlib.show": (False, _validate_boolean),
-    "stats.credible_interval": (0.94, _validate_probability),
+    "stats.hpd_interval": (0.94, _validate_probability),
```
OriolAbril, line 243: I don't really know if anybody else apart from me is using rcparams, so I am not really opposed to just changing them right away, however, if possible it would be a good idea to have a warning if an rcfile has `stats.credible_interval` telling them to update. I could add some `deprecated` list for rcparams or something similar.

arviz/utils.py
OriolAbril: It is the other way around, what used to happen was:
* `credible_interval="auto"` -> use rcParams
* `credible_interval=None` -> hide hpd line (only implemented in plot_posterior because everywhere else (like hpd or plot_forest) it did not really make sense to hide 
* otherwise use value, has to be float between 0 and 1 ...

arviz/rcparams.py
canyon289: Oh this is a good point. I will add

Also I use them btw! They are great

OriolAbril: I base my assumption on the fact that no issue about bad rcparams behaviour has been opened and I find it hard to believe there are actually no issues at all :thinking:

OriolAbril:
There are still some credible intervals present, we should probably do a grep ignorecase allowing space and _ between them. Example: forestplot uses the var name internally which is clearly no big deal (i.e. [L348](https://github.com/canyon289/arviz/blob/hpd_interval/arviz/plots/backends/bokeh/forestplot.py#L348)) but also labels as credible interval ([L420](https://github.com/canyon289/arviz/blob/hpd_interval/arviz/plots/backends/bokeh/forestplot.py#L420)). Example is bokeh, matplotlib probably does the same

canyon289:
> There are still some credible intervals present, we should probably do a grep ignorecase allowing space and _ between them. Example: forestplot uses the var name internally which is clearly no big deal (i.e. [L348](https://github.com/canyon289/arviz/blob/hpd_interval/arviz/plots/backends/bokeh/forestplot.py#L348)) but also labels as credible interval ([L420](https://github.com/canyon289/arviz/blob/hpd_interval/arviz/plots/backends/bokeh/forestplot.py#L420)). Example is bokeh, matplotlib probably does the same

Ahh there were so many. I was messing up my search command. Thanks you @OriolAbril

canyon289:
@OriolAbril For this one its inconsistent between methods. That being said I'm happy to help standardize

What do you think about 

    hpd_interval=None -> use rcParams
    hpd_interval='hide' -> hide hpd line (only implemented in plot_posterior because everywhere else (like hpd or plot_forest) it did not really make sense to hide
    otherwise use value, has to be float between 0 and 1 ...


image

OriolAbril:
None to use rcParams, "hide" to not plot line amd value otherwise is great, my comment was more on the error message, "hide" is not replacing "auto"

canyon289:
Ah got it. Thats ok I think as there is that exception that makes it obvious? I decided not to replace it because then downstream a bunch of stuff would have to be changed. I figure I can get away with this as I don't think a lot of people are using 'auto' explicitly

arviz/stats/stats.py
```diff
@@ -1117,7 +1123,7 @@ def summary(
         hpd_lower, hpd_higher = xr.apply_ufunc(
             _make_ufunc(hpd, n_output=2),
```
OriolAbril, line 1124: I think this is no longer necessary, `hpd` uses `apply_ufunc` and should work straight away, I am actually quite amazed this still works. I'll post an issue

arviz/stats/stats.py
```diff
@@ -1155,7 +1161,7 @@ def summary(
         circ_hpd_lower, circ_hpd_higher = xr.apply_ufunc(
```
OriolAbril, line 1161: same as above

arviz/utils.py
```diff
@@ -601,3 +601,25 @@ def get_coords(data, coords):
...
+    """Replace credible_interval with hpd_interval and to warns of be deprecation."""
+    warnings.warn(
+        (
+            "Keyword argument credible_interval has been deprecated "
+            "Please replace with hpd_interval"
+        ),
+    )
+
+    if isinstance(credible_interval, str) and credible_interval == "auto":
+        raise Exception("Argument value 'auto' has been renamed to 'hide'")
```
OriolAbril, line 616: the exception should be different. people who was using `"auto"` (probably nobody) now have to use `None`, its people using `None` whose plots will change and should modify the code to use `"hide"`. I don't see any way to warn people who used to pass `None` other than release notes. Maybe in the deprecation warning? it is only relevant for plot_posterior though.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 * plot_ppc animation: improve docs and error handling (#1162)
 
 ### Deprecation
+* `credible_interval` argument replaced by `hpd_interval`throughout with exception of `plot_loo_pit`
```
OriolAbril, line 38: there should be also a note on the difference in behaviour of `plot_posterior` so users know they have to change `None` to `"hide"`

arviz/utils.py
canyon289: Ohhhhhhhhhhhhhhhhhh that makes sense. I 100% agree with you. I will add a warning

canyon289: Ok now thinking about it I'm stuck. because if I check for `None` then a warning will be raised by default since that's now the default arg, which I think will happen for most users. but if I dont then the users who were specifying `None ` explicitly will get confused, but I think that's a smaller subset.

Do you think the docs will help if people get confused without needing to make it as clear during runtime?

OriolAbril: True, only alternative that comes to mind is setting the default of `credible_interval` to `"deprecated"` or some other new and unrelated value to clearly distinguish when are users actually passing a value to it, not sure if it is a good idea or if it will be even more confusing though.

aloctavodia:
> I have also one concern about highest _posterior_ density, I like hpd because one cal always extend it depending on the ocasion to posterior, prior... I have encountered people very confused to me doing prior checks and using HPD to make sure the values of some parameter lied inside a sensible region. I'd advocate for using HPD everywhere and maybe extend the acronym only once and explain that it does not need to be applied to posterior necessarily.

You have a good point, what about using the more general term Highest Density Interval (HDI)?

canyon289:
> > I have also one concern about highest _posterior_ density, I like hpd because one cal always extend it depending on the ocasion to posterior, prior... I have encountered people very confused to me doing prior checks and using HPD to make sure the values of some parameter lied inside a sensible region. I'd advocate for using HPD everywhere and maybe extend the acronym only once and explain that it does not need to be applied to posterior necessarily.
> 
> You have a good point, what about using the more general term Highest Density Interval (HDI)?

Oh boy. I hate that you're *right* after I made all the code changes!!!! What should we do, rename everything to HDI, leave HPD, something different?

I can go through and rename everything again, but hopefully only one more time :D

aloctavodia:
Sorry for being late, haha. I think we should rename every instance of hpd to hdi and instead of credible_interval use something like hdi_prob or hdi_probability, or hdip

AlexAndorra:
Ha ha sorry to pile on @aloctavodia's and @OriolAbril's answers @canyon289, but I think HDI is indeed more appropriate. Replacing `credible_interval` kwarg by `hdi_prob` looks very good to me 👌 
However, let's make sure we all agree _before_ you make the changes and only have to do that one time.

canyon289:
Releasing this is going to break pymc3. We have to coordinate changes on master
image

AlexAndorra:
Ah indeed, it's like the warmup PR... This means PyMC has to pin the latest ArviZ, right? This will already be that casr anyway, to enable the new `return_inference_data` kwarg on `sample`.
How can we make the tests pass in the meantime though?

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * Integrate jointplot into pairplot, add point-estimate and overlay of plot kinds (#1079)
 * New grayscale style. This also add two new cmaps `cet_grey_r` and `cet_grey_r`. These are perceptually uniform gray scale cmaps from colorcet (linear_grey_10_95_c0) (#1164)
 * Add warmup groups to InferenceData objects, initial support for PyStan (#1126) and PyMC3 (#1171)
+* `hdi_prob` will not plot hpd if argument `hide` is passed. Previously `credible_interval` would omit HPD if `None` was passed  (#1176)
```
ahartikainen, line 21: What do we do for the rest of the hpd mentions?

canyon289: I need to keep going through and fixing it. I got tired yesterday but will start again today

canyon289:
Alright so I didnt change `plot_hpd` and I left hpd verbiage the same in the Loo diagnostic plots because in that diagnostic it is indeed the hpd. There were a lot of references to HPD and credible interval so I hope I got them all.

If this looks good are you guys ok merging and then following up any ones missed in prs?

AlexAndorra:
> If this looks good are you guys ok merging and then following up any ones missed in prs?

I agree -- it's hard to reach the target with only one arrow on this kind of PRs 🎯 
Thanks a lot for your work on this 👍

arviz/plots/hpdplot.py
```diff
@@ -3,15 +3,16 @@
...
 from scipy.signal import savgol_filter
 
-from ..stats import hpd
+from ..stats import hdi
 from .plot_utils import get_plotting_function, matplotlib_kwarg_dealiaser
 from ..rcparams import rcParams
+from ..utils import credible_interval_warning
 
 
 def plot_hpd(
```
aloctavodia, line 12: I think we should also change this to `plot_hpi` (with a deprecation warning)

canyon289: Added issue. Will merge this then go for that one

PR #1185, opened 2020-05-15
fix ridgeplot failure for discrete variables

aloctavodia:
Just a quick fix.

### DIFF
--- a/arviz/plots/backends/matplotlib/forestplot.py
+++ b/arviz/plots/backends/matplotlib/forestplot.py
@@ -250,9 +250,12 @@ class PlotHandler:
             Width of line on border of ridges
         alpha : float
             Transparency of ridges
-        kind : string
+        ridgeplot_kind : string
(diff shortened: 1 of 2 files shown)


aloctavodia:
@agustinaarroyuelo we may want to revisit this in the future to support quantiles for discrete variables.

canyon289:
Ideally test would be added

aloctavodia:
I was thinking adding tests for discrete variables would be a good fit to @agustinaarroyuelo's GSOC project ;-)

canyon289:
Ok lets keep this open until tests are added. @agustinaarroyuelo if you have questions on how to add tests let me know!

aloctavodia:
This should be merged before the next release, because ridgeplot fails for discrete variables. I will add the test, then we can discuss with Agustina how to improve or increase coverage for discrete and circular variables.

canyon289:
LGTM. Opened an issue for remaining tests https://github.com/arviz-devs/arviz/issues/1191

PR #1189, opened 2020-05-18
Improve scripts readme

aloctavodia:
Fix typo and add explanation for container.ps1


canyon289:
Ah crap I had fixed this locally but not pushed. Im sorry

PR #1190, opened 2020-05-18
Keep replacing hpd with hdi

### LINKED ISSUE #1186: Change `plot_hpd` to `plot_hdi` (with a deprecation warning)
I think we should also change this to `plot_hdi` (with a deprecation warning). It is OK with me if we do it in another PR, but before the next release.

_Originally posted by @aloctavodia in https://github.com/arviz-devs/arviz/pull/1176_

aloctavodia:
Fix #1186


canyon289:
You are too kind @aloctavodia. An `hpd` function needs to exist or pymc3 import breaks tests. Likely need one anyway with a deprecation warning

arviz/plots/loopitplot.py
```diff
@@ -210,14 +210,14 @@ def plot_loo_pit(
...
         x_vals = np.linspace(0, 1, len(loo_pit_kde))
-        if use_hpd:
-            if hpd_kwargs is None:
-                hpd_kwargs = {}
-            hpd_kwargs.setdefault("color", to_hex(hsv_to_rgb(light_color)))
-            hpd_fill_kwargs = hpd_kwargs.pop("fill_kwargs", {})
-            hpd_fill_kwargs.setdefault("label", "Uniform HPD")
-            hpd_kwargs["fill_kwargs"] = hpd_fill_kwargs
-            hpd_kwargs["credible_interval"] = credible_interval
+        if use_hdi:
```
canyon289, line 213: for loo pit I left it hpd because this particular case it is hpd. I don't think loo pit makes sense for anything bu the posterior. Not opposed to changing it but wanted to get your thoughts

arviz/plots/posteriorplot.py
```diff
@@ -184,7 +184,7 @@ def plot_posterior(
 
         >>> az.plot_posterior(data, var_names=['mu'], kind='hist')
 
-    Change size of HPD interval
+    Change size of highest density interval
```
canyon289, line 187: Thanks for catching my misses

arviz/plots/loopitplot.py
aloctavodia: that's true, but having hpd and hdi seems to me like a source of potential confusion for users or even ourselves in the future

arviz/plots/posteriorplot.py
aloctavodia: I did not remove any function (at least not on purpose, haha) I just marked `plot_hpd` as deprecated in favor of `plot_hdi`. And I replaced the strings `hpd` to `hdi`.

canyon289: Ah yes you're right I see it now

arviz/plots/hdiplot.py
```diff
@@ -131,6 +133,11 @@ def plot_hpd(
...
     # TODO: Add backend kwargs
-    plot = get_plotting_function("plot_hpd", "hpdplot", backend)
-    ax = plot(**hpdplot_kwargs)
+    plot = get_plotting_function("plot_hdi", "hdiplot", backend)
+    ax = plot(**hdiplot_kwargs)
     return ax
+
+
+def plot_hpd(*args, **kwargs):
+    warnings.warn("plot_hdi has been deprecated, please use plot_hdi", DeprecationWarning)
```
canyon289, line 142: Pydocstyle complaining about this one method. Maybe add disable or deprecation comment to get a pass

canyon289:
LGTM merging

PR #1201, opened 2020-05-22
Labeling restructure

### LINKED ISSUE #984: integrate index_origin rcParam
## Tell us about it
In some cases, users may prefer to visualize the results using 1 indexed values instead of python's 0 indexed ones. Part of #792.

## Thoughts on implementation

The `summary` function already has an `index_origin` argument, but does not use the rcParam value as default. Moreover, the labels used for plots should also be updated to allow 1 indexed visualization (they all call `make_label` and [`selection_to_string`](https://github.com/arviz-devs/arviz/blob/f4f68d98e4a9a16470b66992f7428ce9b3cfb549/arviz/plots/plot_utils.py#L275)). 

#981 is also related to this.

### LINKED ISSUE #1091: Use coordinate names in posterior summary()
Plotting functions such as `plot_posterior()` and `plot_trace()` identify coordinates in variables by their coordinate names. But `summary()` does not. Using coordinate names in the summary() method would result in clearer output.

### Code to demonstrate:
```
import pystan
import arviz
import numpy

print(arviz.__version__)

code = """
data {
    int<lower=1> N;
    int<lower=1> P;
    vector[N] y;
    matrix[N,P] X;
}
parameters {
    vector[P] beta;
    real alpha;
    real<lower=0> sigma;
}
model {
    beta ~ std_normal();
    alpha ~ std_normal();
    sigma ~ cauchy(0, 1) T[0,];
    y ~ normal_id_glm(X, alpha, beta, sigma);
}
"""

# Compile model
model = pystan.StanModel(model_code=code)

#### Data
# Dimensions
P = 4
N = 200
# Parameters
alpha = 1
beta = numpy.array(range(P)) - (P - 1)/2
sigma = 2
# Data
rg = numpy.random.default_rng()
X = rg.normal(size=(N,P))
y = numpy.matmul(X, beta) + alpha + rg.normal(scale=sigma, size=N)

# Sample
fit = model.sampling(
    data={'N':N, 'P':P, 'X':X, 'y':y},
    init='random', n_jobs=1, 
    pars=['alpha','beta','sigma'])

samples = arviz.from_pystan(fit,
                            observed_data='y',
                            constant_data='X',
                            coords={'observation':list(range(200)), 
                                    'covariate':['A','B','C','D']}, 
                            dims={'y':['observation'], 
                                  'beta':['covariate'], 
                                  'X':['observation','covariate']})

arviz.plot_posterior(samples)
arviz.plot_trace(samples)
print(arviz.summary(samples))
```

### Output
`0.6.1`
Figure_1
Figure_2
```
          mean     sd  hpd_3%  hpd_97%  mcse_mean  mcse_sd  ess_mean  ess_sd  ess_bulk  ess_tail  r_hat
alpha    1.026  0.144   0.752    1.301      0.002    0.001    5294.0  5294.0    5258.0    2582.0    1.0
beta[0] -1.504  0.148  -1.800   -1.239      0.002    0.001    5550.0  5293.0    5570.0    3159.0    1.0
beta[1] -0.531  0.153  -0.825   -0.252      0.002    0.002    5597.0  5040.0    5582.0    3062.0    1.0
beta[2]  0.396  0.140   0.143    0.671      0.002    0.001    5700.0  5498.0    5682.0    3168.0    1.0
beta[3]  1.086  0.155   0.801    1.390      0.002    0.001    5714.0  5514.0    5717.0    3156.0    1.0
sigma    2.073  0.104   1.895    2.287      0.001    0.001    5101.0  5024.0    5168.0    3039.0    1.0
```

### Desired output
(plots unchanged)
```
          mean     sd  hpd_3%  hpd_97%  mcse_mean  mcse_sd  ess_mean  ess_sd  ess_bulk  ess_tail  r_hat
alpha    1.026  0.144   0.752    1.301      0.002    0.001    5294.0  5294.0    5258.0    2582.0    1.0
beta[A] -1.504  0.148  -1.800   -1.239      0.002    0.001    5550.0  5293.0    5570.0    3159.0    1.0
beta[B] -0.531  0.153  -0.825   -0.252      0.002    0.002    5597.0  5040.0    5582.0    3062.0    1.0
beta[C]  0.396  0.140   0.143    0.671      0.002    0.001    5700.0  5498.0    5682.0    3168.0    1.0
beta[D]  1.086  0.155   0.801    1.390      0.002    0.001    5714.0  5514.0    5717.0    3156.0    1.0
sigma    2.073  0.104   1.895    2.287      0.001    0.001    5101.0  5024.0    5168.0    3039.0    1.0
```
Alternatively, a pandas multiindex could be used:
```
            mean     sd  hpd_3%  hpd_97%  mcse_mean  mcse_sd  ess_mean  ess_sd  ess_bulk  ess_tail  r_hat
alpha      1.026  0.144   0.752    1.301      0.002    0.001    5294.0  5294.0    5258.0    2582.0    1.0
beta    A -1.504  0.148  -1.800   -1.239      0.002    0.001    5550.0  5293.0    5570.0    3159.0    1.0
        B -0.531  0.153  -0.825   -0.252      0.002    0.002    5597.0  5040.0    5582.0    3062.0    1.0
        C  0.396  0.140   0.143    0.671      0.002    0.001    5700.0  5498.0    5682.0    3168.0    1.0
        D  1.086  0.155   0.801    1.390      0.002    0.001    5714.0  5514.0    5717.0    3156.0    1.0
sigma      2.073  0.104   1.895    2.287      0.001    0.001    5101.0  5024.0    5168.0    3039.0    1.0
```

### LINKED ISSUE #1499: arviz.summary() indexes parameters from 0, instead of taking the index described in coordinates
**Describe the bug**
Stan indexes from 1. When calling arviz.summary(), the parameters are indexed from 0, regardless of the coordinates specified in the arviz.from_pystan() call.

**To Reproduce**
```
X = np.random.rand(100,9)
y = (X*0.1).sum(axis=1) + np.random.normal(0,1,100)

reg_data = {'X': X,
            'y': y,
            'N': X.shape[0],
            'K': X.shape[1]}
reg_code = """
data {
    int<lower=0> N; // number of samples
    int<lower=0> K; // number of predictors
    vector[N] y; // response variable
    matrix[N, K] X; // predictor matrix
}
parameters {
    real alpha;
    vector[K] beta; // coefficients for predictors
    real<lower=0> sigma;
}
transformed parameters {
    vector[N] mu;
    mu = alpha + X * beta;
}
model {
    y ~ normal(mu, sigma);
}
generated quantities {
    vector[N] log_lik;
    vector[N] y_hat;
    
    for (i in 1:N) {
        log_lik[i] = normal_lpdf(y[i] | mu[i], sigma);
        y_hat[i] = normal_rng(mu[i], sigma);
    }

}
"""
# fit the model
sm = ps.StanModel(model_code=reg_code)
fit = sm.sampling(data=reg_data, iter=5000, chains=4)

# convert pystan to arviz
fit_az = az.from_pystan(posterior=fit,
                      posterior_predictive="y_hat",
                      observed_data=["y"],
                      log_likelihood={"y": "log_lik"},
                      coords={"chain": np.arange(4),
                              "draw": np.arange(2500),
                              "var": np.arange(1,X.shape[1]+1)},
                      dims={'alpha': ["chain","draw"],
                            'beta': ["chain","draw", "var"],
                            'sigma': ["chain","draw"]})

# plot_forest() correctly inherits the labels from the fit_az object
axes = az.plot_forest(
    fit_az,
    kind="ridgeplot",
    var_names=["alpha", "beta", "sigma"],
    combined=True,
    figsize=(9, 4),
    hdi_prob=0.95
)
plt.show()

# but summary just indexes the betas from 0, even though stan calls them beta[1], beta[2]...beta[9]
az.summary(fit_az, var_names=["alpha", "beta", "sigma"])
```

**Expected behavior**
az.summary() should number the parameters 1-9, as is specified in the coords, in the same way that plot_forest does.

**Additional context**
I think this is particular to pystan, because other packages may just index from 0 like python does.

OriolAbril:
## Description
Enforcing and maintaining `index_origin` at label display has proved to be plenty of work and to not give the best of results. I propose to use `index_origin` at [InferenceData creation](https://github.com/arviz-devs/arviz/blob/master/arviz/data/base.py#L103) and then always display coordinate values. I think this will simplify the code significantly without many drawbacks. Moreover, this will make all representations of the data coherent, from xarray html repr to ArviZ plots.

Therefore, the aims of this PR are:
* [x] Use coordinate values in `az.summary`, fixes #1091 and fixes #1499
  * [x] Required restructuring `make_label` and `xarray_var_iter` to prevent cyclic imports

image

* [x] Remove `index_origin` from stats and plotting functions. I basically affects 2 functions.
  * [x] `az.summary` 
  * [ ] `utils.flatten_inference_data_to_dict` and `to_cds` -> do in another PR, probably better to wait until we have sound converters for dict and dataframe? @percygautam @ahartikainen 
* [x] Use index_origin in default [coord generation](https://github.com/arviz-devs/arviz/blob/master/arviz/data/base.py#L103). Closes #984
* [x] Think (maybe implement) ways for users to customize generated labels options that seem interesting to have ~~are `var_name: coord_val`, `var_name [coord_val]`, `var_name [dim: coord_val]` and the like. Maybe a `label_str.format(var_name=var_name, dim=dim, coord_val=coord_val)` is enough?~~ Proposal below

## Proposal: Labeler class
I propose to use a labeler class with a handful of methods to cover the different labeling tasks. The base task is done with 2 functions, 1st apply the dim+coord labeling function looping over all present dims and 2nd combining the result of this function with the variable name. This could be a single `make_label` function, but I think a class will be better for extensibility, moreover, a class can also handle labeling when multiple models are present (i.e. plot_forest) and the different needs between plot labels and summary labels.

## Checklist


ahartikainen:
I'm not sure about this.

I kind of want to use 0-based indexing with pystan, and just have a summary with "correct" index.

Would this force me to use 1-based indexing?

OriolAbril:
Do you only want to see 1 based indexing in summary? or in plots too? 

Under the proposal above the coordinate values would be 1-base indexes, to use 0-based indexing you'd have to use `isel` instead of `sel`. Otherwise, to me it feels very weird to not be able to use the labels shown in summary to index which is what currently happens with index_origin=1.

ahartikainen:
Yeah, I forget isel. Yeah let's go forward with this

OriolAbril:
There are still some `from_xyz` functions to update, but I think the bulk of the PR is done. I think it is particularly relevant for users of `index_origin` which would mostly be @ahartikainen and @sethaxen. I think this will force ArviZ.jl to do a couple updates (I can help) and I do not expect too many things to break but it would be great to double check and more importantly to make sure this approach will help better integrate index_origin argument.

---

Tasks that will be left pending for other PRs are:
* update utils.flatten_inference_data_to_dict and to_cds to use coord labels (if possible)
* add function to customize label format
* testing?
  * Not sure if a couple tests on `generate_dims_coords` with index origin would be enough or if `from_xyz` should also be tested. any preference/idea @canyon289 ?

canyon289:
@OriolAbril Can we create issue tickets for the remaining things and merge? Would hate for you to get stuck with a merge conflict as so many files have changed

OriolAbril:
I could be merged, but this introduces quite some changes in labeling and index_origin usage that affect most of the library. I think it would be better to wait until it has been properly reviewed. 

Most of the changes are reorganizing imports, so rebasing should be automatic or not too difficult, I don't mind a couple rebases until it gets some reviews

arviz/data/base.py
```diff
@@ -127,46 +139,62 @@ def numpy_to_data_array(ary, *, var_name="data", coords=None, dims=None):
...
-    if "draw" not in dims:
+    if "draw" not in dims and "draw" in default_dims:
         dims = ["draw"] + dims
-    if "chain" not in dims:
+    if "chain" not in dims and "chain" in default_dims:
         dims = ["chain"] + dims
 
-    if "chain" not in coords:
+    if "chain" not in coords and "chain" in default_dims:
         coords["chain"] = utils.arange(n_chains)
```
ahartikainen, line 186: chains should still start from 0?

ahartikainen: Or should it also start at index_origin

arviz/sel_utils.py
```diff
@@ -0,0 +1,224 @@
...
+        Whether to position the coordinates' label "below" (default) or "beside"
+        the name of the variable
+
+    Returns
+    -------
+    label
+        A text representation of the label
+    """
+    if selection:
+        sel = selection_to_string(selection)
```
ahartikainen, line 43: just thinking about this

Let's say we have `theta[3]` with coords ["Ferrari", "Fiat", "Ford"]

Could we create labels as (can index origin work with these?)

    "theta[0] Ferrari"
    "theta[1] Fiat"
    "theta[2] Ford"

or 

    "theta[0]\nFerrari"
    "theta[1]\nFiat"
    "theta[2]\nFord"

For this to happen, do we need to update selection (`xarray_sel_iter`?). And will index_origin work when coordinates are explicitly set?

arviz/sel_utils.py
```diff
@@ -0,0 +1,224 @@
...
+        if position == "below":
+            base = "{}\n{}"
```
ahartikainen, line 45: This looks bad with numbers, but good with strings

arviz/sel_utils.py
```diff
@@ -0,0 +1,224 @@
...
+        elif position == "beside":
+            base = "{}[{}]"
```
ahartikainen, line 47: this is correct with numbers, "bad" with strings

arviz/sel_utils.py
```diff
@@ -0,0 +1,224 @@
...
+    for var_name in var_names:
+        if var_name in data:
+            new_dims = _dims(data, var_name, skip_dims)
+            vals = [purge_duplicates(data[var_name][dim].values) for dim in new_dims]
+            dims = _zip_dims(new_dims, vals)
+            if reverse_selections:
+                dims = reversed(dims)
+
+            for selection in dims:
+                yield var_name, selection
```
ahartikainen, line 141: this should somehow return also the indices (and respect the index origin)

arviz/sel_utils.py
```diff
@@ -0,0 +1,224 @@
...
+        data_to_sel = {data.name: data}
+
+    for var_name, selection in xarray_sel_iter(
+        data,
+        var_names=var_names,
+        combined=combined,
+        skip_dims=skip_dims,
+        reverse_selections=reverse_selections,
+    ):
+        yield var_name, selection, data_to_sel[var_name].sel(**selection).values
```
ahartikainen, line 184: selection should be tuple with `iloc` and `loc` selection methods, right?

arviz/stats/stats.py
```diff
@@ -1045,6 +1040,11 @@ def summary(
         Dimensions specification for the variables to be used if the ``fmt`` is ``'xarray'``.
     credible_interval: float, optional
         deprecated: Please see hdi_prob
+    order
+        deprecated: order is now ignored.
```
ahartikainen, line 1044: Why is order removed?

ahartikainen: or is there a way to select order for unwrapped values?

arviz/sel_utils.py
OriolAbril: I want this to be possible and should be if we support custom labeling function, I am not sure what would be a good way though. 

I do think we will have to update xarray_sel_iter or use xarray_sel_iter to create a function targeted to labeling. In addition to the index, I also would like to expose the dimension so that `theta\nbrand: Ferrari`, `theta\nbrand: Fiat` are also possible. It may also be interesting to keep coordinate values simple (maybe abbreviations) and then have some attribute `label` (and `labels` attr in coordinates) so that users could select variable `theta` using `theta` but set its name to `$\theta$` for plots to show the greek letter. This gets crazy quite fast though. Eventually, `var_name`, `dim`/`coord_name`, `coord_value`, `coord_idx` and maybe `var_label`, `dim_label` and `coord_label` would be exposed when creating labels so custom functions can play with them to get the desired labels.

Regarding `coord_idx`, I think it should not include `index_origin` to get the same value that would be used by `isel`, but we should have an example on adding `index_origin` to `coord_idx` inside the labeller function.

arviz/sel_utils.py
OriolAbril: yep, any ideas on how to handle this are very welcome, having these 2 different defaults (one for plots and another for summary) also makes it harder (at least for me) to think about sensible ways to change the labeling function

arviz/sel_utils.py
OriolAbril: I agree on returning indices, I would not enforce index_origin here (the coordinate label would respect index_origin if integer). I personally don't like having indices that do not correspond to the ones used to index/slice and therefore, as I said above, I would use regular indices and eventually (and optionally) add `index_origin`

arviz/sel_utils.py
OriolAbril: that sounds good, how about `yield var_name, selection, iselection, data_to_sel...`?

arviz/stats/stats.py
OriolAbril: I think that having informative labels, order is not really necessary (also it is not present in any plot and nobody has complained). High dimension and confusing cases should probably better use `dim: coord_value` format or something like this. 

One of the goals here is to make summary and plots use the same labeling and between removing order from summary or adding order to plots I went for removing order from summary.

arviz/sel_utils.py
ahartikainen: That sounds more robust way to handle this

arviz/sel_utils.py
ahartikainen: that option sounds good

OriolAbril: I was looking into this and I realized that having indexes is currently close to impossible, I think it would be better to move this to an issue and also discuss what we want to be possible in a lab meeting.

One of the first operations done is `dataset.sel()`, so the correct index information is lost before calling `xarray_sel_iter`.

arviz/stats/stats.py
rpgoldman: Weren't we talking about returning xarray's of axes (or other backend entities) as a way to do this?  TBH, I don't know if it's possible to move around the subplots in matplotlib -- if this can be done, then the xarray solution would work.

canyon289:
Hi, are there things we can do to help get this one merged? Looks like a lot of work

OriolAbril:
I think the current code is mostly ok and is an improvement over the actual one, not sure it's the way to go though, it still has some clear limitations and it's hard to customize the label formatting.

ahartikainen:
this can wait, right

OriolAbril:
I have rebased the PR and continued the work started here. I proposed to use a Labeller class. The latest commit adds the base class and some subcasses I think will be interesting. I have added some screenshots below, please review.

## BaseLabeller (used by default)

image

## DimCoordLabeller

image

## NoRepeatLabeller

image

## MapLabeller

image

arviz/labels.py
```diff
@@ -0,0 +1,70 @@
...
+
+class MapLabeller(BaseLabeller):
+    def __init__(self, var_name_map=None, dim_map=None, coord_map=None):
+        self.var_name_map = {} if var_name_map is None else var_name_map
+        self.dim_map = {} if dim_map is None else dim_map
+        self.coord_map = {} if coord_map is None else coord_map
+
+    def dim_coord_to_str(self, dim, coord_val, coord_idx):
+        dim_str = self.dim_map.get(dim, dim)
+        coord_str = self.coord_map.get(coord_val, coord_val)
```
OriolAbril, line 54: This could also be `coord_str = self.coord_map.get(dim, {}).get(coord_val, coord_val)`, it won't be as conveninet when there is only one dimension present, but for multidim it may be the only way

arviz/sel_utils.py
OriolAbril, line 141:
```suggestion
            dims = _zip_dims(new_dims, vals)
            idims = _zip_dims(new_dims, [np.arange(len(v)) for v in vals])
            if reverse_selections:
                dims = reversed(dims)
                idims = reversed(idims)

            for selection in dims:
                yield var_name, selection, iselection
```

@ahartikainen I think this will correctly return the index, not sure if it will reduce performance noticeably though. And it has an important caveat which can't be overcome for now (but I think labeller class can fix it) as it's a construction issue as I commented at some point before. One of the first steps is to call `get_coords` to subset the data if `coords` arg is present. And we would take the positional indexes from there, not from the original dataset. 

long story short, when coords are passed, the shown positional indexes would work for `dataset.sel(coords).isel(...)` but will not necessarily work with `dataset.isel(...)`. I think its fine if we document it, and there may be ways to overcome that with labeller class.

ahartikainen: I think you need

    for selection, iselection in zip(dims, idims)

arviz/plots/bpvplot.py
```diff
@@ -93,6 +96,8 @@ def plot_bpv(
         For example, `data_pairs = {'y' : 'y_hat'}`
         If None, it will assume that the observed data and the posterior/prior
         predictive data have the same variable name.
+    labeller : labeller instance, optional
+        Class providing the method `make_label_vert` to generate the labels in the plot titles.
```
OriolAbril, line 100: Do we want to allow only labeller objects? Or strings too and have some kind of mapping/`getattr(az.labels, string)`?

arviz/plots/backends/bokeh/bpvplot.py
```diff
@@ -175,12 +176,12 @@ def plot_bpv(
...
-            xlabel = "{} / {}".format(var_name, pp_var_name)
-        else:
-            xlabel = var_name
+        # if var_name != pp_var_name:
+        #     xlabel = "{} / {}".format(var_name, pp_var_name)
+        # else:
+        #     xlabel = var_name
         _title = Title()
-        _title.text = xlabel
+        _title.text = labeller.make_label_vert(var_name, sel, isel)
```
OriolAbril, line 184: Here matplotlib and bokeh behaved diferently. I have modified bokeh to match mpl, but we can also add a new method to the labellers to take the possible 2 var_names into account. cc @aloctavodia

OriolAbril:
Updated several plots to see how labeling behaved in different ranges, and while I was at it I added a legend argument to `plot_forest`. I have set it to true by default, which combined with #1527 leaves this as default:

image

The rest of the plots work more or less like summary above but with plots, note that the screenshots have the wrong default though, we are now using `var_name[coord]`,  `var_name[dim: coord]` or `var_name[dim#idx]`.

sethaxen:
> I think it is particularly relevant for users of `index_origin` which would mostly be @ahartikainen and @sethaxen. I think this will force ArviZ.jl to do a couple updates (I can help) and I do not expect too many things to break but it would be great to double check and more importantly to make sure this approach will help better integrate index_origin argument.

What kinds of things do you think might need to be updated on the ArviZ.jl side?

OriolAbril:
> What kinds of things do you think might need to be updated on the ArviZ.jl side?

There are two changes that I think will affect the Julia side, on the user side it should be backwards compatible in the sense that the noticeable changes will be fixing issues.

1. index_origin things. Up until this PR, when no coord values were provided, the converters generated automatic integer values starting from 0. Then, plots and stats had an `index_origin` parameter that added 1 to interger coord values so that even though the coord values started at 0, plot labels would start at 1. 

From here on, users can choose what to use when labeling, coords, indexes, nothing at all... and the default is set to always use coord values. index_origin is then enforced at idata creation. Therefore, stan users will set index_origin=1 and will get coordinate values starting from 1. They can then subset their data, chains, draws... with `.sel` using 1-based indexing or with `.isel` using 0-based indexing.

To keep showing 1 based indexing in plots, the julia converters should be updated to enforce index_origin at creation.

2. is about index based labeling. Up until now, the info was simply not available to plotting functions, now it is, which means that  `xarray_var_iter` and `xarray_sel_iter` return 4 elements now instead of 3. 

I don't know if these are used at all on the julia side, but if they are everything will break :/

If you point me to the relevant pieces of code I can help with updating

sethaxen:
> 1. index_origin things. Up until this PR, when no coord values were provided, the converters generated automatic integer values starting from 0. Then, plots and stats had an `index_origin` parameter that added 1 to interger coord values so that even though the coord values started at 0, plot labels would start at 1.
> 
> From here on, users can choose what to use when labeling, coords, indexes, nothing at all... and the default is set to always use coord values. index_origin is then enforced at idata creation. Therefore, stan users will set index_origin=1 and will get coordinate values starting from 1. They can then subset their data, chains, draws... with `.sel` using 1-based indexing or with `.isel` using 0-based indexing.
> 
> To keep showing 1 based indexing in plots, the julia converters should be updated to enforce index_origin at creation.

Nice! Currently ArviZ.jl handles this by setting the `data.index_origin` rcParam upon loading arviz: 
https://github.com/arviz-devs/ArviZ.jl/blob/v0.4.11/src/setup.jl#L55
If the user doesn't supply `index_origin` upon creation, does it default to the rcParam, or is the rcParam going away?

> 1. is about index based labeling. Up until now, the info was simply not available to plotting functions, now it is, which means that  `xarray_var_iter` and `xarray_sel_iter` return 4 elements now instead of 3.
> 
> I don't know if these are used at all on the julia side, but if they are everything will break :/

No, we don't use these on the Julia side, though ArviZPlots currently does. Should be easy to update though; I'll follow how you do that for arviz's plots implementations.

OriolAbril:
> If the user doesn't supply index_origin upon creation, does it default to the rcParam, or is the rcParam going away?

The rcParam is kept so users can set the default there and get the right coord values without having to set index_origin every single time they call the converter.

sethaxen:
> The rcParam is kept so users can set the default there and get the right coord values without having to set index_origin every single time they call the converter.

Okay, then it sounds like for ArviZ.jl no changes will be necessary. I'll add to our converters the new `index_origin` argument and update ArviZPlots.jl accordingly when a new release includes these changes.

OriolAbril:
Ready to merge as soon as tests pass!

PR #1207, opened 2020-05-24
test on both pymc3 master and release

OriolAbril:
## Description
We missed a bug that could have been caught testing on latest pymc3 release, given that we do have 2 external CI jobs, I have configured it so that one runs on master and another on latest release.

## Checklist

### DIFF
--- a/.azure-pipelines/azure-pipelines-external.yml
+++ b/.azure-pipelines/azure-pipelines-external.yml
@@ -10,11 +10,12 @@ jobs:
   timeoutInMinutes: 360
   strategy:
     matrix:
-      Python_37_Latest:
+      Python_37_Latest_PyMC3_gh:
         python.version: 3.7
         pystan.version: "latest"
         pymc3.version: "github"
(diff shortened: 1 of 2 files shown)


.azure-pipelines/azure-pipelines-external.yml
```diff
@@ -10,11 +10,12 @@ jobs:
   timeoutInMinutes: 360
   strategy:
     matrix:
-      Python_37_Latest:
+      Python_37_Latest_PyMC3_gh:
```
ahartikainen, line 13: Why gh can't be in special group?

OriolAbril: it can, I just set it in the first group because the requirement we list is github version (and for now we only tested on gh version) so I assumed it was the "official" dependency (first group) and the latest release was the extra check (group below). 

It will be cleaner the other way round as you say though. I'll wait to check everything works (to be sure no extra changes are needed) before pushing so I don't cancel the run

PR #1212, opened 2020-05-27
Governance (without finance portion)

canyon289:
Establish governance.

This one is a little bit different than other prs.
* The text here should not conflict with what in the doc we agreed upon below (it should be copy and paste with some pymc3 bits)
https://docs.google.com/document/d/1cCCyW-hnwbAIm9ixuyqS8fAzQABvGMNdAxPpJjYsnOM/edit#

* This will be open for a week to allow anyone to comment or dissent

Once this PR is merged a second PR will open with the specifics for the election.


GOVERNANCE.md
```diff
@@ -0,0 +1,189 @@
...
+The Project
+===========
+
+The ArviZ Project (The Project) is an open source software project
+affiliated with the 501c3 NumFocus Foundation. The goal of The Project is to
+develop open source software and deploy open and public websites and services
+for reproducible, exploratory and interactive computing. The Software developed
+by The Project is released under the Apache 2 open source license,
+developed openly and hosted in public GitHub repositories under the
+[GitHub organization](https://github.com/arviz-devs/arviz). Examples of
```
OriolAbril, line 13: I think this should be link to org page, not ArviZ repo

GOVERNANCE.md
```diff
@@ -0,0 +1,189 @@
...
+The Project is developed by a team of distributed developers, called
+Contributors. Contributors are individuals who have contributed code,
+documentation, designs or other work to one or more Project repositories.
+Anyone can be a Contributor. Contributors can be affiliated with any legal
+entity or none. Contributors participate in the project by submitting,
+reviewing and discussing GitHub Pull Requests and Issues and participating in
+open and public Project discussions on GitHub, Slack, and Gitter chat rooms. The foundation of Project participation is openness
+and transparency.
+
+There have many Contributors to the Project, whose contributions are listed in the logs of other repositories of the PyMC and projects.
```
OriolAbril, line 26: ArviZ?

GOVERNANCE.md
```diff
@@ -0,0 +1,189 @@
...
+-----------------------
+
+* N Person steering council (Must be odd, will be decided in first election)
+* Core Contributors 
+* General Contributors
+
+
+Random Variables Council
+------------------------
+The Project will have a Steering Council that consists of Project Contributors
```
OriolAbril, line 72: To be nominated, I think the requirement was to be core dev for a year, which is what is written in eligibility. I think here should be core dev too

GOVERNANCE.md
```diff
@@ -0,0 +1,189 @@
...
+#### One time Decisions
+* The first election will also include a vote for the number of Council Members, options being 5 and 7. A simple majority will indicate preference.
+  * At the conclusion of the vote this section should be removed
+
+
+#### Voting criteria for future elections
+Voting for first election is restricted to establish stable governance, and to defer major decision to elected leaders
+* For the first election only the folks in Slack can vote (excluding GSOC students)
+* In the first year, the council must determine voting eligibility for future elections between two criteria:
+  * Those with commit bits
```
OriolAbril, line 181: I'd say core developers instead of people wit commit bits to stick to the bodies defined above

GOVERNANCE.md
```diff
@@ -0,0 +1,189 @@
...
+
+-   Openness & Transparency
+-   Active Contribution
+-   Institutional Neutrality
+
+Traditionally, project leadership was unstructured but primarily driven
+by a subset of Core Contributors whose active and consistent
+contributions have been recognized by their receiving “commit rights” to the
+Project GitHub repositories. In general all Project decisions are made through
+consensus among the Core Developers with input from the Community.
```
OriolAbril, line 52: I think all references use "core contributors" except this one. I prefer "core developers" or development team, which I also find to clarify the distinctions with general contributors by using a different name. However, core contributors is also ok. We should stick with the same name always though

AlexAndorra: I agree we should stick to the same name. I think I remember that Core Contributors was used because it's defined in a larger sense than just the number of commits (which is closer to Core Dev definition).
It was in the gdoc but I didn't see it here. So I think this role should be clarified, as well as General Contributors, as Oriol wrote.

GOVERNANCE.md
canyon289: Its not defined. Its one of the actions the council needs to take once we elect them

GOVERNANCE.md
canyon289: Agree with consistent verbiage.

Two questions. I'll do one at a time.
The doc mostly uses contributors because "contributing" is a wider scope than "developing". We're inconsistent in the Google Doc as well, but to nail it down here. Developer or contributor? I'm in favor of contributor

GOVERNANCE.md
```diff
@@ -0,0 +1,189 @@
...
+* In the event of a tie there will be a runoff election for the tied candidates. To avoid further ties and discriminate more among the tied candidates, this vote will be held by Majority Judgment: for each candidate, voters judge their suitability for office as either Excellent, Very Good, Good, Acceptable, Poor, or Reject. Multiple candidates may be given the same grade by a voter. The candidate with the highest median grade is the winner.
+* If more than one candidate has the same highest median-grade, the MJ winner is discovered by removing (one-by-one) any grades equal in value to the shared median grade from each tied candidate's total. This is repeated until only one of the previously tied candidates is currently found to have the highest median-grade.
+* In the ties are still present after this second round, the winner will be chosen at random. Each person tied will pick a number between 0 and 100, and a random integer will be generated from random.org. The person with the closet circular distance will be selected.
+* At the conclusion of voting, all the results will be posted.
```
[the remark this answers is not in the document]
canyon289: There will be an election.md for each election in a directory. The election PR will make that more clear and once we have it I can link it here

GOVERNANCE.md
```diff
@@ -0,0 +1,189 @@
...
+* If this is a whole-council vote, then it was necessarily called by Core contributors (since Council members can’t remove the whole Council) and only Core contributors vote. The vote is deemed successful if at least two thirds of voters express a lack of confidence.
+* If this is a single-member vote called by Council Members, only Council Members vote, and the vote is deemed successful if at least half the voters express a lack of confidence. Council Members also have the possibility to call for the whole core team to vote (i.e Council members and Core contributors), although this is not the default option. The threshold for successful vote is also at 50% of voters for this option.
+* If a single-member vote succeeds, then that member is removed from the council and the resulting vacancy can be handled in the usual way. If a whole-council vote succeeds, the council is dissolved and a new council election is triggered immediately.
+
+#### Ejecting core contributors
+* Core contributors can be ejected through a simple majority vote.
```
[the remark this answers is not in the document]
canyon289: Added with what I think we agreed upon

GOVERNANCE.md
AlexAndorra: In favor of "Contributor" too, as it encompasses more what "contributing" means

GOVERNANCE.md
AlexAndorra: Yeah that's pretty wide, but maybe it's ok for the first election? IIUC, refining the definition will be on the first Council's to-do lis.
In any case, it's indeed closer to "Core contributor" than "Core Dev" -- related to our discussion above.

OriolAbril: Edited my comment to say core contributors. The main point is to stick with the bodies defined in the community architecture section where we define steering council, core contributors and general contributors

GOVERNANCE.md
OriolAbril: core contributor is prefectly fine too, I have edited my answers to reduce confusion

GOVERNANCE.md
```diff
@@ -0,0 +1,189 @@
...
+Traditionally, project leadership was unstructured but primarily driven
+by a subset of Core Contributors whose active and consistent
+contributions have been recognized by their receiving “commit rights” to the
+Project GitHub repositories. In general all Project decisions are made through
+consensus among the Core Developers with input from the Community. 
+
+While this approach has served us well, as the Project grows and faces more
+legal and financial decisions and interacts with other institutions, we see a
+need for a more formal governance model. Moving forward The Project leadership
+will consist of a Random Variables Council. We view this governance model as
```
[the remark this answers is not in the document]
canyon289: :)

GOVERNANCE.md
canyon289: Everything should say developers now. Leaving this comment open unti I'm sure I didnt mess up

canyon289:
Added definition for core contributor and confirmation process, although it pretty to be honest

GOVERNANCE.md
```diff
@@ -173,21 +173,30 @@ Each voter can vote zero or more times, once per each candidate. As this is not
...
   * The contributing community at large
 
+### Core Contributors
+Core Contributors are those who have provided consistent and meaningful contributions to ArviZ.
+These can be, but are not limited to, code contributions, community contributions, tutorial 
+development etc. 
+
+#### Core Contributor Nominations and Confirmation Process
+Current Core Contributors can nominate candidates for consideration by the council. The council
+can make the determination for acceptance with a process of their choosing.
```
AlexAndorra, line 194: - This seems vaguer than necessary -- I think we have more informative priors than this 😉 
I mean: the Council will probably accept proposed Core Contributors through the same system of election as the one used to elect / reject the Council, right? So maybe let's just say this here.
- Also, I think we never talked about that: are there any upper and lower bounds on the number of Core Contributors?

OriolAbril: There should be no election for core contributors to join as such, otherwise the strain on core devs and contributors would be too significant. One of the goals is to keep governance as little work as possible.

OriolAbril: I would not set a maximum for core contributors

AlexAndorra: But how are Core Contributors accepted then? Informal votes?
A simple yes/no vote as is done for the first round of Council elections doesn't seem too big of a strain to me, especially if you do it in batch -- e.g every quarter end, all nominated candidates are voted in or out by the Council. Looks more straightforward and transparent to me than doing it gradually and informally 🤷‍♂️ 
No need to be as formal as for the Council though (with a PR and so on) -- a vote on Slack should be sufficient.

OriolAbril: Current core contributors propose people to join as core contributors and council decides as they see fit (I think not being specific on selection criteria and so on is intended in order to allow any kind of contributor to become core, not only people sending prs).

Regarding voting and so on, the normal decision rules for council would be applied (are they defined above? ).

AlexAndorra: > I think not being specific on selection criteria and so on is intended in order to allow any kind of contributor to become core, not only people sending prs.

Completely agree with that -- was just wondering about the voting process.

GOVERNANCE.md
canyon289: Yes............. (of course I mess up in the comment)

GOVERNANCE.md
canyon289: I would agree there is no voting process. None of the other projects Ive seen have had one, Even python. Someone is just like Hey Alex/Oriol etc is great lets give them a commit bit or w/e.

That and I think the council should make the decision based on the factors at the time, not a rigid process setup now. Thats my 2 cents

AlexAndorra: Ow ok! I think we're trying to maximize different utility functions, but that makes sense too and I'm not opposed to it 👌

GOVERNANCE.md
```diff
@@ -0,0 +1,202 @@
...
+
+All members of the Council shall disclose to the rest of the
+Council any conflict of interest they may have. Members with a conflict of
+interest in a particular issue may participate in Council discussions on that
+issue, but must recuse themselves from voting on the issue. 
+
+### Council Selection Process
+
+#### Eligibility
+* Must be core contributor for at least one year
```
OriolAbril, line 131: I commented above, here the requirement is being core contributor for at least a year, above it is being general contributor for at least a year. I think the above wording has to be changed to say core contributor

GOVERNANCE.md
```diff
@@ -0,0 +1,204 @@
...
+
+The Project Community consists of all Contributors and Users of the Project.
+Contributors work on behalf of and are responsible to the larger Project
+Community and we strive to keep the barrier between Contributors and Users as
+low as possible.
+
+The Project is formally affiliated with the 501c3 NumFOCUS Foundation
+([http://numfocus.org](http://numfocus.org)),  NumFOCUS is the
+only legal entity that has a formal relationship with the project (see
+Institutional Partners section below).
```
OriolAbril, line 37: There is no section below yet

GOVERNANCE.md
```diff
@@ -0,0 +1,204 @@
...
+* Make decisions about the overall scope, vision and direction of the project.
+
+Note that each individual council member does not have the power to unilaterally wield these responsibilities, but the council as a whole must jointly make these decisions. In other words, Council Members are first and foremost Core Contributors, but only when needed they can collectively make decisions for the health of the project.
+
+ArviZ will be holding its first election to determine its initial council in the coming weeks and this document will be updated.
+
+### Private communications of the Council
+
+Unless specifically required, all Council discussions and activities will be
+public and done in collaboration and discussion with the Project Contributors
```
OriolAbril, line 105: public as in github/discourse or public as in slack (all core contributors would then have access, but general contributors would not)

GOVERNANCE.md
canyon289: Youre right removed

GOVERNANCE.md
canyon289: Updated wording. Let me know if it sounds better

GOVERNANCE.md
```diff
@@ -101,8 +101,9 @@ ArviZ will be holding its first election to determine its initial council in the
 ### Private communications of the Council
 
 Unless specifically required, all Council discussions and activities will be
-public and done in collaboration and discussion with the Project Contributors
-and Community. The Council will have a private mailing list that will be used
+between public (Github, gitter), and partially public channels (Slack)
+and done in collaboration and discussion with the Project Contributors
```
AlexAndorra, line 105: I saw you replaced "Project Contributors" by "Core Contributors" above -- maybe let's do it here too?

GOVERNANCE.md
```diff
@@ -0,0 +1,204 @@
...
+* If more than one candidate has the same highest median-grade, the MJ winner is discovered by removing (one-by-one) any grades equal in value to the shared median grade from each tied candidate's total. This is repeated until only one of the previously tied candidates is currently found to have the highest median-grade.
+* If ties are still present after this second round, the winner will be chosen at random. Each person tied will pick a number between 0 and 100, and a random integer will be generated from random.org. The person with the closet circular distance will be selected.
+* At the conclusion of voting, all the results will be posted.
+* The decision about who can vote after the first election is deferred to the council. See below for details
+
+#### Length of Tenure and Reverification
+* Council members term limits are 4 years, after which point their seat will come up for reelection.
```
aloctavodia, line 152: Should the council be renewed by half every two years? This could be useful to preserve the know-how the council could have acquired. Probably the election dynamics will make it higly likely that the council updates partially, even if we do not enforce it, but anyway.
It seems the council members can be reelected indefinitely, should this be made explicit?

canyon289: I'm fine with 2 years! If anyone disagrees please make it known now. Council Members can be reelected indefinitely (which is by design but arbitrary an can be changed). I'll make it explicit

aloctavodia: OK, but my main point is about renewing in halfs or renewing the entire council

canyon289:
Merging this PR to establish base test for governance, but this *does not implement our governance model*. Talking to NumFOCUS we need to include a section on how we will handle finances. This will be added in a follow up PR.

In parallel the election PR https://github.com/arviz-devs/arviz/pull/1216 needs to be approved and merged, after which we will start the election process contained in that doc. At the conclusion of all that we ail have governance

PR #1214, opened 2020-05-29
fix bug in plot_loo_pit with use_hdi

OriolAbril:
## Checklist


aloctavodia:
My original intention was to create a separated PR, but maybe we can go with this.

This makes two important changes. compute the kde over the data range (not over 0, 1). This help to spots problems, as the KDE is "not adding data". Compare our [example](https://arviz-devs.github.io/arviz/generated/arviz.plot_loo_pit.html) in the gallery, with the new version.

 lala.

This also introduce a theoretical HDI, instead of extimating the HDI from the samples.

lala_hdi

maybe we should enforce the zero y-bottom-limit.

aloctavodia:
Sorry for the mess I am closing this in favor of #1215

PR #1215, opened 2020-05-29
change kde limits and hdi computation

aloctavodia:

This makes two important changes. compute the kde over the data range (not over 0, 1). This help to spots problems, as the KDE is "not adding data". Compare our [example](https://arviz-devs.github.io/arviz/generated/arviz.plot_loo_pit.html) in the gallery, with the new version.

 lala.

This also introduce a theoretical HDI, instead of estimating the HDI from the samples.

lala_hdi

maybe we should enforce the zero y-bottom-limit.


arviz/plots/loopitplot.py
```diff
@@ -206,20 +206,19 @@ def plot_loo_pit(
...
             fill_kwargs.setdefault("label", "{:.3g}% credible interval".format(credible_interval))
     else:
-        loo_pit_kde, _, _ = _fast_kde(loo_pit, xmin=0, xmax=1)
+        loo_pit_kde, xmin, xmax = _fast_kde(loo_pit)
 
         unif = np.random.uniform(size=(n_unif, loo_pit.size))
-        x_vals = np.linspace(0, 1, len(loo_pit_kde))
+        x_vals = np.linspace(xmin, xmax, len(loo_pit_kde))
         if use_hdi:
+            n_obs = len(loo_pit_kde)
```
OriolAbril, line 214: Why `len(loo_pit_kde)` and not `loo_pit.size`? Shouldn't the trend "the more observations the more restrictive hdi should be"?

aloctavodia: The asymptotic distribution should be uniform on the interval 0 and 1. If you do not have values covering that range then the model is miss-calibrated, even if it looks flat in a subinterval. I think not setting the range between 0 and 1. Makes this problem even easier to spot. 
You are right it should be loo_pit.size.

OriolAbril: I was also underestimating the normalization effect, loo pit kde generally ends up clearly above the upper limit of unif kdes

CHANGELOG.md
```diff
@@ -11,10 +11,14 @@
 ### Documentation
 
 ## v0.8.3 (2020 May 28)
+### New features
+* loo-pit plot. The kde is computed over the data interval (this could be shorter than [0, 1]). The hdi is computed analitically (#1215)
```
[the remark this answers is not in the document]
aloctavodia: oops

OriolAbril:
Is this ready to merge?

aloctavodia:
Yes.

Issue #1066, opened 2020-02-12
InferenceData extensions [Discussion]


OriolAbril:
`InferenceData` objects are central to ArviZ, and even though a common subset of tasks using InferenceData can be done directly with ArviZ plotting and stats functions, any task that deviates from this becomes more and more convoluted and long. 

The aim of this issue is to start a discussion about new capabilities to add to `InferenceData` and generate a proposal (which will be added to [xarray_examples](https://github.com/arviz-devs/xarray_examples) for discussion with xarray team). 

I also think there are several groups of functions, if it may help start brainstroming or generating different proposals per group. Ideas on all levels are welcome!

### Straightforward extensions to `xr.Dataset` methods
`.sel` is a good example of this. I think several methods could fit in this category and very roughly follow a similar pattern:

```
def idata_extension(self, groups, ... , **kwargs):
    for group in groups:
        if group not in self._groups:
            raise Error
        # some kind of check to make method as convenient as possible
        # an example is sel using only the dimensions present in current group to index
        dataset = getattr(self, group)
        setattr(self, group, datasel.method(**kwargs)
```

In addition to `groups` we should think about other ArviZ specific args, common in most functions and not passed to xarray. Maybe `inplace` and/or `copy`? 

Also, `groups` could accept groups and some _metagroups_ so that one keyword represents several proper groups. We could go as far as adding the metagroups dict in rcParams. One metagroup example could be `"posteriors" -> ("posterior", "sample_stats", "log_likelihood", "posterior_predictive")`

Some ideas of functions that could fit in this category are:
* `.isel`
* `stack` and `unstack`
* `rename`, `rename_dims` and `rename_vars`
* `.load` and other dask related methods like `chunk` would be interesting after ArviZ starts becoming Dask friendly.

Many [dataset methods](http://xarray.pydata.org/en/stable/generated/xarray.Dataset.html) make sense to extend, so I think we should focus on the ones that solve more issues on our side. For example, if we make an extension to apply_ufunc compatible with inference data or extend the map method, the mean, median, max... are not really necessary, only convenient, whereas other methods may have no alternative. 

Commenting the ones you expect to use the most seems like a good start to choose where to begin with.

## Specific inference data methods
This category requires a much more detailed and custom implementation. Some examples that would fall here are:
* InferenceData html repr
[...]

PR #1217, opened 2020-06-01
html representation for jupyter notebooks

percygautam:
## Description
This PR implements the html representation for arviz.InferenceData object in jupyter notebooks. Related to discussion in #1066.

## Checklist


arviz/data/inference_data.py
```diff
@@ -133,6 +133,18 @@ def __repr__(self):
             msg += "\n\nWarmup iterations saved ({}*).".format(WARMUP_TAG)
         return msg
 
+    def _repr_html_(self):
+        """Make html representation of object"""
```
canyon289, line 137: Would change object to `InferenceData` so this docstring can be read without having to refer to class name

arviz/data/inference_data.py
```diff
@@ -133,6 +133,18 @@ def __repr__(self):
...
+        xr.set_options(display_style="html")
+        html_template, element_template, css_template = html_templates(self)
+        template, elements = "", ""
+        for data_group in self._groups:
+            func = getattr(self, data_group)
+            elements += element_template.format(data_group, func._repr_html_())
+        template += html_template.format(elements)
```
canyon289, line 144: Are these `+=` operators adding strings together? If so could this be changed to another method of string concat? The reason I ask is that `+=` is typically used for numbers or other things and it's confusing to see it used as strings.

Its typically not used for string concats for the reasons detailed in the article below

https://wiki.python.org/moin/PythonSpeed/PerformanceTips#String_Concatenation

arviz/utils.py
```diff
@@ -664,3 +665,35 @@ def credible_interval_warning(credible_interval, hdi_prob):
 
     hdi_prob = credible_interval
     return hdi_prob
+
+
+def html_templates():
```
canyon289, line 670: I would change this python function to a class that of structure. Reason for this is that

1. Function doesn't do anything but return values
2. Defining objects in a function scope makes it hard to "get at them" without running function
3. In future use case we only want one of these things we still need to deal with the other 2 because with a function call we always get them
4. In the code where this is used it'll make it more clear that computation likely isn't happening but were just "grabbing a template"



```
HTMLTemplates:
    html_template = xyz
```

As a sidenote DataClasses would be great but they were introduced in 3.7 so if we use them here they will break ArviZ for Python 3.6 users

arviz/data/inference_data.py
OriolAbril: I would also rename `func` to `ds`, `group` or `data` so the name is related to the object it stores. Maybe something like:

```
for group in self._groups_all:
    ds = getattr(...
```

I'd also use `_groups_all` given that groups start hidden and users can expand the desired ones

arviz/data/inference_data.py
```diff
@@ -133,6 +133,18 @@ def __repr__(self):
             msg += "\n\nWarmup iterations saved ({}*).".format(WARMUP_TAG)
         return msg
 
+    def _repr_html_(self):
+        """Make html representation of object"""
+        xr.set_options(display_style="html")
```
OriolAbril, line 138: I am not sure about this. html repr is the default for xarray now (we could require a version where this is already default) so this will generally have no effect, and if it were not the default, it means the user has specifically changed it so undoing the change in the background and with no option to override it seems like a bad idea.

How does xarray choose between html repr and "regular" print? Maybe we can check xarray defaults us too to choose one or the other?

arviz/data/inference_data.py
```diff
@@ -133,6 +133,18 @@ def __repr__(self):
...
+        html_template, element_template, css_template = html_templates(self)
+        template, elements = "", ""
```
OriolAbril, line 140: I'd rename the `template` string to something more suited, maybe `html_repr`?

arviz/utils.py
```diff
@@ -664,3 +665,35 @@ def credible_interval_warning(credible_interval, hdi_prob):
...
+              <div class='xr-header'>
+                <div class="xr-obj-type">arviz.InferenceData</div>
+              </div>
+              {}
+            </div>
+            """
+    element_template = """
+            <ul class="xr-sections">
+            <li class = "xr-section-item">
+                  <input id="xr-dataset-coord_{0}" class="xr-section-summary-in" type="checkbox">
```
OriolAbril, line 682: given that the group name is added to the `id` I am guessing the id is arbitrary and the class is what defines the format used. Is that so? If so, can this be renamed so it becomes more clear what is happening? 

If I am not mistaken, in `{1}` we put the whole html_repr of the xarray dataset corresponding to the group at hand, so having xr, dataset and coord but no az, idata nor group in id section is confusing.

arviz/data/inference_data.py
percygautam: @canyon289 Thanks for the suggestion. I'd keep it in mind for the future. I also learnt a lot of other tricks too from the article :)

arviz/utils.py
percygautam: Okay, I understand. I've changed it to class.

arviz/data/inference_data.py
percygautam: I understand the issue, I tried to solve it by saving the default style and reapplying after the display of html_repr.
xarray is doing like [this](https://github.com/jsignell/xarray/blob/1d960933ab252e0f79f7e050e6c9261d55568057/xarray/core/common.py#L138) The default option is "text".

arviz/utils.py
percygautam: Yes, the id is arbitrary. I left it as it is when I first tinkered with xarray's code and forgot to change it. I've changed it.

percygautam:
@canyon289 Yes the xarray's repr is in html if we set the option. This PR is making the whole inferenceData object dynamic. The result can be seen in this jupyter notebook: inhttps://nbviewer.jupyter.org/github/percygautam/arviz-examples/blob/master/HTML%20repr.ipynb

arviz/data/inference_data.py
OriolAbril: I would mimic xarray behaviour. If display style is text we should not print the html repr. 

The default in latest xarray is already html https://github.com/pydata/xarray/blob/master/xarray/core/options.py#L23 so we should probably require xarray>=0.15.1. I think we should have html repr by default so requiring this or later xarray version would make it default.

arviz/utils.py
```diff
@@ -664,3 +665,34 @@ def credible_interval_warning(credible_interval, hdi_prob):
...
+    html_template = """
+            <div>
+              <div class='xr-header'>
+                <div class="xr-obj-type">arviz.InferenceData</div>
+              </div>
+              {}
+            </div>
+            """
+    element_template = """
+            <ul class="xr-sections">
```
OriolAbril, line 680: Currently the vertical spacing between groups is very large. I think this could be due to `ul` being part of the element template as this is defining a different list for each group. Can you try moving it to the html template so that we have a single list (only one `ul` defined in html template) and the groups are then elements of this single list (one `li` per group)

arviz/data/inference_data.py
percygautam: Okay, I understand. I was reading the fork of the guy who implemented the html_repr for xarray a long time ago.

OriolAbril:
LGTM, we can merge after updating a couple notebooks.

canyon289:
I am going to request tests before merge. IMO all code should be tested if it can, and this is no exception :)

arviz/data/inference_data.py
```diff
@@ -125,14 +126,38 @@ def __init__(self, **kwargs):
...
         msg = "Inference data with groups:\n\t> {options}".format(
             options="\n\t> ".join(self._groups)
         )
         if self._groups_warmup:
             msg += "\n\nWarmup iterations saved ({}*).".format(WARMUP_TAG)
         return msg
 
+    def _repr_html_(self):
+        """Make html representation of InferenceData object."""
+        from xarray.core.options import OPTIONS
```
canyon289, line 139: Why is this import here? Shouldnt it be at the top of the module?

arviz/utils.py
```diff
@@ -664,3 +665,32 @@ def credible_interval_warning(credible_interval, hdi_prob):
 
     hdi_prob = credible_interval
     return hdi_prob
+
+
+class HtmlTemplate:
+    """Contains html templates."""
```
canyon289, line 671: Expand docstring. Contains html templates for what?

Would change this to something along the lines of contains html templates for InferenceData repr

arviz/utils.py
```diff
@@ -664,3 +665,32 @@ def credible_interval_warning(credible_interval, hdi_prob):
...
+                <div class="xr-obj-type">arviz.InferenceData</div>
+              </div>
+              <ul class="xr-sections">
+              {}
+              </ul>
+            </div>
+            """
+    element_template = """
+            <li class = "xr-section-item">
+                  <input id="idata_{0}" class="xr-section-summary-in" type="checkbox">
```
canyon289, line 685: Instread of using numbers for string interpolation, using named arguments would make this easier to read and harder to mess up

arviz/utils.py
```diff
@@ -664,3 +665,32 @@ def credible_interval_warning(credible_interval, hdi_prob):
...
+                  <label for="idata_{0}" class = "xr-section-summary">{0}</label>
+                  <div class="xr-section-inline-details"></div>
+                  <div class="xr-section-details">
+                      <ul id="xr-dataset-coord-list" class="xr-var-list">
+                          <div style="padding-left:2rem;">{1}<br></div>
+                      </ul>
+                  </div>
+            </li>
+            """
+    specific_style = ".xr-wrap{width:700px!important;}"
```
canyon289, line 695: No change required but just curious what is so specific about this style?

canyon289:
@percygautam this is already looking great and youre on the right track. Hope you dont mind my additional comments :)

arviz/utils.py
percygautam: The xarray uses the style `min-width:400px; max-width:700px;` for sections. For single xarray dataset max-width is used, while when used in sub-section min-width is used.
So, we are overwriting our own style over xarray's style.

percygautam:
> I am going to request tests before merge. IMO all code should be tested if it can, and this is no exception :). I would add a test that calls `_repr_html_` directly and tests that the html string is returned.
> 
> I would also add a tests the correct repr is returned without an explicit call. I assume this will require some mocking and to "trick" the non jupyter notebook test to provide the html repr.

I have added the basic test. How do I add the second test? I looked through but couldn't find the way.

arviz/data/inference_data.py
OriolAbril: could it be you modified that and rolled back the changes? I have a vague memory of `f"<pre>{escape(repr(self))}</pre>"` but now I an not sure. Either way, this clearly has to be changed to use the xarray default instead of overriding it

arviz/utils.py
```diff
@@ -664,3 +665,36 @@ def credible_interval_warning(credible_interval, hdi_prob):
 
     hdi_prob = credible_interval
     return hdi_prob
+
+
+class HtmlTemplate:
+    """Contains html templates for InferenceData repr."""
```
OriolAbril, line 671: I think you'll need this to keep both black and numpydoc happy: https://github.com/arviz-devs/arviz/blob/master/arviz/rcparams.py#L28

arviz/utils.py
OriolAbril: This `ul`/`li` change is also marked as resolved but it is not and again I have a vague memory of seeing the `ul` moved to html instead of element template :/

percygautam:
Actually I did revert some commits as I have forgot to gitignore the cache of notebooks and pushed them. I'll check and make all the changes again.

arviz/data/inference_data.py
percygautam: Yes, that's the case. I apologize for these discrepancies. I have corrected the ones I found.

arviz/tests/base_tests/test_data.py
```diff
@@ -456,6 +456,13 @@ def test_map(self, use):
         )
         assert np.allclose(idata_map.posterior.mu, idata.posterior.mu)
 
+    def test_repr_html(self):
+        idata = load_arviz_data("centered_eight")
```
canyon289, line 460: NIT, test should have docstring

arviz/tests/base_tests/test_data.py
```diff
@@ -456,6 +456,13 @@ def test_map(self, use):
...
+        html = idata._repr_html_()  # pylint: disable=protected-access
+
+        assert html is not None
```
canyon289, line 463: This is gerat

arviz/utils.py
canyon289: Cool!

arviz/data/inference_data.py
```diff
@@ -125,14 +127,38 @@ def __init__(self, **kwargs):
...
         msg = "Inference data with groups:\n\t> {options}".format(
             options="\n\t> ".join(self._groups)
         )
         if self._groups_warmup:
             msg += "\n\nWarmup iterations saved ({}*).".format(WARMUP_TAG)
         return msg
 
+    def _repr_html_(self):
+        """Make html representation of InferenceData object."""
+        display_style = OPTIONS["display_style"]
```
canyon289, line 140: > > I am going to request tests before merge. IMO all code should be tested if it can, and this is no exception :). I would add a test that calls `_repr_html_` directly and tests that the html string is returned.
> > I would also add a tests the correct repr is returned without an explicit call. I assume this will require some mocking and to "trick" the non jupyter notebook test to provide the html repr.
> 
> I have added the basic test. How do I add the second test? I looked through but couldn't find the way.

This is going to be  fun one. Making tests is sometimes a puzzle and harder than the actual programming itself, but important if you want you feature to be stable and not broken by any others.

In the test you'll have to monkeypatch `OPTIONS` to so when the test executes the value is `text` to test that this if condition works. Just being up front here it's going to be hard and confusing, but youre going to learn a lot more about python.

I don't want to block this whole PR for that one thing though. We can merge but I really would like you to add this test in a followup PR. At a minimum create an issue so we can track


https://docs.pytest.org/en/latest/monkeypatch.html

OriolAbril: I think it will not be necessary to monekypatch because the display style option is already customizable similarly to an rcParam. We can start the test by setting the `display_style` to `text`/`html`, make sure the correct repr is returned and then finish the test by going back to the original displaystyle (much like your original implementation of the repr @percygautam).

arviz/tests/base_tests/test_data.py
```diff
@@ -456,6 +456,13 @@ def test_map(self, use):
...
+        assert "<div" in html
```
OriolAbril, line 464: I would make this assert more specific to check all required elements are present (at least partially). i.e. for html repr to work, we need our custom part (we could check for group names being present for example), the dataset repr should be present too (we could check for variable names of for Coordinate section title being present) and also style inherited from xarray (we could test a generic line/command from the style to assert is is correclty being appended to the repr)

arviz/data/inference_data.py
canyon289: Its typically bad practice to have tests rely on global state like rcparam although in this case lets just do it cause monkeypatching would be annoying :D

percygautam: I am okay with both options. We can currently implement the test using rcParams, and create the issue to track the monkeypatch test and merge this PR. I will implement the test later.
The `html_repr` really facilitates to check other methods that I am adding, so it'll be better to merge this one as soon as possible.

canyon289: Agree with merging as soon as possible! Seems like youre only a couple linting errors away :)

canyon289:
Add changelog and this one should be good to go

OriolAbril:
@percygautam can you rebase and rerun notebook? `from_pyjags` has just been added so it is one more case for which to show the new repr

PR #1219, opened 2020-06-02
Added support for PyJAGS in ArviZ

michaelnowotny:
This pull request adds functionality to translate posterior samples generated by PyJAGS to an ArviZ inference data object via a new public function 'from_pyjags'. A set of unit tests checks that a round trip translation from PyJAGS to ArviZ and back retains the same information. A new section has been added to the documentation notebook illustrating the use ArviZ in conjunction with PyJAGS.

## Checklist


StanczakDominik:
With respect to the Pylint failure: StackOverflow at https://stackoverflow.com/a/54045715/4417567 and existing arviz code such as https://github.com/arviz-devs/arviz/blob/34b96712b948e88b1640ae4c755749099bec3bec/arviz/tests/base_tests/test_data.py#L1 both seem to suggest disabling that particular linting. There's another workaround at https://stackoverflow.com/a/57015304/4417567 with the `name` parameter to `fixture`, but it seems to be rather more complicated.

ahartikainen:
Hi, thanks for the PR.

I think  `io_pyjags.py` could follow similar structure as other `_io.py` files. This will make it easier to expand the functionality and add special handling for different groups.

For e.g

https://github.com/arviz-devs/arviz/blob/master/arviz/data/io_pystan.py

Create a converter class
```
class PyStanConverter:
    """Encapsulate PyStan specific logic."""

    def __init__(
        self,
        *,
        posterior=None,
        posterior_predictive=None,
        predictions=None,
        prior=None,
        prior_predictive=None,
        observed_data=None,
        constant_data=None,
        predictions_constant_data=None,
        log_likelihood=None,
        coords=None,
        dims=None,
        save_warmup=None,
    ):
        self.posterior = posterior
        self.posterior_predictive = posterior_predictive
        self.predictions = predictions
        self.prior = prior
        self.prior_predictive = prior_predictive
        self.observed_data = observed_data
        self.constant_data = constant_data
        self.predictions_constant_data = predictions_constant_data
        self.log_likelihood = log_likelihood
        self.coords = coords
        self.dims = dims
        self.save_warmup = rcParams["data.save_warmup"] if save_warmup is None else save_warmup

        import pystan  # pylint: disable=import-error

        self.pystan = pystan
```

Then each group is handled with a class method which return xarray datasets

```
    @requires("posterior")
    def posterior_to_xarray(self):
        """Extract posterior samples from fit."""
        posterior = self.posterior
       ....
        data, data_warmup = get_draws(posterior, ignore=ignore, warmup=self.save_warmup)

        return (
            dict_to_dataset(data, library=self.pystan, coords=self.coords, dims=self.dims),
            dict_to_dataset(data_warmup, library=self.pystan, coords=self.coords, dims=self.dims),
        )
```

Then finally a method to transform class to InferenceData

```
    def to_inference_data(self):
        """Convert all available data to an InferenceData object.
        Note that if groups can not be created (i.e., there is no `fit`, so
        the `posterior` and `sample_stats` can not be extracted), then the InferenceData
        will not have those groups.
        """
        data_dict = self.data_to_xarray()
        return InferenceData(
            save_warmup=self.save_warmup,
            **{
                "posterior": self.posterior_to_xarray(),
                "sample_stats": self.sample_stats_to_xarray(),
                "log_likelihood": self.log_likelihood_to_xarray(),
                "posterior_predictive": self.posterior_predictive_to_xarray(),
                "predictions": self.predictions_to_xarray(),
                "prior": self.prior_to_xarray(),
                "sample_stats_prior": self.sample_stats_prior_to_xarray(),
                "prior_predictive": self.prior_predictive_to_xarray(),
                **({} if data_dict is None else data_dict),
            },
        )
```

To extract library specific data can use "external" functions as is the case in PyStan which return dictionaries containing data

```
def get_draws(fit, variables=None, ignore=None, warmup=False):
    """Extract draws from PyStan fit."""
    ....
    return data, data_warmup
```

```` 
def get_sample_stats(fit, warmup=False):
    """Extract sample stats from PyStan fit."""
    ...
    return data, data_warmup
```

And `from_xyz` function then can be defined as 

def from_pystan(
    posterior=None,
    *,
    posterior_predictive=None,
    predictions=None,
    prior=None,
    ...
):
    """Convert PyStan data into an InferenceData object.
    For a usage example read the
    :doc:`Cookbook section on from_pystan </notebooks/InferenceDataCookbook>`
    Parameters
    ----------
    posterior : StanFit4Model or stan.fit.Fit
        PyStan fit object for posterior.
    posterior_predictive : str, a list of str
        Posterior predictive samples for the posterior.
    predictions : str, a list of str
        Out-of-sample predictions for the posterior.
    prior : StanFit4Model or stan.fit.Fit
        PyStan fit object for prior.
    ...
    Returns
    -------
    InferenceData object
    """
        return PyStanConverter(
            posterior=posterior,
            posterior_predictive=posterior_predictive,
            predictions=predictions,
            prior=prior,
            ...
        ).to_inference_data()
```

**Comment by ahartikainen** [comment]
Also, notice that PyStan returns also the warmup iterations, but other libs can return only posterior samples. So the structure might be a bit different.

File: arviz/data/io_pyjags.py [hunk 1]
```diff
@@ -0,0 +1,326 @@
...
+    elif isinstance(variables, str):
+        variables = [variables]
+
+    if not isinstance(variables, Sequence):
+        raise TypeError('variables must be of type Sequence or str')
+
+    variables = tuple(variables)
+
+    if warmup_iterations > 0:
+        warmup_samples, actual_samples =\
```
ahartikainen, line 115: You can just put everything to oneline and then let black handle multiline version.

arviz/data/io_pyjags.py
```diff
@@ -0,0 +1,326 @@
...
+def get_draws(pyjags_samples: tp.Dict[str, np.ndarray],
+              variables: tp.Optional[tp.Union[str, tp.Iterable[str]]] = None,
+              warmup: bool = False,
+              warmup_iterations: int = 0) \
+        -> tp.Tuple[tp.Dict[str, np.ndarray], tp.Dict[str, np.ndarray]]:
+    """
+
+    Parameters
+    ----------
+    pyjags_samples: a didctionary mapping variable names to NumPy arrays of MCMC
```
ahartikainen, line 90:
```suggestion
    pyjags_samples: a dictionary mapping variable names to NumPy arrays of MCMC
```

arviz/data/io_pyjags.py
```diff
@@ -0,0 +1,326 @@
...
+            number_of_chains, chain_length = chains.shape
+            chains = chains.reshape((number_of_chains, chain_length, 1))
+
+        variable_name_to_samples_map[variable_name] = \
+            np.swapaxes(chains, 0, 2)
+
+    return variable_name_to_samples_map
+
+
+def from_pyjags(posterior: tp.Optional[tp.Dict[str, np.ndarray]] = None,
```
ahartikainen, line 295: add `dims` and `coords` option too, this way idata can contain metadata for needed dimensions/items.

Extra groups can be added with varnames option (e.g. move certain values from posterior to prediction groups (not sure how pyjags handle these.

Is there a way to calculate elementwise log_likelihood for data?

Also data for example can be added as a dict.

arviz/data/io_pyjags.py
```diff
@@ -0,0 +1,326 @@
...
+           (parameter_dimension, chain_length, number_of_chains)
+
+    save_warmup: whether or not to return warmup draws if any
+    warmup_iterations: how many iterations are for warmup
+
+    Returns
+    -------
+    An Arviz inference data object
+    """
+    return (PyJAGSConverter(posterior=posterior,
```
ahartikainen, line 322: Here too, you can let black to handle multiline

arviz/data/io_pyjags.py
```diff
@@ -35,7 +35,7 @@ def __init__(
         try:
             import pyjags  # pylint: disable=import-error
             self.pyjags = pyjags
-        except Exception as e:
+        except Exception:
```
StanczakDominik: That's pretty broad. Maybe it would make sense to just catch `ImportError` here?

michaelnowotny: You are right, the linter had complained about this. I changed the exception type to ModuleNotFoundError

arviz/data/io_pyjags.py
michaelnowotny: thanks for catching that

arviz/data/io_pyjags.py
michaelnowotny: I tried black but don't like the way it formats things. Among other things, it often goes above 80 characters per line. Is using black mandatory for ArviZ?

ahartikainen: Yes. And we use 100 char per line.

arviz/data/io_pyjags.py
michaelnowotny: I did not find a way to automatically get the log-likelihood from JAGS.

arviz/data/io_pyjags.py
```diff
@@ -0,0 +1,320 @@
...
+    Convert PyJAGS posterior samples to an ArviZ inference data object.
+
+    Takes a python dictionary of samples that has been generated by the sample
+    method of a model instance and returns an Arviz inference data object.
+
+    Parameters
+    ----------
+    posterior: a dictionary mapping variable names to NumPy arrays containing
+               posterior samples with shape
+               (parameter_dimension, chain_length, number_of_chains)
```
OriolAbril, line 294:
```suggestion
    posterior: dict of {str : array_like}, optional
        a dictionary mapping variable names to NumPy arrays containing posterior samples 
        with shape (parameter_dimension, chain_length, number_of_chains)
```

Docstrings should follow [numpydoc style](https://github.com/arviz-devs/arviz/blob/master/CONTRIBUTING.md#docstring-formatting-and-type-hints). similar comment for arguments below

arviz/data/io_pyjags.py
```diff
@@ -0,0 +1,320 @@
...
+            is the name of the dimension, the values are the index values.
+    dims: dict[str, List(str)]
+          A mapping from variables to a list of coordinate names for the variable.
+
+    save_warmup: whether or not to return warmup draws if any
+    warmup_iterations: how many iterations are for warmup
+
+    Returns
+    -------
+    An Arviz inference data object
```
OriolAbril, line 311:
```suggestion
    InferenceData
```

numpydoc style again, first should go the type of the returned object, then indented the description (which is optional and in this case I don't think is necessary)

arviz/tests/external_tests/test_data_pyjags.py
```diff
@@ -0,0 +1,147 @@
...
+    _convert_arviz_samples_dictionary_to_pyjags_samples_dictionary,
+    _extract_samples_dictionary_from_arviz_inference_data,
+    from_pyjags,
+)
+
+
+@pytest.fixture()
+def pyjags_samples_dict() -> tp.Dict[str, np.ndarray]:
+    return {
+        "b": np.array(
```
OriolAbril, line 20: `np.random.normal(size=(...))` can be used to reduce the lines used

arviz/tests/external_tests/test_data_pyjags.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def verify_equality_of_numpy_values_dictionaries(
+    dict_1: tp.Dict[tp.Any, np.ndarray], dict_2: tp.Dict[tp.Any, np.ndarray]
+) -> bool:
+    if dict_1.keys() != dict_2.keys():
+        return False
+
+    for key in dict_1.keys():
+        if not np.all(dict_1[key] == dict_2[key]):
```
OriolAbril, line 100: `np.allclose` should be more robust

arviz/tests/external_tests/test_data_pyjags.py
```diff
@@ -0,0 +1,147 @@
...
+        arviz_inference_data_from_pyjags_samples_dict
+    )
+
+    assert verify_equality_of_numpy_values_dictionaries(
+        arviz_samples_dict_from_pyjags_samples_dict,
+        arviz_samples_dict_from_arviz_inference_data_from_pyjags_samples_dict,
+    )
+
+
+def test_from_pyjags(
```
OriolAbril, line 133: I does not look like it is testing `from_pyjags`

arviz/data/io_pyjags.py
michaelnowotny: I ran black with 100 characters per line which produces lines that are longer than 100. This causes pylint to complain:
arviz/tests/external_tests/test_data_pyjags.py:110:0: C0301: Line too long (138/100) (line-too-long)
arviz/tests/external_tests/test_data_pyjags.py:123:0: C0301: Line too long (130/100) (line-too-long)
arviz/tests/external_tests/test_data_pyjags.py:137:0: C0301: Line too long (130/100) (line-too-long)
arviz/tests/external_tests/test_data_pyjags.py:141:0: C0301: Line too long (115/100) (line-too-long)

If I run black with 80 or 90 characters per line it also produces lines that are longer than 100.

If I manually insert line breaks to meet the 100 character limit then black complains that the file is not formatted the way it would format it. Is there a way to either disable the pylint check for characters per line or the black check for individual files?

arviz/data/io_pyjags.py
```diff
@@ -0,0 +1,320 @@
...
+    variable_names: the variables in the dictionary to use; if None use all
+
+    Returns
+    -------
+    A tuple of two pyjags samples dictionaries in PyJAGS format
+    """
+    if variable_names is None:
+        variable_names = tuple(pyjags_samples.keys())
+
+    assert isinstance(variable_names, tuple)
```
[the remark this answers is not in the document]
michaelnowotny: It is not necessary. The code would work with an iterable. But it is more expressive. Since a tuple is immutable, It conveys the fact that the function does not change the input.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Change Log
 
 ## v0.x.x Unreleased
+* Added support for PyJAGS via the function 'from_pyjags' in the module arviz.data.io_pyjags. Unit tests have been added in arviz.tests.external_tests.test_data_pyjags. A section has been added to the documentation at InferenceDataCookbook.ipynb illustrating the use of ArviZ in conjunction with PyJAGS.
```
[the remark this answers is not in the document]
michaelnowotny: Ok, done that

arviz/data/io_pyjags.py
```diff
@@ -0,0 +1,320 @@
...
+
+
+class PyJAGSConverter:
+    """Encapsulate PyJAGS specific logic."""
+
+    def __init__(
+        self,
+        *,
+        posterior: tp.Optional[tp.Dict[str, np.ndarray]] = None,
+        prior: tp.Optional[tp.Dict[str, np.ndarray]] = None,
```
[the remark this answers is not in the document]
michaelnowotny: Changed Dict to Mapping

arviz/tests/external_tests/test_data_pyjags.py
michaelnowotny: I thought about that but since the actual data are being copied, they should be left unchanged. There are no computations happening and the data type should be left intact.

arviz/tests/external_tests/test_data_pyjags.py
michaelnowotny: It is intended to be a roundtrip test. It starts off with a PyJAGS dictionary, which is converted to InferenceData using `from_pyjags` (in the fixture arviz_inference_data_from_pyjags_samples_dict) and then converted back into a PyJAGS dictionary before being compared to the original data.

OriolAbril: Ohh, I see now, sorry. Maybe it can be renamed to `test_from_pyjags_round` or something like that? 

Also, I think it is even more important to test that the generated inference data has everything it should have. We have a helper function [check_multiple_attrs](https://github.com/arviz-devs/arviz/blob/master/arviz/tests/helpers.py#L169) to ease this kind of checks which is used in the other tests on converter functions

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### New features
 * loo-pit plot. The kde is computed over the data interval (this could be shorter than [0, 1]). The hdi is computed analitically (#1215) 
+* Added support for PyJAGS via the function `from_pyjags` in the module arviz.data.io_pyjags.
```
OriolAbril, line 7:
```suggestion
* Added support for PyJAGS via the function `from_pyjags` in the module arviz.data.io_pyjags. (#1219)
```

CHANGELOG.md
```diff
@@ -13,6 +14,7 @@
 ### Deprecation
 
 ### Documentation
+* A section has been added to the documentation at InferenceDataCookbook.ipynb illustrating the use of ArviZ in conjunction with PyJAGS.
```
OriolAbril, line 17:
```suggestion
* A section has been added to the documentation at InferenceDataCookbook.ipynb illustrating the use of ArviZ in conjunction with PyJAGS. (#1219)
```

arviz/data/io_pyjags.py
```diff
@@ -0,0 +1,323 @@
...
+
+    coords: dict[str, iterable]
+            A dictionary containing the values that are used as index. The key
+            is the name of the dimension, the values are the index values.
+
+    dims: dict[str, List(str)]
+          A mapping from variables to a list of coordinate names for the variable.
+
+    save_warmup: whether or not to return warmup draws if any
+    warmup_iterations: how many iterations are for warmup
```
OriolAbril, line 310: follow numpydoc style

ahartikainen:
If possible, we should add 8schools example for the pyjags too.

arviz/data/io_pyjags.py
michaelnowotny: Ok, it is added.

OriolAbril:
> If possible, we should add 8schools example for the pyjags too.

It has already been added to the notebook: https://github.com/michaelnowotny/arviz/blob/master/doc/notebooks/InferenceDataCookbook.ipynb

I have a couple comments on the notebook.

The import of xarray and setting xarray style as html are not necessary. 

Redefining the `eight_school_data` dictionary should not be necessary either (as well as arviz and numpy import). We should probably move the definition to the first cell along with arviz and numpy imports so it is clear the data is common for all libraries.

For consistency, black-like style should be used. For example:

```
jags_prior_model = pyjags.Model(
    code=eight_school_prior_model_code, 
    data={"J": 8}, 
    chains=4, 
    threads=4,
    chains_per_thread=1
)
```

And finally, I'd change the name of `idata` to `pyjags_data` and remove the plots and summary, simply print the idata object to follow the other libraries example. 
```

File: arviz/data/io_pyjags.py (hunk 16 above)
Comment (michaelnowotny) [reply arviz/data/io_pyjags.py:310]: > If possible, we should add 8schools example for the pyjags too.

A pyjags implementation of the eight schools example is in the inference data cookbook.

Comment (michaelnowotny) [reply arviz/data/io_pyjags.py:310]: I removed the pyjags requirement again as the azure dev-ops system tries to install the package which fails since it requires the unix package jags underneath (which would have to be installed via apt-get install on Debian Linux and derivatives). Here is the error message:
    Package jags was not found in the pkg-config search path.
    Perhaps you should add the directory containing `jags.pc'
    to the PKG_CONFIG_PATH environment variable
    No package 'jags' found
    Error while executing pkg-config: Command '['pkg-config', '--libs', '--cflags', 'jags']' returned non-zero exit status 1.

**Comment by michaelnowotny** [comment]
> > If possible, we should add 8schools example for the pyjags too.
> 
> It has already been added to the notebook: https://github.com/michaelnowotny/arviz/blob/master/doc/notebooks/InferenceDataCookbook.ipynb
> 
> I have a couple comments on the notebook.
> 
> The import of xarray and setting xarray style as html are not necessary.
> 
> Redefining the `eight_school_data` dictionary should not be necessary either (as well as arviz and numpy import). We should probably move the definition to the first cell along with arviz and numpy imports so it is clear the data is common for all libraries.
> 
> For consistency, black-like style should be used. For example:
> 
> ```
> jags_prior_model = pyjags.Model(
>     code=eight_school_prior_model_code, 
>     data={"J": 8}, 
>     chains=4, 
>     threads=4,
>     chains_per_thread=1
> )
> ```
> 
> And finally, I'd change the name of `idata` to `pyjags_data` and remove the plots and summary, simply print the idata object to follow the other libraries example.

I made the suggested changes to the cookbook but was not able to successfully rerun the cmdstan portion, which is now displaying an error message. Could one of you take the notebook and rerun this particular section once everything else is set for release?

File: arviz/tests/external_tests/test_data_pyjags.py (hunk 10 above)
Comment (michaelnowotny) [reply arviz/tests/external_tests/test_data_pyjags.py]: Ok, I took care of it

File: arviz/data/io_pyjags.py (hunk 16 above)
Comment (ahartikainen) [reply arviz/data/io_pyjags.py:310]: can we do 

    conda install -c conda-forge jags

Comment (ahartikainen) [reply arviz/data/io_pyjags.py:310]: Oh wait, I don't know if we use conda or not in devops.

But adding 

    apt-get update
    apt-get install jags

before line 46 in `.azure-pipelines/azure-pipelines-external.yml` should work

Comment (michaelnowotny) [reply arviz/data/io_pyjags.py:310]: Ok, I added the installation via apt-get and re-added pyjags as a requirement.

File: arviz/tests/external_tests/test_data_pyjags.py (hunk 8 above)
Comment (michaelnowotny) [reply arviz/tests/external_tests/test_data_pyjags.py]: Good idea, I incorporated this into the tests.

**Comment by michaelnowotny** [comment]
The tests are passing now.

**Comment by OriolAbril** [comment]
> Could one of you take the notebook and rerun this particular section once everything else is set for release?

We are rerunning all notebooks in  #1217 so this should be no problem.

File: arviz/data/io_pyjags.py [hunk 17]
```diff
@@ -0,0 +1,319 @@
...
```
OriolAbril, line 306: this should follow numpydoc style, otherwise the API page will not be rendered correctly

```suggestion
   save_warmup : bool, optional
       Save warmup iterations InferenceData object. If not defined, use default defined by the rcParams.
    warmup_iterations: int, optional
        Number of warmup iterations
```

arviz/tests/external_tests/test_data_pyjags.py
```diff
@@ -0,0 +1,99 @@
...
+        arviz_dict_from_idata_from_pyjags_dict
+    )
+
+    assert verify_equality_of_numpy_values_dictionaries(
+        pyjags_samples_dict, pyjags_dict_from_arviz_idata
+    )
+
+
+@pytest.mark.parametrize("warmup_iterations", [0, 5])
+def test_inference_data_attrs(pyjags_samples_dict, warmup_iterations: int):
```
OriolAbril, line 90: divide this test into a couple or 3 tests.  

1. `test_from_pyjags` same code as here removing `warmup_iterations` and adding also prior group
2. `test_from_pyjags_warmup` same code as here (prior not necessary), test warmup inclusion. I am not sure what cases are interesting though, warmup_iterations > 0 and save_warmup = True for sure, but warmup_iterations > 0 and save_warmup = False? maybe that should be an error or a warning? and iter=0 + True?
3. `test_from_pyjags_warning/error`? (to be added depending on what we decide regarding the question above)

This here I think is the main unresolved issue.

arviz/data/io_pyjags.py
```diff
@@ -0,0 +1,319 @@
...
+
+    if variables is None:
+        variables = list(pyjags_samples.keys())
+    elif isinstance(variables, str):
+        variables = [variables]
+
+    if not isinstance(variables, Sequence):
+        raise TypeError("variables must be of type Sequence or str")
+
+    variables = tuple(variables)
```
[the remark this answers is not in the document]
michaelnowotny: I reduced the type check to an iterable which includes sets and lists but would be perfectly fine removing it altogether.

I removed the assert and the try except block around the pyjags import

arviz/tests/external_tests/test_data_pyjags.py
michaelnowotny: Not sure I correctly understand the intention behind everything. 

1.) I added prior samples. I am now including all the cases with and without prior/posterior. I am not sure creating a test-case without the argument warmup_iterations is required if we include a case where we set the parameter to zero, since it defaults to zero anyway. Once again, I am not sure if I understood what you meant.

2.) I created all of the combinations:
    - save_warmup True/False/None
    - warmup_iterations: 0, 5
    The combination warmup_iterations=5 and save_warmup=False is legitimate in my opinion. Perhaps the warmup iterations use a lot of memory and the user decides not to keep them. But it is still important for the converter to not include the warmup iterations for subsequent analysis using ArviZ.

   The combination of save_warmup = True and warmup_iterations = False simply returns an empty dictionary. We could also return a dictionary of empty NumPy arrays. 

What do you think?

arviz/tests/external_tests/test_data_pyjags.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+
+    assert verify_equality_of_numpy_values_dictionaries(
+        PYJAGS_POSTERIOR_DICT, pyjags_dict_from_arviz_idata
+    )
+
+
+@pytest.mark.parametrize("posterior", [None, PYJAGS_POSTERIOR_DICT])
+@pytest.mark.parametrize("prior", [None, PYJAGS_PRIOR_DICT])
+@pytest.mark.parametrize("save_warmup", [True, False, None])
```
OriolAbril, line 78:
```suggestion
@pytest.mark.parametrize("save_warmup", [True, False])
```

None is not really necessary as it will be converted to either true or false depending on the rcParam

arviz/tests/external_tests/test_data_pyjags.py
```diff
@@ -0,0 +1,95 @@
...
+        prior=prior,
+        save_warmup=save_warmup,
+        warmup_iterations=warmup_iterations,
+    )
+    test_dict = {}
+    if prior is not None:
+        test_dict["prior"] = ["b", "int"]
+
+    if posterior is not None:
+        test_dict["posterior"] = ["b", "int"]
```
OriolAbril, line 92:
```suggestion
    warmup_prefix = "" if save_warmup and warmup_iterations else "~"
    test_dict = {
        f"{"~" if posterior is None else ""}posterior": ["b", "int"],
        f"{"~" if posterior is None else ""}prior": ["b", "int"],
        f"{warmup_prefix}warmup_posterior": ["b", "int"],
    }
```

This approach will also check the presence of the warmup posterior group and it explicitly checks non presence too (if the prior group were present when it should not, test will fail). I have seen that samples are splitted into warmup and samples not only for posterior but for prior too. Are prior samples not generated straight away via forward sampling that warmup/burnin would be needed?

arviz/tests/external_tests/test_data_pyjags.py
OriolAbril: Thanks for all the explanation, everything makes sense. Returning an empy dictionary is perfect and would make `InferenceData` initializer to not create the group at all (which is what we want, the other converters do not create empty groups on purpose)

arviz/tests/external_tests/test_data_pyjags.py
michaelnowotny: Ok, I incorporated this

arviz/tests/external_tests/test_data_pyjags.py
michaelnowotny: Thanks for following up. Glad we are on the same page.

arviz/tests/external_tests/test_data_pyjags.py
michaelnowotny: Made a few more fixes. Integration tests are running now.

OriolAbril:
Sorry, big fingers on mobile

arviz/data/io_pyjags.py
```diff
@@ -0,0 +1,321 @@
...
+    coords=None,
+    dims=None,
+    save_warmup=None,
+    warmup_iterations: int = 0,
+) -> InferenceData:
+    """
+    Convert PyJAGS posterior samples to an ArviZ inference data object.
+
+    Takes a python dictionary of samples that has been generated by the sample
+    method of a model instance and returns an Arviz inference data object.
```
OriolAbril, line 284:
```suggestion
    method of a model instance and returns an Arviz inference data object. 
    For a usage example read the
    :doc:`Cookbook section on from_pyjags </notebooks/InferenceDataCookbook>`
```

This will make it easier for users to find example usage.

doc/api.rst
```diff
@@ -104,6 +104,7 @@ Data
     from_numpyro
     from_pystan
     from_tfp
+    from pyjags
```
OriolAbril, line 107:
```suggestion
    from_pyjags
```

michaelnowotny:
> LGTM, I don't want to block merging.
> 
> I am still curious about `warmup_prior` existence, does this mean that jags does not use forward sampling to get prior samples? Also why are posterior and prior warmups the same?

Good question. As far as I can see JAGS itself has no functionality to explicitly sample from the prior. However one can simply create a model that leaves out the likelihood (and doesn't use any data). I have included such a separate prior model in the eight schools example in the inference data cookbook. Theoretically, one could directly sample hierarchically from the graph of (hyper)priors. One would not need to resort to Metropolis-Hastings (or even Gibbs) sampling. But I do not know whether JAGS realized that and samples directly. The possibility of warmup iterations seems like more robust move.

More modern MCMC packages such as PyMC3 have features to sample from all kinds of distributions associated with Bayesian analysis including predictive distributions and so it makes sense for ArviZ to support those. JAGS is basically a Gibbs sampler that draws from a target distribution described by a graphical model. It is not a Turing complete language like Stan that lets you generate quantities and solve ODEs while sampling.

PR #1220, opened 2020-06-03
Add finances section

canyon289:
Added a very basic generic stub. Looked at other projects but most don't really have this section filled out. @fonnesbeck gave me some tips, and I looked at matplotlibs and it says to refer to a section that doesn't exist :)

https://github.com/matplotlib/governance/blob/master/governance.md

In general though I think there's two broad paths. We let the council decide how they want to do this, or we try and make strict now so the process is consistent across councils. Please share opinions if you have any.

The outcome of this is that we need to have a general process as well as responsibility holder, It doesnt need to happen before this PR is merged, but it does need to be communicated when we apply for financial sponsorship

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -93,7 +93,7 @@ Council Members will have the responsibility of
 * Make decisions when regular community discussion doesn’t produce consensus on an issue in a reasonable time frame.
 * Make decisions about strategic collaborations with other organizations or individuals.
 * Make decisions about the overall scope, vision and direction of the project.
-* Developing finding sources or spending money
+* Developing funding sources or spending money.
 
 The council may choose to delegate these responsibilities to sub committees. If so they must update this document to make the delegation clear.
 
@@ -205,4 +205,3 @@ can make the determination for acceptance with a process of their choosing.
 #### Core Contributor Responsibilities
 * Enforce code of conduct
 * Maintain a check against Council
-
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -93,11 +93,12 @@ Council Members will have the responsibility of
 * Make decisions when regular community discussion doesn’t produce consensus on an issue in a reasonable time frame.
 * Make decisions about strategic collaborations with other organizations or individuals.
 * Make decisions about the overall scope, vision and direction of the project.
-* Developing funding sources or spending money.
+* Developing funding sources
+* Deciding how to disburse funds
 
 The council may choose to delegate these responsibilities to sub-committees. If so, Council members must update this document to make the delegation clear.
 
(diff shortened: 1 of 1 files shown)


GOVERNANCE.md
```diff
@@ -93,6 +93,9 @@ Council Members will have the responsibility of
 * Make decisions when regular community discussion doesn’t produce consensus on an issue in a reasonable time frame.
 * Make decisions about strategic collaborations with other organizations or individuals.
 * Make decisions about the overall scope, vision and direction of the project.
+* Developing finding sources or spending money
```
AlexAndorra, line 96:
```suggestion
* Developing funding sources or spending money.
```

GOVERNANCE.md
```diff
@@ -93,6 +93,9 @@ Council Members will have the responsibility of
...
+
+The council may choose to delegate these responsibilities to sub committees. If so they must update this document to make the delegation clear.
```
AlexAndorra, line 98:
```suggestion
The council may choose to delegate these responsibilities to sub-committees. If so, Council members must update this document to make the delegation clear.
```

aloctavodia:
I also tried to find information related to this in other project without luck :-( I agree with @AlexAndorra  that at this point we can keep this vague and high-level.

canyon289:
Thanks @AlexAndorra for edting directly. The only thing I think I want to avoid is where the council is biased to give themselves money. Maybe the most we dictate is that there is a sub committee composed of at least 2 people not on the council that makes the proposals, and council approves. That way its separates power. Havent thought through this entirely but throwing the idea out for now

AlexAndorra:
Good point -- this also ensure information about finances and projects go around smoothly in the community and doesn't stay at the Council level. 
You can even do the other way around: the Council makes the proposal, and the sub-committee composed of at least 2 people not on the Council approves. That way, the Council can't filibuster until it gets its way or resort to _quid pro quo_.

aloctavodia:
I like the idea of the Council making the proposals and someone else approving. But instead of a sub-committee what about every core-contributor. This could be done over slack with a simple poll... or just asking if someone is opposed.

GOVERNANCE.md
```diff
@@ -93,8 +93,12 @@ Council Members will have the responsibility of
 * Make decisions when regular community discussion doesn’t produce consensus on an issue in a reasonable time frame.
 * Make decisions about strategic collaborations with other organizations or individuals.
 * Make decisions about the overall scope, vision and direction of the project.
+* Developing funding sources
+* Deciding how to disburse funds
```
aloctavodia, line 97: Should we add something about the council needing to consult the core developers (basically my comment on slack)?

canyon289: Added

canyon289:
Merging, We can keep editing until Council is formed

PR #1222, opened 2020-06-04
Draft plot_bpv (bayesian p-value)

aloctavodia:
This is a new kind of posterior predictive check plot, focused toward p_values. This a super early draft (the code is still a mess). I  have doubts about the arguments names, suggestions are more than welcome. What do you think of these examples. We could also add other like ecdf, but I prefer to concentrate first on getting these right.

BTW, the first too are the same as loo-pit, but using the data instead of the IS weights. 

A few examples
kind="u_value", reference="analytical". We want a uniform distribution over the [0, 1] interval
u_value_analytical_None

kind="u_value", reference="samples" We want a uniform distribution over the [0, 1] interval
u_value_samples_None

kind="p_value", reference="analytical", t_stat=None. We want a symmetric distribution centered at 0.5 
p_value_analytical_None

kind="p_value", reference="samples", t_stat=None We want a symmetric distribution centered at 0.5
p_value_samples_None

kind="t_stat", reference=None, t_stat="median", other values are "mean", "std", quantiles can be passed as string of number between 0 and 1. Finally the user can pass an arbitrary function. The density is the distribution of sampled T_stat. The dot represent the mean of the observed T_stat. The legend shows the bayesian p_value (bpv)
t_stat_None_median


aloctavodia:
The basic elements are in place. So this is ready for review.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * `plot_trace` now supports multiple aesthetics to identify chain and variable
   shape and support matplotlib aliases (#1253)
 * `plot_hdi` can now take already computed HDI values (#1241)
+* `plot_bpv`. A new plot intented to plot Bayesian p-values (#1222)
```
ColCarroll, line 13:
```suggestion
* `plot_bpv`. A new plot for Bayesian p-values (#1222)
```

arviz/plots/bpvplot.py
```diff
@@ -0,0 +1,280 @@
...
+):
+    """
+    Plot Bayesian p-value for observed data and Posterior/Prior predictive.
+
+    Parameters
+    ----------
+    data : az.InferenceData object
+        InferenceData object containing the observed and posterior/prior predictive data.
+    kind : str
+        Type of plot to display (u_value, p_value, t_stat). Defaults to u_value.
```
ColCarroll, line 49: Can you add what `u_value`, `p_value`, `t_stat` are (or references)

arviz/plots/bpvplot.py
```diff
@@ -0,0 +1,280 @@
...
+    t_stat : str, float, or callable
+        T statistics to compute from the observations and predictive distributions. Allowed strings
+        are "mean", "median" or "std". Defaults to "median". Alternative a quantile can be passed
+        as a float (or str) in the interval (0, 1). Finally a user defined function is also
+        acepted, see examples section for details.
+    bpv : bool
+        If True (default) add the bayesian p_value to the legend when kind = t_stat.
+    mean : bool
```
ColCarroll, line 57: Not a strong feeling, but maybe change to `plot_mean`?

arviz/plots/bpvplot.py
```diff
@@ -0,0 +1,280 @@
...
+        Whether or not to plot the mean T statistic. Defaults to True.
+    reference : str
+        How to compute the distributions used as reference for u_values or p_values. Allowed values
+        are "analytical" (default) and "samples". Use `None` to do not plot any reference.
```
ColCarroll, line 61:
```suggestion
        are "analytical" (default) and "samples". Use `None` to do not plot any reference. Defaults to "samples".
```

arviz/plots/bpvplot.py
```diff
@@ -0,0 +1,280 @@
...
+    n_ref : int, optional
+        Number of reference distributions to sample when `reference=samples`
```
ColCarroll, line 63:
```suggestion
        Number of reference distributions to sample when `reference=samples`. Defaults to 100.
```

arviz/plots/plot_utils.py
```diff
@@ -695,3 +695,25 @@ def _dealiase_sel_kwargs(kwargs, prop_dict, idx):
             {prop: props[idx] for prop, props in prop_dict.items()}, "plot"
         ),
     }
+
+
+def is_valid_quantile(value):
+    """Check if value is a number between 0 and 1."""
+    try:
+        value = float(value)
+        return 0 < value < 1
```
ColCarroll, line 704: I think you want this return _after_ the except, since you are watching for a ValueError while casting to float, right?

examples/bokeh/bokeh_plot_bpv_tstat.py
```diff
@@ -0,0 +1,10 @@
+"""
+Bayesian p-value Posterior plot
+===============================
+
+_thumb: .6, .5
+"""
+import arviz as az
+
+data = az.load_arviz_data("regression1d")
+az.plot_bpv(data, kind="t_stat", t_stat="0.5", backend="bokeh")
```
ColCarroll, line 10:
```suggestion
az.plot_bpv(data, kind="t_stat", t_stat=0.5, backend="bokeh")
```

examples/matplotlib/mpl_plot_bpv_tstat.py
```diff
@@ -0,0 +1,15 @@
...
+
+_thumb: .6, .5
+"""
+import matplotlib.pyplot as plt
+import arviz as az
+
+az.style.use("arviz-darkgrid")
+
+data = az.load_arviz_data("regression1d")
+az.plot_bpv(data, kind="t_stat", t_stat="0.5")
```
ColCarroll, line 13:
```suggestion
az.plot_bpv(data, kind="t_stat", t_stat=0.5)
```

examples/bokeh/bokeh_plot_bpv_tstat.py
```diff
@@ -0,0 +1,10 @@
+"""
+Bayesian p-value Posterior plot
```
ColCarroll, line 2: Should this have a different name from the other, to indicate it is using a T stat?

examples/matplotlib/mpl_plot_bpv_tstat.py
```diff
@@ -0,0 +1,15 @@
+"""
+Bayesian p-value Posterior plot
```
ColCarroll, line 2: Same question on naming here

examples/bokeh/bokeh_plot_bpv_tstat.py
aloctavodia: Passing a string is actually valid

arviz/plots/plot_utils.py
aloctavodia: I think this works ok.

aloctavodia:
maybe @ahartikainen wants to improve the bokeh code in a future PR ;-)

arviz/plots/plot_utils.py
canyon289: This will work as written

arviz/plots/bpvplot.py
```diff
@@ -0,0 +1,294 @@
...
+    show=None,
+):
+    """
+    Plot Bayesian p-value for observed data and Posterior/Prior predictive.
+
+    Parameters
+    ----------
+    data : az.InferenceData object
+        InferenceData object containing the observed and posterior/prior predictive data.
+    kind : str
```
OriolAbril, line 48: Nit: I would modify this following numpydoc advise:

> When a parameter can only assume one of a fixed set of values, those values can be listed in braces, with the default appearing first:
>
>     order : {'C', 'F', 'A'}
>        Description of `order`.

arviz/plots/bpvplot.py
```diff
@@ -0,0 +1,294 @@
...
+    t_stat : str, float, or callable
+        T statistics to compute from the observations and predictive distributions. Allowed strings
+        are "mean", "median" or "std". Defaults to "median". Alternative a quantile can be passed
+        as a float (or str) in the interval (0, 1). Finally a user defined function is also
+        acepted, see examples section for details.
+    bpv : bool
+        If True (default) add the bayesian p_value to the legend when kind = t_stat.
+    plot_mean : bool
+        Whether or not to plot the mean T statistic. Defaults to True.
+    reference : str
```
OriolAbril, line 68: same as above

arviz/plots/bpvplot.py
```diff
@@ -0,0 +1,294 @@
...
+        How to compute the distributions used as reference for u_values or p_values. Allowed values
+        are "analytical" (default) and "samples". Use `None` to do not plot any reference.
+        Defaults to "samples".
+    n_ref : int, optional
+        Number of reference distributions to sample when `reference=samples`. Defaults to 100.
+    hdi_prob: float, optional
+        Probability for the highest density interval for the analytical reference distribution when
+        computing u_values. Should be in the interval (0, 1]. Defaults to
+        0.94.
```
OriolAbril, line 77: defaults to `stats.hdi_prob` rcParam

arviz/plots/bpvplot.py
```diff
@@ -0,0 +1,294 @@
...
+        InferenceData object containing the observed and posterior/prior predictive data.
+    kind : str
+        Type of plot to display ("p_value", "u_value", "t_stat"). Defaults to u_value.
+        For "p_value" we compute p := p(y* ≤ y | y). This is the probability of the data y being
+        larger or equal than the predicted data y*. The ideal value is 0.5 (half the predictions
+        below and half above the data).
+        For "u_value" we compute pi := p(yi* ≤ yi | y). i.e. like a p_value but per observation yi.
+        This is also known as marginal p_value. The ideal distribution is uniform. This is similar
+        to the LOO-pit calculation/plot, the difference is than in LOO-pit plot we compute
+        pi = p(yi* r ≤ yi | y-i ), where y-i, is all other data except yi.
```
OriolAbril, line 56: I think there is one extra `r` before the lower or equal than sign

arviz/plots/bpvplot.py
```diff
@@ -0,0 +1,294 @@
...
+        its own array of plot areas (and return it).
+    backend : str, optional
+        Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
+    plot_ref_kwargs :  dict, optional
+        Extra keyword arguments to control how reference is represented. Passed to `plt.plot` or
+        `plt.axhspan`(when `kind=u_value` and `reference=analytical`).
+    backend_kwargs : bool, optional
+        These are kwargs specific to the backend being used. For additional documentation
+        check the plotting method of the backend.
+    group : {"prior", "posterior"}, optional
```
OriolAbril, line 129: along the idea of comments above, this should be inverted to show the default first

arviz/plots/bpvplot.py
```diff
@@ -0,0 +1,294 @@
...
+        List of dimensions to flatten in posterior_predictive/prior_predictive. Only flattens
+        across the coordinates specified in the coords argument. Defaults to flattening all
+        of the dimensions. Dimensions should match flatten excluding dimensions for data_pairs
+        parameters. If flatten is defined and flatten_pp is None, then `flatten_pp=flatten`.
+    legend : bool
+        Add legend to figure. By default True.
+    ax : numpy array-like of matplotlib axes or bokeh figures, optional
+        A 2D array of locations into which to plot the densities. If not supplied, Arviz will create
+        its own array of plot areas (and return it).
+    backend : str, optional
```
OriolAbril, line 121: same as above

arviz/plots/bpvplot.py
```diff
@@ -0,0 +1,294 @@
+"""Bayesian p-value Posterior/Prior predictive plot."""
+import numpy as np
+from matplotlib.colors import to_hex
```
OriolAbril, line 3: we can use `az.plots.plot_utils.vectorized_to_hex`. It basically calls `matplotlib.colors.to_hex` but works with lists of colors too (not relevant here) and we could at some point (far future) make a version of the function not requiring matplotlib, i.e. try importing from matplotlib and if not possible use bokeh version (as far as I know bokeh has no function that does this yet)

arviz/plots/plot_utils.py
```diff
@@ -695,3 +695,25 @@ def _dealiase_sel_kwargs(kwargs, prop_dict, idx):
...
+def sample_reference_distribution(dist, shape):
+    """Generate samples from a scipy distribution with a given shape."""
+    x_ss = []
+    densities = []
+    dist_rvs = dist.rvs(size=shape)
+    for idx in range(shape[1]):
+        density, xmin, xmax = _fast_kde(dist_rvs[:, idx])
+        x_s = np.linspace(xmin, xmax, len(density))
+        x_ss.append(x_s)
+        densities.append(density)
```
OriolAbril, line 718: we should check memory performance and see if its possible to preallocate. Not sure how it will perform, but specially the append looks suspicious.

arviz/tests/base_tests/test_plots_bokeh.py
```diff
@@ -966,3 +967,18 @@ def test_plot_rank(models, kwargs):
 def test_plot_dist_comparison_warn(models):
     with pytest.raises(NotImplementedError, match="The bokeh backend.+Use matplotlib bakend."):
         plot_dist_comparison(models.model_1, backend="bokeh")
+
+
```
OriolAbril, line 971: I think everything will work with multidimensional data (that is `chain, draw, *shape` with `shape` being multidimensional) thanks to the skip_dims and careful reshaping, however, it would be great to test on multidimensional models too. Should be the same as test below but using `multidim_models` fixture instead, probably a couple cases are enough for multidim.

arviz/plots/bpvplot.py
```diff
@@ -0,0 +1,294 @@
...
+        For "p_value" we compute p := p(y* ≤ y | y). This is the probability of the data y being
+        larger or equal than the predicted data y*. The ideal value is 0.5 (half the predictions
+        below and half above the data).
+        For "u_value" we compute pi := p(yi* ≤ yi | y). i.e. like a p_value but per observation yi.
+        This is also known as marginal p_value. The ideal distribution is uniform. This is similar
+        to the LOO-pit calculation/plot, the difference is than in LOO-pit plot we compute
+        pi = p(yi* r ≤ yi | y-i ), where y-i, is all other data except yi.
+        For "t_stat" we compute := p(T(y)* ≤ T(y) | y) where T is any T statistic. See t_stat
+        argument below for details of available options.
+    t_stat : str, float, or callable
```
OriolAbril, line 59: How about adding a `"identity"` option too to encourage users to calculate T statistics manually and store them in idata object and then use these new variables as variable names with `t_stat="identity"`? 

I know it is already possible to do using `t_stat=lambda x: x`, hence the _encourage_ above. This would be helpful if T statistic function was expensive to calculate or for convenience to use T statistic with both `plot_bpv` and `loo_pit`

arviz/plots/plot_utils.py
aloctavodia: my memory is foggy but I think that was my first attempt, but then I changed because the size of the grid used by _fast_kde is not fixed and this make cause trouble. That should be fixed with the new upcoming kde. Nevertheless I agree this is something we need to revisit.

aloctavodia:
Thanks for the comments @OriolAbril!

PR #1226, opened 2020-06-07
Add core contributors

canyon289:
Adding core contributors per process detailed in https://github.com/arviz-devs/arviz/blob/master/elections/ArviZ_2020.md

If there is any comment or question make a comment in this PR

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -201,6 +201,7 @@ Current Core Contributors can nominate candidates for consideration by the counc
 can make the determination for acceptance with a process of their choosing.
 
 #### Current Core Contributors
+* Alex Andorra (@AlexAndorra)
 * Osvaldo Martin (@aloctavodia)
 * Oriol Abril Pla (@OriolAbril) 
 * Du Phan (@fehiepsi)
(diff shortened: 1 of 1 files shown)


GOVERNANCE.md
```diff
@@ -201,7 +201,14 @@ Current Core Contributors can nominate candidates for consideration by the counc
...
 
 #### Current Core Contributors
-* Will be updated with Core Contributor list during first election
+* Alex Andorra (@AlexAndorra)
+* Colin Carroll (@ColCarroll)
+* Ari Hartikainen (@ahartikainen)
+* Ravin Kumar (@canyon289)
+* Osvaldo Martin (@aloctavodia)
+* Du Phan (@fehiepsi)
+* Oriol Abril Pla (@OriolAbril)
```
OriolAbril, line 210: I did not notice this was alphabetical order at first, we have two last names and I am still young enough to try and keep them both :see_no_evil:. It will probably be better to reduce confusion to use `Oriol Abril-Pla`

canyon289:
Alright no dissents so merging it all. Congratulations everyone

PR #757, opened 2019-07-18
WIP Import coords and dims from pymc3

aseyboldt:
*Depends on https://github.com/pymc-devs/pymc3/pull/3551 *
This allows arviz to read dims and coords from pymc3 when they are defined.

PR #1228, opened 2020-06-08
Automatically take coords and dims from PyMC3 model

michaelosthege:
This PR supersedes #757 (I can't push to Adrians branch) and resolves conflicts with master. (Rebases over almost a year of commits!)

When https://github.com/pymc-devs/pymc3/pull/3551 is merged, this PR enables ArviZ to automatically take `coords` and `dims` from the PyMC3 model.


OriolAbril:
Do you prefer to wait until PR is merged in PyMC3 and add tests here or merge this now and add test in follow-up PR once pymc3 one has been merged?

AlexAndorra:
I think we can wait and tests here -- the PyMC PR should be ready soon

OriolAbril:
is the new functionality only for dims or for coords too? 

LGTM, I think we can merge like this and open an issue so we remember to extend the test to coords too.

michaelosthege:
@OriolAbril the way how I understand it is that dims/coords go hand in hand. The keys are the dims, the values to coords.

OriolAbril:
Ohh, I read too past and missed the `pm.Model(coords=coords)`

kyleabeauchamp:
I think the current pymc3  <> arviz coords integration is broken in the current pymc3 release because it is using an arviz version that does not have this code present.

See https://github.com/pymc-devs/pymc3/pull/3977 and https://github.com/pymc-devs/pymc3/issues/3976 for details

michaelosthege:
> I think the current pymc3  <> arviz coords integration is broken in the current pymc3 release because it is using an arviz version that does not have this code present.
> See pymc-devs/pymc3#3977 and pymc-devs/pymc3#3976 for details

Broken in the sense that the feature is not there, right? ArviZ 0.8.3 doesn't automatically take the coords, because this PR came in after 0.8.3.
The PyMC3 example notebooks (which are not yet rendered for https://docs.pymc.io) were run with ArviZ master, which may cause some confusion.

The real issue here is that we're still sitting on some technical debt in both projects when it comes to releasing & doc-building.

PR #1229, opened 2020-06-08
[WIP] Start adding Dask support

nitishp25:
## Description

Start adding Dask compatibility starting from `diagnostics`/`stats` ufuncs.
## Checklist

- [ ] Code style correct (follows pylint and black guidelines)

### DIFF
--- a/arviz/stats/diagnostics.py
+++ b/arviz/stats/diagnostics.py
@@ -63,7 +63,7 @@ def bfmi(data):
 
 
 def ess(
-    data, *, var_names=None, method="bulk", relative=False, prob=None, **kwargs,
+    data, *, var_names=None, method="bulk", relative=False, prob=None, dask_kwargs=None,
 ):
     r"""Calculate estimate of the effective sample size (ess).
 
@@ -95,8 +95,8 @@ def ess(
         `ress = ess / n`
     prob : float, or tuple of two floats, optional
         probability value for "tail", "quantile" or "local" ess functions.
-    **kwargs : kwargs passed to `xarray.apply_ufunc`, optional
-        See :function:`xarray:xarray.apply_ufunc` for accepted values.
+    dask_kwargs : dict, optional
+        Dask related kwargs passed to :func:`~arviz.wrap_xarray_ufunc`.
 
     Returns
     -------
@@ -197,11 +197,11 @@ def ess(
     ufunc_kwargs = {"ravel": False}
     func_kwargs = {"relative": relative} if prob is None else {"prob": prob, "relative": relative}
     return _wrap_xarray_ufunc(
-        ess_func, dataset, ufunc_kwargs=ufunc_kwargs, func_kwargs=func_kwargs, **kwargs
+        ess_func, dataset, ufunc_kwargs=ufunc_kwargs, func_kwargs=func_kwargs, dask_kwargs=dask_kwargs
     )
 
 
-def rhat(data, *, var_names=None, method="rank", **kwargs):
+def rhat(data, *, var_names=None, method="rank", dask_kwargs=None):
     r"""Compute estimate of rank normalized splitR-hat for a set of traces.
 
     The rank normalized R-hat diagnostic tests for lack of convergence by comparing the variance
@@ -228,8 +228,8 @@ def rhat(data, *, var_names=None, method="rank", **kwargs):
         - "folded"
         - "z_scale"
         - "identity"
-    **kwargs : kwargs passed to `xarray.apply_ufunc`, optional
-        See :function:`xarray:xarray.apply_ufunc` for accepted values.
+    dask_kwargs : dict, optional
+        Dask related kwargs passed to :func:`~arviz.wrap_xarray_ufunc`.
 
     Returns
     -------
@@ -308,11 +308,11 @@ def rhat(data, *, var_names=None, method="rank", **kwargs):
     ufunc_kwargs = {"ravel": False}
     func_kwargs = {}
     return _wrap_xarray_ufunc(
-        rhat_func, dataset, ufunc_kwargs=ufunc_kwargs, func_kwargs=func_kwargs, **kwargs
+        rhat_func, dataset, ufunc_kwargs=ufunc_kwargs, func_kwargs=func_kwargs, dask_kwargs=dask_kwargs
     )
 
 
-def mcse(data, *, var_names=None, method="mean", prob=None, **kwargs):
+def mcse(data, *, var_names=None, method="mean", prob=None, dask_kwargs=None):
     """Calculate Markov Chain Standard Error statistic.
 
     Parameters
@@ -333,8 +333,8 @@ def mcse(data, *, var_names=None, method="mean", prob=None, **kwargs):
 
     prob : float
         Quantile information.
-    **kwargs : kwargs passed to `xarray.apply_ufunc`, optional
-        See :function:`xarray:xarray.apply_ufunc` for accepted values.
+    dask_kwargs : dict, optional
+        Dask related kwargs passed to :func:`~arviz.wrap_xarray_ufunc`.
 
     Returns
     -------
@@ -397,7 +397,7 @@ def mcse(data, *, var_names=None, method="mean", prob=None, **kwargs):
     ufunc_kwargs = {"ravel": False}
     func_kwargs = {} if prob is None else {"prob": prob}
     return _wrap_xarray_ufunc(
-        mcse_func, dataset, ufunc_kwargs=ufunc_kwargs, func_kwargs=func_kwargs, **kwargs
+        mcse_func, dataset, ufunc_kwargs=ufunc_kwargs, func_kwargs=func_kwargs, dask_kwargs=dask_kwargs
     )
 
 
--- a/arviz/stats/diagnostics.py
+++ b/arviz/stats/diagnostics.py
@@ -63,14 +63,7 @@ def bfmi(data):
 
 
 def ess(
-    data,
-    *,
-    var_names=None,
-    method="bulk",
-    relative=False,
-    prob=None,
-    dask="forbidden",
-    output_dtypes=None
+    data, *, var_names=None, method="bulk", relative=False, prob=None, **kwargs,
 ):
     r"""Calculate estimate of the effective sample size (ess).
 
@@ -102,11 +95,8 @@ def ess(
         `ress = ess / n`
     prob : float, or tuple of two floats, optional
         probability value for "tail", "quantile" or "local" ess functions.
-    dask : "forbidden" or "parallelized"
-        Defaults to "forbidden". Use "parallelized" if passing dask arrays to enable parallelization.
-        Must also provide `output_dtypes` argument if dask="parallelized".
-    output_dtypes : list of dtypes
-        Only used if dask=’parallelized’.
+    **kwargs : kwargs passed to `xarray.apply_ufunc`, optional
+        See :function:`xarray:xarray.apply_ufunc` for accepted values.
 
     Returns
     -------
@@ -207,11 +197,11 @@ def ess(
     ufunc_kwargs = {"ravel": False}
     func_kwargs = {"relative": relative} if prob is None else {"prob": prob, "relative": relative}
     return _wrap_xarray_ufunc(
-        ess_func, dataset, dask, output_dtypes, ufunc_kwargs=ufunc_kwargs, func_kwargs=func_kwargs
+        ess_func, dataset, ufunc_kwargs=ufunc_kwargs, func_kwargs=func_kwargs, **kwargs
     )
 
 
-def rhat(data, *, var_names=None, method="rank", dask="forbidden", output_dtypes=None):
+def rhat(data, *, var_names=None, method="rank", **kwargs):
     r"""Compute estimate of rank normalized splitR-hat for a set of traces.
 
     The rank normalized R-hat diagnostic tests for lack of convergence by comparing the variance
@@ -238,11 +228,8 @@ def rhat(data, *, var_names=None, method="rank", dask="forbidden", output_dtypes
         - "folded"
         - "z_scale"
         - "identity"
-    dask : "forbidden" or "parallelized"
-        Defaults to "forbidden". Use "parallelized" if passing dask arrays to enable parallelization.
-        Must also provide `output_dtypes` argument if dask="parallelized".
-    output_dtypes : list of dtypes
-        Only used if dask=’parallelized’.
+    **kwargs : kwargs passed to `xarray.apply_ufunc`, optional
+        See :function:`xarray:xarray.apply_ufunc` for accepted values.
 
     Returns
     -------
@@ -321,11 +308,11 @@ def rhat(data, *, var_names=None, method="rank", dask="forbidden", output_dtypes
     ufunc_kwargs = {"ravel": False}
     func_kwargs = {}
     return _wrap_xarray_ufunc(
-        rhat_func, dataset, dask, output_dtypes, ufunc_kwargs=ufunc_kwargs, func_kwargs=func_kwargs
+        rhat_func, dataset, ufunc_kwargs=ufunc_kwargs, func_kwargs=func_kwargs, **kwargs
     )
 
 
-def mcse(data, *, var_names=None, method="mean", prob=None, dask="forbidden", output_dtypes=None):
+def mcse(data, *, var_names=None, method="mean", prob=None, **kwargs):
     """Calculate Markov Chain Standard Error statistic.
 
     Parameters
@@ -346,11 +333,8 @@ def mcse(data, *, var_names=None, method="mean", prob=None, dask="forbidden", ou
 
     prob : float
         Quantile information.
-    dask : "forbidden" or "parallelized"
-        Defaults to "forbidden". Use "parallelized" if passing dask arrays to enable parallelization.
-        Must also provide `output_dtypes` argument if dask="parallelized".
-    output_dtypes : list of dtypes
-        Only used if dask=’parallelized’.
+    **kwargs : kwargs passed to `xarray.apply_ufunc`, optional
+        See :function:`xarray:xarray.apply_ufunc` for accepted values.
 
     Returns
     -------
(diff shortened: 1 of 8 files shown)


arviz/stats/diagnostics.py
```diff
@@ -65,7 +65,16 @@ def bfmi(data):
...
-def ess(data, *, var_names=None, method="bulk", relative=False, prob=None):
+def ess(
+    data,
+    *,
+    var_names=None,
+    method="bulk",
+    relative=False,
+    prob=None,
+    dask="forbidden",
+    output_dtypes=None
```
OriolAbril, line 76: it would be better to use `**kwargs`, this way we don't have to keep the docstrings and defaults up to date with xarray (and we can directly link to their docs: https://arviz-devs.github.io/arviz/developer_guide.html#documentation) and if extra kwargs were needed (or if new arguments were added by xarray) everything would still work

canyon289: Agreed

nitishp25: I thought of that too but then got confused by the `kwargs` argument used by `xr.apply_ufunc`, now I know the difference between `kwargs` and `**kwargs` :sweat_smile:

arviz/stats/diagnostics.py
```diff
@@ -65,7 +65,9 @@ def bfmi(data):
     return _bfmi(dataset.energy)
 
 
-def ess(data, *, var_names=None, method="bulk", relative=False, prob=None):
+def ess(
```
canyon289, line 68: Instead of setting kwargs can we use decorators that signify dask? Otherwise kwargs can be anything and itll be hard to trace whats going on

canyon289: Btw I know Oriol said use **kwargs so this is a slightly different change in direction but since this is a major architectural change want to explore a couple options before we settle on one, and after seeing this one realize some of the downsides

nitishp25: I see. So for that decorator we could have `dask.enable(**kwargs)` and `dask.disable` in the `dask` class to control just like numba and also take dask kwargs from user to use for the decorator?

canyon289: That would be ideal because its much more obvious, but its just an idea I dont know for sure if it will work. Would welcome feedback on the feasability

nitishp25: I've added a sample class and decorator for `wrap_xarray_ufunc`. This is just to see if this is what we want :)

canyon289:
Thanks @nitishp25 I'll review over weekend which is 48 hours from now. My preliminary feedback is to change `**kwargs` to `dash_kwargs` so its clear which arguments are coming from dask wrapper

nitishp25:
> Thanks @nitishp25 I'll review over weekend which is 48 hours from now. My preliminary feedback is to change `**kwargs` to `dask_kwargs` so its clear which arguments are coming from dask wrapper,

I've separated `dask_kwargs` for `wrap_xarray_ufunc` and should we still keep the regular `**kwargs` for stats/diagnostics functions if the user wants to pass them to `xr.apply_ufunc` or are they unnecessary?

arviz/stats/diagnostics.py
```diff
@@ -195,10 +199,12 @@ def ess(data, *, var_names=None, method="bulk", relative=False, prob=None):
...
     ufunc_kwargs = {"ravel": False}
     func_kwargs = {"relative": relative} if prob is None else {"prob": prob, "relative": relative}
-    return _wrap_xarray_ufunc(ess_func, dataset, ufunc_kwargs=ufunc_kwargs, func_kwargs=func_kwargs)
+    return _wrap_xarray_ufunc(
+        ess_func, dataset, ufunc_kwargs=ufunc_kwargs, func_kwargs=func_kwargs, **kwargs
+    )
 
 
-def rhat(data, *, var_names=None, method="rank"):
+def rhat(data, *, var_names=None, method="rank", **kwargs):
```
canyon289, line 207: Per your previous question I would not include `**kwargs` here. The reason is that previously the API was easy to introspect, but now anything could be in kwargs and the flexibility comes at the cost of readability. If this namspace is used solely for ask I would say either add the dask keyword args explicitly, or if the dask args are not known ahead of time change this to `dask_kwargs` so throughout the method is easy to see what is related to ask.

Let me know if this seems crazy, happy to take feedback!

OriolAbril: I would argue against adding Dask kwargs explicitly as it is still an area very actively developed in both xarray and dask. If you look at the [blame](https://github.com/pydata/xarray/blame/master/xarray/core/computation.py#L771) of dask related kwargs, you'll see the `meta` kwarg was added only 5 months ago for example. And yet, `apply_ufunc` with dask does not support multiple outputs with dask (I think) but will do so in the near future which could add even another dask related kwarg. 

Adding them explicitly would force us to keep up to date with xarray modifications (not only of new kwargs but also when defaults are modified).

OriolAbril: I think the ideal behaviour would be having the default `dask_kwargs` in `Dask` class but exposing `dask_kwargs` too to user functions. Using `dask_kwargs` instead of `**kwargs` I think is a good idea because even though these kwargs can theoretically be used to pass non dask parameters, it is a bad idea to pass non dask parameters. `dask_kwargs` helps getting this accross.

Regarding documentation (so things don't end up being too obscure) I think the ideal would be linking to `wrap_xarray_ufunc` or `xarray.apply_ufunc` and also to a guide about dask usage. The same way `xarray.DataArray.sel` has examples in API docs but still links to the Indexing section of the User Guide.

canyon289: Agree

canyon289:
> > Thanks @nitishp25 I'll review over weekend which is 48 hours from now. My preliminary feedback is to change `**kwargs` to `dask_kwargs` so its clear which arguments are coming from dask wrapper,
> 
> I've separated `dask_kwargs` for `wrap_xarray_ufunc` and should we still keep the regular `**kwargs` for stats/diagnostics functions if the user wants to pass them to `xr.apply_ufunc` or are they unnecessary?

Unncessary and sometimes harmful because it becomes impossible to know what the api actually is. like @OriolAbril was mentioning above, lets remove the generic kwargs entirely so its clear what goes into the method

nitishp25:
@canyon289 @OriolAbril I think I have now made the changes according to the comments, please have a look :)
If everything's fine I can start adding tests, benchmarks and user guide.

nitishp25:
> Looking good, I do like the decorator approach and I think it could even allow further customization such as `numba.guvectorize` :thinking: (definitely in another PR though).

That sounds interesting to me too, given the difficulties I faced while trying to implement `dask.delayed` in `make_ufunc` :/

canyon289:
@nitishp25 Glanced over this and its looking good. Give me a couple more days to do a full review. Apologies on the delay

canyon289:
Just scoping to the code here it looks good, except I don't see Dask added to the requirements anywhere. Should it be added?

In terms of the idea of dask itself, now that the basic architecture is in have we run a speed test to ensure the benefit's we were hoping for materialize? I believe the two were faster calculation of statistics. Has this been proven even if just locally?

The reasons I ask is that, the idea of Dask is good. So now that the wrapper is written it would be good to do a preliminary test to ensure the benefits were looking for show up in a "scrappy" test. This way it proves that the additional complexity added to the codebase in this PR is worth the squeeze.

Hope that makes sense. Thank you for going through the iteration. I like where you got this!

OriolAbril:
I expect some speed-up thanks to Dask with the current changes, however the most critical improvement is on memory usage. It would be interesting to add a couple of benchmarks on this, I think asv allows memory profiling too. These could be added in another PR though if prefered.

> Just scoping to the code here it looks good, except I don't see Dask added to the requirements anywhere. Should it be added?

True, dask should be added to https://github.com/arviz-devs/arviz/blob/master/requirements-optional.txt

nitishp25:
Yes, I'll add the tests in this PR and the speed and memory benchmarks in the next PR

canyon289:
> I expect some speed-up thanks to Dask with the current changes, however the most critical improvement is on memory usage. It would be interesting to add a couple of benchmarks on this, I think asv allows memory profiling too. These could be added in another PR though if prefered.
> 
> > Just scoping to the code here it looks good, except I don't see Dask added to the requirements anywhere. Should it be added?
> 
> True, dask should be added to https://github.com/arviz-devs/arviz/blob/master/requirements-optional.txt

@OriolAbril @nitishp25 I dont think we need a full pr, but I would like to see at least a local test or an ad hoc notebook confirming the hypothesis that dask will improve memory usage before we merge. This way we can confirm the extra complexity and maintenance burden that will be added by dask integration will worth it.

Doesnt have to be anything crazy, just a couple screenshots in this PR would be sufficient

ahartikainen:
what means memory use?

We want dask to run stuff parallel.

100k parameters --> ess / rhat or maybe summary

OriolAbril:
> what means memory use?
> 
> We want dask to run stuff parallel.
> 
> 100k parameters --> ess / rhat or maybe summary

We can use Dask to tackle two challenges: computation on large datasets where fitting them into memory and operating with them is challenging and parallelizing functions. (I copied this from several messages in slack so some things may not be completely relevant here, I only skimmed to check everything made sense globally)

**Regarging large data:** I think that the best way to tackle large data is allowing `dask="parallelized"` in stats, diagnostics and so on. In that case, the ufuncs passed to `apply_ufunc` must be single threaded (which is our current implementation already) and would therefore not be too much work. 

There are still several challenges to overcome such as getting [`_multichain_statistics`](https://github.com/arviz-devs/arviz/blob/master/arviz/stats/diagnostics.py#L939) to work given that when using dask, multiple outputs are not supported. 

It is also direct to keep this feature optional with no extra work at all. 

@nitishp25 can you check what happens when dask is enabled? Does `az.summary` work?

**Regarding parallelization:** the way to general speed ups in not too large data seems to be trying to parallelize the loop in the ufuncs with things like numba.guvectorize or dask.delayed which look much harder work and I don't see as straightforward to keep as an optional feature, it also has the challenge of working properly with the large data case where multithreading is not allowed.

 I think it would be great to see what alternatives are actually feasible and could be implemented together so we can then choose the final implementation, there are several extra point to take into account in addition to pure speed ups when it comes to choose the final implementation in order to keep the package maintainable and with a sensible set of dependencies

--- 

This PR tackles the first challenge only. 

Therefore the expected improvements are being able to process 4Gb datasets with 600Mb or RAM (numbers totally invented, let's see what happens actually with the benchmarks) and speeding up computations on large arrays  (with small arrays a slight overhead is to be expected) as chunks are processed in parallel with `dask="parallelized"`. _Note that for arrays stored as numpy arrays/single chunk dask arrays, using `dask="parallelized"` provides literally no speed up_, it cannot parallelize the loop over non core dims we have in `make_ufunc`, we have to take care of that ourselves as said above.

canyon289:
@OriolAbril My ask is now that this code is written we try out Challenge 1 and see if intended effect does indeed work before merging this PR. What I imagine is a notebook where three things are tried without dask, and then Dask is enabled and those three things now are "magically" better. Nothing formal, a local single time benchmark is fine. It just would be good to see the data before committing this code to master

OriolAbril:
> What I imagine is a notebook where three things are tried without dask, and then Dask is enabled and those three things now are "magically" better.

Definitely, if I recall correctly @nitishp25 already ran some benchmarks that were included in his GSoC proposal so I think it should not be too much work to gather them in a notebook and make sure we get the expected result with the code in this PR.

canyon289:
> > What I imagine is a notebook where three things are tried without dask, and then Dask is enabled and those three things now are "magically" better.
> 
> Definitely, if I recall correctly @nitishp25 already ran some benchmarks that were included in his GSoC proposal so I think it should not be too much work to gather them in a notebook and make sure we get the expected result with the code in this PR.

I think a I remember those too, running those based off the changes in this branch is all we need

nitishp25:
Okay, I'll add screenshots from the notebook here.

canyon289:
@nitishp25 I would just add the whole notebook to a gist and the link :) That would be better than screenshots

mrocklin:
I'm excited to see some work here.  I am curious to see how this is used in practice.

OriolAbril:
One example motivating Dask integration in my case is being able to use posterior predictive checks and information criteria for model comparison with astrophysical models such as exoplanet, which is an area where Bayesian methods are quite common, see [exoplanet](https://github.com/exoplanet-dev/exoplanet) library for example. 

In these scenarios, the models generally have a low to moderate number of parameters as they rely heavily on physics to be well informed so, even if relatively slow compared to models with the same number of parameters, running the inference step does not require a lot of memory and can be done with personal computers. Most importantly, even when inference is run on HPCs results can be postprocessed and analyzed locally. Unlike the model parameters however, the number of observations can easily be of the order 1e5 or higher. When multiplied by the number of samples in each fit, computation starts to be challenging memory-wise for analysis requiring both pointwise data per observation and per sample. Both posterior predictive checks and information criteria are examples of such analysis.

I think that both posterior predictive checks and information criteria can be really useful to analyze and criticize the models, and using Dask would allow researchers to continue running the analysis from their personal computers. With more and more satellite data being available, I think that this issue will extend to many analysis and even to different fields.

ahartikainen:
Hi, remember to `git pull` before working with the branch (I fixed merge conflicts)

OriolAbril:
I mentioned this PR in a [post on PyMC discourse](https://discourse.pymc.io/t/memory-spike-at-the-end-of-the-mcmc-sampling/5669?u=oriolabril), I figured it'd be interesting to share here.

nitishp25:
That's nice! Sorry for the long period of inactivity, I'll try to get this done this week

nitishp25:
Here is the link for the [memory benchmark notebook](https://gist.github.com/nitishp25/883384c62e56a75ddbc58df94e4aa8a0), unfortunately `az.summary()` doesn't work yet because it needs separate `dask_kwargs` for each function it uses. Should we add these specific `dask_kwargs` args separately for each function for summary or would it be too complicated?

OriolAbril:
I am not surprised there are issues with `summary`. I'll try to look into it someday, but may take a while until I can properly work on this.

Regarding the benchmark, can you add an example with hdi and with loo_pit? They may require extra arguments in addition to output_dtypes and I am not completely sure they will work as for example loo_pit returns a tuple

nitishp25:
> I am not surprised there are issues with summary. I'll try to look into it someday, but may take a while until I can properly work on this.

I think that creating specific arguments for all the functions that `summary` uses would be too complicated for the user and would require the user to know the flow of data in `summary` so we would have to come up with something else maybe.

> Regarding the benchmark, can you add an example with hdi and with loo_pit? They may require extra arguments in addition to output_dtypes and I am not completely sure they will work as for example loo_pit returns a tuple

Alright, I'll test them.

OriolAbril:
> I think that creating specific arguments for all the functions that `summary` uses would be too complicated for the user and would require the user to know the flow of data in `summary` so we would have to come up with something else maybe?

I thought summary used the multichain statistics to compute everything at once so they would only have to provide one set of arguments which should be feasible if properly documented, maybe hdi is computed outside of this :thinking:

nitishp25:
> I thought summary used the multichain statistics to compute everything at once so they would only have to provide one set of arguments which should be feasible if properly documented, maybe hdi is computed outside of this thinking

Ohh right, yes `hdi` is not a part of `_multichain_statistics`.
I've added an example of `hdi` in the gist, for `loo`, `loo_pit` and `summary` I think we should wait for xarray to support multiple outputs with `dask="parallelized"` so that they can work.

arviz/stats/stats.py
```diff
@@ -1670,7 +1691,7 @@ def apply_test_function(
     func_kwargs: mapping, optional
         Passed as is to ``func``
     wrap_data_kwargs, wrap_pp_kwargs: mapping, optional
-        kwargs passed to ``az.stats.wrap_xarray_ufunc``. By default, some suitable input_core_dims
+        kwargs passed to :func:`~arviz.wrap_xarray_ufunc`. By default, some suitable input_core_dims
```
OriolAbril, line 1694: Thanks! And thanks for using clickable links everywhere :smile:

OriolAbril (approved):
> I've added an example of hdi in the gist

Thanks! Looks great. If I may, only one comment for when writing the tutorial notebook, I would not add the `output_sizes` arg to the Dask class for hdi and I'd use the opportunity to show how one can pass dask kwargs to the functions while still using the kwargs defined in the class.

> for loo, loo_pit and summary I think we should wait for xarray to support multiple outputs with dask="parallelized" so that they can work.

Sounds good! Are you tracking the PR? Otherwise I will subscribe to the notifications to make sure we know when to start working on this. In the meantime we can think about creation of idata objects using Dask Arrays. I can see several alternatives, not sure which to choose.

nitishp25:
> Thanks! Looks great. If I may, only one comment for when writing the tutorial notebook, I would not add the output_sizes arg to the Dask class for hdi and I'd use the opportunity to show how one can pass dask kwargs to the functions while still using the kwargs defined in the class.

Ah yes, had that in mind but forgot. Will take care while making the notebook.

> Are you tracking the PR? Otherwise I will subscribe to the notifications to make sure we know when to start working on this.

It was merged last month and xarray 0.16.1 releases just a few hours ago! But can we update the code and start using it right now? Will the args be incompatible with older versions of xarray using lastest version of arviz? (I think tests also run on older versions?)

nitishp25:
> Thanks! Looks great. If I may, only one comment for when writing the tutorial notebook, I would not add the output_sizes arg to the Dask class for hdi and I'd use the opportunity to show how one can pass dask kwargs to the functions while still using the kwargs defined in the class.

Ah yes, had that in mind but forgot. Will take care while making the notebook.

> Are you tracking the PR? Otherwise I will subscribe to the notifications to make sure we know when to start working on this.

It was merged last month and xarray 0.16.1 releases just a few hours ago! But can we update the code and start using it right now? Will the args be incompatible with older versions of xarray using lastest version of arviz? (I think tests also run on older versions?)

OriolAbril:
> But can we update the code and start using it right now? Will the args be incompatible with older versions of xarray using lastest version of arviz? (I think tests also run on older versions?)

I don't think we test against old xarray versions, but the arguments used here are all part of the public api, there should not be any breaking changes until xarray 1.0.0. That being said however, we may want to merge that and take care about the multiple output stats functions in another PR.

nitishp25:
Thanks! This is ready now right?

OriolAbril:
Yes, I added a warning that it's still a work in progress given that multiple outputs do now work yet. Unless someone opposes I'll merge in ~24 hours

ahartikainen:
Great job, I will merge this in.

ahartikainen:
@nitishp25 thanks for the PR!

mrocklin:
I'm glad to see this in.  Please reach out to Dask folks at
github.com/dask/dask/issues/new if you all run into any issues or have
questions

PR #1230, opened 2020-06-09
Test azurepipelines 1.0.0rc3

canyon289:
Trying new azure-pipelines pytest plugin cause this tweet told me to

https://twitter.com/anthonypjshaw/status/1270205981616529408


OriolAbril:
Not sure about what has to be reported but it looks like everything worked with Azure itself and codecov but we got some new warnings in summary and tests tab now says we are running 366 instead of 20,822. Seems to be an issue with passing results from pytest to Azure, not with skipping tests: https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=2707&view=logs&j=5b3317c5-94d9-53ec-dbbc-47b80eb52f0e&t=e6aab551-00b0-57ae-3e52-afeb5a52978f&l=85

PR #1236, opened 2020-06-11
Fix keyword ignores

canyon289:
## Description
hpd forwarding to hdi has a bug where it ignores user arguments

Here's a small example
```
def function_1(value="a", **kwargs):
    function_2(value="TheWrongVal", **kwargs)

def function_2(value=None, **kwargs):
    print(value)


function_1("a")
function_1("test_val")
```
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
   objects. (#1218)
 * Added `log_likelihood` argument to `from_pyro` and a warning if log likelihood cannot be obtained (#1227)
 * Skip tests on matplotlib animations if ffmpeg is not installed (#1227)
+* Fix hpd bug where arguments were being ignored (#1236)
 
 ### Deprecation
 * Using `from_pymc3` without a model context available now raises a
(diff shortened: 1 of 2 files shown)


canyon289:
Note to self. I also need to add a test to ensure HDI values are being calculated to correct values, not just some value

OriolAbril:
Should we make a patch release for this? When are we planning to remove `hpd`?

canyon289:
We probably should do patch because it breaks existing functionality, unless we are removing hpd and making a breaking change soon

OriolAbril:
Releasing a patch sounds good, just wanted to check. Should it be done on current master or adding only this commit to previous patch release? Not sure if html repr and other brand new and not much tested functionality should go in the patch, none of this is breaking but it is really new.

canyon289:
Oh man this is a good point. Maybe we just cut new release then in shorter timeframe. Will think on it today reply here

canyon289:
Alright i dont think a more complciated branching structure is worth it. I think we merge and make the next release, but I also wont do it until governance is over because there's a bunch going on with that, so 3 weeks or so from me.

If someone else wants to cut a release though I support it

PR #1241, opened 2020-06-13
[WIP] fix hdi false positive warning

OriolAbril:
## Description
hdi used to support 2d data at most, so `plot_hdi` converted the data to 2d. Thanks to #1117 this is not necessary anymore and triggers this warning.

I have also done a couple of updates to `summary` also as a followup of #1117. I want to note that now `az.hdi` and `xr.Dataset.quantile` return an object with the same format and we can therefore use the same pattern as hdi now uses to report also equal tail intervals.

## Checklist


AlexAndorra:
Thanks for fixing this! I had it on my to do list for weeks now and never
found the time 🤦‍♂️
Will review tomorrow 👌

Le dim. 14 juin 2020 à 00:47, codecov[bot] <notifications@github.com> a
écrit :

arviz/plots/hdiplot.py
```diff
@@ -34,8 +36,11 @@ def plot_hdi(
     ----------
     x : array-like
         Values to plot.
-    y : array-like
+    y : array-like, optional
         Values from which to compute the hdi. Assumed shape (chain, draw, \*shape).
+        Only optional if hdi_data is present.
+    hdi_data : array_like, optional
+        Hdi values to use.
```
AlexAndorra, line 43: I think the description should be a bit more explicit, especially w.r.t the difference with `y`. As such, I fear this could be mistaken for `hdi_prob`

arviz/plots/hdiplot.py
```diff
@@ -13,14 +13,16 @@
...
+    y=None,
     hdi_prob=None,
+    hdi_data=None,
     color="C1",
     circular=False,
     smooth=True,
     smooth_kwargs=None,
     fill_kwargs=None,
     plot_kwargs=None,
+    hdi_kwargs=None,
```
AlexAndorra, line 25: I don't any description for this kwarg below

arviz/plots/hdiplot.py
```diff
@@ -139,5 +150,5 @@ def plot_hdi(
 
 
 def plot_hpd(*args, **kwargs):  # noqa: D103
-    warnings.warn("plot_hdi has been deprecated, please use plot_hdi", DeprecationWarning)
+    warnings.warn("plot_hpd has been deprecated, please use plot_hdi", DeprecationWarning)
```
AlexAndorra, line 153: Good catch!

arviz/plots/hdiplot.py
OriolAbril: definitely, I have in mind to extend docs, also add a `See also` section pointing to `hdi` and maybe one example.

arviz/plots/hdiplot.py
```diff
@@ -34,75 +37,124 @@ def plot_hdi(
...
+        :func:`scipy:scipy.signal.savgol_filter` for details.
+    fill_kwargs : dict, optional
+        Keywords passed to :meth:`mpl:matplotlib.axes.Axes.fill_between`
+        (use fill_kwargs={'alpha': 0} to disable fill) or to
+        :meth:`bokeh:bokeh.plotting.figure.Figure.patch`.
+    plot_kwargs : dict, optional
+        HDI limits keyword arguments, passed to :meth:`mpl:matplotlib.axes.Axes.plot` or
+        :meth:`bokeh:bokeh.plotting.figure.Figure.patch`.
+    hdi_kwargs : dict, optional
+        Keyword arguments passed to :func:`~arviz.hdi`. Ingnored if ``hdi_data`` is present.
```
AlexAndorra, line 67:
```suggestion
        Keyword arguments passed to :func:`~arviz.hdi`. Ignored if ``hdi_data`` is present.
```

OriolAbril:
Ready to merge, if nobody complains I'll merge tomorrow

OriolAbril:
I guess there are a couple points that could be a bit more problematic:
* I am setting the coordinate values in `hdi` dim to `["lower", "higher"]` which does not include the hdi_prob info (this diverges from summary df for instance, I find `hdi_3%` a little confusing). I have included `hdi_prob` as an attribute in case it were needed.
* passing 2d data to `hdi` will work and be backwards compatible but I have added a FutureWarning because it does not comply with ArviZ `chain, draw, *shape` convention. In a future release we should remove FutureWarning and enforce the convention to make hdi behave coherently with all other ArviZ functions

Extra easter egg: now `az.hdi`'s behaviour is very similar to the point that using `hdi_data` allows to pass the result of `Dataset.quantile` to `az.plot_hdi` (a transpose may be needed to make sure the quantile/hdi dimension is the last one)

AlexAndorra:
That looks good to me 👌 Can be included in 0.8.4

PR #1244, opened 2020-06-15
Capitalized HDI everywhere it is used in text

fonnesbeck:
HDI is an acronym, so it should be capitalized. Currently, it is "HDI" in some places and "hdi" in others (e.g. `plot_forest`):

Screenshot 2020-06-15 11 33 35

This PR makes capitalization consistent.

### DIFF
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -227,4 +227,3 @@ def plot_forest(
     plot = get_plotting_function("plot_forest", "forestplot", backend)
     axes = plot(**plot_forest_kwargs)
     return axes
-``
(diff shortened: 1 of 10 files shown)


arviz/plots/forestplot.py
```diff
@@ -227,3 +227,4 @@ def plot_forest(
     plot = get_plotting_function("plot_forest", "forestplot", backend)
     axes = plot(**plot_forest_kwargs)
     return axes
+``
```
AlexAndorra, line 230: Looks like a typo -- I think this is what's causing the tests to fail

canyon289:
Things look good. Merging. Thanks @fonnesbeck

PR #1245, opened 2020-06-16
Added functionality to incorporate the log-likelihood computed by JAGS into ArviZ

michaelnowotny:
JAGS users can manually compute the log-likelihood at each datapoint. This pull requests adds functionality to ArviZ that extracts MCMC samples of the log_likelihood at each data point from a PyJAGS samples dictionary. This enables the calculation of WAIC. An example has been added to the inference data cookbook that illustrates both how to manually implement the calculation of the log-likelihood in JAGS as well as how to incorporate this likelihood in the conversion of PyJAGS sample dictionaries to ArviZ inference data objects.

## Checklist


arviz/tests/external_tests/test_data_pyjags.py
```diff
@@ -16,6 +20,41 @@
 PYJAGS_PRIOR_DICT = {"b": np.random.randn(3, 10, 3), "int": np.random.randn(1, 10, 3)}
 
 
+EIGHT_SCHOOL_DATA = {
+    "J": 8,
+    "y": np.array([28.0, 8.0, -3.0, 7.0, -1.0, 1.0, 18.0, 12.0]),
+    "sigma": np.array([15.0, 10.0, 16.0, 11.0, 9.0, 11.0, 10.0, 18.0]),
+}
```
OriolAbril, line 27: this data is already available as a fixture: https://github.com/arviz-devs/arviz/blob/master/arviz/tests/helpers.py#L21 it should be possible to import and add as function argument to use it from helpers.

arviz/tests/external_tests/test_data_pyjags.py
```diff
@@ -99,3 +138,55 @@ def test_inference_data_attrs(posterior, prior, save_warmup, warmup_iterations:
...
+        save_warmup=True,
+        warmup_iterations=NUMBER_OF_WARMUP_SAMPLES,
+    )
+
+
+def test_waic(pyjags_data: az.InferenceData):
+    waic = az.waic(pyjags_data)
+
+    assert -31.0 < waic.waic < -30.0
+    assert 0.75 < waic.p_waic < 0.90
```
OriolAbril, line 192: I think we should test groups and variables presence using `check_multiple_attrs` instead of calling waic. If necessary there can be more assertions in addition to `check_multiple_attrs` to check correct dims, shapes...

arviz/tests/external_tests/test_data_pyjags.py
```diff
@@ -1,6 +1,10 @@
+# pylint: disable=no-member, invalid-name, redefined-outer-name
 import typing as tp
 import numpy as np
 import pytest
+import pyjags
+
+import arviz as az
```
OriolAbril, line 7: we try to use `from arviz import from_pyjags` style imports

arviz/data/io_pyjags.py
```diff
@@ -297,6 +310,10 @@ def from_pyjags(
                prior samples with shape
                (parameter_dimension, chain_length, number_of_chains)
 
+    log_likelihood_name: str, optional
+                         the variable name in the posterior samples dictionary representing the
+                         log-likelihood at each observation
```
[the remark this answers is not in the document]
canyon289: If possible can these be changed so they match numpy style docstrings, including spacing?  The reason is our docs get autobuilt using Sphinx and following a consistent style, including spacing helps avoid issues when the html is generated!

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

arviz/tests/external_tests/test_data_pyjags.py
```diff
@@ -16,6 +20,41 @@
 PYJAGS_PRIOR_DICT = {"b": np.random.randn(3, 10, 3), "int": np.random.randn(1, 10, 3)}
 
 
+EIGHT_SCHOOL_DATA = {
+    "J": 8,
+    "y": np.array([28.0, 8.0, -3.0, 7.0, -1.0, 1.0, 18.0, 12.0]),
+    "sigma": np.array([15.0, 10.0, 16.0, 11.0, 9.0, 11.0, 10.0, 18.0]),
+}
+
+EIGHT_SCHOOL_PRIOR_MODEL_CODE = """
```
canyon289, line 29: This is a great PR so please dont take any of these comments as major criticism. Just as someone that is trying to understand this code and keep it working and alive as much as possible.

This eight schools code is only used in one place. It would be helpful if it were moved to `jags_prior_model` so its clear its not used anywhere else. The global makes it seem like its reused

arviz/tests/external_tests/test_data_pyjags.py
```diff
@@ -16,6 +20,41 @@
...
+model {
+    mu ~ dnorm(0.0, 1.0/25)
+    tau ~ dt(0.0, 1.0/25, 1.0) T(0, )
+    for (j in 1:J) {
+        theta_tilde[j] ~ dnorm(0.0, 1.0)
+    }
+}
+"""
+
+EIGHT_SCHOOL_POSTERIOR_MODEL_CODE = """
```
canyon289, line 39: Same comment for this posterior code as made on the prior code

arviz/tests/external_tests/test_data_pyjags.py
```diff
@@ -99,3 +138,55 @@ def test_inference_data_attrs(posterior, prior, save_warmup, warmup_iterations:
     fails = check_multiple_attrs(test_dict, arviz_inference_data_from_pyjags_samples_dict)
     print(fails)
     assert not fails
+
+
+@pytest.fixture()
+def jags_prior_model() -> pyjags.Model:
+    return pyjags.Model(
```
canyon289, line 145: It would help others if `pyjags.Model(` was assigned to a var, and that var was returned, because it makes it possible to insert a debugger and understand what is being generated with `pyjags.Model(`. Returning the computed value immediately makes it harder to interrupt the execution flow and introspect

canyon289:
If possible can you try copying one of the patterns from the other libraries? Particularly this `load_cached_model` approach.

What this does is cache the model outputs so on subsequent reruns, inference doesnt need to be run again. Additionally the consistency will make it easier to read and maintain over time.

https://github.com/arviz-devs/arviz/blob/master/arviz/tests/external_tests/test_data_pymc.py#L34
https://github.com/arviz-devs/arviz/blob/master/arviz/tests/external_tests/test_data_emcee.py#L76

Cant tell you how much we appreciate this pull request and are thrilled that PyJags users will gain benefit from ArviZ as well. Thank you!

CHANGELOG.md
```diff
@@ -3,9 +3,11 @@
 ## v0.x.x Unreleased
 
 ### New features
+* loo-pit plot. The kde is computed over the data interval (this could be shorter than [0, 1]). The hdi is computed analitically (#1215)
```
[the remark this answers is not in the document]
michaelnowotny: I am removing it. Thanks for catching this!

CHANGELOG.md
```diff
@@ -3,9 +3,11 @@
 ## v0.x.x Unreleased
 
 ### New features
+* loo-pit plot. The kde is computed over the data interval (this could be shorter than [0, 1]). The hdi is computed analitically (#1215) 
 * loo-pit plot. The kde is computed over the data interval (this could be shorter than [0, 1]). The hdi is computed analitically (#1215)
 * Added `html_repr` of InferenceData objects for jupyter notebooks. (#1217)
 * Added support for PyJAGS via the function `from_pyjags` in the module arviz.data.io_pyjags. (#1219)
+* Added support to incorporate the log-likelihood generated by JAGS into ArviZ
```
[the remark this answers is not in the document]
michaelnowotny: Yes, I agree.

CHANGELOG.md
```diff
@@ -23,6 +25,7 @@
 ### Documentation
 * A section has been added to the documentation at InferenceDataCookbook.ipynb illustrating the use of ArviZ in conjunction with PyJAGS. (#1219)
 * Fixed inconsistent capitalization in `plot_hdi` docstring (#1221)
+* Amended the documentation at InferenceDataCookbook.ipynb to show how to calculate the log-likelihood in PyJAGS and how to use the information in ArviZ
```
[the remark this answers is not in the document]
michaelnowotny: I agree

arviz/data/io_pyjags.py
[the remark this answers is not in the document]
michaelnowotny: Changed the indent and pushed a commit.

arviz/tests/external_tests/test_data_pyjags.py
michaelnowotny: Ok, I changed that.

arviz/tests/external_tests/test_data_pyjags.py
michaelnowotny: Ok, I made the change. Please let me know if this is not what you had in mind!

arviz/tests/external_tests/test_data_pyjags.py
michaelnowotny: Ok, I moved both the prior and the posterior model code to the functions where they are used.

arviz/tests/external_tests/test_data_pyjags.py
michaelnowotny: Done

arviz/tests/external_tests/test_data_pyjags.py
michaelnowotny: Ok, I pushed these changes.

michaelnowotny:
> If possible can you try copying one of the patterns from the other libraries? Particularly this `load_cached_model` approach.
> 
> What this does is cache the model outputs so on subsequent reruns, inference doesnt need to be run again. Additionally the consistency will make it easier to read and maintain over time.
> 
> https://github.com/arviz-devs/arviz/blob/master/arviz/tests/external_tests/test_data_pymc.py#L34
> https://github.com/arviz-devs/arviz/blob/master/arviz/tests/external_tests/test_data_emcee.py#L76
> 
> Cant tell you how much we appreciate this pull request and are thrilled that PyJags users will gain benefit from ArviZ as well. Thank you!

Hi Ravin,

thank you for your feedback. I agree with most of what you suggest. However, I disagree with this idea. It takes less than 100ms to construct and estimate the model using PyJAGS (including the warmup). Whenever something is cached its creates the question of cache invalidation. I think the introduction of this 'side-effect' is not worth it in this particular instance. As for the argument of maintainability: I think creating these classes for testing will be introduce unnecessary complexity (i.e. be an over engineered solution).

I am happy to be contributing to ArviZ. I think it is an awesome library that saves hours and days of development time since it absolves the user from programming their own plotting and diagnostics functions.

BTW: If anyone here is competent in compiling C extensions on Windows, I am looking for someone to help me bring PyJAGS to Windows. The problem is two-fold: 
1.) To locate the JAGS DLL PyJAGS uses Pkg-Config on unix. This is a less severe issue which could be circumvented by distributing the JAGS DLL directly with PyJAGS or asking the user to set an environment variable such as JAGS_HOME pointing to the location of JAGS on their system.
2.) The more severe problem is that I have not found a way for Python to tell the Visual Studio linker where to find the JAGS DLL. Part of the problem might be that Windows DLLs seem to have a different naming pattern than Unix shared object (which are prefixed by 'lib').

In any case, I had not been able to come up with a solution but it seems silly to exclude windows users from using PyJAGS especially given the fact that JAGS and RJAGS run on Windows.

If anyone has experience with Python C extensions on Windows, please step forward.

arviz/data/io_pyjags.py
```diff
@@ -18,14 +18,20 @@ class PyJAGSConverter:
     def __init__(
         self,
         *,
-        posterior: tp.Optional[tp.Mapping[str, np.ndarray]] = None,
+        posterior: tp.Optional[tp.Dict[str, np.ndarray]] = None,
         prior: tp.Optional[tp.Mapping[str, np.ndarray]] = None,
+        log_likelihood_name: tp.Optional[str] = None,
```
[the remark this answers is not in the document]
michaelnowotny: Thanks for the suggestion. I shall take a look at this and get back to you.

canyon289:
@michaelnowotny Thank you for all these changes and the correspondence, things are looking great.

On the caching that makes sense, I didn't realize pyjags was so fast. If that's the case we can avoid caching.

As for the for classes, in most other cases I would agree with you! Classes aren't always needed in pytest, and with the simplicity of pyjags the organization they afford isn't big in this one module. In sampler pytest modules however what helps a lot is consistency, and because all the other samplers follow the class pattern it reduces the code reading burden on the maintainer, which for open source is incredibly important given no one is paid (unfortunately). 

Given that would you be amenable to using the class data pattern and using the eight school params fixture??
https://github.com/arviz-devs/arviz/blob/master/arviz/tests/external_tests/test_data_pymc.py#L33
https://github.com/arviz-devs/arviz/blob/master/arviz/tests/external_tests/test_data_numpyro.py#L25
https://github.com/arviz-devs/arviz/blob/master/arviz/tests/external_tests/test_data_pystan.py#L34

arviz/data/io_pyjags.py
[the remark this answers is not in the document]
michaelnowotny: I added the functionality. Thanks for bringing this to my attention.

arviz/tests/external_tests/test_data_pyjags.py
michaelnowotny: I added a check for the presence of the log-likelihood via `check_multiple_attrs`. I think that it does not hurt (and is probably a good idea) to check that what is in the likelihood attribute is actually the likelihood and not some other data grabbed from the posterior. Please let me know if this is ok for you or you would prefer to have it removed.

michaelnowotny: I added a check for the presence of the log-likelihood via `check_multiple_attrs`. I think that it does not hurt (and is probably a good idea) to check that what is in the likelihood attribute is actually the likelihood and not some other data grabbed from the posterior. Please let me know if this is ok for you or you would prefer to have it removed.

michaelnowotny:
Ok, I refactored the test code to use classes.

arviz/tests/external_tests/test_data_pyjags.py
```diff
@@ -29,73 +43,184 @@ def verify_equality_of_numpy_values_dictionaries(
...
+            log_likelihood={"y": "log_like"},
+            save_warmup=True,
+            warmup_iterations=NUMBER_OF_WARMUP_SAMPLES,
+        )
+
+    def test_waic(self, pyjags_data):
+        waic_result = waic(pyjags_data)
+
+        assert -31.0 < waic_result.waic < -30.0
+        assert 0.75 < waic_result.p_waic < 0.90
```
OriolAbril, line 201: I am continuing the conversation from previous comment here. 

I do understand it makes sense to check not only the presence of the log likelihood group but its values too, however, I don't think this should be done using waic because there are several issues unrelated to `from_pyjags` that could cause the test to fail, if possible we should get the test to fail only if what they are testing for fails. To put one example, we recently modified the default scale in which waic and loo return the results, if we had been using waic to test log likelihood data we would have had to update both waic tests _and_ all converter tests. 

I'd suggest one of the following approaches:
* if random seeding is trustworthy and implemented, checking the first row or something like this aganist some hardcoded values could work
* using some simpler and not subject to change quantity like the mean
* given that we do have posterior and log likelihood samples in the inference data, use the samples in the posterior and the observed data fixture to check the first row of log likelihood values

michaelnowotny: Thank you for the explanation. That makes sense. I removed the test for waic and am now only testing for the presence of the likelihood group via `check_multiple_attrs`.

arviz/data/io_pyjags.py
```diff
@@ -288,24 +311,32 @@ def from_pyjags(
...
-               (parameter_dimension, chain_length, number_of_chains)
+        a dictionary mapping variable names to NumPy arrays containing
+        prior samples with shape
+        (parameter_dimension, chain_length, number_of_chains)
+
+    log_likelihood: dict of {str: str}, list of str or str, optional
+        Pointwise log_likelihood for the data. log_likelihood is extracted from the
+        posterior. It is recommended to use this argument as a dictionary whose keys
+        are observed variable names and its values are the variables storing log
+        likelihood arrays in the Stan code. In other cases, a dictionary with keys
```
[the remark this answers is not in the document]
michaelnowotny: Thanks for catching that. I made the change.

michaelnowotny:
Thanks for the input and code reviews!

canyon289:
@michaelnowotny Thank you so much for bearing with us and adding pyjags compatibility! Please feel free to open issue tickets, or add extra prs or whatever else in the future. Your assistance work here was much appreciated

PR #1246, opened 2020-06-16
Change the default scatter_kwargs in pairplot

fehiepsi:
Fix the issue: the grid is overwriting points in pair plot. For example, in [the gallery example](https://arviz-devs.github.io/arviz/examples/matplotlib/mpl_plot_pair.html), blue points are separated into two pieces at grid lines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Added `log_likelihood` argument to `from_pyro` and a warning if log likelihood cannot be obtained (#1227)
 * Skip tests on matplotlib animations if ffmpeg is not installed (#1227)
 * Fix hpd bug where arguments were being ignored (#1236)
+* Change the default `zorder` of scatter points from `0` to `0.6` in `plot_pair` (#1246)
 
 ### Deprecation
 * Using `from_pymc3` without a model context available now raises a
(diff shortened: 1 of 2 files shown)


OriolAbril:
Thanks! I have been thinking about why this was added and it could be to put the dots behind 2d KDE and hexbin histogram. Could you check that with this zorder dots are hidden by KDE filled contour? This example https://arviz-devs.github.io/arviz/examples/matplotlib/mpl_plot_pair_point_estimate.html shows both dots and kde.

I think the ideal order would be (from top/front to bottom/back):
1. divergences
1. kde/histbin
1. scatter
1. grid lines

fehiepsi:
Hi @OriolAbril, it seems to work correctly to me. Here is the image.

kde

fehiepsi:
I think the default zorder of grid is 0.5 (I just guess from [this doc](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.axes.Axes.set_axisbelow.html#matplotlib.axes.Axes.set_axisbelow)) so I set scatter zorder to 0.6 by default. It seems to work correct now.

kde

OriolAbril:
Sounds and looks great, thanks! I'll merge after adding to the changelog (I can do it myself if you prefer)

fehiepsi:
Thanks for reviewing and suggestions, @OriolAbril!

PR #1247, opened 2020-06-17
Use constrained layout in all styles

ColCarroll:
This adds `constrained_layout=True` to all arviz styles. This was experimental as of matplotlib 2.2, we use >3.0. I've been using this for my plots for a few years with no problem.  I've included an example below. Notice the text boxes do not overlap with the subplots. 

Running 

```python
import arviz as az

az.style.use("arviz-darkgrid")

centered = az.load_arviz_data("centered_eight")
az.plot_trace(centered, var_names=["theta", "mu", "tau"], divergences=True);
```

Here is before the change:
image

Here is after the change:

image

### DIFF
--- a/arviz/plots/styles/arviz-darkgrid.mplstyle
+++ b/arviz/plots/styles/arviz-darkgrid.mplstyle
@@ -5,6 +5,7 @@
 figure.figsize: 7.2, 4.8
 figure.dpi: 100.0
 figure.facecolor: white
+figure.constrained_layout.use: True
 text.color: .15
 axes.labelcolor: .15
 legend.frameon: False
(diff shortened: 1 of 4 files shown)


OriolAbril:
Should we then remove `plot.matplotlib.constrained_layout` rcParam? https://github.com/arviz-devs/arviz/blob/master/arvizrc.template#L45 (I am not sure if it is enforced anywhere)

AlexAndorra:
Depending on what you guys think about Oriol's comment above, I think this is ready to go into 0.8.4

aloctavodia:
As pointed by Oriol `constrained_layout=True` is setted in arviz rcParams and is enforced [here](https://github.com/arviz-devs/arviz/blob/a6ea9e5bdbe1a5ea0c4e11fd0644cffb72d54934/arviz/plots/backends/matplotlib/__init__.py#L15). And  by default. @ColCarroll is this not working as default for you?

ColCarroll:
It might be - I was confused since `plt.rcParams['figure.constrained_layout.use']` remains `False`, but I can see that the returned plot has constrained_layout=True.

A remaining benefit of this change is that just _using_ the arviz stylesheet will apply a constrained layout, even if not using an arviz plot.

PR #1254, opened 2020-06-20
added methods from xr.Dataset

percygautam:
## Description
Related to #1066 

## Checklist


percygautam:
Added the methods: `isel`, `stack`, `unstack`, `rename`, `rename_vars`, `rename_dims`

percygautam:
Tasks:

- update 'unstack' method to include `dim`
- add tests
- add more methods

percygautam:
Hi all, I have created the wrapper function for adding xr.Dataset methods:
```
    def extend_xr_method(func):

        @functools.wraps(func)
        def wrapped(*args, **kwargs):
            _filter = kwargs.pop('filter_groups', None)
            _groups = kwargs.pop('groups', None)
            _inplace = kwargs.pop('inplace', False)
            self = args[0]
            args = list(args)

            out = self if _inplace else deepcopy(self)

            groups = self._group_names(_groups, _filter)
            for group in groups:
                xr_data = getattr(out, group)
                args[0] = xr_data
                xr_data = func(*args, **kwargs)
                setattr(out, group, xr_data)

            return None if _inplace else out

        return wrapped
```

Please review, so I can implement it.

arviz/tests/base_tests/test_data.py
```diff
@@ -323,6 +323,18 @@ def test_addition(self):
...
 
+    def test_extend_xr_method(self):
+        arr = xr.DataArray(
+            data=np.ones((2, 3)),
+            dims=["x", "y"],
+            coords={"x": range(2), "y": range(3), "a": ("x", [3, 4])},
+        )
+        ds = xr.Dataset({"v": arr})
+        dataset = convert_to_inference_data(ds)
+        assert dataset.set_index(x="a").posterior == ds.set_index(x="a")
```
OriolAbril, line 334: I think this should be a function from `xarray.testing`, probably `assert_identical`

arviz/data/inference_data.py
```diff
@@ -173,6 +175,30 @@ def __delattr__(self, group):
...
+            groups = self._group_names(_groups, _filter)  # pylint: disable=protected-access
+            for group in groups:
+                xr_data = getattr(out, group)
+                args[0] = xr_data
+                xr_data = func(*args, **kwargs)  # pylint: disable=not-callable
+                setattr(out, group, xr_data)
+
+            return None if _inplace else out
+
+        return wrapped
```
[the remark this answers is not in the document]
OriolAbril: Also, the `_wrap_xarray_method` was a bad skeleton to test things like this (which is why it is private unlike `map`), it should be removed now.

arviz/data/inference_data.py
```diff
@@ -324,6 +355,311 @@ def sel(
...
+        filter_groups: {None, "like", "regex"}, optional, default=None
+            If `None` (default), interpret groups as the real group or metagroup names.
+            If "like", interpret groups as substrings of the real group or metagroup names.
+            If "regex", interpret groups as regular expressions on the real group or
+            metagroup names. A la `pandas.filter`.
+        inplace: bool, optional
+            If ``True``, modify the InferenceData object inplace,
+            otherwise, return the modified copy.
+        **kwargs: mapping
+            It must be accepted by Dataset.isel().
```
OriolAbril, line 384: modify to use intersphinx, see https://arviz-devs.github.io/arviz/developer_guide.html#hyperlinks

arviz/data/inference_data.py
```diff
@@ -324,6 +355,311 @@ def sel(
...
+        self, groups=None, filter_groups=None, inplace=False, **kwargs,
+    ):
+        """Perform an xarray selection on all groups.
+
+        Loops groups to perform Dataset.isel(key=item)
+        for every kwarg if key is a dimension of the dataset.
+        One example could be performing a burn in cut on the InferenceData object
+        or discarding a chain. The selection is performed on all relevant groups (like
+        posterior, prior, sample stats) while non relevant groups like observed data are
+        omitted. See :meth:`xarray.Dataset.isel <xarray:xarray.Dataset.isel>`
```
OriolAbril, line 368: simplify to 

    :meth:`xarray:xarray.Dataset.isel`

the example is `idata.sel` is pedagogical to show how would one use custom hyperlink text.

arviz/data/inference_data.py
OriolAbril: Noticed the line:

    We can therefore link to matplotlib docs on Axes.plot from any docstring using :meth:`matplotlib.axes.Axes.plot`.

is wrong, it should be:

    We can therefore link to matplotlib docs on Axes.plot from any docstring using :meth:`mpl:matplotlib.axes.Axes.plot`.

arviz/data/inference_data.py
```diff
@@ -324,6 +355,311 @@ def sel(
...
+        groups = self._group_names(groups, filter_groups)
+
+        out = self if inplace else deepcopy(self)
+        for group in groups:
+            dataset = getattr(self, group)
+            kwarg_dict = {}
+            for key, value in kwargs.items():
+                if not set(value).difference(dataset.dims):
+                    kwarg_dict[key] = value
+            dataset = dataset.stack(**kwarg_dict)
```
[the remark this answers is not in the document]
percygautam: Yeah, you're probably right. I have done the changes for the stack. Should I also correct other methods too, or wait for the lab-meeting?

arviz/data/inference_data.py
OriolAbril: For the record, I sent a PR yesterday to fix this, everything should be correct now. There is surely plenty of room for improvement though

arviz/data/inference_data.py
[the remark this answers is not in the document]
OriolAbril: I'd wait for lab meeting

arviz/data/inference_data.py
```diff
@@ -173,6 +175,53 @@ def __delattr__(self, group):
...
+    def extend_xr_method(func):  # pylint: disable=no-self-argument
+        """Wrapper to add methods to InferenceData Class"""
+
+        @functools.wraps(func)
+        def wrapped(*args, **kwargs):
+
+            _filter = kwargs.pop("filter_groups", None)
+            _groups = kwargs.pop("groups", None)
+            _inplace = kwargs.pop("inplace", False)
+            self = args[0]
```
ahartikainen, line 187: should the first arg just be `self`

arviz/data/inference_data.py
```diff
@@ -173,6 +175,53 @@ def __delattr__(self, group):
     def _groups_all(self):
         return self._groups + self._groups_warmup
 
+    def extend_xr_method(func):  # pylint: disable=no-self-argument
```
ahartikainen, line 178: why this method doesn't need a self argument?

arviz/data/inference_data.py
```diff
@@ -173,6 +175,53 @@ def __delattr__(self, group):
...
+        """Wrapper to add methods to InferenceData Class"""
+
+        @functools.wraps(func)
+        def wrapped(*args, **kwargs):
+
+            _filter = kwargs.pop("filter_groups", None)
+            _groups = kwargs.pop("groups", None)
+            _inplace = kwargs.pop("inplace", False)
+            self = args[0]
+            args = list(args)
```
ahartikainen, line 188: Can you give a small comment about the args order? What should be there?

The first args[0] is self, but also data in later steps?

arviz/data/inference_data.py
```diff
@@ -173,6 +175,53 @@ def __delattr__(self, group):
...
+                setattr(out, group, xr_data)
+
+            return None if _inplace else out
+
+        description = """\
+        This method is extended from xarray.Dataset methods. For more info see :meth:`xarray:xarray.Dataset.{method_name}`
+        """.format(
+            method_name=func.__name__
+        )
+        params = """\
```
ahartikainen, line 206: is that \ needed?

arviz/data/inference_data.py
```diff
@@ -173,6 +175,53 @@ def __delattr__(self, group):
...
+            filter_groups: {None, "like", "regex"}, optional, default=None
+                If `None` (default), interpret groups as the real group or metagroup names.
+                If "like", interpret groups as substrings of the real group or metagroup names.
+                If "regex", interpret groups as regular expressions on the real group or
+                metagroup names. A la `pandas.filter`.
+            inplace: bool, optional
+                If ``True``, modify the InferenceData object inplace,
+                otherwise, return the modified copy.
+            """
+        wrapped.__doc__ = description + params
```
ahartikainen, line 221: Is there any whitespace left between desc and params?

arviz/data/inference_data.py
```diff
@@ -173,6 +175,53 @@ def __delattr__(self, group):
     def _groups_all(self):
         return self._groups + self._groups_warmup
 
+    def extend_xr_method(func):  # pylint: disable=no-self-argument
+        """Wrapper to add methods to InferenceData Class"""
+
+        @functools.wraps(func)
```
ahartikainen, line 181: what does this do?

arviz/data/inference_data.py
```diff
@@ -324,6 +378,322 @@ def sel(
...
+        **kwargs: mapping
+            It must be accepted by :meth:`xarray:xarray.Dataset.stack`.
+
+        Returns
+        -------
+        InferenceData
+            A new InferenceData object by default.
+            When `inplace==True` perform selection in place and return `None`
+
+        """
```
ahartikainen, line 466: Can we add example for `draw, chain --> samples`

arviz/data/inference_data.py
```diff
@@ -324,6 +378,322 @@ def sel(
...
+            metagroup names. A la `pandas.filter`.
+        inplace: bool, optional
+            If ``True``, modify the InferenceData object inplace,
+            otherwise, return the modified copy.
+
+        Returns
+        -------
+        InferenceData
+            A new InferenceData object by default.
+            When `inplace==True` perform selection in place and return `None`
```
ahartikainen, line 515:
```suggestion
            When `inplace==True` perform selection in-place and return `None`
```

arviz/data/inference_data.py
```diff
@@ -324,6 +378,322 @@ def sel(
...
+        Loops groups to perform Dataset.rename(name_dict)
+        for every key in name_dict if key is a dimension/data_vars of the dataset.
+        The renaming is performed on all relevant groups (like
+        posterior, prior, sample stats) while non relevant groups like observed data are
+        omitted. See :meth:`xarray:xarray.Dataset.rename`
+
+        Parameters
+        ----------
+        name_dict: dict
+            Dictionary whose keys are current variable or dimension names
```
ahartikainen, line 546: is coordinates also included in dimensions

arviz/data/inference_data.py
```diff
@@ -324,6 +378,322 @@ def sel(
...
+        groups: str or list of str, optional
+            Groups where the selection is to be applied. Can either be group names
+            or metagroup names.
+        filter_groups: {None, "like", "regex"}, optional, default=None
+            If `None` (default), interpret groups as the real group or metagroup names.
+            If "like", interpret groups as substrings of the real group or metagroup names.
+            If "regex", interpret groups as regular expressions on the real group or
+            metagroup names. A la `pandas.filter`.
+        inplace: bool, optional
+            If ``True``, modify the InferenceData object inplace,
```
ahartikainen, line 557:
```suggestion
            If ``True``, modify the InferenceData object in-place,
```

Should we use in-place in all occurrences?

arviz/data/inference_data.py
ahartikainen: Is this function meant for general users or only developers / advanced users?

OriolAbril: I would make it private

arviz/data/inference_data.py
```diff
@@ -324,6 +378,322 @@ def sel(
...
+            dataset = getattr(self, group)
+            valid_keys = set(name_dict.keys()).intersection(dataset.dims)
+            dataset = dataset.rename_dims({key: name_dict[key] for key in valid_keys})
+            setattr(out, group, dataset)
+        if inplace:
+            return None
+        else:
+            return out
+
+    # extending xarray.Dataset methods
```
OriolAbril, line 683: I would remove this comment and the one below

arviz/data/inference_data.py
```diff
@@ -324,6 +378,322 @@ def sel(
...
+    set_index = extend_xr_method(xr.Dataset.set_index)
+    get_index = extend_xr_method(xr.Dataset.get_index)
+    reset_index = extend_xr_method(xr.Dataset.reset_index)
+    set_coords = extend_xr_method(xr.Dataset.set_coords)
+    reset_coords = extend_xr_method(xr.Dataset.reset_coords)
```
OriolAbril, line 688: can you add some more methods? The ones from the subset I am using the most or I think will be used the most but that are still missing are: `assign`, `assign_coords`, `sortby`

Some SO answers [1](https://stackoverflow.com/questions/61827061/xarray-make-two-dataarrays-in-the-same-dataset-use-the-same-coordinate-system), [2](https://stackoverflow.com/questions/59504320/how-do-i-subdivide-refine-a-dimension-in-an-xarray-dataset) of ArviZy calculations heavily requiring xarrray.

arviz/data/inference_data.py
```diff
@@ -324,6 +378,322 @@ def sel(
...
+
+    # extending apply methods
+    mean = extend_xr_method(xr.Dataset.mean)
+    median = extend_xr_method(xr.Dataset.median)
+    min = extend_xr_method(xr.Dataset.min)
+    max = extend_xr_method(xr.Dataset.max)
+    cumsum = extend_xr_method(xr.Dataset.cumsum)
```
OriolAbril, line 695: Here I think `sum` and `quantile` could be useful too

arviz/tests/base_tests/test_data.py
```diff
@@ -323,6 +324,18 @@ def test_addition(self):
         fails = check_multiple_attrs(test_dict, new_idata)
         assert not fails
 
+    def test_extend_xr_method(self):
+        arr = xr.DataArray(
+            data=np.ones((2, 3)),
+            dims=["x", "y"],
+            coords={"x": range(2), "y": range(3), "a": ("x", [3, 4])},
+        )
+        ds = xr.Dataset({"v": arr})
```
OriolAbril, line 333: I think it will be easier to initialize `ds` with `from_dict`, especially because I think that the created inference data should have several groups, otherwise we cannot test that the method is applied to all desired groups and **not** applied to the excluded groups. For example with 3 groups, the method can be applied to two of them and not to the third, and there are not too much groups to test each of the 3 manually. Having said that, maybe it makes more sense to create a helper function to compare an inference data to a list of datasets or something like this. I can try to think about some useful api for the helper.

arviz/tests/base_tests/test_data.py
```diff
@@ -323,6 +324,18 @@ def test_addition(self):
...
+        dataset = convert_to_inference_data(ds)
+        assert_identical(dataset.set_index(x="a").posterior, ds.set_index(x="a"))
+        dataset.set_index(x="a", inplace=True)
+        assert_identical(dataset.posterior, ds.set_index(x="a"))
```
OriolAbril, line 337: kudos on testing both inplace values, however, I think it would make the test easier to follow if a fixture was used for that. `test_sel` can be a template on `inplace`, not so much on the helper I was talking about above.

arviz/data/inference_data.py
percygautam: I have changed it.

arviz/data/inference_data.py
percygautam: Done. Thanks!!

arviz/data/inference_data.py
percygautam: It's not needed. I was just experimenting with the docstring.

arviz/data/inference_data.py
percygautam: No, I will correct it.

arviz/data/inference_data.py
percygautam: This is used to create wrapper function which wraps around func. You can check it out [here](https://docs.python.org/3/library/functools.html#functools.wraps).

arviz/data/inference_data.py
percygautam: I will add examples to all the added methods. Thanks!

arviz/data/inference_data.py
percygautam: Yes they are included.

arviz/data/inference_data.py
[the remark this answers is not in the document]
percygautam: Should I also add the description of the xarray method in `see also`?

arviz/tests/base_tests/test_data.py
percygautam: Yeah, I am planning to add a `equals` method to InferenceData, similar to `xarray's` one.

arviz/data/inference_data.py
```diff
@@ -173,6 +175,56 @@ def __delattr__(self, group):
     def _groups_all(self):
         return self._groups + self._groups_warmup
 
+    def _extend_xr_method(func):  # pylint: disable=no-self-argument
```
OriolAbril, line 178: I think self should also go here

arviz/data/inference_data.py
```diff
@@ -324,6 +381,325 @@ def sel(
...
+    assign_coords = _extend_xr_method(xr.Dataset.assign_coords)
+    sortby = _extend_xr_method(xr.Dataset.sortby)
+
+    mean = _extend_xr_method(xr.Dataset.mean)
+    median = _extend_xr_method(xr.Dataset.median)
+    min = _extend_xr_method(xr.Dataset.min)
+    max = _extend_xr_method(xr.Dataset.max)
+    cumsum = _extend_xr_method(xr.Dataset.cumsum)
+    sum = _extend_xr_method(xr.Dataset.sum)
+    quantile = _extend_xr_method(xr.Dataset.quantile)
```
OriolAbril, line 701: and these ones would become `self._extend...`

arviz/data/inference_data.py
percygautam: I don't think this will work like this. I think as `_extend_xr_method` is a helper function, this can be defined outside the class(maybe in `base.py`)?

arviz/data/inference_data.py
percygautam: this is returning error `self not defined`

arviz/data/inference_data.py
OriolAbril: I don't know, maybe ask in slack to get some advise on this? I am confused by the fact that the `wrapper` uses `_group_names` which is a private method (and I think it should be) so it should be part of idata class, maybe the wrapped method definitions should go in `__init__`? :confused:

percygautam:
@OriolAbril multiple tests are failing for this PR, can you help?

arviz/data/inference_data.py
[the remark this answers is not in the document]
percygautam: Hey, I don't remember what was decided in the lab-meeting. Is it discussed? What to do in this case?

OriolAbril:
I don't have time to fix mpl right now so I temporally pinned the version to <3.3 until we can make ArviZ code compatible with 3.3, I think this should fix the issues and get ci to pass again. see #1305

arviz/tests/base_tests/test_data.py
```diff
@@ -323,6 +324,26 @@ def test_addition(self):
         fails = check_multiple_attrs(test_dict, new_idata)
         assert not fails
 
+    @pytest.mark.parametrize("inplace", [True, False])
+    def test_extend_xr_method(self, inplace):
+        data = np.random.normal(size=(4, 500, 8))
+        dataset = from_dict(
```
OriolAbril, line 330: i'd use `idata` or `data`, `dataset` is very confusing as it seems like this is an xarray dataset

arviz/tests/base_tests/test_data.py
```diff
@@ -323,6 +324,26 @@ def test_addition(self):
...
+            posterior={"a": data[..., 0], "b": data},
+            prior={"a": data[..., 0], "b": data},
+            posterior_predictive={"a": data[..., 0], "b": data},
+        )
```
OriolAbril, line 334: I think this can be made a fixture, which can then be used with all methods, no need to create the idata every time. see https://github.com/arviz-devs/arviz/pull/1315/files#diff-98d6c3c4d2b318798dcd501599fa5127 with the `data_random`. `test_sel` could also use it so its code can be simplified

arviz/data/inference_data.py
```diff
@@ -324,6 +331,319 @@ def sel(
...
+        for group in groups:
+            dataset = getattr(self, group)
+            valid_keys = set(name_dict.keys()).intersection(dataset.dims)
+            dataset = dataset.rename_dims({key: name_dict[key] for key in valid_keys})
+            setattr(out, group, dataset)
+        if inplace:
+            return None
+        else:
+            return out
+
```
OriolAbril, line 629: more methods to add: chunk, unify_chunks, load, compute, persist

arviz/tests/base_tests/test_data.py
```diff
@@ -325,34 +335,23 @@ def test_addition(self):
...
-        data = np.random.normal(size=(4, 500, 8))
-        dataset = from_dict(
-            posterior={"a": data[..., 0], "b": data},
-            prior={"a": data[..., 0], "b": data},
-            posterior_predictive={"a": data[..., 0], "b": data},
-        )
-        dataset_copy = dataset
+    def test_extend_xr_method(self, data_random, inplace):
+        idata = data_random
+        idata_copy = idata
```
OriolAbril, line 342: I think idata copy will not be a copy which means that the asserts below `assert_identical` are not asserting what we want to check, instead `assert_identical(idata_copy.posterior, idata.posterior)` would work, and therefore the `sum` is actually having no effect. 

In addition to using `.copy()` (I think idata already has the method), it is probably better to not use `sum()` but to use `sum(dim="draw")`, therefore in the situation I am talking about above, test would fail due to `draw` dim not being present

percygautam: Okay, I understand. But I don't think we have the `.copy` method for idata (yet).

arviz/tests/base_tests/test_data.py
```diff
@@ -77,6 +79,18 @@ def no_remote_data(monkeypatch, tmpdir):
     )
 
 
+@pytest.fixture
```
OriolAbril, line 82: what was the issue with module scope? I think all the other fixtures use it.

arviz/tests/base_tests/test_data.py
```diff
@@ -77,6 +79,18 @@ def no_remote_data(monkeypatch, tmpdir):
...
+def data_random():
+    data = np.random.normal(size=(4, 500, 8))
+    idata = from_dict(
+        posterior={"a": data[..., 0], "b": data},
+        sample_stats={"a": data[..., 0], "b": data},
+        observed_data={"b": data[0, 0, :]},
+        posterior_predictive={"a": data[..., 0], "b": data},
+    )
+    return idata
```
OriolAbril, line 91: In views of `extend` method which will require multiple inference data objects, I propose the following pattern used in https://github.com/arviz-devs/arviz/blob/master/arviz/tests/helpers.py#L43:

```
def create_data_random(groups=None, seed=10):
    if groups is None:
        groups = ["posterior", "sample_stats", "observed_data", "posterior_predictive"]
    rng = np.random.default_rng(seed)
    data = rng.normal(size=(4, 500, 8))
    idata_dict = dict(
        posterior={"a": data[..., 0], "b": data},
        sample_stats={"a": data[..., 0], "b": data},
        observed_data={"b": data[0, 0, :]},
        posterior_predictive={"a": data[..., 0], "b": data},
        prior={"a": data[..., 0], "b": data},
        prior_predictive={"a": data[..., 0], "b": data},
    )
    idata = from_dict(**{group: ary for group, ary in idata_dict.items() if group in groups})
    return idata

@pytest.fixture(scope="module")
def data_random():
    idata = create_data_random()
    return idata
```

arviz/tests/base_tests/test_data.py
percygautam: The tests were failing for some cases in tests which used `inplace` like `isel`, as `inplace=True`is changing the data and datarandom was being generated only once if the scope is module. I changed the scope to function (default) so that data is generated everytime.

arviz/tests/base_tests/test_data.py
percygautam: Yeah, good idea!

arviz/data/inference_data.py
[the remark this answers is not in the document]
percygautam: @OriolAbril What about this one?

PR #1271, opened 2020-06-30
[Testing Do not merge] Add more verbosity to doc builds

canyon289:
## Description
Trying to figure out why master branch doc build is failing

Edit:
Figured it out. Here's the fix https://github.com/arviz-devs/arviz/pull/1272

PR #1272, opened 2020-06-30
Add missing ax that is breaking docs

canyon289:
Found out docs were broken because in the bokeh examples you have to return ax

Exploratory PR is here
https://github.com/arviz-devs/arviz/pull/1271


canyon289:
It wasnt obviou so it wasnt your fault, I still need to figure out way to raise a nicer warning, or least write a doc or something.

PR #1276, opened 2020-07-01
Update Council number

canyon289:
## Description
Update council number pending election results


canyon289:
Alright been a couple of days so merging!

PR #1281, opened 2020-07-05
distplot will convert list into numpy array automatically...

### LINKED ISSUE #1277: Automatic conversion of list type to ndarray
## Tell us about it
It would be nice if list types were automatically converted to ndarray in `az.plot_dist`

image

Shreyaansh-darkside:
## Description
fixes #1277

## Checklist

### DIFF
--- a/arviz/plots/distplot.py
+++ b/arviz/plots/distplot.py
@@ -2,6 +2,7 @@
 """Plot distribution as histogram or kernel density estimates."""
 
 import xarray as xr
+import numpy as np
 from .plot_utils import get_plotting_function, matplotlib_kwarg_dealiaser
 from ..numeric_utils import get_bins
 from ..data import InferenceData
@@ -145,6 +146,8 @@ def plot_dist(
 
         >>> az.plot_dist(b, rug=True, quantiles=[.25, .5, .75], cumulative=True)
     """
+    values = np.array(values) if str(type(values)) == "<class 'list'>" else values
+
     if isinstance(values, (InferenceData, xr.Dataset)):
         raise ValueError(
             "InferenceData or xarray.Dateset object detected,"


arviz/plots/distplot.py
```diff
@@ -145,6 +146,8 @@ def plot_dist(
 
         >>> az.plot_dist(b, rug=True, quantiles=[.25, .5, .75], cumulative=True)
     """
+    values = np.array(values) if str(type(values)) == "<class 'list'>" else values
```
canyon289, line 149: Thank you for the contribution1 

For feedback instead of comparing the string of the object could `isinstance` be used instead? See this stackoverflow answer for further details
https://stackoverflow.com/questions/1549801/what-are-the-differences-between-type-and-isinstance

As a matter of personal preference as well could this be expanded to be multi lines. Its clever to keep it on one line but harder to read

Shreyaansh-darkside:
Thanks for your valuable suggestion... I will look into it😃😃

arviz/plots/distplot.py
canyon289: Just to note this as well, please add a test to the testsuite to double check this functionality! If you need help let us know

Shreyaansh-darkside: Sir this was my first contribution, so i want to learn how to add test to the testsuite..

canyon289: Youre doing great! The first PR is the hardest so just bear with the comments and we'll get there. 

We use pytest with arviz. The docs on this website will show what those look like 
https://docs.pytest.org/en/stable/

When you get a handle of how pytest works you'll want to add a test that's similar to this one. The idea is to test the the list conversion works as expected. By writing these tests we can be sure ArviZ functionality is working as intended, both for this PR but also in the future
https://github.com/arviz-devs/arviz/blob/master/arviz/tests/base_tests/test_plots_matplotlib.py#L374

ahartikainen: I think this line should be replaced with

    values = np.asarray(values)

It does basically the same checks in it, but works with tuples too.

canyon289:
@Shreyaansh-darkside Still working on this one?

Samasth:
@canyon289 If this has not yet been done, I can add the test cases as this is very much related to the tests I added for dist_plot bins error.

canyon289:
@Shreyaansh-darkside Will keep this open for another 2 weeks, but if there is no more activity will close

canyon289:
@Shreyaansh-darkside This PR now has a merge conflict and is stale. I'm going to close

PR #1284, opened 2020-07-07
Add new KDE function.

tomicapretto:
## Description

This pull requests replaces the function `_fast_kde()` with a new function called `kde()` that can be used to estimate density functions for both linear and circular data.

### Major changes

* Everything related to KDE now lives in `kde_utils.py` instead of `numeric_utils.py`. The new function requires so many utilities itself that it is worth a new file. 
* The old function `_fast_kde()` used to return `density`, `xmin` and `xmax`, which required creating a new grid after estimating the density. 
The new function returns `grid` and `pdf`. It can also return the estimated bandwidth by calling `kde()` with the argument `bw_return` set to `True`.
* The new function `kde()` supports circular data via its boolean argument `circular`.
* The argument `bw` used to represent the bandwidth scaling factor in the old function `_fast_kde()`. Now it represents either the bandwidth method or the bandwidth itself.
It is still possible to pass a bandwidth scaling factor via the argument `bw_fct` which is 1 by default.
* `kde()` supports adaptive bandwidth for linear data. This is enabled by setting `adaptive` to `True`.
* `kde()` supports several bandwidth estimation methods (see below).
* The old `_fast_kde()` used to perform boundary correction without extending the observed domain in all the cases. 
This is also the default for `kde()` but these settings can be changed with the arguments `bound_correction`, `extend`, `extend_fct`, and `custom_lims`.

### Bandwidth estimation methods supported

For linear data:

* Scott's rule (`bw = 'scott'`).
* Silverman's rule (`bw = 'silverman'`).
* Improved Sheather-Jones plug-in method (`bw = 'isj'`).
* Expermiental, which is the average of Silverman's rule and the ISJ method (`bw = 'experimental'`). This is the default.

For circular data:

* Taylor's rule (`bw = 'taylor'`). It's like the Scott's rule of the circular world.

### Comments:

* `kde_utils.py` does not contain anything related to 2D KDE yet. It would be good to put all that here too.
* The `circular` argument is not passed to higher level functions yet. This is something to do in the near future.
* All the functions that used to call `_fast_kde()` now call `kde()` and have been updated to receive the new objects.

## Plots

Updating the KDE function impacts in many different plots. Here is an example that compares the old version versus the new one. Original is [here](https://arviz-devs.github.io/arviz/examples/matplotlib/mpl_plot_posterior.html).

**Old:**
old

**New:**
new

We also have the following comparison which inclues the new version, the old version and the KDE offered by scipy. 
5000 samples are used in all the cases.

comparison

## Checklist


arviz/kde_utils.py
```diff
@@ -0,0 +1,812 @@
...
+        t_j = np.power((c1 * (c2 / (N * f))), (2.0 / (3.0 + 2.0 * j)))
+        f = np.sum(k_sq ** j * a_sq * np.exp(-k_sq * np.pi ** 2.0 * t_j))
+        f *= 0.5 * np.pi ** (2 * j)
+
+    out = t - (2 * N * np.pi ** 0.5 * f) ** (-0.4)
+    return out
+
+
+def _root(function, N, args, x):
+    # The idea here was borrowed from KDEpy implementation.
```
aloctavodia, line 292: not clear what the idea actually is

arviz/kde_utils.py
```diff
@@ -0,0 +1,812 @@
...
+    # The right bound is at most 0.01
+    found = 0
+    N = max(min(1050, N), 50)
+    tol = 10e-12 + 0.01 * (N - 50) / 1000
+
+    while found == 0:
```
aloctavodia, line 298: `while not found` is a little bit more pythonic

arviz/kde_utils.py
```diff
@@ -0,0 +1,812 @@
...
+            bw, res = brentq(function, 0, 0.01, args=args, full_output=True,
+                             disp=False)
+            found = 1 if res.converged else 0
+        except ValueError:
+            bw = 0
+            tol *= 2.0
+            found = 0
+        if bw <= 0:
+            warnings.warn(
+                "Improved Sheather-Jones did not converge to a positive value. "
```
aloctavodia, line 309: This is something we have to discuss, do we want to potentially freak out users with these warnings? Do we want to silent them by default but still let the option to turn-on warnings? 
what about having the warnings but adding a section to the educational resources explaining the details of our KDE methods their use, limitations, etc (This is not something we need to do right now, not something that @tomicapretto must do).

arviz/kde_utils.py
```diff
@@ -0,0 +1,812 @@
...
+    while found == 0:
+        try:
+            bw, res = brentq(function, 0, 0.01, args=args, full_output=True,
+                             disp=False)
+            found = 1 if res.converged else 0
+        except ValueError:
+            bw = 0
+            tol *= 2.0
+            found = 0
+        if bw <= 0:
```
aloctavodia, line 307: This could be combined with the next condition, right? `if bw <= 0 or tol  >= 0` or putted in other way is it important for the user to distinguish between these two convergence errors?

arviz/kde_utils.py
```diff
@@ -0,0 +1,812 @@
...
+    Returns
+    -------
+    x : 1-D numpy array
+        If no error is thrown, a 1 dimensional array of
+        sample data from the variable for which a density estimate is desired.
+
+    """
+    if not isinstance(x, (list, np.ndarray)):
+        raise ValueError((
+            f"`x` is of the wrong type.\n"
```
aloctavodia, line 345: Maybe we can remove this first line. the second and third are clear enough

arviz/kde_utils.py
```diff
@@ -0,0 +1,812 @@
...
+            f"Can't produce a density estimator for {type(x)}.\n"
+            f"Please input a numeric list or numpy array."
```
aloctavodia, line 347: what about a list with numbers?

arviz/kde_utils.py
```diff
@@ -0,0 +1,812 @@
...
+    x : Object whose type is checked before computing the KDE.
+
+    Returns
+    -------
+    x : 1-D numpy array
+        If no error is thrown, a 1 dimensional array of
+        sample data from the variable for which a density estimate is desired.
+
+    """
+    if not isinstance(x, (list, np.ndarray)):
```
aloctavodia, line 343: This will fail for tuples, not sure we should allow them, just saying.

arviz/kde_utils.py
```diff
@@ -0,0 +1,812 @@
...
+        raise TypeError(
+            "Elements of `custom_lims` must be numeric or None, not bool.")
+
+    if custom_lims[0] is None:
+        custom_lims[0] = x_min
+
+    if custom_lims[1] is None:
+        custom_lims[1] = x_max
+
+    types = (int, float, np.integer, np.float)
```
aloctavodia, line 400: maybe just combine this with the next line.

arviz/kde_utils.py
```diff
@@ -0,0 +1,812 @@
...
+        Number of bins
+    grid_min: float
+        Minimum value of the grid
+    grid_max: float
+        Maximum value of the grid
+
+    """
+
+    # Set up number of bins. Should enable larger grids?
+    if grid_len > 4096:
```
aloctavodia, line 456: The kde_linear and kde_circular functions are setting grid_len=512.  Shouldn't then we just let the user pass the numbers they want and add a mention in the docstring about "reasonable ranges"

arviz/kde_utils.py
```diff
@@ -0,0 +1,812 @@
...
+
+    if bw_return:
+        return grid, pdf, bw
+    else:
+        return grid, pdf
+
+
+def kde_circular(
+    x,
+    bw="default",
```
aloctavodia, line 622: Could we have a different name? "circular scott", "taylor"...

aloctavodia:
thanks @tomicapretto for your contribution. This is a big one so I made a first round of revision and I will do a another one ASAP.

aloctavodia:
BTW I really like that in the example you posted the new kde is more wiggly than the current version :-)

arviz/kde_utils.py
```diff
@@ -0,0 +1,812 @@
...
+
+
+# Misc utils
+def vonmises_pdf(x, mu, kappa):
+    assert kappa > 0, "Argument 'kappa' must be positive."
+    pdf = 1 / (2 * np.pi * ive(0, kappa)) * np.exp(np.cos(x - mu) - 1) ** kappa
+    return pdf
+
+
+def circular_mean(x):
```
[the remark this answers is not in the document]
tomicapretto: Seems like custom function is faster (it was not my main goal when I implemented it, I just didn't know scipy had its own function)
Screenshot from 2020-07-17 00-31-40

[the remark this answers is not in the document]
tomicapretto: Seems like it is faster, but I implemented it just because I didn't know scipy had its own function.
Screenshot from 2020-07-17 00-31-40

arviz/kde_utils.py
tomicapretto: I'm not aware if people uses tuples for this kind of work. In my opinion they should not we allowed but that is just because I feel it is very unnatural. I think that we should accept it only if Arviz accepts tuples in similar circumstances.

arviz/kde_utils.py
tomicapretto: I think I'm missing something. The current implementation does accept a numeric list (which is a list with numbers, isn't it?). Do you mean something different?

arviz/kde_utils.py
tomicapretto: Something like this?

```python
 all_numeric = all(isinstance(i, (int, float, np.integer, np.float)) for i in custom_lims)
```

arviz/kde_utils.py
tomicapretto: Since Scott's rule of thumb is named after the guy who proposed the Gaussian rule of thumb, I guess Taylor's rule of thumb is fine for this Von mises rule of thumb in the circular setting. Then `taylor` sounds good to me. Thanks for the suggestion.

arviz/kde_utils.py
aloctavodia: Sorry for being unclear. I suggest to write f"Please input a list with numbers or numpy array."

arviz/kde_utils.py
aloctavodia: I agree tuples will be a little bit unnatural in this context.

aloctavodia:
It seems you need to rebase, there are a lot of commits unrelated to this PR

tomicapretto:
Comaprison between new version, old version and scipy kde. 5000 samples are used in all the cases.
comparison

CHANGELOG.md
```diff
@@ -3,9 +3,13 @@
 ## v0.x.x Unreleased
 ### New features
 * Added InferenceData dataset containing circular variables (#1265)
+* Replaced `_fast_kde()` with `kde()` which now also supports circular variables
```
aloctavodia, line 6: The sentence fits in a single line

CHANGELOG.md
```diff
@@ -3,9 +3,13 @@
...
+via the argument `circular` (#1284).
 
 ### Maintenance and fixes
 * plot_posterior: fix overlap of hdi and rope (#1263)
+* All the functions that used to call `_fast_kde`() now use `kde()` and have
```
aloctavodia, line 11: The sentence fits in a single line

arviz/kde_utils.py
```diff
@@ -187,11 +186,21 @@ def get_bw(x, bw, grid_counts=None, x_std=None, x_range=None):
         raise ValueError((
             f"Unrecognized `bw` argument.\n"
             f"Expected a positive numeric or one of the following strings:\n"
-            f"{list(BW_METHODS.keys())}."))
+            f"{list(BW_METHODS_LINEAR.keys())}."))
     return bw
 
 
-# Misc utils
+# Misc utils ------------------------------------------------------------------------
```
aloctavodia, line 193: remove the dashes

arviz/kde_utils.py
```diff
@@ -495,6 +504,84 @@ def kde(x, circular=False, **kwargs):
     grid : Gridded numpy array for the x values.
     pdf : Numpy array for the density estimates.
     bw: optional, the estimated bandwidth.
+
```
aloctavodia, line 507: Probably a better place for these examples is `plot_kde`. Also a circular example could be added to the example gallery

CHANGELOG.md
aloctavodia: Also in general use the 100 limit to break lines.

arviz/kde_utils.py
tomicapretto: Could it be tackled in another PR? 
I've updated `plot_kde()` and the other functions to use `kde()` but they are not passing extra arguments to `kde()` yet. When I tried to do it I saw it required to modify several things related to how the call is done and I preferred to keep this PR more focus on the KDE itself.

aloctavodia: yes, but open an issue as a reminder.

arviz/kde_utils.py
```diff
@@ -0,0 +1,895 @@
...
+    Takes angles in radians and normalize them to [-pi, pi) or [0, 2 * pi)
+    depending on `pi_centered`.
+    """
+    if pi_centered:
+        return x % (2 * np.pi)
+    else:
+        return (x + np.pi) % (2 * np.pi) - np.pi
+    
+def vonmises_pdf(x, mu, kappa):
+    assert kappa > 0, "Argument 'kappa' must be positive."
```
OriolAbril, line 205: I would use an if + raising value error here. I don't think an invalid kappa should raise an assertion error: https://docs.python.org/3/library/exceptions.html#ValueError

arviz/kde_utils.py
```diff
@@ -0,0 +1,895 @@
...
+
+
+def circular_mean(x):
+    sinr = np.sum(np.sin(x))
+    cosr = np.sum(np.cos(x))
+    mean = np.arctan2(sinr, cosr)
+    return mean
+
+
+def a1inv(x):
```
OriolAbril, line 217: another candidate for conditional jit

arviz/kde_utils.py
OriolAbril: shoudn't this use a boolean? instead of 0 or 1?

arviz/kde_utils.py
```diff
@@ -0,0 +1,895 @@
...
+                "Improved Sheather-Jones did not converge. "
+                "Using Silverman's rule instead.",
+                Warning
+            )
+            bw = (bw_silverman(x) / np.ptp(x)) ** 2
+    return bw
+
+# KDE Utilities ---------------------------------------------------------------------
+
+def check_type(x):
```
OriolAbril, line 335: I am not sure about what differences in behaviour has this versus some `np.asarray().flatten()` or `np.array().flatten()` (or maybe atleast_1d is needed?), I understand we want to make sure we get a numpy array and that the array is 1d right?

arviz/kde_utils.py
OriolAbril: You can link to plot_kde with rst to make sure people reading this docstring know to go there for examples: https://arviz-devs.github.io/arviz/developer_guide.html#hyperlinks They should both probably refer one another in the See also section.

arviz/kde_utils.py
```diff
@@ -0,0 +1,895 @@
...
+    else:
+        grid_min = x_min
+        grid_max = x_max
+    return grid_min, grid_max, grid_len
+
+# KDE Functions --------------------------------------------------------------------
+
+def kde(x, circular=False, **kwargs):
+    """
+    1 dimensional density estimation.
```
OriolAbril, line 488: again not sure if numpydoc lint will complain, we generally start docstrings in the same line as `"""`

arviz/kde_utils.py
```diff
@@ -0,0 +1,895 @@
...
+    pdf : Numpy array for the density estimates.
+    bw: optional, the estimated bandwidth.
+    """
+
+    # Check `x` is from appropiate type
+    x = check_type(x)
+
+    # Assert `bw_fct` is numeric and positive
+    assert isinstance(bw_fct, (int, float))
+    assert bw_fct > 0
```
OriolAbril, line 668: similar comment to one above, these two should probably be type error and value error respectively, right now they'd raise an assertion error without any custom message

OriolAbril: the isinstance int, float will also give problems with numpy dtypes i.e. float32

arviz/kde_utils.py
```diff
@@ -0,0 +1,895 @@
...
+import numpy as np
+from scipy.fftpack import fft
+from scipy.optimize import fsolve
+from scipy.optimize import brentq
+from scipy.signal import gaussian, convolve
+from scipy.special import ive
+
+from .stats.stats_utils import histogram
+
+# Bandwidth functions ---------------------------------------------------------------
```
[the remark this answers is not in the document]
tomicapretto: I think it was already corrected.

arviz/kde_utils.py
```diff
@@ -0,0 +1,895 @@
...
+# Linear KDE
+def bw_scott(x, x_std=None):
+    """
+    Scott's Rule
+    """
```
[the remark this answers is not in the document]
tomicapretto: Thanks for pointing this out. I'm going to update all similar case.

arviz/kde_utils.py
```diff
@@ -0,0 +1,895 @@
...
+    assert kappa > 0, "Argument 'kappa' must be positive."
+    pdf = 1 / (2 * np.pi * ive(0, kappa)) * np.exp(np.cos(x - mu) - 1) ** kappa
+    return pdf
+
+
+def circular_mean(x):
+    sinr = np.sum(np.sin(x))
+    cosr = np.sum(np.cos(x))
+    mean = np.arctan2(sinr, cosr)
+    return mean
```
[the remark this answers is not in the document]
tomicapretto: I am aware of Scipy version but I decided to keep this one because it is faster. As it is used now, things like nan checking are done in `kde_circular()` which is the function that calls `circular_mean()`. 
I guess you mention checks and all that because you're considering using the function in other parts of arviz.
I think there are two alternatives here:

* Implement a `circular_mean()` function that is going to be used everywhere in arviz, instead of Scipy version.
* Use Scipy version in other parts of arviz and keep this function for the circular kde since it is faster and safety checks have already been performed when `circular_mean()` is called. 

What is your opinion on this?

BTW, I'm not aware if Scipy `circmean()` is currently being used in other parts of arviz, I just guess this is the default alternative.

arviz/kde_utils.py
```diff
@@ -0,0 +1,895 @@
...
+    bw = 0.9 * a * len(x) ** (-0.2)
+    return bw
+
+
+def bw_isj(x, grid_counts=None, x_range=None):
+    """
+    Improved Sheather and Jones method as explained in [1]
+    This is an internal version pretended to be used by the KDE estimator.
+    When used internally computation time is saved because things like minimums,
+    maximums and the grid are pre-computed.
```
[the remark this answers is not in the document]
tomicapretto: Thanks, updating it!

arviz/kde_utils.py
```diff
@@ -0,0 +1,895 @@
...
+    """
+    if isinstance(bw, bool):
+        raise ValueError((
+            f"`bw` must not be of type `bool`.\n"
+            f"Expected a positive numeric or one of the following strings:\n"
+            f"{list(BW_METHODS_LINEAR.keys())}."))
+    if isinstance(bw, (int, float)):
+        if bw < 0:
+            raise ValueError(f"Numeric `bw` must be positive.\nInput: {bw:.4f}.")
+    elif isinstance(bw, str):
```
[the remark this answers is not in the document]
tomicapretto: Positive integer/float values are just returned. This is a simplified version

```python
def bw_check(bw):
    if isinstance(bw, bool):
        return "Bool"
    if isinstance(bw, (int, float)):
        if bw < 0:
            return "bw < 0"
    elif isinstance(bw, str):
        return "string_bw"
    else:
        return "unrecognized"
    return bw

print(bw_check(True))
print(bw_check(-1))
print(bw_check('abc'))
print(bw_check([1, 2]))
print(bw_check(1.5))

# Bool
# bw < 0
# string_bw
# unrecognized
# 1.5
```

arviz/kde_utils.py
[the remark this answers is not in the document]
OriolAbril: I think we use a circmean (not sure which) in summary with circular variables, I think we should have a single function used in both places. 

I am fine with keeping the fast one, given that it covers our use case and we have tests checking it gives the same result as scipy one.

arviz/kde_utils.py
tomicapretto: My concern with `np.asarray().flatten()` is that it works for something like `l = [[1, 2], [3, 4]]` but does not work for `l = [[1, 2], [3, 4, 5]]`.

```python
l = [[1, 2], [3, 4]]
np.array(l).flatten()
# array([1, 2, 3, 4])

l = [[1, 2], [3, 4, 5]]
np.array(l).flatten()
# array([list([1, 2]), list([3, 4, 5])], dtype=object)
```

With "works" I mean to return a 1d array of numbers. In the second case it is a 1d array but made of lists.
If I do `np.asfarray(x).flatten()` it works correctly in the first case and raises an error in the second.

I have the following ideas:

* Call `x = np.asfarray(x).flatten()` instead of the current `x = np.asfarray(x)` plus dimension number checking. If dimension of `x` is larger than 1 two things can happen. It is either successfully and silently flattened or an error is raised. 
* Call `x = np.asfarray(x)`. If dimension of `x` is larger than 1D and similar to the second example, it will raise an error. Otherwise it won't. Then check dimension number. If this is larger than 1, call `flatten()` method and raise a warning to inform the user the original object `x` was not 1-dimensional but was successfully converted. 
* Like the previous idea but without raising any warning if successfully flattening.

Personally, I'm more towards being more informative and signalling error/warnings. I think that if the user accidentally passes an object whose dimension is larger than 1 we should not resolve the situation silently - we can't be sure this is his/her original intention.
However, since this is my first contribution to the project and there may be other internal conventions that I'm not aware of,  I think it is good to follow your advise in this issue. Please let me know what would you do here.

arviz/kde_utils.py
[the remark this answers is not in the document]
tomicapretto: If you agree I can add the following function to `numeric_utils.py` and make use of it here. 
I will also add a test to check the result matches the one obtained with scipy (as far as I've manually tested, it does).

```python
def circular_mean(x, na_rm=False):   
    if len(x[np.isnan(x)]) > 0:
        if na_rm:
            x = x[~np.isnan(x)]
        else:
            return np.nan
    
    sinr = np.sum(np.sin(x))
    cosr = np.sum(np.cos(x))
    mean = np.arctan2(sinr, cosr)
    
    return mean
```

arviz/kde_utils.py
tomicapretto: I can replace it with the following

```python
if not isinstance(bw_fct, (int, float, np.integer, np.floating)):
    raise TypeError(f"`bw_fct` must be a positive number, not an object of {type(bw_fct)}.")
    
if not bw_fct > 0:
    raise ValueError(f"`bw_fct` must be a positive number, not {bw_fct}.")
```

arviz/kde_utils.py
```diff
@@ -0,0 +1,895 @@
...
+        bw_fun = select_bw_method(bw)
+        bw_lower = bw.lower()
+        if bw_lower == "isj":
+            bw = bw_fun(x, grid_counts, x_range)
+        elif bw_lower in ["scott", "silverman"]:
+            bw = bw_fun(x, x_std)
+        elif bw_lower == "experimental":
+            bw = bw_fun(x, grid_counts, x_std, x_range)
```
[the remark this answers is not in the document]
tomicapretto: This is one of the things I'm still not sure about. If we use the same signature for all the functions, I think they will look odd for a final user. But, on the other hand, are we expecting `bw_*` functions to be used by a final user? I think that if they are going to be used internally, we should use a common call signature so we avoid all these ifs and elifs. 

Another idea that just comes to my mind. Create two versions of every function. One is internal and the other is ready for the end user. The internal has the common signature with ignored args. Example: 

```python

def bw_silverman(x):
    return _bw_silverman(x)

def _bw_silverman(x, x_std=None, grid_counts=None, x_range=None):
    if x_std is None:
        x_std = np.std(x)
    q75, q25 = np.percentile(x, [75, 25])
    x_iqr = q75 - q25
    a = min(x_std, x_iqr / 1.34)
    bw = 0.9 * a * len(x) ** (-0.2)
    return bw
```

What are your opinions on this? cc @aloctavodia

arviz/kde_utils.py
ahartikainen: Doesn't floating work with python ints too?

arviz/kde_utils.py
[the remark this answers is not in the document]
OriolAbril: Sounds good, I would use `np.any` instead of slicing and then checking length, and can you open an issue on using numba for this? There are other functions that could benefit from it

arviz/kde_utils.py
OriolAbril: I think the second case raising an error is to be expected, and silently flattening is perfectly ok and coherent with the rest of ArviZ. 

`kde` will generally not be called by users, they will use plot_kde, plot_posterior... I think that kde, like plot_kde and plot_dist who also take an array input, is expected to generate a single kde even if the input is multidimensional (i.e. scalar variables will generally be stored as 2d arrays with (chain, draw) dims). To get multiple kdes, plot_posterior, plot_density... should be used

arviz/kde_utils.py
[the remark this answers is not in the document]
aloctavodia: what about

def circular_mean(x, na_rm=False):   
    if na_rm:
        x = x[~np.isnan(x)]
    
    sinr = np.sum(np.sin(x))
    cosr = np.sum(np.cos(x))
    mean = np.arctan2(sinr, cosr)
    
    return mean

arviz/kde_utils.py
tomicapretto: Is this what you mean? It returns `False`
```python
isinstance(1, np.floating)
# False
```

arviz/kde_utils.py
tomicapretto: Hmmm, I think I used two different warnings because I was interested in that. But if I think twice I see you are right and it is not something important to the user.

arviz/kde_utils.py
ahartikainen: Haha, my bad: with python floats

aloctavodia:
I think this is (almost) ready to merge, we can keep improving the code and its integration to the rest of ArviZ in future PRs. The only missing part is to add a `DeprecationWarning` indicating users calling `_fast_kde` to call instead `kde`. Also not sure if we should keep the dash or not. i.e. we should replace `_fast_kde` with `kde` or `_kde`. Thoughts @OriolAbril, @ahartikainen?

ahartikainen:
Do we want users to call kde? At least for now our kde code has been only for internal use, and I think there is a good reason for that --> there is a lot of coordination between input <<-->> output

For example scipy kde can calculate pdf etc, can predict any point user wants. Our is purely for plotting.

aloctavodia:
I sometimes call _fast_kde, but I think in general we do not want user to call it. So I would like to keep the dash.

arviz/numeric_utils.py
```diff
@@ -1,11 +1,24 @@
 """Numerical utility functions for ArviZ."""
 import warnings
 import numpy as np
-from scipy.signal import convolve, convolve2d
+from scipy.signal import convolve2d
 from scipy.sparse import coo_matrix
 
 from .stats.stats_utils import histogram
 from .utils import _stack, _dot, _cov
+from .kde_utils import _kde
```
OriolAbril, line 9: I think this is a cyclic import

tomicapretto: I moved things from kde_utils.py to numeric_utils.py (as suggested) but that gives us this cyclic import. It works well because I don't use a relative import in kde_utils, but I don't feel it is very elegant/proper. Any suggestions on how to improve it?

canyon289:
@tomicapretto I really appreciate your very detailed initial PR message with lots of graphs and explanations for code changes at a human level and plots for comparison. It helps quite a bit to help people like me understand whats going on and difference is, again very much appreciate the effort you put in!

arviz/numeric_utils.py
OriolAbril: given that `_fast_kde` is a private method, I think we can leave it in `numeric_utils` raising an error and pointing to `_kde`

OriolAbril: I am not sure if we should have `_kde` available again as `az._kde` though

aloctavodia: I think we should as we used to have _fast_kde

arviz/kde_utils.py
```diff
@@ -0,0 +1,863 @@
...
+    if not isinstance(x, (list, np.ndarray)):
+        raise ValueError((
+            f"Can't produce a density estimator for {type(x)}.\n"
+            f"Please input a list with numbers or numpy array."
+        ))
+    # Will raise an error if `x` can't be casted to numeric or flattened to one dimension.
+    try:
+        x = np.asfarray(x).flatten()
+    except Exception as e:
+        print(
```
OriolAbril, line 309: If we go with this, I'd use warning or logging here for this message.

arviz/kde_utils.py
```diff
@@ -0,0 +1,863 @@
...
+    x : 1-D numpy array
+        If no error is thrown, a 1 dimensional array of
+        sample data from the variable for which a density estimate is desired.
+
+    """
+    if not isinstance(x, (list, np.ndarray)):
+        raise ValueError((
+            f"Can't produce a density estimator for {type(x)}.\n"
+            f"Please input a list with numbers or numpy array."
+        ))
```
OriolAbril, line 304: I would remove this. I think it limits the usabiity without too much to gain. What if the input is a dask array? or a pytorch tensor? Moreover, I agree that a tuple is not natural in this context, but arguably, lists are not too natural either. 

Eventually we only want to make sure `x` is a numpy array, having informative error messages is good, but our users already know numpy so I think we can trade a little on this front for flexibility and maintainability.

arviz/kde_utils.py
```diff
@@ -0,0 +1,863 @@
...
+    """
+
+    # Set up number of bins.
+    if grid_len < 100:
+        grid_len = 100
+    grid_len = int(grid_len)
+
+    # Set up domain
+    # `custom_lims` overrides `extend`
+    # `bound_correction` overrides `extend`
```
OriolAbril, line 421: I'd move this to the docstring

arviz/kde_utils.py
```diff
@@ -0,0 +1,863 @@
...
+            "to a 1 dimensional float array."
+        )
+        raise e
+
+    x = x[np.isfinite(x)]
+
+    if x.size == 0:
+        raise ValueError("`x` does not contain any finite number.")
+    if x.size == 1:
+        raise ValueError("`x` is of length 1. Can't produce a KDE with only one data point.")
```
OriolAbril, line 320: should 2 data points be allowed? maybe we should set a threshold somewhere and between 2 and threshold issue a warning?

aloctavodia: I think it should be allowed. Both because this was the previous behavior and because it's the minimum number for the method to work. We could have threshold in other places, like we already have in plot_ppc or as we could add to plot_dist_comparison

arviz/kde_utils.py
tomicapretto: So the idea would be not to check the type of the input and let numpy signal the error, if necessary, when casting to numpy array?

arviz/kde_utils.py
tomicapretto: An initial version threw a warning when the size was below 30 or 50 but then I decided to remove it. I think I did it because of the examples using eight schools data (which would have caused warnings everywhere). I don't have any problem in adding warnings, but first I would like to check if others have any opinions on this.

arviz/kde_utils.py
OriolAbril: yes, numpy is the base library for arrays in python and all the other play well with it, `np.array(ary)` will generally work for dask, jax, xarray, pytorch, tensorflow... objects

tomicapretto: Great, makes much sense. I'll change it.

arviz/kde_utils.py
```diff
@@ -297,17 +297,12 @@ def _check_type(x):
...
-            f"Please input a list with numbers or numpy array."
-        ))
     # Will raise an error if `x` can't be casted to numeric or flattened to one dimension.
     try:
         x = np.asfarray(x).flatten()
     except Exception as e:
-        print(
-            "The following exception occurred while trying to convert `x`",
+        warnings.warn(
+            "The following exception occurred while trying to convert `x`" +
```
OriolAbril, line 305:
```suggestion
            "The following exception occurred while trying to convert `x`"
```

tomicapretto:
@aloctavodia, @OriolAbril, @ahartikainen and @canyon289 thank you all for the reviews and the work!!

ahartikainen:
Thanks for the PR!

tomicapretto:
Hi @joelostblom thanks for the words of recognition! It was a very challenging and fun work. 

The sustain behind the experimental bandwidth was a simulation study that I performed. I wrote a report and uploaded it to my _old_ webpage. Then I migrated it and that was lost (sorry!). But I went to the repository and found it in an old commit. So have a look at the `report.html` here https://github.com/tomicapretto/webpage/tree/5c5df838ed90fcb676d26904885740faf148fd0b/content/files.

joelostblom:
Super, thank you so much for digging that up! That also cleared up a misunderstanding I had around how the adaptive kernel works. It's interesting that Scott and Silverman are used so widely although they appear to have clear oversmoothing drawbacks, especially for sample sizes <~5,000 - 10,000 observations. Thanks again!

PR #1289, opened 2020-07-13
Add data storage to project description

kyleabeauchamp:
I think the xarray-backed trace data definition is actually a key part of the project and probably deserves to live in the one-liner project description.


OriolAbril:
Good idea, thanks! Can you also update https://github.com/arviz-devs/arviz/blob/master/doc/index.rst so docs also show it?

kyleabeauchamp:
Done, see #1290

PR #1290, opened 2020-07-13
Update index.rst

kyleabeauchamp:
Same as #1289 but in index.rst


canyon289:
Thanks @kyleabeauchamp!

PR #1223, opened 2020-06-04
Add to_dict method to InferenceData object

percygautam:
## Description
This PR adds the `to_dict` method to InferenceData object. 

## Checklist

- [x] Follows [official](https://github.com/arviz-devs/arviz/blob/master/CONTRIBUTING.md#pull-request-checklist) PR format
- [x] New features are properly documented (with an example if appropriate)?
- [x] Includes new or updated tests to cover the new feature
- [x] Code style  correct (follows pylint and black guidelines)
- [x] Changes are listed in [changelog](https://github.com/arviz-devs/arviz/blob/master/CHANGELOG.md#v0xx-unreleased)

PR #1300, opened 2020-07-16
Add custom methods for InferenceData

### LINKED ISSUE #1239: add merge/add group function
## Tell us about it
`az.concat` currently raises an error if there are repeated groups which is perfectly fine, however, there are situations where one wants to add the different groups to the original idata object and discard the repeated groups. 

related to #1066 

## Thoughts on implementation

several options, some of them complementary:

1. add an `idata.merge` function so that idata wiht groups post, sample_stats, obs_data can be merged with idata with groups post_pred and obs_data discarding the obs_data in one of the two (could have left/right modes)
2. add an `idata.add_groups` so that we can pass it dicts (calls dict_to_dataset) or datasets as kwargs and they are added to idata (I think kwargs is important to allow special usecases and support any group_name, alternatively a mapping {group_name: ds/dict} could be used)
3. make `idata.concat` check equality of datasets with isclose on both vals and coords (but not on attrs! they will surely have different creation time)

percygautam:
## Description
In continuation with the #1223 and #1254 . Related to #1066 . FIx #1239 

## Checklist

### DIFF
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -12,6 +12,7 @@ import numpy as np
 import xarray as xr
 from xarray.core.options import OPTIONS
 
+from .base import dict_to_dataset
 from ..utils import _subset_list, HtmlTemplate
 from ..rcparams import rcParams
 
@@ -323,6 +324,59 @@ class InferenceData:
             return None
         else:
             return out
+
+    def add_groups(self, group_dict=None, coords=None, dims=None, **kwargs):
+        """Add new groups to InferenceData object.
+
+        Parameters
+        ----------
+        group_dict: dict[str] -> dict, xr.Dataset
+            Groups to be added
+        coords : dict[str] -> ndarray
+            Coordinates for the dataset
+        dims : dict[str] -> list[str]
+            Dimensions of each variable. The keys are variable names, values are lists of
+            coordinates.
+        **kwargs: mapping
+            The keyword arguments form of group_dict. One of group_dict or kwargs must be provided.
+
+        """
+        if group_dict is None:
+            group_dict = {}
+
+        group_dict.update(kwargs)
+        if not group_dict:
+            raise ValueError("One of group_dict or kwargs must be provided.")
+        for group, dataset in group_dict.items():
+            if group in self._groups_all:
+                warnings.warn(
+                    "{group} group already exists. Continuing without adding the group {group}".format(
+                        group=group
+                    ),
+                    UserWarning,
+                )
+            if group not in SUPPORTED_GROUPS_ALL:
+                warnings.warn(
+                    "{} group is not defined in the InferenceData scheme".format(group), UserWarning
+                )
+            if dataset is None:
+                continue
+            elif isinstance(dataset, dict):
+                dataset = dict_to_dataset(dataset, coords=coords, dims=dims)
+            elif not isinstance(dataset, xr.Dataset):
+                raise ValueError(
+                    "Arguments to add_groups() must be xarray Datasets or dicts"
+                    "(argument '{}' was type '{}')".format(group, type(dataset))
+                )
+            if not group.startswith(WARMUP_TAG):
+                if dataset:
+                    setattr(self, group, dataset)
+                    self._groups.append(group)
+            elif group.startswith(WARMUP_TAG):
+                if dataset:
+                    setattr(self, group, dataset)
+                    self._groups_warmup.append(group)
+        return None
 
     def _group_names(self, groups, filter_groups=None):
         """Handle expansion of group names input across arviz.
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -345,12 +345,13 @@ class InferenceData:
         group_dict = either_dict_or_kwargs(group_dict, kwargs, "add_groups")
         if not group_dict:
             raise ValueError("One of group_dict or kwargs must be provided.")
+        repeated_groups = [group for group in group_dict.keys() if group in self._groups]
+        if repeated_groups:
+            raise ValueError("{} group(s) already exists.".format(repeated_groups))
         for group, dataset in group_dict.items():
-            if group in self._groups_all:
-                raise ValueError("{} group already exists.".format(group))
             if group not in SUPPORTED_GROUPS_ALL:
                 warnings.warn(
-                    "{} group is not defined in the InferenceData scheme".format(group), UserWarning
+                    "The group {} is not defined in the InferenceData scheme".format(group), UserWarning
                 )
             if dataset is None:
                 continue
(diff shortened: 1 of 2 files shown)


percygautam:
Added and tested the method `add_groups` to add groups to the idata. Following is the usage examples of the method:
```
data = np.random.normal(size=(4, 500, 8))
idata = from_dict(
    posterior={"a": data[..., 0], "b": data},
    sample_stats={"a": data[..., 0], "b": data},
    observed_data={"b": data[0, 0, :]},
    posterior_predictive={"a": data[..., 0], "b": data},
)
idata.add_groups({"prior": {"a": data[..., 0], "b": data}}) #one way
idata.add_groups({"new_group": idata.posterior}) #other way
```

arviz/data/inference_data.py
```diff
@@ -324,6 +325,59 @@ def sel(
         else:
             return out
 
+    def add_groups(self, group_dict=None, coords=None, dims=None, **kwargs):
+        """Add new groups to InferenceData object.
+
+        Parameters
+        ----------
+        group_dict: dict[str] -> dict, xr.Dataset
```
OriolAbril, line 333:
```suggestion
        group_dict: dict of {str : dict or xarray.Dataset}, optional
```

OriolAbril: The above is following the guidance on pandas docstring guide, however, pandas docs also use the pattern `dict of str -> dict or xarray.Dataset`. Here I kind of prefer the first alternative due to the or, in general both are fine.

arviz/data/inference_data.py
```diff
@@ -324,6 +325,59 @@ def sel(
...
+            Dimensions of each variable. The keys are variable names, values are lists of
+            coordinates.
+        **kwargs: mapping
+            The keyword arguments form of group_dict. One of group_dict or kwargs must be provided.
+
+        """
+        if group_dict is None:
+            group_dict = {}
+
+        group_dict.update(kwargs)
```
OriolAbril, line 347: I would try to mimic xarray, may be even use [this same function](https://github.com/pydata/xarray/blob/1be777fe725a85b8cc0f65a2bc41f4bc2ba18043/xarray/core/utils.py#L256). This implementation behaves differently from xarray in the sense that kwargs overwrite group_dict whereas in xarray both of them being present raises an error

OriolAbril: see for example [isel](https://github.com/pydata/xarray/blob/master/xarray/core/dataarray.py#L1042) as an example of how dict/kwargs are implemented

arviz/data/inference_data.py
```diff
@@ -324,6 +325,59 @@ def sel(
...
+        if not group_dict:
+            raise ValueError("One of group_dict or kwargs must be provided.")
+        for group, dataset in group_dict.items():
+            if group in self._groups_all:
+                warnings.warn(
+                    "{group} group already exists. Continuing without adding the group {group}".format(
+                        group=group
+                    ),
+                    UserWarning,
+                )
```
OriolAbril, line 357: isn't a `continue` needed here to actually skip the group? I am more inclined towards erroring though, more or less like concat does.

arviz/data/inference_data.py
```diff
@@ -324,6 +325,59 @@ def sel(
...
+            if group not in SUPPORTED_GROUPS_ALL:
+                warnings.warn(
+                    "{} group is not defined in the InferenceData scheme".format(group), UserWarning
+                )
+            if dataset is None:
+                continue
+            elif isinstance(dataset, dict):
+                dataset = dict_to_dataset(dataset, coords=coords, dims=dims)
```
OriolAbril, line 365: this will not work for groups without `chain`, `draw` dims. I think a possible fix could be provided by #1201, until then maybe documenting that is enough

arviz/data/inference_data.py
```diff
@@ -324,6 +325,59 @@ def sel(
...
+                    "(argument '{}' was type '{}')".format(group, type(dataset))
+                )
+            if not group.startswith(WARMUP_TAG):
+                if dataset:
+                    setattr(self, group, dataset)
+                    self._groups.append(group)
+            elif group.startswith(WARMUP_TAG):
+                if dataset:
+                    setattr(self, group, dataset)
+                    self._groups_warmup.append(group)
```
OriolAbril, line 378:
```suggestion
            if dataset:
                setattr(self, group, dataset)
                if group.startswith(WARMUP_TAG):
                    self._groups_warmup.append(group)
                else:
                    self._groups.append(group)
```

arviz/tests/base_tests/test_data.py
```diff
@@ -480,6 +480,28 @@ def test_repr_html(self):
         assert escape(idata.__repr__()) in html
         xr.set_options(display_style=display_style)
 
+    def test_add_groups(self):
```
OriolAbril, line 483: I would split that into several functions, one function to test "normal" behaviour, one for error, another for warning...

arviz/tests/base_tests/test_data.py
```diff
@@ -480,6 +480,28 @@ def test_repr_html(self):
...
+        idata = from_dict(
+            posterior={"a": data[..., 0], "b": data},
+            sample_stats={"a": data[..., 0], "b": data},
+            observed_data={"b": data[0, 0, :]},
+            posterior_predictive={"a": data[..., 0], "b": data},
+        )
+        with pytest.raises(ValueError):
+            idata.add_groups()
+        idata.add_groups({"prior": {"a": data[..., 0], "b": data}})
+        assert "prior" in idata._groups
```
OriolAbril, line 494: assert both `prior` is in `_groups` and that `prior` is an attribute of `idata`, also a warmup group should also be checked

arviz/data/inference_data.py
percygautam: Sorry, it was intended to add continue here. My mistake. Will raise error!

arviz/data/inference_data.py
percygautam: Should I add a warning for this?

OriolAbril: sounds good

percygautam:
Added and tested the method `merge` to merge two idata objects. Following is the usage examples of the method:
```
data = np.random.normal(size=(4, 500, 8))
idata = from_dict(
    posterior={"a": data[..., 0], "b": data},
    sample_stats={"a": data[..., 0], "b": data},
    observed_data={"a": data[0, 0, :]},
    posterior_predictive={"a": data[..., 0], "b": data},
)
idata2 = from_dict(
    prior={"a": data[..., 0], "b": data},
    prior_predictive={"a": data[..., 0], "b": data},
    observed_data={"b": data[0, 0, :]},
)
idata.merge(idata2) 
idata.merge(idata2, join=right)
```

arviz/data/inference_data.py
```diff
@@ -324,6 +326,136 @@ def sel(
...
+                )
+            if dataset:
+                setattr(self, group, dataset)
+                if group.startswith(WARMUP_TAG):
+                    self._groups_warmup.append(group)
+                else:
+                    self._groups.append(group)
+        return None
+
+    def merge(self, other, join="left"):
```
OriolAbril, line 431: I like the function, not sure about the name though, I know it was my proposal in the issue, but I think we can find a better one, merge will probably be confusing

arviz/data/inference_data.py
```diff
@@ -324,6 +326,136 @@ def sel(
...
+        **kwargs: mapping
+            The keyword arguments form of group_dict. One of group_dict or kwargs must be provided.
+
+        """
+        group_dict = either_dict_or_kwargs(group_dict, kwargs, "add_groups")
+        if not group_dict:
+            raise ValueError("One of group_dict or kwargs must be provided.")
+        for group, dataset in group_dict.items():
+            if group in self._groups_all:
+                raise ValueError("{} group already exists.".format(group))
```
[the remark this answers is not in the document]
percygautam: Yeah, makes sense. Thanks!

arviz/data/inference_data.py
```diff
@@ -324,6 +326,136 @@ def sel(
...
+                )
+            if dataset:
+                setattr(self, group, dataset)
+                if group.startswith(WARMUP_TAG):
+                    self._groups_warmup.append(group)
+                else:
+                    self._groups.append(group)
+        return None
+
+    def add_groups(self, group_dict=None, coords=None, dims=None, **kwargs):
```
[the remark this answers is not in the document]
percygautam: Yeah, no idea how!!

arviz/data/inference_data.py
percygautam: Yes. How does `join` or `extend` sound?

OriolAbril: I like extend!

arviz/data/inference_data.py
```diff
@@ -627,6 +629,91 @@ def rename_dims(self, name_dict=None, groups=None, filter_groups=None, inplace=F
...
+    def extend(self, other, join="left"):
+        """Merge InferenceData object with other InferenceData object.
+
+        Parameters
+        ----------
+        other : InferenceData object
+            InferenceData to be added
+        join : str, optional
+            defines the type of join. Can be either 'left' or 'right'.
+            Default is 'left'.
```
OriolAbril, line 699:
```suggestion
        join : {'left', 'right'}, default 'left' 
            Defines how the two decide which group to keep when the same group is
            present in both objects. 'left' will discard the group in ``other`` whereas 'right' 
            will keep the group in ``other`` and discard the one in ``self``.
```

arviz/data/inference_data.py
```diff
@@ -627,6 +629,91 @@ def rename_dims(self, name_dict=None, groups=None, filter_groups=None, inplace=F
...
+
+        """
```
OriolAbril, line 701: Can you add a See also section linking to `concat` and to `add_groups`. Similar thing for the other two so that they link each other.

arviz/tests/base_tests/test_data.py
```diff
@@ -584,6 +584,99 @@ def test_repr_html(self):
...
         xr.set_options(display_style=display_style)
 
+    def test_add_groups(self):
+        data = np.random.normal(size=(4, 500, 8))
+        idata = from_dict(
+            posterior={"a": data[..., 0], "b": data},
+            sample_stats={"a": data[..., 0], "b": data},
+            observed_data={"b": data[0, 0, :]},
+            posterior_predictive={"a": data[..., 0], "b": data},
+        )
```
OriolAbril, line 594: can you use the fixture from the other PR to simplify and reduce duplicated code? You can use the fixture and (for `extend`) create `idata2 = create_...` to get a 2nd idata which is different from the fixture

arviz/data/inference_data.py
```diff
@@ -627,6 +629,91 @@ def rename_dims(self, name_dict=None, groups=None, filter_groups=None, inplace=F
...
+        if not isinstance(other, InferenceData):
+            raise ValueError("Merging is possible between two InferenceData objects only.")
```
OriolAbril, line 703: I think we should avoid merging to describe what `extend` does. Maybe extending or combining

arviz/tests/base_tests/test_data.py
percygautam: Yeah, I will. I was waiting for other PR to merge and then forget altogether.

percygautam:
I think we can merge this and add other custom methods as the need arise.

PR #1307, opened 2020-07-19
Partial mpl 3.3 fixes

ahartikainen:
## Description

There are certain errors we need to fix

- [x] Fix vlines/hlines ymin array size **(compatible with mpl<3.3.0)**
- [x] Fix empty subplots error with figure call **(compatible with mpl<3.3.0)**
~~- [ ] Fix duplicate default parameter problem **(compatible with mpl<3.3.0)**~~
~~- [ ] fix changed return types (need to be careful to be compatible with previous versions of mpl)~~

I think the first two are fixed, but we need to update and fix the last point by thinking about how we define the default parameters. For example test are failing due to both `c` and `color` are defined. This used to be ignored, but in 3.1 it started to give a warning, and now it raises an exception.

Is there any mpl function we can use or do we want to manually create one. (There are not too many defaults we currently use). We also should use the main kw for each function. For example in color case, most mpl functions use color, but some use colors (e.g. vlines).

- `linewidth`, `linewidths`, `lw` --> `linewidth`/`linewidths`
- `color`, `colors`, `c` --> `color`, `colors`

edit.

This PR only fixes first two bugs.

### DIFF
--- a/arviz/plots/backends/matplotlib/jointplot.py
+++ b/arviz/plots/backends/matplotlib/jointplot.py
@@ -33,7 +33,7 @@ def plot_joint(
     }
     if ax is None:
         # Instantiate figure and grid
-        fig, _ = plt.subplots(0, 0, figsize=figsize, **backend_kwargs)
+        fig = plt.figure(figsize=figsize, **backend_kwargs)
         grid = plt.GridSpec(4, 4, hspace=0.1, wspace=0.1, figure=fig)
 
         # Set up main plot
--- a/arviz/plots/backends/matplotlib/pairplot.py
+++ b/arviz/plots/backends/matplotlib/pairplot.py
@@ -113,7 +113,7 @@ def plot_pair(
                 # Instantiate figure and grid
                 widths = [2, 2, 2, 1]
                 heights = [1.4, 2, 2, 2]
-                fig, _ = plt.subplots(0, 0, figsize=figsize, **backend_kwargs)
+                fig = plt.figure(figsize=figsize, **backend_kwargs)
                 grid = plt.GridSpec(
                     4,
                     4,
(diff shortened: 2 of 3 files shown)


OriolAbril:
I think using [`matplotlib_kwarg_dealiaser`](https://github.com/arviz-devs/arviz/blob/master/arviz/plots/plot_utils.py#L657) function should fix the issue, we just have to make sure all kwargs are properly dealiased and that we always set the defaults with the "default" key.

ahartikainen:
Will that work also for bokeh keywords? Or when backend is bokeh? Or do we even care in that case?

If that function passes "unknown" keywords, we probably can use it, otherwise use it only for mpl?

Will that remove duplicates or only transform them to the correct key?

OriolAbril:
As far as I know, bokeh has no aliases and uses different arguments (i.e. line_color instead of color),thus, bokeh kwargs are left unmodified by dealiaser function and there is a test on that.

Regarding duplicates no idea actually but given that our function is a wrapper on matplotlib's, the behaviour should be the same as mpl 3.3

ahartikainen:
Ok, then just let add call for it before we specify defaults.

ahartikainen:
Oh yes, animation has some bug (some function now returns a line not a list)

ahartikainen:
I increased the mpl kwargs fix, but not sure what should be our logic of using it. What functions add defaults and what functions don't?

I also increased the flexibility a bit for it.

ahartikainen:
This is starting to become a really ugly fix. I think we need to update how we handle the defaults.

I think we need to rething that. Let's move mpl defaults under mpl code? And then move bokeh defaults to bokeh code?

Is there a possibility that backend mpl code doesn't call main funcs anymore but calls only the backend mpl code, same for the bokeh.

Then this would mean that the main functions could be really simple.

canyon289:
I agree, I think having backend specific code in "main" plotting methods is a bad thing. It should all be in backend code

ahartikainen:
I will merge the small fixes. The rest need another PR.

Issue #339, opened 2018-10-09
summary for circular variables


aloctavodia:
Summary used to have a `circ_var_names` argument in order to specify for which variables to compute circular statistics instead of the regular ones (#45). This argument was replaced by `include_circ`, this accept booleans, when `True` it will compute the regular statistics plus the circular statistics for all variables. I think we should revert this and have `circ_var_names` again, or maybe some other alternative.

The ideal behavior will be to automatically compute the proper statistics based on the variable type, but at least for the moment that information is not available for ArviZ

andy6975:
Is this issue open for work? I would like to work on this. Thanks!

aloctavodia:
Yeah, this is still an issue. Thanks for your offer.

smit-s:
@aloctavodia  If this issue is still open, can I work on it please?

aloctavodia:
Is still open, I think @agustinaarroyuelo  was planning to work on this as part of the GSoC (but she also is proposing many other tasks).

smit-s:
No problem at all. I will pick up regression plot and time-series.

PR #1313, opened 2020-07-22
Adapt summary function to circular variables 

agustinaarroyuelo:
## Description

In this PR I changed the `az.summary` function to display statistics for circular variables more neatly. It is related to issue #339. 

`torsionals = az.load_arviz_data('glycan_torsion_angles')`
`az.summary(torsionals, circ_var_names=['tors'])` 

outputs:

sc_summary

## Checklist
- [X] Includes new or updated tests to cover the new feature

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -1270,7 +1270,7 @@ def summary(
 
         if kind != "diagnostics":
             for metric, circ_stat in zip(
-                # Replace the first 5 statistics
+                # Replace only the first 5 statistics for their circular equivalent
                 metrics[:5],
                 (circ_mean, circ_sd, circ_hdi_lower, circ_hdi_higher, circ_mcse),
             ):
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -1178,51 +1178,38 @@ def summary(
         hdi_higher = hdi_post.sel(hdi="higher", drop=True)
 
     if circ_var_names:
-        circ_mean = []
-        circ_sd = []
-        circ_mcse = []
-        circ_hdi_lower = []
-        circ_hdi_higher = []
+        nan_policy = "omit" if skipna else "propagate"
+        circ_mean = xr.apply_ufunc(
+            _make_ufunc(st.circmean),
+            posterior,
+            kwargs=dict(high=np.pi, low=-np.pi, nan_policy=nan_policy),
+            input_core_dims=(("chain", "draw"),),
+        )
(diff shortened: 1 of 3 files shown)


arviz/stats/stats.py
```diff
@@ -992,7 +992,7 @@ def summary(
     fmt: str = "wide",
     kind: str = "all",
     round_to=None,
-    include_circ=None,
+    circ_var_names=None,
```
OriolAbril, line 995: I think we should keep `include_circ` for now and raise a deprecation warning when used to ease transition

agustinaarroyuelo:
Thanks for your comments @OriolAbril and @ahartikainen. I updated the PR description. As you can see, circular statistics are placed only where the circular variables are, so no need for extra columns. 

Maybe waiting for the new example to update the doctrings will be better but I will check with @aloctavodia

ahartikainen:
That looks great.

This PR doesn't need to address this, but should we add a column for `is_circular` and use True for variables that are?
Normal summary doesn't need this, but maybe when we use circular variables it might be a good idea.

Also (I forgot already this) are those in degrees or in radians?

arviz/stats/stats.py
```diff
@@ -1258,17 +1265,16 @@ def summary(
...
-                "circular_hdi_{:g}%".format(100 * (1 - alpha / 2)),
-                "circular_mcse",
-            )
-        )
+
+    if circ_var_names:
+
+        if kind != "diagnostics":
+            for metric, circ_stat in zip(
+                metrics[:5], (circ_mean, circ_sd, circ_hdi_lower, circ_hdi_higher, circ_mcse)
```
canyon289, line 1273: Whats 5 here? Would add comment for the magic number

aloctavodia: Agree about the comment. 
The 5 is related to the "stats" columns, the diagnostics are computed the same for linear and circular variables. Actually, we have to think if the diagnostics also need to be changed for circular variables (not part of this PR).

aloctavodia:
Not sure about having a "true column" for circular variables. I think a better approach is to have that information included in the InferenceData object.

agustinaarroyuelo:
> That looks great.
> 
> This PR doesn't need to address this, but should we add a column for `is_circular` and use True for variables that are?
> Normal summary doesn't need this, but maybe when we use circular variables it might be a good idea.
> 
> Also (I forgot already this) are those in degrees or in radians?

This angles are in radians

PR #1323, opened 2020-07-27
Move contributing blurb up

canyon289:
Make contributing blurb more prominent


canyon289:
Self merging so this deploys today

PR #1326, opened 2020-07-29
new gray scale

aloctavodia:
This is a small modification of the C0-C3  colors in arviz-grayscale. The C3 color was a little bit difficult to see specially if printed.


canyon289:
Rebased on fixed master

PR #1327, opened 2020-07-29
reduce namespace pollution

aloctavodia:
Gets rid of potentially confusing/annoying things like `az.os`. Not sure if there is a more pythonic way to do it.

### DIFF
--- a/arviz/__init__.py
+++ b/arviz/__init__.py
@@ -2,27 +2,28 @@
 """ArviZ is a library for exploratory analysis of Bayesian models."""
 __version__ = "0.9.0"
 
-import os
-import logging
-from matplotlib.pyplot import style, register_cmap
-from matplotlib.colors import LinearSegmentedColormap
+from os import path as _path
+import logging as _logging
+from matplotlib.pyplot import style
+from matplotlib.pyplot import register_cmap as _register_cmap
 
+from matplotlib.colors import LinearSegmentedColormap as _LinearSegmentedColormap
 
 # add ArviZ's styles to matplotlib's styles
-arviz_style_path = os.path.join(os.path.dirname(__file__), "plots", "styles")
-style.core.USER_LIBRARY_PATHS.append(arviz_style_path)
+_arviz_style_path = _path.join(_path.dirname(__file__), "plots", "styles")
+style.core.USER_LIBRARY_PATHS.append(_arviz_style_path)
 style.core.reload_library()
 
 # Configure logging before importing arviz internals
-_log = logging.getLogger("arviz")
+_log = _logging.getLogger("arviz")
 
 
-if not logging.root.handlers:
-    handler = logging.StreamHandler()
-    formatter = logging.Formatter("%(name)s - %(levelname)s - %(message)s")
-    handler.setFormatter(formatter)
-    _log.setLevel(logging.INFO)
-    _log.addHandler(handler)
+if not _logging.root.handlers:
+    _handler = _logging.StreamHandler()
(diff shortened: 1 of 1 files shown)


canyon289:
I think `__all__` is the python way to do it and ensure its actually hidden not just hidden under underscore. I can try in a bit

ahartikainen:
`__all__` is for `from x import *`-

`_item` is a better way

aloctavodia:
Right `__all__` will not work for this. So @ahartikainen are you saying this is the way to go?

ColCarroll:
You can also use `del os`

ColCarroll:
(or use functions that do the importing internally)

canyon289:
> You can also use `del os`

If we do this why bother importing it at all?

ColCarroll:
Import `os`, do what you need with it, then delete it. It will no longer be in the namespace.

ahartikainen:
wait, why does that work? Each function is created at import step?

canyon289:
> wait, why does that work? Each function is created at import step?

Im also confused how this works with other modules that then import OS or whatever but I was willing to believe Colin it worked and the tests seem to pass so it looks like it works

I like the `__all__` solution better because that one is much more intentional but I'm also not the one writing this PR so I'll approve whatever is step in good direction

aloctavodia:
I hope this clarifies this PR.

Suppose we have a package with the name foo consisting of two modules:

__init__.py
```
import os
from .do import *

def hi():
    print(os.getcwd())
hi()

del os
```

and do.py

``ìmport os``

After importing `foo` we will get the working directory printed on the screen and we will have access to the function `foo.hi` and the module `foo.do`, but not `foo.os` which will be available only as `foo.do.os`.  If we instead use the  `__all__` pattern `foo.os` will be still available.

The effect of the `del` command is restricted to the module  in which is called, the function `foo.hi` will run only once (when foo is imported) and then will fail because `os` is no longer available at the `foo.` level. But we will be able to call `foo.do.os`

ahartikainen:
I don't get that second part. We can not call `foo.hi` after the import?

aloctavodia:
No, but we don't want to call it. I just add it to the exemplify the local effect of `del`. In practice we will also want to remove `hi` from the namespace.

ahartikainen:
I'm still confused. So what you are saying:

after `del mod` we can not use funcs with `mod`

aloctavodia:
that's right. so you can use del to remove everything that you do not want the use to call from arviz like `os` or `np`. But you can still keep things like `style`.

ahartikainen:
Still doesn't make sense

So why would we want to remove `np` from namespace if functions using `np` would fail?

aloctavodia:
In my example the user can still use `os` either if they call `foo.do.os` or if they do  `import os`. The only thing del is doing is removing `os` from `foo.` 

What happens in a namespace stays in that namespace.

ahartikainen:
Yeah, so `foo.hi` should still work, after the `del os` is called.

aloctavodia:
`hi` can only be executed once before removing `os`. that example is precisely to show that `del` is removing `os` from `__intit__.py` but not affecting the `os` from `do`.

ahartikainen:
Good, I get it now. :D

canyon289:
Any remaining code pushes? If not I want to

* Merge this PR
* Open an issue ticket to remind us to change this `__all__` so we can make namespace both for ArviZ both explicit and even cleaner

ahartikainen:
the __all__ does not work for this case (or does it?)

try this

    import az
    az. <-- press tab --> what do you see?

aloctavodia:
This is ready to merge

PR #1329, opened 2020-07-30
Fix ci pylint fail on pytest not callable error on markers

canyon289:
pytest said they fixed the pylint error but we're still seeing it. Kicking off CI to run see if unpinning works

https://docs.pytest.org/en/stable/changelog.html
image


canyon289:
Merging it. Rebase your prs people

PR #1330, opened 2020-07-31
Simplify CmdStan filereading

ahartikainen:
## Description

Simplify file reading

1. Scan file and extract comments
2. Read data with `pandas`

Also extract warmup if it exists. 

## Checklist


ahartikainen:
Ok, I think this is ready. Will merge in 48h if no one reviews this.

arviz/data/io_cmdstan.py
```diff
@@ -584,11 +589,11 @@ def _process_configuration(comments):
...
             )
             key = (
-                "warmup_time"
+                "warmup_time_seconds"
                 if "(Warm-up)" in comment
-                else "sampling_time"
+                else "sampling_time_seconds"
                 if "(Sampling)" in comment
-                else "total_time"
+                else "total_time_seconds"
```
OriolAbril: Now PyMC3 also stores this info as attributes, should we include some attributes in the schema similarly to sample_stats? In any case, any attribute would always be allowed, but if added to schema, some common ones like these three would have a name independent of the sampling backend

ahartikainen: Yeah, time sounds something we could store. Also init, inv_metric and sampling size.

We can refine this in a later PR (and also parse the rest of the values)

PR #1335, opened 2020-08-04
General fixes (isort, plotting)

ahartikainen:
Fix bokeh errors (docs fail to build) introduced in #1312 

Here are the list of changes

1. `isort` compatibility (`pyproject.toml` file, it can be optional)
2. Combine `numeric_utils` and `kde_utils` into `density_utils` and moved it under the `stats/density_utils`
3. Rename `_kde` to `kde` and added it to the main namespace (it is nice nifty function)
4. most plotting functions now use `create_axes_grid` function (which call `plt.subplots` / `bkp.figure`)
5. `create_axes_grid` is moved under `plots.backends.<backend>.__init__.py` so both backend have independent functions
5. `backend_kwargs` is used to set properties for these figure-functions (needed defaults are set where necessary -> figsize, squeeze, +other)
6.  mpl: backend_kwargs + squeeze --> follow `plt.subplots` behaviour --> np.ravel can be used to iterate through axes (even when there is one axis)
7. bokeh: squeeze = True --> `create_axes_grid` returns figure object if rows=1 and cols=1; also dpi, inch, pixel figsize transformations are moved under the create_axes_grid --> and finally the function returns only ndarray / figure object(s) >> these are the same as basically the axis object in mpl
8. `dealiase_sel_kwargs` are next to `create_axes_grid` with a bit different behaviours for bokeh and mpl (normalize kwargs for mpl; this failed with bokeh kwargs in mpl 3.3)
9. bokeh can not handle hex color with alpha --> line_alpha etc for bokeh figures and normal colors without alpha
10. this was done in previous pr but both mpl + bokeh backend plot funcs take the same kwargs even when they are not used now
11. update a lot of the tests --> not for better tbh, we need to explicitly test that the correct things are returned (now it is not true)

I hope I didn't break anything with the changes

Also TBD:
1. `_scale_fig_size` also under the `plots/backend/<backend>/__init__.py` so we can handle bokeh and mpl differently 
2. Bokeh should move to use pixel values --> height, width for the main funcs which are "pixel" values; we can make this work also for mpl (bokeh --> use height, width if given else figsize; mpl figsize --> then height + width)
3. Bokeh should probably move to use scaled (screen size; dynamical size) images --> I mean if we want to use bokeh, why to use small images?

ps. If you need help we merge conflicts, I can help 😄


canyon289:
I skimmed this, its hard to review it all but everything i saw made logical sense. Very inclined to just merge

ahartikainen:
Yeah, if anything is broken, let's fix it.

PR #1336, opened 2020-08-04
Circular traceplot for matplotlib

agustinaarroyuelo:
## Description

In this PR I am adding argument `circ_var_names ` to `plot_trace` in order to obtain a plot like this:

traceplot_circ

## Checklist


arviz/plots/backends/matplotlib/traceplot.py
```diff
@@ -67,8 +70,12 @@ def plot_trace(
         If True adds a rugplot. Defaults to False. Ignored for 2D KDE. Only affects continuous
         variables.
     lines : tuple or list
-        list of tuple of (var_name, {'coord': selection}, [line_positions]) to be overplotted as
+        List of tuple of (var_name, {'coord': selection}, [line_positions]) to be overplotted as
         vertical lines on the density and horizontal lines on the trace.
+    circ_var_names : list
```
aloctavodia, line 75: This should be string, or list of strings. the same as `var_names`

arviz/plots/backends/matplotlib/traceplot.py
```diff
@@ -67,8 +70,12 @@ def plot_trace(
...
+        List of circular variables to account for when plotting KDE.
+    degrees : bool
```
aloctavodia, line 77: maybe this should be "circ_var_units" and defaults to radians. 

Additionally. I am ok with having degrees or radians for now. But we may want to contemplate other options like allowing negative values, like going from -180 to 180 instead of from 0 to 360 (and the same for radians).

arviz/plots/backends/matplotlib/traceplot.py
```diff
@@ -193,9 +203,9 @@ def plot_trace(
     plot_kwargs.setdefault("linewidth", linewidth)
 
     if axes is None:
-        _, axes = plt.subplots(len(plotters), 2, squeeze=False, figsize=figsize, **backend_kwargs)
+        fig = plt.figure(tight_layout=True, figsize=figsize)
```
aloctavodia, line 206: We should not use `tight_layout` as is not compatible with `constrained_layout=True` that is the arviz's default.

arviz/plots/backends/matplotlib/traceplot.py
```diff
@@ -273,94 +252,147 @@ def plot_trace(
...
+                xticks = get_bins(value)
+                axes.set_xticks(xticks[:-1])
+            axes.set_yticks([])
+            axes.set_title(make_label(var_name, selection), fontsize=titlesize, wrap=True, y=1)
+            axes.tick_params(labelsize=xt_labelsize)
+
+            xlims = axes.get_xlim()
+            ylims = axes.get_ylim()
+
+            if divergences and not is_circular:
```
aloctavodia, line 320: This mean circular variables will not have divergences?

arviz/plots/backends/matplotlib/traceplot.py
```diff
@@ -210,57 +220,26 @@ def plot_trace(
                     invalid_var_names, all_var_names
                 )
             )
-
+    # pylint: disable=too-many-nested-blocks
     for idx, (var_name, selection, value) in enumerate(plotters):
-        value = np.atleast_2d(value)
+        for idy in range(2):
```
aloctavodia, line 225: The idy variable is introduced because axes can have only one projection so we can not mix polar and regular axes, right? So we need to create and access them as separated objects, right?

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -179,33 +181,33 @@ def test_plot_trace_legend(compact, combined):
...
-    assert axes.shape == (3, 2)
+    assert len(axes.figure.axes) == 6
 
 
 @pytest.mark.parametrize("kwargs", [{"var_names": ["mu", "tau"], "lines": [("hey", {}, [1])]}])
 def test_plot_trace_invalid_varname_warning(models, kwargs):
     with pytest.warns(UserWarning, match="valid var.+should be provided"):
         axes = plot_trace(models.model_1, **kwargs)
-    assert axes.shape
+    assert axes.figure
```
aloctavodia, line 210: why it is better to check `axes.figure` instead of just `axes`?

ahartikainen: why would that work? axes is numpy array?

But, true `ax.figure` kind of test that mpl axis is created.

arviz/plots/backends/matplotlib/traceplot.py
agustinaarroyuelo: Yes, that is right

arviz/tests/base_tests/test_plots_matplotlib.py
agustinaarroyuelo: Hi, axes is not an np.array anymore. It is a matplotlib AxesSubplot object.

ahartikainen: oh yes, this is traceplot 👍

PR #1345, opened 2020-08-11
More docker fixes

canyon289:


canyon289:
Alright well docker seems fixed. Now just our doc builds are broken because were missing a plot
image

scripts/Dockerfile
```diff
@@ -32,7 +32,7 @@ RUN apt-get update && apt-get install -y git build-essential pandoc vim ffmpeg \
 
 # Copy needed scripts
 COPY $SRC_DIR /opt/arviz/requirements*.txt
-COPY $SRC_DIR /opt/arviz/scripts
+COPY $SRC_DIR /opt/arviz
```
ahartikainen, line 35: But we don't want to import all at this point --> docker cache will fail everytime something is changed in the codebase

canyon289: ah i see. Ok let me try and figure this out

ahartikainen:
This is the problem.

Last line should install arviz into the env, but apparently it installs it inside base. How do we activate env --> install to env with pip

scripts/Dockerfile
canyon289: Fixed

canyon289:
> This is the problem.
> 
> Last line should install arviz into the env, but apparently it installs it inside base. How do we activate env --> install to env with pip

Ok I try and fix this too before I sleep

scripts/Dockerfile
```diff
@@ -60,5 +60,5 @@ RUN find /opt/arviz/ -type f -print0 | xargs -0 dos2unix
 RUN rm -f arviz/tests/saved_models/*.pkl
 RUN find -type d -name __pycache__ -exec rm -rf {} +
 
-#  Install editable using the setup.py
+# Install editable using the setup.py
 RUN /bin/bash -c "source /root/.bashrc" && pip install -e .
```
canyon289, line 64: Why do we need to install editable here each time the containe runs? We could do this earlier in the installation and it would still work?

ahartikainen: Because docker user should be able to change the source --> build docker quickly and run shell / tests.

And when it is editable mode, user can "dynamically" change src without cd to site-packages.

ahartikainen: Also it is not everytime container runs but at build step

ahartikainen:
Now it works and docs can be built

canyon289:
Thanks ari

PR #1349, opened 2020-08-13
Add JSON converter

percygautam:
## Description
As per discussed in the channel, this PR implements the methods `from_json` and `to_json`.

## Checklist

### DIFF
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -17,8 +17,7 @@ from xarray.core.utils import either_dict_or_kwargs
 
 from ..rcparams import rcParams
 from ..utils import HtmlTemplate, _subset_list
-from .base import _extend_xr_method, dict_to_dataset
-from .io_dict import from_dict
+from .base import _extend_xr_method, dict_to_dataset, _make_json_serializable
 
 try:
     import ujson as json
@@ -252,25 +251,6 @@ class InferenceData:
             empty_netcdf_file.close()
         return filename
 
-    @staticmethod
-    def from_json(filename):
-        """Initialize object from a json file.
-
-        Parameters
-        ----------
-        filename : str
-            location of json file
-
-        Returns
-        -------
-        InferenceData object
-        """
-
-        with open(filename, "rb") as file:
-            idata_dict = json.load(file.read())
-
-        return from_dict(**idata_dict, save_warmup=True)
-
     def to_dict(self, groups=None, filter_groups=None):
         """Convert InferenceData to a dictionary following xarray naming conventions.
 
@@ -339,7 +319,7 @@ class InferenceData:
         str
             Location of json file
         """
-        idata_dict = self.to_dict()
+        idata_dict = _make_json_serializable(self.to_dict())
 
         with open(filename, "w") as file:
             json.dump(idata_dict, file, **kwargs)
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -10,7 +10,6 @@ from datetime import datetime
 from html import escape
 
 import netCDF4 as nc
-import json
 import numpy as np
 import xarray as xr
 from xarray.core.options import OPTIONS
@@ -20,6 +19,12 @@ from ..rcparams import rcParams
 from ..utils import HtmlTemplate, _subset_list
 from .base import _extend_xr_method, dict_to_dataset
 from .io_dict import from_dict
+
+try:
+    import ujson as json
+except ImportError:
+    # Can't find ujson using json
+    import json
 
 SUPPORTED_GROUPS = [
     "posterior",
@@ -261,7 +266,7 @@ class InferenceData:
         InferenceData object
         """
 
-        with open(filename, "r") as file:
+        with open(filename, "rb") as file:
             idata_dict = json.load(file.read())
 
         return from_dict(**idata_dict, save_warmup=True)
@@ -327,7 +332,7 @@ class InferenceData:
         filename : str
             Location to write to
         kwargs : dict
-            kwargs passed to json.dumps()
+            kwargs passed to json.dump()
 
         Returns
         -------
@@ -335,10 +340,9 @@ class InferenceData:
             Location of json file
         """
         idata_dict = self.to_dict()
-        serialised_data = json.dumps(idata_dict, **kwargs)
 
         with open(filename, "w") as file:
-            file.write(serialised_data)
+            json.dump(idata_dict, file, **kwargs)
 
         return filename
 
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -10,6 +10,7 @@ from datetime import datetime
 from html import escape
 
 import netCDF4 as nc
+import json
 import numpy as np
 import xarray as xr
 from xarray.core.options import OPTIONS
@@ -18,6 +19,7 @@ from xarray.core.utils import either_dict_or_kwargs
 from ..rcparams import rcParams
 from ..utils import HtmlTemplate, _subset_list
 from .base import _extend_xr_method, dict_to_dataset
+from .io_dict import from_dict
 
 SUPPORTED_GROUPS = [
     "posterior",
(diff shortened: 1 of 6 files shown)


arviz/data/inference_data.py
```diff
@@ -10,6 +10,7 @@
 from html import escape
 
 import netCDF4 as nc
+import json
```
ahartikainen, line 13: We should try to import first in try-except

    import ujson as json

There are actually couple of drop-in placements for json. Should we have a function that decides which of the libs is used?

arviz/data/inference_data.py
```diff
@@ -245,6 +247,25 @@ def to_netcdf(self, filename, compress=True, groups=None):
...
+        filename : str
+            location of json file
+
+        Returns
+        -------
+        InferenceData object
+        """
+
+        with open(filename, "r") as file:
+            idata_dict = json.load(file.read())
```
ahartikainen, line 265: We should check would it be faster to read json in `rb` vs `r`

(Probably doesn't matter, ujson is muchmuch faster)

arviz/data/inference_data.py
```diff
@@ -298,6 +319,29 @@ def to_dict(self, groups=None, filter_groups=None):
...
+        kwargs : dict
+            kwargs passed to json.dumps()
+
+        Returns
+        -------
+        str
+            Location of json file
+        """
+        idata_dict = self.to_dict()
+        serialised_data = json.dumps(idata_dict, **kwargs)
```
ahartikainen, line 338: Should we just use `.dump`?

This will create a copy of the data in-memory, `dump` might use less ram.

percygautam:
During testing, I am getting an error `TypeError: Object of type 'ndarray' is not JSON serializable`. I have found [this](https://pynative.com/python-serialize-numpy-ndarray-into-json/) solution. Is it alright? Should I implement it?

ahartikainen:
I think we should just use orjson, given that it is much faster than the default lib and supports numpy types.

Adding that to requirements is not a problem in my opinion

cc @OriolAbril @canyon289 @ColCarroll @AlexAndorra @aloctavodia

ahartikainen:
So my previous suggestion of using ujson can be replaced with orjson, and let's make it dependency

ahartikainen:
Also, orjson doesn't have load method, but I think we can still use it (so your original implementation was fine, haha, sorry for changing it)

arviz/tests/base_tests/test_data.py
```diff
@@ -1111,6 +1111,23 @@ def test_empty_inference_data_object(self):
...
 
 
+class TestJSON:
+    def test_json_converters(self, models):
+        idata=models.model_1
+
+        filepath=os.path.realpath("test.json")
+        idata.to_json(filepath)
+
+        idata_copy = from_json("test.json")
```
ahartikainen, line 1121: should this use `filepath`

arviz/data/inference_data.py
```diff
@@ -339,7 +319,7 @@ def to_json(self, filename, **kwargs):
         str
             Location of json file
         """
-        idata_dict = self.to_dict()
+        idata_dict = _make_json_serializable(self.to_dict())
 
         with open(filename, "w") as file:
```
ahartikainen, line 324: would "wb" work here?

arviz/tests/base_tests/test_data.py
percygautam: oh, I must have missed this

arviz/data/inference_data.py
percygautam: It's returning the error `TypeError: a bytes-like object is required, not 'str'`.

arviz/data/io_json.py
```diff
@@ -0,0 +1,56 @@
...
+        See `convert_to_inference_data`
+
+    Returns
+    -------
+    str
+        filename saved to
+    """
+    inference_data = convert_to_inference_data(data, group=group, coords=coords, dims=dims)
+    file_name = inference_data.to_json(filename)
+    return file_name
```
OriolAbril, line 56: Here, now that `to_netcdf` has already been around for a while, I would actually diverge from its behaviour. I think that having grooups, coords and dims here is very confusing, we have had several issues and questions about people undestanding that `group="prior"` means that only the prior is stored, similar thing for coords. 

I personally think we can force users to convert manually before calling `to_json` and have `to_json` accept only inferencedata objects. cc @ahartikainen @aloctavodia 

Otherwise, I would modify the docsting to be _extremely_ clear on this. All three parameters should be explicit in that they are ignored if the input is inferencedata and they should link to convert_to_inference_data

arviz/tests/base_tests/test_data.py
```diff
@@ -1111,6 +1112,23 @@ def test_empty_inference_data_object(self):
...
 
+class TestJSON:
+    def test_json_converters(self, models):
+        idata = models.model_1
+
+        filepath = os.path.realpath("test.json")
+        idata.to_json(filepath)
+
+        idata_copy = from_json(filepath)
+        for group in idata._groups_all:  # pylint: disable=protected-access
```
OriolAbril, line 1123: should we add an `__iter__` method to inferencedata or use a dict_like behaviour? looping over groups seems like something that should be possible and part of the public api. cc @ahartikainen @aloctavodia @canyon289 @AlexAndorra

arviz/data/io_json.py
ahartikainen: I'm fine with this suggestion

arviz/tests/base_tests/test_data.py
ahartikainen: Yes, that sounds reasonable.

OriolAbril: Cool, ai think dict like is what will be more intuitive so users can choose items, keys or values, maybe renaming keys to groups?

It could be done in another PR, just thought of this here

percygautam: Yes, I'll implement that!

arviz/data/io_json.py
```diff
@@ -0,0 +1,57 @@
...
+from .io_dict import from_dict
+
+try:
+    import ujson as json
+except ImportError:
+    # Can't find ujson using json
+    import json
+
+
+def from_json(filename):
```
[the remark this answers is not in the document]
percygautam: Actually that was causing circular import error as we are using `from_dict` and it imports InferenceData.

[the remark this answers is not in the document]
OriolAbril: Not sure I follow, ideally, inference_data will NOT import io_json, only io_json will import inference_data.py, so io_json does not even need to import json. I think that should avoid circular imports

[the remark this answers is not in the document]
percygautam: In that case, `from_json` will be implemented in InferenceData, which would require the import of `from_dict` in `inference_data.py`.  As `from_dict` import InferenceData, that would still raise circular import error.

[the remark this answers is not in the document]
OriolAbril: Ooohh, true, I completely forgot from_dict was not an inferencedata method :sweat_smile: (it's to_dict only? none of them?) This is fine then

ahartikainen (approved):

OriolAbril (approved):
I think we can merge this after a couple changes:

Do we have to add `ujson` to `requirements-optional`? Do we have to explain in `from/to_json` docs that `ujson` is used if found and if not `json` is used?

ahartikainen:
Yes to both.

somewhere so users can see it clearly?

docs + docstring (or what did you mean with docs?)

OriolAbril:
docstring is probably enough

arviz/data/io_json.py
```diff
@@ -30,7 +32,9 @@ def from_json(filename):
 def to_json(idata, filename):
     """Save dataset as a json file.
 
-    WARNING: Only idempotent in case `data` is InferenceData
+    It uses ujson, and uses json if it is not available.
```
ahartikainen, line 35:
```suggestion
    Function utilizes ujson -module to accelerate the process if available.
```

ahartikainen (commented):
Maybe something like this?

We don't need to mention default json module.

arviz/data/io_json.py
```diff
@@ -0,0 +1,52 @@
...
+    with open(filename, "rb") as file:
+        idata_dict = json.load(file)
+
+    return from_dict(**idata_dict, save_warmup=True)
+
+
+def to_json(idata, filename):
+    """Save dataset as a json file.
+
+    It uses ujson, and uses json if it is not available.
```
ColCarroll, line 35:
```suggestion
    Will use the faster `ujson` (https://github.com/ultrajson/ultrajson) if it is available.
```

arviz/data/io_json.py
```diff
@@ -0,0 +1,52 @@
...
+
+    WARNING: Only idempotent in case `data` is InferenceData.
```
ColCarroll, line 37:
```suggestion
    WARNING: Only idempotent in case `idata` is InferenceData.
```

arviz/data/io_json.py
ahartikainen: @percygautam  maybe use this one?

percygautam: okay sure

PR #1356, opened 2020-08-19
add __iter__ method

percygautam:
## Description
Related to discussion in #1349.

## Checklist


percygautam:
I have added `__iter__`. Should I also add keys, values, and items, or maybe some other methods?

arviz/data/inference_data.py
```diff
@@ -177,6 +177,11 @@ def __delattr__(self, group):
     def _groups_all(self):
         return self._groups + self._groups_warmup
 
+    def __iter__(self):
```
ahartikainen, line 180: Do we wan't to expose this with __iter__ or with `.keys`, `.items`, `.values`?

Or how dict does it?

arviz/data/inference_data.py
```diff
@@ -177,6 +177,11 @@ def __delattr__(self, group):
...
+        for group in self._groups_all:  # pylint: disable=protected-access
+            dataset = getattr(self, group)
```
ahartikainen, line 182: I think we can assume all groups are found as an attribute, so nothing needs to be changed here.

If user add stuff `_groups_all` let it fail.

arviz/data/inference_data.py
```diff
@@ -177,6 +177,11 @@ def __delattr__(self, group):
     def _groups_all(self):
         return self._groups + self._groups_warmup
 
+    def __iter__(self):
+        for group in self._groups_all:  # pylint: disable=protected-access
```
ahartikainen, line 181: Are these still needed? `.pylintrc` should have a ignore for these.

percygautam: Yeah, it ignores all these. I have added it as a habit. WIll remove.

arviz/data/inference_data.py
percygautam: In my opinion, `__iter__` will be more useful. Maybe we can add a `.groups` function (like keys) which will be more intuitive . What are your thoughts on this @OriolAbril ?

OriolAbril: I think dict like will be more flexible. 

What will iter loop over? Group names? Xarray datasets? A tuple of (group name, dataset)? I think all 3 can have their uses, hence my preference for the dict like approach, as I said somewhere, renaming the `keys` to something else, maybe `groups`. __iter __ could do the same as keys as it happens for dicts or we could eventually decide on forcing expliciteness and have __iter __ raise an error explaining the 3 alternatives

percygautam: Ok, I'll do that.

percygautam: Should I add`filter_groups` to the values?

OriolAbril: I think its probably better to keep the functions as close as possible as dict counterparts, so no extra arguments. 

It is true however that there is no clear way of subsetting an inferencedata on a group basis other than deleting groups. Should there be a way to do that? or doing `for ... in idata... if <condition>` is enough?

arviz/data/inference_data.py
```diff
@@ -178,9 +178,23 @@ def _groups_all(self):
...
+            yield group
+
+    def groups(self):
+        return self._groups_all
+
+    def values(self):
+        datasets = []
+        for group in self._groups_all:
+            datasets.append(getattr(self, group))
+        return datasets
```
OriolAbril, line 191: for `groups` a list is fine because we already have that same list. here however, I think using generators will perform better

```suggestion
        for group in self._groups_all:
            yield getattr(self, group)
```

arviz/data/inference_data.py
```diff
@@ -178,9 +178,23 @@ def _groups_all(self):
...
+
+    def items(self):
+        item = []
+        for group in self._groups_all:
+            item.append((group, getattr(self, group)))
+        return item
```
OriolAbril, line 197: same comment as above.

```suggestion
        for group in self._groups_all:
            yield (group, getattr(self, group))
```

arviz/data/inference_data.py
percygautam: I think adding extra arguments with default values should not be much of a problem as we can still use `groups()` similar to `keys()`. I think this will be a cleaner way to subset the idata groups.

OriolAbril: I have a slight preference towards having a subsetting function and the looping ones separated because is that setting, one can do any of `idata.<subsetting>.items()`, `idata.<subsetting>` or `idata.items()` whereas adding the subsetting inside `items()` and `groups` would not allow pure subsetting yet. But again, it's a slight preference, let's see what other people thinks cc @ahartikainen @aloctavodia @AlexAndorra

OriolAbril (approved):
LGTM, will merge once CI passes

ahartikainen (approved):

ahartikainen:
Will merge after the CI is done

PR #1357, opened 2020-08-20
Add attr  and name to idata

percygautam:
## Description
Related to the discussion being held in the channel.

## Checklist


percygautam:
How these changes are incorporated in `concat`. Also, briefly explain the `name` attribute?

ahartikainen:
For concat, either we make everything a list or list only when needed, same for name.

    idata1.attrs["a"] = 1
    idata2.attrs["a"] = 1

    -> idata["a"] == [1, 1] or idata["a"] == 1

    idata1.attrs["a"] = 1
    idata2.attrs["a"] = 2

    -> idata["a"] == [1, 2]

ahartikainen:
I think that might be the simplest option?

percygautam:
I have added `attrs` to concat as suggested. I was modifying `from_dict` to include attrs, but it already contains `attrs` argument, that point to `dataset.attrs`. Should I change already existing attrs to `dataset_attrs`? Or something else?

ahartikainen:
I think attrs in from_dict should point to the main attrs (this what you add right now)

percygautam:
Will this not change the existing API? Also, if we do so, what should the `attrs` be named? (Maybe `group_attrs` or `dataset_attrs`?)

ahartikainen:
I think existing api is not yet released?

That sounds a good idea.

We could also add kwargs for `from_dict` and then get attrs for each group if needed?

So `attrs` could then mean the idata attrs. And each group might have `{groupname}_attrs`?

OriolAbril:
Another question is whether we should modify current code to store arviz/ppl version... as idata attrs instead of repeating them in every group.

arviz/data/io_dict.py
```diff
@@ -335,6 +368,8 @@ def from_dict(
         A mapping from variables to a list of coordinate values for predictions.
     attrs : dict
         A dictionary containing attributes for different groups.
+    kwargs : dict
+        A dictionary containing group attrs.
```
OriolAbril, line 372: Accepted kwargs should be docummented.

ahartikainen:
Do we want to get this in the next release?

ahartikainen:
This PR can wait to next release. It will change many io functions.

ahartikainen (approved):

PR #1359, opened 2020-08-21
Separation plot

agustinaarroyuelo:
## Description

The [separation plot](https://onlinelibrary.wiley.com/doi/abs/10.1111/j.1540-5907.2011.00525.x) is a really interesting and simple way for assessing a model's fit when the outcome is binary. After fitting, model predictions are sorted and represented as vertical lines. When adding a color scheme that identifies the positive and negative class, for a good fit, we should see most of the instances of the positive class on the right hand side of the plot. That is, where the highest valued predictions are located.

``` python
idata = az.load_arviz_data('classification10d')
ax = az.plot_separation(
    idata=idata, 
    y='outcome',
    y_hat_line=True,
    expected_events=True,
    figsize=(10, 1),
)
```

separationplot

~It would be great if you could suggest how to choose default colors for this plot. Finally, I would like to know your overall comments on this code.~

## Checklist

### DIFF
--- a/arviz/plots/backends/matplotlib/separationplot.py
+++ b/arviz/plots/backends/matplotlib/separationplot.py
@@ -47,6 +47,8 @@ def plot_separation(
 
     exp_events_kwargs.setdefault("color", "k")
     exp_events_kwargs.setdefault("marker", "^")
+    exp_events_kwargs.setdefault("s", 100)
+    exp_events_kwargs.setdefault("zorder", 2)
 
     backend_kwargs = {
         **backend_kwarg_defaults(),
@@ -102,9 +104,8 @@ def plot_separation(
 
     for i, loc in enumerate(locs):
         positive = not y[idx][i] == 0
-        label = "Positive class" if positive else "Negative class"
         alpha = 1 if positive else 0.3
-        ax.bar(loc, 1, width=width, label=label, alpha=alpha, **plot_kwargs)
+        ax.bar(loc, 1, width=width, alpha=alpha, **plot_kwargs)
 
     if y_hat_line:
         ax.plot(np.linspace(0, 1, len(y_hat)), y_hat[idx], label=label_y_hat, **y_hat_line_kwargs)
--- a/arviz/plots/backends/matplotlib/separationplot.py
+++ b/arviz/plots/backends/matplotlib/separationplot.py
@@ -2,18 +2,16 @@
 import warnings
 import matplotlib.pyplot as plt
 import numpy as np
-import xarray as xr
 
-from ....data import InferenceData
 from ...plot_utils import _scale_fig_size
 from . import backend_kwarg_defaults, backend_show, create_axes_grid
 
 
 def plot_separation(
-    idata,
     y,
     y_hat,
     y_hat_line,
+    label_y_hat,
     expected_events,
     figsize,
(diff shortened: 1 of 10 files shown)


arviz/plots/separationplot.py
```diff
@@ -0,0 +1,112 @@
...
+
+def plot_separation(
+    idata=None,
+    y=None,
+    y_hat=None,
+    y_hat_line=False,
+    expected_events=False,
+    figsize=None,
+    textsize=None,
+    color=None,
```
OriolAbril, line 14: should we set `color="C0"` here and then convert to hex before passing it to the backend specific function? This is more a philosophical question about whether we want this behaviour than anything else.

arviz/plots/backends/matplotlib/separationplot.py
```diff
@@ -0,0 +1,131 @@
...
+
+    if y_hat_line:
+        ax.plot(np.linspace(0, 1, len(y_hat)), y_hat[idx], label=label_y_hat, **y_hat_line_kwargs)
+
+    if expected_events:
+        expected_events = int(np.round(np.sum(y_hat)))
+        ax.scatter(y_hat[idx][expected_events - 1], 0, label="Expected events", **exp_events_kwargs)
+
+    if legend and expected_events or y_hat_line:
+        handles, labels = plt.gca().get_legend_handles_labels()
```
OriolAbril, line 119: why gca instead of using ax? not sure how is this different from `ax.legend()` :sweat_smile:

arviz/plots/separationplot.py
```diff
@@ -0,0 +1,112 @@
...
+        Text size for labels. If None it will be autoscaled based on figsize.
+    color : str, optional
+        Color to assign to the postive class. The negative class will be plotted using the
+        same color and an `alpha=0.3` transparency.
+    legend : bool, optional
+        Show the legend of the figure.
+    ax: axes, optional
+        Matplotlib axes or bokeh figures.
+    plot_kwargs : dict, optional
+        Additional keywords passed to ax.bar for separation plot.
```
OriolAbril, line 54:
```suggestion
        Additional keywords passed to :meth:`mpl:matplotlib.axes.Axes.bar` or :meth:`bokeh:bokeh.plotting.Figure.vbar` for separation plot.
```

doc/api.rst
```diff
@@ -35,6 +35,7 @@ Plots
     plot_ppc
     plot_rank
     plot_trace
+    plot_separation
```
OriolAbril, line 38:
```suggestion
    plot_separation
    plot_trace
```

arviz/plots/separationplot.py
aloctavodia: Yeah, I think we want this everywhere, not only here.

OriolAbril (approved):
Added a small comment to try to minimize duplicated code, but it's note necessary to do, can be merged as is.

I love the PR and having the separation plot on ArviZ. I have been asked a couple times for posterior predictive plots for binary outcomes in Discourse, if you are up for it, it could be interesting to make a blogpost about it, how to use it, how to interpret it and share it in pymc and stan discourses

aloctavodia:
thanks @agustinaarroyuelo!

PR #1364, opened 2020-08-25
add new colors to arviz-styles

aloctavodia:

All new colors have only 5 colors (instead of 10), most of them are created by choosing a single color and then decreasing the value of alpha. They are the color equivalent of our current grayscale. The exception are royish, viridish and plamish. Which are based on the colormaps fire from colorcet, and viridis and plasma from matplotlib.

Black is included as the C4/C9 color.

test_cb_white

test_cb_darkgrid

This also introduce a small fix to plot_ppc. To first allow to change the color and second to better work with the grayscale style and the new colors.


AlexAndorra (approved):
This is nice, thanks @aloctavodia ! Can't wait to use that in new examples 🎉

canyon289:
I like all the ish

examples/matplotlib/mpl_styles.py
```diff
@@ -20,18 +20,27 @@
...
+    ["arviz-white", "arviz-viridish"],
+    ["arviz-white", "arviz-plasmish"],
 ]
 
-fig = plt.figure(figsize=(12, 12))
+fig = plt.figure(figsize=(12, 25))
 for idx, style in enumerate(style_list):
-    with az.style.context(style):
-        ax = fig.add_subplot(3, 2, idx + 1, label=idx)
+    with az.style.context(style, after_reset=True):
```
canyon289, line 37: What's after reset?

aloctavodia: reset style settings to the default values and then apply the new style.

PR #1372, opened 2020-09-08
Disable KDE domain narrower than data limits

tomicapretto:
## Description
Currently, the KDE returns a pdf with an area under the curve smaller than 1 if a narrower custom domain is specified.
This PR fixes that. 

**Old**

old

**New**

now_1
new_2

### DIFF
--- a/arviz/stats/density_utils.py
+++ b/arviz/stats/density_utils.py
@@ -8,6 +8,7 @@ from scipy.optimize import brentq
 from scipy.signal import convolve, convolve2d, gaussian  # pylint: disable=no-name-in-module
 from scipy.sparse import coo_matrix
 from scipy.special import ive  # pylint: disable=no-name-in-module
+from scipy.integrate import simps # pylint: disable=no-name-in-module
 
 from ..utils import _cov, _dot, _stack, conditional_jit
 
@@ -640,7 +641,10 @@ def _kde_linear(
 
     if cumulative:
         pdf = pdf.cumsum() / pdf.sum()
-
+	
+    if grid[0] > x_min or grid[-1] < x_max:
+    	pdf = pdf / simps(pdf, grid)
+    	
     if bw_return:
         return grid, pdf, bw
     else:
--- a/arviz/stats/density_utils.py
+++ b/arviz/stats/density_utils.py
@@ -8,7 +8,6 @@ from scipy.optimize import brentq
 from scipy.signal import convolve, convolve2d, gaussian  # pylint: disable=no-name-in-module
 from scipy.sparse import coo_matrix
 from scipy.special import ive  # pylint: disable=no-name-in-module
-from scipy.integrate import simps # pylint: disable=no-name-in-module
 
 from ..utils import _cov, _dot, _stack, conditional_jit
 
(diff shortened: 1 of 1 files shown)


OriolAbril:
Maybe there should be a warning when this happens? I had not thought about draws outside the kde bounds before seeing the PR but now I'm thinking in many cases the reason there are draws outside the bounds could be a typo. I guess the warning would depend on how much of a good/bad practice we think this is

tomicapretto:
I've just added **WIP** to title since there are tests that have to be modified.

tomicapretto:
> Maybe there should be a warning when this happens? I had not thought about draws outside the kde bounds before seeing the PR but now I'm thinking in many cases the reason there are draws outside the bounds could be a typo. I guess the warning would depend on how much of a good/bad practice we think this is

So here there is a new idea:
* If the user specifies a domain narrower than the observed, drop observations outside the specified bounds and estimate the density as usual without having to make any corrections. Plus, throw a warning because this may be an error.

* If the user specified a domain wider than the observed, just estimate the density with the specified bounds without dropping anything (it just doesn't make sense) and do not throw any warning.

Opinions? @aloctavodia @OriolAbril

OriolAbril:
Do you have an example in mind where this could happen? 

I had never considered this situation before and I am debating between a warning or directly an error.

tomicapretto:
You are right. I don't know why I considered that and honestly I don't have a good reason to make it a warning instead of an error. 
I've just checked Scipy and KDEpy implementations and they don't allow it (they don't allow custom limits at all).  R base `density()` function allows narrower domains but the resulting area is smaller than 1.

So, I think that if someone has a good reason to make it a warning, we can discuss it. If not, we make it an error.

tomicapretto:
In addition, previous `_fast_kde()` would raise an assertion error when the limits were narrower than observed data.

ahartikainen:
If there is going to be kde narrower than data limits, then we should either drop the data outside the limits or then calculate normalization based on the datalimits, but just show the narrower limits --> this probably needs complicated handling if we want to follow same procedure / x-density for the narrower limits.

aloctavodia:
If we can not think of why we need custom defined limits a better choice could be to remove this feature.

tomicapretto:
Now you get an error if you try to use the KDE with a custom limit that does not include all data points. Note it still
works if the boundary is wider (I assumed we agree on it since it is more reasonable)

``` python
import numpy as np
from arviz.stats.density_utils import kde

x, y = kde(np.random.normal(size=500))
plt.plot(x, y);
```
image

```python
x, y = kde(np.random.normal(size=500), custom_lims = [-5, 5])
plt.plot(x, y);
```
image

```python
x, y = kde(np.random.normal(size=500), custom_lims = [-2, 2])
# [...]
# ValueError: Some observations are outside `custom_lims` boundaries.
```

ahartikainen:
There could be reasons for having smaller-than-data limits.

Can we enable these:
- don't drop data outside limits --> do normal kde procedure, but let's just calculate "correct" number of points so when we slice output to "limits" it will contain similar amount of points as it would when the limits is wider than data.
- drop data outside limits --> boolean rule as a keyword.

Do you get my point about "don't drop, just pad and then slice"?

tomicapretto:
I think I understand. But in that case, the density near the boundaries will be influenced by points outside them. My concern is that since you are not viewing those points, you may end up thinking they don't affect the result at all.

ahartikainen:
We need to be explicit about the options. (E.g. description in docstring)

This will be "advanced" feature.

aloctavodia (approved):

aloctavodia:
I think this is a good solution. Still unsure why we should allow customs limits smaller than data limits. I think if for some reason a  user want to do this, they should just trim the data before passing it to ArviZ's functions.

PR #1375, opened 2020-09-09
Fix violin plot nan handling

tomicapretto:
## Description

If you have two variables of different length in the same pandas data frame and attempt to get a violinplot with both varaibles, the lines indicating the credible interval and interquartile range will be missed for the group with fewer observations.

**Code**

```python
import arviz as az
import numpy as np
import pandas as pd

np.random.seed(2935)
a = np.random.normal(6, 2.5, 300)
b = np.random.normal(8, 2, 200)
df = pd.DataFrame(
    {
        'Group': ['a'] * 300 + ['b'] * 200,
        'Val':np.hstack([a, b])
    }
)
az.plot_violin({'a':a, 'b':b})
```

**Old**
old1

**New**
new1

**Note:** The old plot also uses the old KDE function so that explains the difference in the estimation.


OriolAbril:
I think something darker is happening here. I like the fix but would like to get the root cause which I suspect is `hdi` or the conversion to inferencedata.

The first step done by plot_violin is converting the dict to inferencedata, as the arrays are 1d, their dimension is interpreted as the draw dimension, so I'm guessing the resulting idata has two variables with shape 300, and b has 100 nans at the end. I am not really sure why this works nor that this is what happens so we should doble check.

Then hdi is called with an array filled with nans, there is the possibility that hdi returns nans whenever nans are present in the input array and then as the hdi pims are nans, the line is not plotted. If that were the case, similar issues could happen with plot_posterior for example with nans present and we should try to fix hdi instead.

I seem to recall hdi working with nans, so it's probably np.percentile what returns the nans, in which case the mask could be applied only to the percentile call or maybe there is an argument to ignore nans?

I won't be able to try the code until Monday so I'm trying to cover many possibilities to compensate :sweat_smile:

tomicapretto:
According to what I talked to Osvaldo it is the conversion to inferencedata that adds nans because it assumes you have draws from a probability distribution (or at least that is what I understood).
`np.percentile()` does not work with nans. I've just checked what happens with `hdi()` and it seems to fail too.

With  `az.hdi(np.r_[np.random.normal(size=1000), np.nan])` I got `array([-1.59712431, nan])`

OriolAbril:
Can you skipna=True with hdi? if it fails we have to open an issue and fix this

arviz/plots/backends/bokeh/violinplot.py
```diff
@@ -71,8 +71,8 @@ def plot_violin(
             rug_x = -np.abs(np.random.normal(scale=max(dens) / 3.5, size=len(val)))
             ax_.scatter(rug_x, val, **rug_kwargs)
 
-        per = np.percentile(val, [25, 75, 50])
-        hdi_probs = hdi(val, hdi_prob, multimodal=False)
+        per = np.percentile(val[np.isfinite(val)], [25, 75, 50])
```
aloctavodia, line 74: this should be `per = np.nanpercentile(val, [25, 75, 50])`

arviz/plots/backends/bokeh/violinplot.py
```diff
@@ -71,8 +71,8 @@ def plot_violin(
...
+        hdi_probs = hdi(val[np.isfinite(val)], hdi_prob, multimodal=False)
```
aloctavodia, line 75: this should be `hdi_probs = hdi(val, hdi_prob, multimodal=False, skipnan=True)`

aloctavodia (changes requested):

aloctavodia:
The nans are added because InferenceData keep all arrays of the same shape, adding nans if necessary. This will not happen for MCMC samples, but it can happen when you use InferenceData with "actual data" instead of posterior/prior samples. While using ArviZ with data is a little bit of "abuse", I generally do it :-)

arviz/plots/backends/bokeh/violinplot.py
tomicapretto: I think this is still failing. See:
```
val = np.r_[np.random.normal(size=50), np.nan]
az.hdi(val, skipnan=True)

array([-1.77298977,         nan])
```

tomicapretto: Which is reflected here in the original example
image

PR #1379, opened 2020-09-11
Add support for discrete variables in plot_bpv

aloctavodia:
Fix `plot_bpv` to work with discrete variables. A similar fix could be applied to `plot_loopit`

It also add an option (default to false) to plot the mean square error between the uniform distribution and marginal p_value

Old
bpv_p_old
bpv_u_old

New
bpv_p_new
bpv_u_new

data =np.random.binomial(n=1, p=0.5, size=1000)
with pm.Model() as momo:
    p = pm.Beta("p", 1., 1.)

bpv_binary_p_00
bpv_binary_u_00

data = np.random.binomial(n=1, p=0.5, size=50)
with pm.Model() as momo:
    p = pm.Beta("p", 1., 1.)

bpv_binary_p_00
bpv_binary_u_00

data = np.random.binomial(n=1, p=0.5, size=100)
with pm.Model() as momo:
    p = pm.Beta("p", 1., 1.)

bpv_binary_p_00
bpv_binary_u_00

data =np.random.binomial(n=1, p=0.5, size=100)
with pm.Model() as momo:
    p = pm.Beta("p", 100., 1.)

bpv_binary_p_00
bpv_binary_u_00

data =np.random.binomial(n=1, p=0.8, size=100)
with pm.Model() as momo:
    p = pm.Beta("p", 50, 50)
bpv_binary_p_00
bpv_binary_u_00

data = np.random.poisson(2.5, size=1000)
with pm.Model() as momo:
    p = pm.Uniform('p', 0, 10)

bpv_binary_p_00
bpv_binary_u_00

data = np.random.poisson(2.5, size=1000)
with pm.Model() as momo:
    p = pm.Uniform('p', 3, 10)

bpv_binary_p_00
bpv_binary_u_00

data = np.random.poisson(2.5, size=1000)
data[:100] = 4
with pm.Model() as momo:
    p = pm.Uniform('p', 0, 10)

bpv_binary_p_00
bpv_binary_u_00

The two following

 examples are not affected by this PR, but I add them here as reference.

normal model sample size= 1000
bpv_binary_p_00
bpv_binary_u_00

normal model sample size= 100

bpv_binary_p_00
bpv_binary_u_00


OriolAbril:
I understood that the current loo-pit approximation is not appliable to discrete data. I had this in mind to implement some day https://doi.org/10.1111/j.1541-0420.2009.01191.x

Disclaimer: I have yet to review the PR, will do it soon

aloctavodia:
Right, loo-pit is not applicable to discrete data for the same reasons plot_bpv isn't . I will check that paper, thanks!

aloctavodia:
It seems I need to think a little bit more about this.

arviz/plots/backends/matplotlib/bpvplot.py
```diff
@@ -83,6 +88,15 @@ def plot_bpv(
...
         pp_vals = pp_vals.reshape(total_pp_samples, -1)
 
+        if obs_vals.dtype.kind == "i" or pp_vals.dtype.kind == "i":
+            x = np.linspace(0, 1, len(obs_vals))
+            csi = CubicSpline(x, obs_vals)
+            obs_vals = csi(np.linspace(0.001, 0.999, len(obs_vals)))
+
+            x = np.linspace(0, 1, len(pp_vals))
+            csi = CubicSpline(x, pp_vals)
+            pp_vals = csi(np.linspace(0.001, 0.999, len(pp_vals)))
```
OriolAbril, line 98: Not sure about each quantity having an independent interpolation. IIUC, the interpolation is to smooth things out and convert each integer to a set of floats which are close/close-ish to each other but are not the same, here however, in the observed data there will generally not be too many points, so the effect of the interpolation and its smoothing will be quite small, leaving wavy kdes. Here is how it looks for the rugby dataset using the `p value` kind:

image

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 * Add more `plot_parallel` examples ([#1380](https://github.com/arviz-devs/arviz/pull/1380))
 * Bump minimum xarray version to 0.16.1 ([#1389](https://github.com/arviz-devs/arviz/pull/1389)
 * Fix multi rope for `plot_forest` ([#1390](https://github.com/arviz-devs/arviz/pull/1390))
+* Fix `plot_bpv` to work with discrete variables ([#1379](https://github.com/arviz-devs/arviz/pull/1379))
```
OriolAbril, line 30: Personally I would add it to the new features section and say it adds support for discrete variables in plot_bpv.

arviz/plots/backends/matplotlib/bpvplot.py
aloctavodia: You will get a similar plot with continuous values too. This is for a normal model with 60 observed data points.
dada

That's is related to the KDE. If I remember correctly `fast_kde` had an ad-hoc trick to be smoother in the present of a few data points, but was completely trickery so it is not part of our new KDE.

The main problem with discrete values an p-values was that they were completely off, like in the example above where the expected value was ~0.5 but the computed one was ~0.75. interpolation fix that.

How do you suggest using a single interpolation?

OriolAbril (commented):
Will merge after fixing bokeh backend

arviz/plots/backends/matplotlib/bpvplot.py
```diff
@@ -83,6 +88,16 @@ def plot_bpv(
         obs_vals = obs_vals.flatten()
         pp_vals = pp_vals.reshape(total_pp_samples, -1)
 
+        print(obs_vals.shape, pp_vals.shape)
```
ahartikainen, line 91: debug?

ahartikainen:
There are some duplication, but I think we can move things to the main function maybe in another PR when we have time?

arviz/plots/backends/matplotlib/bpvplot.py
aloctavodia: I just want to keep the users informed

arviz/tests/base_tests/test_plots_bokeh.py
```diff
@@ -1037,8 +1037,9 @@ def test_plot_bpv(models, kwargs):
 
 
 def test_plot_bpv_discrete():
-    fake = {"a": np.random.poisson(2.5, 1000)}
-    fake_model = from_dict(posterior_predictive=fake, observed_data=fake)
+    fake_obs = {"a": np.random.poisson(2.5, 100)}
+    fake_pp = {"a": np.random.poisson(2.5, (10, 100))}
```
OriolAbril, line 1041: As a 2 dim array this will be understood as a `(chain, draw)` array, you should add a dimension at the beggining.

arviz/plots/backends/matplotlib/bpvplot.py
```diff
@@ -86,7 +86,8 @@ def plot_bpv(
         pp_var_name, _, pp_vals = pp_plotters[i]
 
         obs_vals = obs_vals.flatten()
-        pp_vals = pp_vals.reshape(total_pp_samples, -1)
+        if pp_vals.ndim > 2:
+            pp_vals = pp_vals.reshape(total_pp_samples, -1)
```
OriolAbril, line 90: I think this will break the ArviZ shape convention and interpret the draw dimension as the observations.

OriolAbril (changes requested):

OriolAbril (approved):

OriolAbril:
Will merge once/if tests pass and release

PR #1380, opened 2020-09-14
[doc] Add plot_parallel examples and fix rankdata use

ahartikainen:
## Description

Add more examples plots. (normal, rankdata and minmax)

Use rankdata with `"average"`.

## Checklist

## Plots

- Default
mpl_plot_parallel

- minmax
mpl_plot_parallel_minmax

- normal
mpl_plot_parallel_normal

- rank
mpl_plot_parallel_rank

### DIFF
--- a/arviz/plots/backends/bokeh/essplot.py
+++ b/arviz/plots/backends/bokeh/essplot.py
@@ -88,7 +88,7 @@ def plot_ess(
 
             values = data[var_name].sel(**selection).values.flatten()
             mask = idata.sample_stats[rug_kind].values.flatten()
-            values = rankdata(values)[mask]
+            values = rankdata(values, method="average")[mask]
             rug_space = np.max(x) * rug_kwargs.pop("space")
             rug_x, rug_y = values / (len(mask) - 1), np.zeros_like(values) - rug_space
 
--- a/examples/matplotlib/mpl_plot_parallel_rank.py
+++ b/examples/matplotlib/mpl_plot_parallel_rank.py
@@ -1,0 +1,17 @@
+"""
+Rank Parallel Plot
+==================
+
+_thumb: .2, .5
+"""
+import matplotlib.pyplot as plt
+
+import arviz as az
+
+az.style.use("arviz-darkgrid")
+
+data = az.load_arviz_data("centered_eight")
+ax = az.plot_parallel(data, var_names=["theta", "tau", "mu"], norm_method="rank")
+ax.set_xticklabels(ax.get_xticklabels(), rotation=70)
+
+plt.show()
--- a/.azure-pipelines/azure-pipelines-docs.yml
+++ b/.azure-pipelines/azure-pipelines-docs.yml
@@ -3,7 +3,7 @@ jobs:
   dependsOn:
     - BaseTests
     - ExternalTests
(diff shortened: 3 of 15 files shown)


arviz/plots/backends/bokeh/essplot.py
```diff
@@ -88,7 +88,7 @@ def plot_ess(
 
             values = data[var_name].sel(**selection).values.flatten()
             mask = idata.sample_stats[rug_kind].values.flatten()
-            values = rankdata(values)[mask]
+            values = rankdata(values, method="average")[mask]
```
aloctavodia, line 91: method="average" is the default, right? So this is making it explicit, right?

examples/matplotlib/mpl_plot_parallel_rank.py
```diff
@@ -0,0 +1,17 @@
...
+"""
+import matplotlib.pyplot as plt
+
+import arviz as az
+
+az.style.use("arviz-darkgrid")
+
+data = az.load_arviz_data("centered_eight")
+ax = az.plot_parallel(data, var_names=["theta", "tau", "mu"], norm_method="rank")
+ax.set_xticklabels(ax.get_xticklabels(), rotation=70)
```
aloctavodia, line 15: :+1:

arviz/plots/backends/bokeh/essplot.py
ahartikainen: yes, nothing really changes

ahartikainen:
Also, see the change in pipelines

https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=3378&view=results

There is published section, where we can download the example figures. What other things should we upload from html pages?
The total build folder is 60+MB, but I does it matter if we upload the whole folder? This would help users/devs to see changes in docs without need to build it locally.

ahartikainen:
Ok, I will set it to upload full build doc, then everything works when needed.

I'm not sure, but I don't think uploading 60mb per successful pipeline is too bad. Users don't need to download them if they don't need it. And pipelines automatically delete old ones. Any opinions/comments?

ahartikainen:
This is ready now.

ahartikainen:
Ok, the build docs size is 93MB.

What if we upload docs dynamically (e.g. user adds [viz] or [doc] to PR header?)

[plot]
[viz]
[doc]
[docs]

ahartikainen:
Now docs are build for every commit, this is a good thing.

It will upload the build docs to artifacts if branch is `master` or some of these can be found in the commit title [plot], [viz], [doc], [docs]

This way user can access/test docs locally without building them (locally).

User needs to download the artifact from the three dots button in the right side of the artifact page

https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=3382&view=artifacts&type=publishedArtifacts

OriolAbril:
This is very cool, thanks! Do you think you can add some lines about this to the developer guide? Or maybe the contributing one? Not really sure what should go into which document yet

OriolAbril (approved):

ahartikainen:
Yes we can add that info. Another PR?

OriolAbril:
> Yes we can add that info. Another PR?

whatever works best for you

ahartikainen:
I will wait with the merge. (still waiting for pymc3 to fix pickling problem)

canyon289:
Were good now

PR #1383, opened 2020-09-16
Make yellow line visible in matplotlib plot_hdi example

chucklesoclock:
## Description
In the [example](https://arviz-devs.github.io/arviz/examples/matplotlib/mpl_plot_hdi.html), there seems to be a desire to show the line from which `y_data_rep` is sampled, but it is not visible because the call to `az.plot_hdi` is after `plt.plot`, and its shaded region covers it. Switching the lines makes the yellow line visible. 

### Current
image
### Possibly Intended
image

## Checklist


ahartikainen (approved):

canyon289:
Thanks for contributing. Restarting the CI job that we've fixed an issue in PyMC3

PR #1386, opened 2020-09-19
edit groups in idata.extend

OriolAbril:
## Description
`idata.extend` does add the groups but does not edit the `_groups` attributes which makes `idata.posterior` to work but `posterior` is not shown when printing nor in the html repr.

## Checklist

### DIFF
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -820,6 +820,10 @@ class InferenceData:
                 )
             dataset = getattr(other, group)
             setattr(self, group, dataset)
+            if group.startswith(WARMUP_TAG):
+                 self._groups_warmup.append(group)
+             else:
+                 self._groups.append(group)
 
     set_index = _extend_xr_method(xr.Dataset.set_index)
     get_index = _extend_xr_method(xr.Dataset.get_index)
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -822,7 +822,7 @@ class InferenceData:
             setattr(self, group, dataset)
             if group.startswith(WARMUP_TAG):
                  self._groups_warmup.append(group)
-             else:
+            else:
                  self._groups.append(group)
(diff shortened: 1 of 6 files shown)


ahartikainen (approved):

canyon289:
Good thing we got out CI working! Seems like theres an indentation error hidden in there

arviz/data/inference_data.py
```diff
@@ -820,6 +820,10 @@ def extend(self, other, join="left"):
                 )
             dataset = getattr(other, group)
             setattr(self, group, dataset)
+            if group.startswith(WARMUP_TAG):
+                 self._groups_warmup.append(group)
+             else:
```
OriolAbril, line 825:
```suggestion
            else:
```

arviz/data/io_dict.py
```diff
@@ -69,11 +69,15 @@ def __init__(
         self.attrs.pop("created_at", None)
         self.attrs.pop("arviz_version", None)
 
-    @requires("posterior")
+    def _init_dict(self, attr_name):
+        dict_or_none = getattr(self, attr_name, None)
```
ahartikainen, line 73: should this return empty dict instead of None?

ahartikainen (approved):

arviz/data/io_dict.py
OriolAbril: yeah, much better, it directly takes care about the if else below, thanks!

OriolAbril: now I realized the if else is what does the actual work, for now all attrs are set, but it's nice to have this here in case we extended io_dict.

ahartikainen: Remember to update the block below too

OriolAbril: The block below needs to check for None. This first one catches attributes that have not been set and the one below catches attributes set to None

PR #1389, opened 2020-09-21
fix css style

ahartikainen:
## Description

Change access to css file.

## Checklist

### DIFF
--- a/arviz/utils.py
+++ b/arviz/utils.py
@@ -696,5 +696,6 @@ class HtmlTemplate:
                   </div>
             </li>
             """
+    _, css_style = xr.core.formatting_html._load_static_files()  # pylint: disable=protected-access
     specific_style = ".xr-wrap{width:700px!important;}"
-    css_template = f"<style> {xr.core.formatting_html.CSS_STYLE}{specific_style} </style>"
+    css_template = f"<style> {css_style}{specific_style} </style>"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 * Update diagnostics to be on par with posterior package ([#1366](https://github.com/arviz-devs/arviz/pull/1366))
 * Use method="average" in `scipy.stats.rankdata` ([#1380](https://github.com/arviz-devs/arviz/pull/1380))
 * Add more `plot_parallel` examples ([#1380](https://github.com/arviz-devs/arviz/pull/1380))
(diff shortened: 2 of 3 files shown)


aloctavodia (approved):
LGTM

arviz/utils.py
```diff
@@ -696,5 +696,6 @@ class HtmlTemplate:
                   </div>
             </li>
             """
+    _, css_style = xr.core.formatting_html._load_static_files()  # pylint: disable=protected-access
```
OriolAbril, line 699: I am fine with enforcing the lazy load but we then have to update requirements to xarray 0.16.1 too, otherwise we need and if/else.

OriolAbril (changes requested):

arviz/utils.py
ahartikainen: True, what should we do? Have our own css file, or copy the lazy load function?

OriolAbril: I gravitate towards requiring xarray 0.16.1 so this code works without problem or having our lazy load function that calls xarray's one if possible and gets css style otherwise.

ahartikainen: Let's update xarray to 0.16.1

OriolAbril (approved):

arviz/utils.py
shoyer: Just to warn you, this line is definitely using an internal xarray function that could go away or be disabled in the future. That's why you had to add the lint comment to disable protected-access :).

If you are OK copying the CSS file into arViz, that would be my preference for the safest/simplest way to do this (without any chance of things breaking in the future).

Otherwise, if for some reason you need to keep it in sync with xarray's version, we could definitely consider adding a public API option for getting this information from xarray. In that case, please open an issue in xarray's issue tracker.

ahartikainen: Yes, we are using internal function. Until we figure out a good way to solve this, I think we are ok with this solution.

xarray should not (and probably are not) worry changing their internal functionality.

ddesroches:
We ran into this issue because our project depends on Pymc3, which relies on Arviz. We are going to work around it by locking the version of XArray to before the breaking change. Your solution to rely on yet another private entry point to Xarray is a recipe for another breakage for yours and all dependent packages. Suggest if you need a coupling to XArray that you PR a new public interface to that package.

ahartikainen:
Ok, I understand. We need a better fix.

pundivived-attention:
So how do we fix it ?

ahartikainen:
This has been fixed in https://github.com/arviz-devs/arviz/pull/1392

PR #1397, opened 2020-09-27
Add missing type hint imports, as suggested by mypy.

obi1kenobi:
A few minor fixes for type hint issues reported by mypy:
```
arviz/stats/stats.py:1133: error: Name 'Dict' is not defined
arviz/stats/stats.py:1133: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Dict")
arviz/data/io_pyro.py:21: error: Name 'Optional' is not defined
arviz/data/io_pyro.py:21: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Optional")
arviz/data/io_numpyro.py:19: error: Name 'Optional' is not defined
arviz/data/io_numpyro.py:19: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Optional")
```

I also replaced use of `builtins.callable` with `typing.Callable` in type hints since `mypy` doesn't consider the lowercase version a valid type:
```
arviz/data/io_pyro.py:22: error: Function "builtins.callable" is not valid as a type
arviz/data/io_pyro.py:22: note: Perhaps you need "Callable[...]" or a callback protocol?
```

Unfortunately, this isn't a comprehensive set of type hint fixes -- `mypy` is still quite unhappy even with these changes applied. I just figured something is better than nothing 😃

### DIFF
--- a/arviz/data/io_numpyro.py
+++ b/arviz/data/io_numpyro.py
@@ -1,6 +1,6 @@
 """NumPyro-specific conversion code."""
 import logging
-from typing import Optional
+from typing import Callable, Optional
 
 import numpy as np
 import xarray as xr
@@ -17,7 +17,7 @@ class NumPyroConverter:
 
     # pylint: disable=too-many-instance-attributes
 
-    model = None  # type: Optional[callable]
+    model = None  # type: Optional[Callable]
     nchains = None  # type: int
     ndraws = None  # type: int
 
--- a/arviz/data/io_numpyro.py
+++ b/arviz/data/io_numpyro.py
@@ -1,5 +1,6 @@
 """NumPyro-specific conversion code."""
 import logging
+from typing import Optional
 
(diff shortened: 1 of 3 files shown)


ahartikainen (approved):

ahartikainen:
We currently don't implement or require type hints, but I think it could be a good addition to our lib. Is there yet a good solution for numpy / xarray related problems (e.g. we would not need to use Any)

OriolAbril (approved):

obi1kenobi:
> Is there yet a good solution for numpy / xarray related problems (e.g. we would not need to use Any)

It seems like some solid progress is being made within the packages themselves: https://github.com/numpy/numpy/issues/7370 It doesn't seem like it's super thorough yet, but it seems quite actively being worked on so it'll hopefully get even better quickly.

PR #1398, opened 2020-09-27
Pass bool instead of int values to pytest.mark.skipif

obi1kenobi:
Resolves the following type errors that mypy points out:
```
arviz/tests/base_tests/test_utils_numba.py:15: error: Argument 1 to "__call__" of "_SkipifMarkDecorator" has incompatible type "int"; expected "Union[str, bool]"  [arg-type]
arviz/tests/base_tests/test_plot_utils.py:203: error: Argument 1 to "__call__" of "_SkipifMarkDecorator" has incompatible type "int"; expected "Union[str, bool]"  [arg-type]
arviz/tests/base_tests/test_plot_utils.py:281: error: Argument 1 to "__call__" of "_SkipifMarkDecorator" has incompatible type "int"; expected "Union[str, bool]"  [arg-type]
arviz/tests/base_tests/test_diagnostics_numba.py:17: error: Argument 1 to "__call__" of "_SkipifMarkDecorator" has incompatible type "int"; expected "Union[str, bool]"  [arg-type]
arviz/tests/base_tests/test_stats_numba.py:18: error: Argument 1 to "__call__" of "_SkipifMarkDecorator" has incompatible type "int"; expected "Union[str, bool]"  [arg-type]
```

### DIFF
--- a/arviz/tests/base_tests/test_plot_utils.py
+++ b/arviz/tests/base_tests/test_plot_utils.py
@@ -23,6 +23,10 @@ from ...utils import get_coords
 from ..helpers import running_on_ci
 
 
+# Check if Bokeh is installed
+bokeh_installed = importlib.util.find_spec("bokeh") is not None  # pylint: disable=invalid-name
+
+
 @pytest.mark.parametrize(
     "value, default, expected",
     [
@@ -200,7 +204,7 @@ def test_filter_plotter_list_warning():
 
 
 @pytest.mark.skipif(
-    (importlib.util.find_spec("bokeh") is None) & ~running_on_ci(),
+    not (bokeh_installed or running_on_ci()),
     reason="test requires bokeh which is not installed",
 )
 def test_bokeh_import():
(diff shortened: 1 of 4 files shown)


arviz/tests/base_tests/test_plot_utils.py
```diff
@@ -23,6 +23,10 @@
 from ..helpers import running_on_ci
 
 
+# Check if Bokeh is installed
+bokeh_installed = importlib.util.find_spec("bokeh") is not None  # pylint: disable=invalid-name
```
obi1kenobi, line 27: I just moved this to the top (it used to be in the middle of the file) and then made all the `skipif` decorators in this file use the same logical expression as the one on line 310. I chose the expression from that line since it seemed the most obvious and explicit to me, but I'm happy to switch to a different one if the maintainers have a different preference.

ahartikainen (approved):

ahartikainen:
Thanks for the PR

PR #1406, opened 2020-10-04
Documentation reorganization

OriolAbril:
## Description
Related to #1331. This PR intends to reorganize the documentation in order to make it easier to extend it. There are several important changes to do this:
* Use MyST for notebooks and markdown files. 
   - [MyST](https://github.com/executablebooks/myst-parser) (and [MyST-NB](https://github.com/executablebooks/MyST-NB)) is a sphinx parser for markdown that also extends markdown in order to allow using sphinx directives from markdown. This allows to keep the schema written in markdown and to use sphinx directive from the example jupyter notebooks. Now for example, notebooks can be referred to using `ref` links and they can use `ref`, `class`, `func`... links to refer to other pages.
* Change the theme to [pydata_sphinx_theme](https://github.com/pandas-dev/pydata-sphinx-theme) to get automatic tables of contents for both page and section at the same time (one in each sidebar)
* Reorganize the folders in which the documentation is stored.
 
todo list:
- ~Fix gallery generator~
- ~Think about pulling relevant docs (i.e. CONTRIBUTING is already included in the docs) such as governance, code of conduct, readme?...~ I think it's fine for now

EDIT: preview: https://oriolabril.github.io/arviz/index.html

## Checklist

- [x] Follows [official](https://github.com/arviz-devs/arviz/blob/master/CONTRIBUTING.md#pull-request-checklist) PR format
- [x] New features are properly documented (with an example if appropriate)?
- [x] Code style  correct (follows pylint and black guidelines)
- [x] Changes are listed in [changelog](https://github.com/arviz-devs/arviz/blob/master/CHANGELOG.md#v0xx-unreleased)

PR #1407, opened 2020-10-05
Prevent wrapping group names in InferenceData HTML repr

sethaxen:
## Description

The HTML repr for `InferenceData` reuses some of the classes of xarray's HTML repr for `xarray.Dataset`.
The `xr-sections` class determines the grid layout of the list of sections in the `Dataset` repr. However, the same class is applied to the list of groups in the `InferenceData` repr. The result is that the same grid layout is applied to both (note that the left-most and two right-most columns are the same width):
Screen Shot 2020-10-04 at 8 50 36 PM
This isn't a problem in the arviz docs or Jupyter notebooks with the default theme, but a different Jupyter theme is used or the `InferenceData` HTML repr is used in a different page, then if the font size is too large, the group names can exceed the fixed width of the grid. This happens in the [ArviZ.jl docs](https://arviz-devs.github.io/ArviZ.jl/v0.4.5/quickstart/#Plotting-with-CmdStan.jl-outputs), because they use a different font size than the arviz docs (thanks to @ColCarroll for pointing that out!)
Screen Shot 2020-10-04 at 9 05 25 PM

This PR adds a `group-sections` subclass to `xr-sections` that is used for the groups list, and in the CSS resets the grid for just that list while keeping all other CSS attributes. The result is that only the grid is changed:
Screen Shot 2020-10-04 at 8 53 33 PM

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Added `__getitem__` magic to InferenceData ([1395](https://github.com/arviz-devs/arviz/pull/1395))
 
 ### Maintenance and fixes
+* prevent wrapping group names in InferenceData repr_html ([1407](https://github.com/arviz-devs/arviz/pull/1407))
 
 ### Deprecation
 
--- a/arviz/static/css/style.css
+++ b/arviz/static/css/style.css
@@ -66,6 +66,10 @@ body.vscode-dark {
   grid-template-columns: 150px auto auto 1fr 20px 20px;
 }
 
+.xr-sections.group-sections {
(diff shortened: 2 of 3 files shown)


sethaxen:
I'd appreciate any suggestions on a good test for this.

sethaxen:
The failures don't seem to be unique to this PR (they also show up in e.g. https://github.com/arviz-devs/arviz/pull/1406).

ColCarroll:
Thanks, Seth! I'm not super worried that we do a great job testing CSS :D

PR #1408, opened 2020-10-05
add group info to summary

ahartikainen:
## Description

Add automatic group handling. Defaults to posterior, but works now with InferenceData without posterior (e.g. prior).

## Checklist

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -990,6 +990,7 @@ def summary(
     data,
     var_names: Optional[List[str]] = None,
     filter_vars=None,
+    group=None,
     fmt: str = "wide",
     kind: str = "all",
     round_to=None,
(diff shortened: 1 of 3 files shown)


ahartikainen:
I think this is ready.

Do you think 1 group is fine or do we want to enable multiple groups?

arviz/stats/stats.py
```diff
@@ -1142,9 +1146,26 @@ def summary(
...
-    var_names = _var_names(var_names, posterior, filter_vars)
-    posterior = posterior if var_names is None else posterior[var_names]
+
+    if isinstance(data, InferenceData):
+        if group is None:
+            if not data.groups():
+                raise TypeError("InferenceData does not contain any groups")
+            if "posterior" in data:
+                dataset = data["posterior"]
+            elif "prior" in data:
```
OriolAbril, line 1156: should we have `kind` default to `stats` in some cases? diagnostics are not really useful for prior samples

OriolAbril (commented):
LGTM, the comment is a minor nit that can be ignored

arviz/stats/stats.py
ahartikainen: We could, unless user uses mcmc to do the prior predictive sampling. (e.g. in Stan defining parameters and only priors --> this can lead to autocorrelation, but could be easier to do with complex models)

OriolAbril: Let's merge like this then

PR #1416, opened 2020-10-13
add warmup for cmdstanpy

ahartikainen:
## Description

Enable to save warmup on CmdStanPy fits.

This assumes that sample is taken with `save_warmup` in CmdStanPy side. If not, then no warmup is saved.

## Checklist


ahartikainen:
I think, I need to update the test data for CmdStanPy (by adding one example with warmup=True).

ahartikainen:
This is ready for review.

ahartikainen:
@mitzimorris could you check the use of some internal variables, are they ok for you.

PR #1442, opened 2020-11-15
avoid repeated warning in summary

### LINKED ISSUE #1441: Repeated warning in summary() when only one chain is sampled
Sometimes I sample just one chain (yeah I know, call the Bayesian police on me). When then calling `az.summary()` I get tons of: `arviz.stats.stats_utils - WARNING - Shape validation failed: input_shape: (1, 2000), minimum_shape: (chains=2, draws=4)` one for each RV, which can be a lot. I think that warning could just be dropped, if only one chain is present I would for example not assume an Rhat value to be calculated but I would still like to see the rest of the summary.

aloctavodia:
Fix  #1441


ahartikainen:
I don't think this is correct fix.

aloctavodia:
Open to suggestions or even alternative PRs

ahartikainen:
So, if I have 2 idata both 1 chain, we should throw that warning everytime a summary is called.

aloctavodia:
Yes, this will handle that case
Captura de pantalla de 2020-11-15 15-05-34

twiecki:
Thanks @aloctavodia !

PR #1448, opened 2020-11-23
[Proposal] Add gray blue style

canyon289:
Proposing adding a new grayblue scale style. This is the exact same as `arviz-grayscale` except the second color is blue rather than another gray.

The purpose is for the ArviZ CRC book which can only use two colors, and gray on gray doesn't look so great. Very open to thoughts on this one, particularly from fellow book authors @aloctavodia @junpenglao and @ahartikainen 

image
image

### DIFF
--- a/arviz/plots/styles/arviz-graybluescale.mplstyle
+++ b/arviz/plots/styles/arviz-graybluescale.mplstyle
@@ -1,0 +1,41 @@
+# This style is based on Seaborn-white parameters, the main differences are
+# Default matplotlib font (no problem using Greek letters!)
+# Larger font size for several elements
+# Colorblind friendly color cycle
+figure.figsize: 7.2, 4.8
+figure.dpi: 100.0
+figure.facecolor: white
+figure.constrained_layout.use: True
+text.color: .15
+axes.labelcolor: .15
+legend.frameon: False
+legend.numpoints: 1
+legend.scatterpoints: 1
+xtick.direction: out
+ytick.direction: out
+xtick.color: .15
+ytick.color: .15
+axes.axisbelow: True
+lines.solid_capstyle: round
+
+axes.labelsize: 15
+axes.titlesize: 16
+xtick.labelsize: 14
+ytick.labelsize: 14
+legend.fontsize: 14
+
+axes.grid: False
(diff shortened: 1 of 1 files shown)


ColCarroll:
I particularly like bigger lines and markers, and this is opinionated, so no offense taken if you don't pick this up, but if you add 

```
lines.linewidth: 3
lines.markerfacecolor: None
lines.markeredgewidth: 2
lines.markersize: 8
```

to the stylesheet in this PR, it ends up looking like this:
image

ColCarroll:
Ah, you used `'.'`, not `'o'` as a marker. Here's the same plot:
image

aloctavodia:
The `arviz-grayscale` style already have blue (#1164). The color cycle is "4 greys" and the "arviz-bue" ["000000", "4a4a4a", "7e7f7f", "b8b8b8", "2a2eec"]. The reason to have the blue as fifth color and not second is that, this provides an automatic greyscale color for most plots  (for example a plot_trace with 4 chains) and the chance to use the blue color as a highlight color when needed.

canyon289:
> The `arviz-grayscale` style already have blue (#1164). The color cycle is "4 greys" and the "arviz-bue" ["000000", "4a4a4a", "7e7f7f", "b8b8b8", "2a2eec"]. The reason to have the blue as fifth color and not second is that, this provides an automatic greyscale color for most plots (for example a plot_trace with 4 chains) and the chance to use the blue color as a highlight color when needed.

For book what about plots like above of two regressions. The gray on gray is harder to read than blue gray. Do you agree disagree with this first point?

If so the second question, is if a dedicated theme is a good idea, or if the color should be manually specified

aloctavodia:
I agree on the first point, but I think it can fixed by manually setting one of the colors to "C4" (or "C2", "C3" and maybe using the "o" marker instead of "."). I do not think a new theme is necessary as we are only changing the order of the colors.
 
The following is speculation but I suspect the editorial will try to reduce the color as much as possible (to reduce costs), so maybe they suggest using markers or a lighter gray like "C2". Although I personally like the touch of color

canyon289:
Thanks colin and osvaldo!

PR #1449, opened 2020-11-23
default credible interval clarification

### LINKED ISSUE #659: Default credible_interval clarification
This isn't exactly a proper issue, but wasn't sure where to put it. (Is there a place for discussion for Arviz like PyMC3's discourse?)

Is there a documented place on the Arviz site or documentation that details the default `credible_interval` of 0.94? I was confused about this until I read in @aloctavodia 's book that, "This is a friendly reminder of the arbitrary nature of the 95% value."

It would be nice to include this somewhere in public documentation. I thought I was missing something about the importance of 0.94.

aloctavodia:
fix #659

Add inline comment in arvizrc.template and a paragraph with a short code example in the ArviZ quickstart guide

### DIFF
--- a/arvizrc.template
+++ b/arvizrc.template
@@ -48,7 +48,7 @@ plot.matplotlib.show         : false       # call plt.show. One of "true", "fals
 
 ### STATS ###
 # rcParams related with statistical and diagnostic functions
-stats.hdi_prob      : 0.94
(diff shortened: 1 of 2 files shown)


hectormz:
Thank you @aloctavodia !

Issue #1295, opened 2020-07-15
Add code for school examples to examples repo


OriolAbril:
Add code used to generate the `centered_eight` and `non_centered_eight` examples to https://github.com/arviz-devs/arviz_example_data.

The code is available at https://arviz-devs.github.io/arviz/notebooks/Introduction.html#Plotting-with-PyMC3-objects and at https://arviz-devs.github.io/arviz/notebooks/InferenceDataCookbook.html#From-PyMC3

sudip-padhye:
If it is still available shall I pick this up?

OriolAbril:
It is still available, I have assigned it to you. 

The [contributing guide](https://github.com/arviz-devs/arviz/blob/master/CONTRIBUTING.md) and the readme in https://github.com/arviz-devs/arviz_example_data should guide you through the process and always feel free to ask questions :)

PR #1450, opened 2020-11-26
Add docstring for local dataset generating code

canyon289:
Adds references to where local datasets can be reproduced

Related to https://github.com/arviz-devs/arviz/issues/1295


ColCarroll (approved):

OriolAbril (approved):
We could also add the link to the docstring of `load_arviz_data`

PR #1452, opened 2020-11-27
automatically get the current axes instance

aloctavodia:

This recover the old behavior, allowing to do this, without the need to explicitly need to pass the axes
 
a= np.random.normal(0, 1, 100)
b= np.random.normal(0, 1, 100)
az.plot_kde(a)
az.plot_kde(b)

test


ahartikainen:
This will get only 1 axis?

Should this go inside create axis grid with some keyword flag?

Can this access also figure -> axes?

aloctavodia:
This will get only one axis but that's the case when you do something like 

az.plot_kde(a)
az.plot_kde(b)

For more complex stuff you will need to pass the axes as usual, as is the case for the rest of our plots.

PR #1460, opened 2020-12-08
Revert despine on bokeh

ahartikainen:
## Description

Remove despine from bokeh violinplot. Each ax (bokeh.figure) is its own object, so we can rearrange and handle each ax individually.

Despining is confusing when user uses other than default for figure location.

## Checklist

### DIFF
--- a/arviz/plots/backends/bokeh/violinplot.py
+++ b/arviz/plots/backends/bokeh/violinplot.py
@@ -58,7 +58,6 @@ def plot_violin(
     else:
         ax = np.atleast_2d(ax)
 
-    current_col = 0
     for (var_name, selection, x), ax_ in zip(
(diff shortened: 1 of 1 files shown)


aloctavodia:
Currently subplots in violinplot are not independent, if you move one the other will moves too.

ahartikainen:
Sure, that is by design. But the subplot order is independent. The default is grid, but there is also possibility to use rectangular (and also do it manually)

aloctavodia:
ok, could we keep despine as optional or could we have it activated by default when "sharey=True"?

ahartikainen:
sharey is doing different thing on bokeh.

We could have some kw if we want

aloctavodia (approved):
LGTM!

PR #1464, opened 2020-12-13
Increase code coverage (stats.density_utils.py)

LegrandNico:
This PR tries to improve testing and code coverage of `arviz.stats.density_utils.py`.
* Add an argument to the `test_plots_kde()` function to cover the case of adaptive KDE.
* Remove the deprecated `_fast_kde()` function


canyon289:
Looks good to me. Thank you for contribution!

PR #1476, opened 2020-12-24
plot_trace, do not plot divergences with rankplots

aloctavodia:
When kind = rank_vlines or rank_bars, and if divergences are present they will be plotted only under the KDE/histogram

Update. A few examples of how it look without this fix

lala_old

lala_old2

bokeh

bokeh2


canyon289:
Why would we not want divergences plotted for the rank diagnostic?

aloctavodia:
What we are doing now is just wrong, divergences are shown at zero when using matplotlib and for the first draws/chains ranks when using bokeh. I think showing divergences with rankplots could be confusing and showing then for the KDEs is enough.

canyon289:
Agree, LGTM

PR #1480, opened 2020-12-25
Add verbiage for unassigning issue tickets

canyon289:
Add a policy so we can unassign people from issue tickets without having to feel bad about it as the expectations are clear and written unambiguously for all


ColCarroll:
I wonder if there's a bot for this? It'd be cool to automatically ping after 12 days, then unassign at 14 with a reasonable message (feel free to pick it back up if you have time!). 

I approve of this, but will wait if others have comments.

AlexAndorra:
+1 for @ColCarroll's suggestions -- if such a bot exists 🤔

aloctavodia:
maybe this is what we want https://github.com/probot/reminders?

canyon289:
We want a conditional robot though right? Like that same robot but with this stale addon?
https://github.com/probot/stale

PR #1484, opened 2021-01-03
Add always-visible titles to the example gallery.

rpgoldman:
Previously the only titles were in tooltips, which made it near impossible to scan the page and find the example you wanted.

Note that this is a WIP if only because there's an Easter Egg in one of the titles that needs fixing!

### DIFF
--- a/doc/sphinxext/gallery_generator.py
+++ b/doc/sphinxext/gallery_generator.py
@@ -13,6 +13,8 @@ import shutil
 import token
 import tokenize
 
+from typing import Optional
+
 import matplotlib
 import matplotlib.pyplot as plt
 from bokeh.io import export_png
@@ -101,7 +103,7 @@ INDEX_TEMPLATE = """
         filter:alpha(opacity=100); /* For IE8 and earlier */
     }}
 
-    .figure span {{
+    span.figure-label {{
         position: absolute;
         display: inline;
         left: 0;
@@ -127,10 +129,24 @@ INDEX_TEMPLATE = """
     }}
 
     .caption {{
-        position: absolue;
+        position: absolute;
         width: 180px;
         top: 170px;
         text-align: center !important;
+    }}
+
+    .figure .gallery-figure-title p {{
+        position: relative;
+        top: 170px;
+        color: black;
+        visibility: visible;
+        text-align: center !important;
+        line-height: normal;
+    }}
+    .figure .gallery-figure-title span {{
+        top: 170px;
+        position: relative;
+        visibility: visible;
     }}
     </style>
 
@@ -143,16 +159,33 @@ Example gallery
 
 """
 
+CONTENTS_ENTRY_TEMPLATE = (
+    ".. raw:: html\n\n"
+    "    <div class='figure align-center'>\n"
+    "    <a href=./{0}/{1}>\n"
+    "    <img src=../_static/{2}>\n"
+    "    <span class='figure-label'>\n"
+    "    <p>{3}</p>\n"
+    "    </span>\n"
+    '    <span class="gallery-figure-title">\n'
+    "      <p>{4}</p>\n"
+    "    </span>\n"
+    "    </a>\n"
+    "    </div>\n\n"
+    "\n\n"
+    ""
+)
+
 
 def create_thumbnail(infile, thumbfile, width=275, height=275, cx=0.5, cy=0.5, border=4):
-    baseout, extout = op.splitext(thumbfile)
+    # baseout, extout = op.splitext(thumbfile)
 
     im = image.imread(infile)
     rows, cols = im.shape[:2]
     size = min(rows, cols)
     if size == cols:
         xslice = slice(0, size)
-        ymin = min(max(0, int(cx * rows - size // 2)), rows - size)
+        ymin = min(max(0, int(cy * rows - size // 2)), rows - size)
         yslice = slice(ymin, ymin + size)
     else:
         yslice = slice(0, size)
@@ -179,6 +212,8 @@ def indent(s, N=4):
 class ExampleGenerator:
     """Tools for generating an example page from a file"""
 
+    _title: Optional[str]
+
     def __init__(self, filename, target_dir, backend, thumb_dir):
         self.filename = filename
         self.target_dir = target_dir
@@ -201,6 +236,12 @@ class ExampleGenerator:
             print("skipping {0}".format(self.filename))
 
     @property
+    def title(self) -> str:
+        if self._title is not None:
+            return self._title
+        return self.modulename
+
+    @property
     def dirname(self):
         return op.split(self.filename)[0]
 
@@ -209,7 +250,7 @@ class ExampleGenerator:
         return op.split(self.filename)[1]
 
     @property
-    def modulename(self):
+    def modulename(self) -> str:
         return op.splitext(self.fname)[0]
 
     @property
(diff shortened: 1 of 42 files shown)


doc/sphinxext/gallery_generator.py
```diff
@@ -143,16 +159,33 @@
...
+    "    </a>\n"
+    "    </div>\n\n"
+    "\n\n"
+    ""
+)
+
 
 def create_thumbnail(infile, thumbfile, width=275, height=275, cx=0.5, cy=0.5, border=4):
-    baseout, extout = op.splitext(thumbfile)
+    # baseout, extout = op.splitext(thumbfile)
```
ColCarroll, line 181: i guess this can just be deleted?

doc/sphinxext/gallery_generator.py
```diff
@@ -200,6 +235,12 @@ def __init__(self, filename, target_dir, backend, thumb_dir):
 
             print("skipping {0}".format(self.filename))
 
+    @property
+    def title(self) -> str:
+        if self._title is not None:
```
ColCarroll, line 240: `Optional[str]` implies a default of `None`? Or do you have to `_title: Optional[str] = None`?

doc/sphinxext/gallery_generator.py
```diff
@@ -282,20 +323,41 @@ def extract_docstring(self):
             break
 
         thumbloc = None
-        for i, line in enumerate(docstring.split("\n")):
+        title: Optional[str] = None
+        ex_title: str = ""
+        for line in docstring.split("\n"):
+            # capture the first non-empty line of the docstring as title
+            if ex_title == "":
+                ex_title = line
```
ColCarroll, line 331: do you want a `continue` here, to avoid setting the title to `_thumb: 0.3, 0.6` or whatever?

examples/matplotlib/mpl_plot_ess_evolution.py
```diff
@@ -3,6 +3,7 @@
 =================
 
 _thumb: .2, .8
+_example_title: Plot evolution of Expected Sample Size (ESS)
```
ColCarroll, line 6: `effective` sample size?

examples/matplotlib/mpl_plot_loo_pit_overlay.py
```diff
@@ -3,6 +3,7 @@
 ====================
 
 _thumb: .5, .7
+_example_title: Plot LOO overlayed on <i>some damn thing</i>.
```
ColCarroll, line 6: watch your language!

ColCarroll (commented):
I found the easter egg! 

Is there a reason to not just use `ex_title` as the `title`? We could even start throwing exceptions or something if it wasn't found. I don't feel strongly either way.

doc/sphinxext/gallery_generator.py
rpgoldman: yes.  Done (will be in next push).

doc/sphinxext/gallery_generator.py
rpgoldman: If I do that, I think I end up with a class property instead of an instance property, don't I?

`Optional[str]` does not imply anything about the value.  I will set it to `None` in the `__init__`.

doc/sphinxext/gallery_generator.py
rpgoldman: Actually, we can still end up setting the title to `_thumb`, even with that change.  To do what you suggest, I have to reorder the clauses in the body of the for loop. I'll do that now.

examples/matplotlib/mpl_plot_loo_pit_overlay.py
rpgoldman: I was hoping you knew what it was overlayed on...

examples/matplotlib/mpl_plot_ess_evolution.py
rpgoldman: Fixed!

rpgoldman:
> I found the easter egg!
> 
> Is there a reason to not just use `ex_title` as the `title`? We could even start throwing exceptions or something if it wasn't found. I don't feel strongly either way.

I put changes in `ex_title` for a couple of reasons:

1. It lets us put something different on the thumbnail for people who are searching for a particular plot.
2. It also lets us have longer titles, if we want, that wouldn't fit on the examples page.

examples/matplotlib/mpl_plot_loo_pit_overlay.py
rpgoldman: ...and whether it's "overlayed" or "overlaid"!

canyon289: overlaid i believe

OriolAbril: I would write something like `Plot LOO-PIT KDE overlaid on uniform distributions KDEs`, but I find it quite hard to summarize that more that what is said in the Examples section of its [API docs](https://arviz-devs.github.io/arviz/api/generated/arviz.plot_loo_pit.html) without loosing so much meaning it can only be understood if you already know what it means

rpgoldman: That one is hugely complex because, per @avehtari's explanation, one needs to know *why* it's important to compare with uniform distributions to understand what the plot is for.

avehtari: Can we assume someone using this would know what are CDF and inverse-CDF? Maybe Wikipedia's PIT article helps https://en.wikipedia.org/wiki/Probability_integral_transform ?

rpgoldman: @avehtari Probably we can make that assumption, in which case we can give this a title that explains it sufficiently for such a person.

Plot LOO overlaid on uniform distributions to assess convergence.

Would that be good?

This does beg the question -- if a person does *not* know the purpose of this plot, where should they look?  API docs? Examples? Or do we need a tutorial of some form?

rpgoldman:
One thing that would be nice to fix before merging: separation plot.

1. the only description of the separation plot cited is from an article hidden behind a paywall.
2. the thumbnail is just a wall of blue.

examples/matplotlib/mpl_plot_loo_pit_overlay.py
OriolAbril: I would say that ideally they would look at the API docs (each example already has a link to the relevant function) if necessary they can go to other resources from there, reference papers, it's page in the [educational materials repo](https://github.com/arviz-devs/Exploratory-Analysis-of-Bayesian-Models)  (which does not exist yet) and other relevant links that may be present.

rpgoldman: I put in the above suggested title -- "Plot LOO overlaid on uniform distributions to assess convergence."  Marking this as resolved.

rpgoldman:
I just pushed a new version with a better-centered matplotlib separation plot.

rpgoldman:
Calling this good enough and clearing the WIP flag.

examples/matplotlib/mpl_plot_loo_pit_overlay.py
avehtari: 1. It's missing a word: Plot LOO "what" overlaid on...
LOO can mean many things, and the confusion could that some would think it's LOO elpds. If you think that probability integral transformation is not well known, how about "LOO predictive ECDF" ?
2. It's not about convergence, but about model checking. So the full sentence could be
"Plot LOO predictive ECDF overlaid on ECDFs of uniform samples to assess predictive calibration"

rpgoldman (commented):
The only one I'm not sure about is the change to `mpl_plot_ess_local` -- the reason I added the phrase that you took out was that the term "local" was not clear. What makes an expected sample size local?

rpgoldman (commented):
Other than that one point, these all seem like clear improvements.

rpgoldman:
These test failures aren't due to this MR.

rpgoldman:
Last thing, if we have the energy, would be to apply these changes to the Bokeh figures, as well.

OriolAbril:
> What makes an expected sample size local?

The quantile ess is calculated with `P(theta < theta_quantile)` (see equation 16 in the reference) whereas the local ess is calculated with `P(theta_quantile < theta < theta_quantile+delta)` (eq 19). From the paper:

> (...) This gives us a local efficiency measure which is more localized than efficiency measure for quantiles and can be used to build intuition about what types of posterior functionals can be computed as illustrated in the examples. While the expectation of a function that only depends on intermediate values can be usually estimated with relative ease, expectations of tail probabilities or other posterior functionals that depend critically on the tail of the distribution will be usually more difficult to estimate. In addition, small probability intervals can be used in practical equivalence testing

reference: https://arxiv.org/pdf/1903.08008.pdf

> if we have the energy, would be to apply these changes to the Bokeh figures

I would wait for that, I want to try simplifying the example gallery using tabs like https://sphinx-panels.readthedocs.io/en/latest/#tabbed-content. After all, the title is the same, the api reference is the same, and it would allow having a single thumbnail and avoid generating bokeh thumbnails which is the main complication when generating the docs and requires weird dependencies.

rpgoldman:
In that case, I think we are ready to merge.

PR #1492, opened 2021-01-09
Resolve a few straightforward mypy type errors.

obi1kenobi:
## Description
This PR is only about improving type hint coverage and quality. It contains no new features and makes no substantive changes to existing functionality, so it should be fully covered by existing tests. This is similar in spirit to #1397 and #1398.

In particular, this PR:
- Applies a `mypy`-suggested type hint to a test helper that requires it in order to type-check.
- Applies `# type: ignore` suppressions to a few instances of a known and still-unresolved `mypy` issue https://github.com/python/mypy/issues/1153

The following `mypy`-reported errors are resolved as a result:
```
arviz/data/io_pystan.py:20: error: Name 'json' already defined (by an import)  [no-redef]
arviz/data/io_json.py:9: error: Name 'json' already defined (by an import)  [no-redef]
arviz/data/base.py:15: error: Name 'json' already defined (by an import)  [no-redef]
arviz/tests/helpers.py:259: error: Incompatible return value type (got "List[str]", expected "List[Union[str, Tuple[str, str]]]")  [return-value]
arviz/tests/helpers.py:259: note: "List" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance
arviz/tests/helpers.py:259: note: Consider using "Sequence" instead, which is covariant
arviz/tests/helpers.py:259: note: Perhaps you need a type annotation for "failed_attrs"? Suggestion: "List[Union[str, Tuple[str, str]]]"
```

Additionally, a similar `mypy` type error related to a conditional `ujson` import that would be uncovered if https://github.com/arviz-devs/arviz/pull/1491 merges is also addressed here.

## Checklist

### DIFF
--- a/arviz/data/base.py
+++ b/arviz/data/base.py
@@ -12,7 +12,9 @@ import xarray as xr
 try:
     import ujson as json
 except ImportError:
-    import json
+    # mypy struggles with conditional imports expressed as catching ImportError:
+    # https://github.com/python/mypy/issues/1153
+    import json  # type: ignore
 
 from .. import __version__, utils
 
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -20,7 +20,9 @@ from .base import _extend_xr_method, _make_json_serializable, dict_to_dataset
 try:
     import ujson as json
 except ImportError:
-    import json
+    # mypy struggles with conditional imports expressed as catching ImportError:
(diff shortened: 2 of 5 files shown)


OriolAbril (approved):

canyon289 (approved):
Thank you for the contribution!

PR #1493, opened 2021-01-10
Minor improvements to type checking and error checks in stats.py.

obi1kenobi:
## Description
- Added a missing error check on the `summary()` function's `kind` parameter. Since the check was missing, that meant that illegal `kind` values could be passed, and cause unexpected, self-inconsistent behavior in the function.
- Improved type hints to use explicit `Literal` types instead of `str`, making errors like the ones mentioned in the prior bullet point less likely to happen as `mypy` would catch them statically.
- Resolved a type error caused by unintentional variable-sharing across unrelated portions of the same function.

As with my other PRs, there are no substantive changes to existing functionality and there is no new functionality, so existing tests should cover everything already.

## Checklist

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -3,7 +3,7 @@
 import warnings
 from collections import OrderedDict
 from copy import deepcopy
-from typing import Any, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import pandas as pd
@@ -25,6 +25,9 @@ from .stats_utils import logsumexp as _logsumexp
 from .stats_utils import make_ufunc as _make_ufunc
 from .stats_utils import stats_variance_2d as svar
 from .stats_utils import wrap_xarray_ufunc as _wrap_xarray_ufunc
+
+if TYPE_CHECKING:
+    from typing_extensions import Literal
 
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


arviz/stats/stats.py
```diff
@@ -1322,12 +1329,12 @@ def summary(
         dfs = []
         for var_name, values in joined.data_vars.items():
             if len(values.shape[1:]):
-                metric = list(values.metric.values)
```
obi1kenobi, line 1325: This `metric` assignment clashed with the `metric` assignment on line 1307 in the original file: `for metric, circ_stat in zip(...)`. This was a type error since the two assignments to `metric` had different types.

Renaming this `metric` to `index_metric` resolves the problem since now the differently-typed values are assigned to different variables.

arviz/stats/stats.py
```diff
@@ -1302,7 +1310,6 @@ def summary(
         metric_names.extend(metrics_names_)
 
     if circ_var_names:
-
```
obi1kenobi, line 1305: This was extraneous whitespace that did not seem to match the style of the surrounding code, so I removed it — hope you don't mind!

arviz/stats/stats.py
```diff
@@ -1262,10 +1269,11 @@ def summary(
...
 
     # Combine metrics
-    metrics = []
-    metric_names = []
+    metrics: List[xr.Dataset] = []
+    metric_names: List[str] = []
     if extend:
-        metrics_names_ = (
+        metrics_: Tuple[xr.Dataset, ...]
+        metrics_names_: Tuple[str, ...] = (
```
obi1kenobi, line 1276: The block below is why I noticed that the `kind` error check was missing, and added it: if `kind` is not one of `"all", "stats", "diagnostics"`, then `metrics_names_` will remain set to the tuple of strings, but `metrics_` will never get assigned — even though it's about to get used at the end of the `if-elif-elif` block. This would have led to a user-unfriendly `UnboundLocalError`.

With the error check I added, supplying an illegal `kind` value is treated the same way as supplying an illegal `order` or `fmt` value: with a `TypeError` containing a descriptive and actionable error message.

OriolAbril:
LGTM, I have added an extra test and will merge if it passes to try and release this week.

PR #1494, opened 2021-01-10
Resolve RcParams MRO confusion and delete unnecessary methods.

obi1kenobi:
## Description
Working to resolve the following issue raised by `mypy`:
```
arviz/rcparams.py:286: error: Cannot determine consistent method resolution order (MRO) for "RcParams"  [misc]
```

It seems that `RcParams` inherits from both `dict` and `MutableMapping`, which is confusing `mypy` because `dict` is also a `MutableMapping`. Additionally, it seems that this confusing MRO situation has led to the need to redefine several of the `MutableMapping` methods explicitly, to avoid MRO-related dispatch issues.

I'm not sure why `RcParams` needs `MutableMapping` _in addition to_ having `dict` as a superclass, so I'm leaving this PR as a draft for now. None of the documentation for the class seems to explain this quirk, so I'm hoping that either test cases surface some leads, or one of the maintainers recalls what the reason might be.

If the tests don't show any issues, and none of the maintainers see a reason to have the multiple inheritance setup and its associated MRO issues, then it may be worth removing `MutableMapping` as this PR does, and clearing up the `mypy` error and the two associated (and currently suppressed) `pylint` errors.

## Checklist

### DIFF
--- a/arviz/rcparams.py
+++ b/arviz/rcparams.py
@@ -6,7 +6,9 @@ import pprint
 import re
 import sys
 import warnings
+from collections.abc import MutableMapping
 from pathlib import Path
+from typing import Any, Dict
 
 import numpy as np
 
@@ -282,7 +284,7 @@ defaultParams = {  # pylint: disable=invalid-name
 }
 
 
-class RcParams(dict):
+class RcParams(MutableMapping):
     """Class to contain ArviZ default parameters.
 
     It is implemented as a dict with validation when setting items.
@@ -292,6 +294,7 @@ class RcParams(dict):
 
     # validate values on the way in
     def __init__(self, *args, **kwargs):
+        self._underlying_storage: Dict[str, Any] = {}
         super().__init__()
         self.update(*args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


OriolAbril:
>

The double inheritance is [copied from matplotlib](https://github.com/matplotlib/matplotlib/blob/4e4410df1f73bff527c93ccb828dbd9b2eae139f/lib/matplotlib/__init__.py#L594) whose rcparams class is the base for ArviZ's one. I probably assumed matplotlib devs had a reason for it and kept going. Having said that, even if there was a reason for matplotlib to do this, ArviZ's rcparams are a simplified version of matplotlib ones so if everything works and all test pass let's simplify this.

obi1kenobi:
I believe the current implementation should work. I removed the `dict` inheritance and simply used an explicit underlying `dict` to store the data — essentially, using composition instead of inheritance. I then implemented the methods `MutableMapping` requires, and rely on `MutableMapping` to implement the rest of the interface on the basis of those methods.

OriolAbril (approved):
Thanks! This looks great

PR #1501, opened 2021-01-15
Add copy() method to InferenceData

rpgoldman:
I was surprised to find that the InferenceData object didn't have a copy method. In `inference_data.py` `copy.deepcopy()` was used. So I have added a copy method that simply calls `deepcopy`.

Possibly we should replace the other uses of deepcopy in this file with invocations of the copy method, which would make things tidier if we later were to change how `InferenceData` objects are to be copied.

This is a matter of style as much as anything: I leave it to the community to decide if it's worth doing.

## Checklist


OriolAbril (approved):

ahartikainen (approved):

ahartikainen:
Needs black formatting.

aloctavodia:
@rpgoldman If you agree we could merge this and leave the replacement of the other uses of deepcopy for another PR.

rpgoldman:
Sure, that is fine. But do you think it’s worth doing that substitution later? If so, I will do it sometime when I’m waiting for some model to sample!

PR #1505, opened 2021-01-19
Avoid overwriting a variable with value of different type.

obi1kenobi:
## Description
Avoids the following `mypy` error:
```
arviz/plots/traceplot.py:183: error: "InferenceData" has no attribute "dims"  [attr-defined]
```
Part of #1496.

The error was happening because the `data` variable was initially of type `InferenceData` but was later overwritten with a value of different type on the line:
```python
data = get_coords(convert_to_dataset(data, group="posterior"), coords)
```
In general, assigning a value of a completely different type to an existing variable will cause both `mypy` and `pylint` to complain about the assignment itself. In this case, the `get_coords()` function itself was not type-hinted, i.e. its return type was considered to be `Any`. This led `mypy` to incorrectly deduce that the return type must have preserved the existing type of `data`, because not doing so would have been an immediate error.

Renaming the assignment on that line to a new variable name avoids the current error as well as the possible future `mypy` or `pylint` errors arising when and if `get_coords()` gets a type hint for its return type.

## Checklist

### DIFF
--- a/arviz/plots/traceplot.py
+++ b/arviz/plots/traceplot.py
@@ -172,19 +172,21 @@ def plot_trace(
     else:
         divergence_data = False
 
-    data = get_coords(convert_to_dataset(data, group="posterior"), coords)
+    coords_data = get_coords(convert_to_dataset(data, group="posterior"), coords)
 
     if transform is not None:
-        data = transform(data)
+        coords_data = transform(coords_data)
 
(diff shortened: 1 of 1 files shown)


ColCarroll:
LGTM -- this is interesting, because `InferenceData` is _not_ an xarray dataset, but it pretends to be. I guess this is actually a helpful distinction, because `data` has properties hanging off it, while `data.posterior` does not.

Issue #1486, opened 2021-01-08
Write "Working with InferenceData" page


OriolAbril:
## Tell us about it
As part of the documentation revamp described in #1331. We should add a page explaining how to work with `InferenceData` and show several common tasks. The guide should cover at least the following topics:

* [ ] modifying, transforming and/or creating a variable within an InferenceData group or groups. See #641 
* [ ]  how to combine/merge/extend multiple InferenceData objects
* [ ]  Stacking dimensions (maybe even reshaping? reshaping is [a bit crazy](https://github.com/arviz-devs/xarray_examples/blob/master/ReshapeDataset.ipynb) though), fancy chain/draw extraction methods from #1469 
* [ ] Some grouping/aggregating of data
* [ ] Coordinate modification. Maybe something like #1461 
* [ ] ...

## Thoughts on implementation

Feel free to tackle parts of the issue instead of generating the whole section at once.

It is also probably a good idea to add a See also section at the bottom linking to other docs (i.e. pymc3 or pystan docs), blogs and other resources that _use_ inferencedata. For example: https://docs.pymc.io/notebooks/multilevel_modeling.html

kenkirito:
can i work on this issue

OriolAbril:
Great, thanks!

kenkirito:
is there any discord server or any community chat or meetings to attain. I am done with the setup. just wanted to know more about the project

OriolAbril:
We have a gitter chat available: https://gitter.im/arviz-devs/community, and there are also questions and discussions about the project at PyMC discourse: https://discourse.pymc.io/ and to a lesser extent at Stan discourse: https://discourse.mc-stan.org/

almostmeenal:
> * [ ]   how to combine/merge/extend multiple InferenceData objects

This sounds good, I recently went through InferenceData docs and they do seem ambiguous and all over the place, the entire issue seems like something that would probably be done over multiple PRs, but I guess to begin with I can take care of this part, I'd like to take it up if noone is working on it at the moment.

OriolAbril:
Great, note that you have to extend the currently existing notebook at https://github.com/arviz-devs/arviz/blob/main/doc/source/getting_started/WorkingWithInferenceData.ipynb

There are some examples on how to combine inferencedata objects at https://arviz-devs.github.io/arviz/api/generated/arviz.concat.html#arviz.concat. 

[...]

PR #1506, opened 2021-01-19
add examples of InferenceData common manipulations 

aloctavodia:
This adds a new notebook with a few examples. Is by no means exhaustive but I think is better to have something than nothing and then it will be easier to keep adding more examples.

This also renames the already existing notebook InferenceDataCookbook to InferenceDataIO and use the name InferenceDataCookbook for the new one.

EDIT: fixes partially issue #1486


OriolAbril (changes requested):
We don't have reviewnb here so I'll do my best to clarify where is each comment referring to.

First of all, the notebook titles have to be updated. The toctree is created using the file names, but sphinx then shows the file _title_ when generating the html docs. Currently both files have InferenceData cookbook as title. I personally would suggest changing both titles. We have had the cookbook on the website for quite a long time and most users are probably familiar with it, I think leaving the same title but with different content will be confusing, changing everything will send the "nice updated docs!" message. How about "creating InferenceData" and "working with InferenceData"?

When calculating the mean in the last example, the first cell has `idata.posterior` instead of `idata.posterior.mean()`.

You'll see that in the cookbook there is a strange `(cookbook)=` thing. This is equivalent to the `.. cookbook:` rst anchors. I would add one to the new notebook to easily link there from other parts of the documentation like the InferenceData api docs

doc/source/getting_started/index.md
```diff
@@ -5,5 +5,13 @@
...
+CreatingInferenceData
+WorkingWithInferenceData
 ```
+
+     4D            2C       2C         4D
+A 1001101 años del 101100 y 101100 del 1001101.
+
+
+44 2C
+77 4D
```
OriolAbril, line 17: looks like something got copy pasted unintentionally

OriolAbril (commented):
I think all `from_xyz` link to `cookbook` anchor on their api docs, i.e. https://github.com/arviz-devs/arviz/blob/idata_examples/arviz/data/io_pymc3.py#L516

doc/source/getting_started/index.md
aloctavodia: haha, at least is not something sensitive like my bank account password, haha

OriolAbril (approved):

PR #1510, opened 2021-01-20
[docs] Update requirements-docs.txt

ColCarroll:
Might fix the docs!


ahartikainen:
Can you add [docs] to title and rerun the CI (so it will build docs)

OriolAbril:
I think docs will be built anyways but will not be uploaded as an artifact when [docs] is missing :thinking: 

Anyway, I was coming here to ask that even tough this will fix the docs, it would be great to wait until after #1512 is merged to make sure it actually works.

PR #1521, opened 2021-01-23
[docs] Tabs for example gallery!

OriolAbril:
## Description
As proposed in #1484, this PR modifies the example gallery to have a single page per type of plot.
Each page has then one tab per backend to ease comparison between backends.

You can see the preview at https://oriolabril.github.io/arviz/examples/index.html, let me know what you think about it.

Also, side note, now only one thumb per example is required, therefore, we can use matplotlib to generate thumbs and forget about exporting bokeh plots as png files and the extra dependencies it requires which I believe will simplify building docs locally.

## Checklist

      PR format?
- [x] Is the fix listed in the [Documentation](https://github.com/arviz-devs/arviz/blob/master/CHANGELOG.md#documentation)
      section of the changelog?

### DIFF
--- a/.azure-pipelines/azure-pipelines-docs.yml
+++ b/.azure-pipelines/azure-pipelines-docs.yml
@@ -53,7 +53,6 @@ jobs:
       # install emcee
       python -m pip install emcee
       python -m pip install .
-      conda install -y geckodriver firefox --channel conda-forge
     displayName: 'Install requirements'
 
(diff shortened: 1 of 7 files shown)


AlexAndorra (approved):
Love it @OriolAbril 😍 
Must have been a bit hard to refactor, from what I saw 😅 But the display is much cleaner now 👌

PR #1527, opened 2021-01-28
Removed repetitive variable names from forest plots of multivariate variables

### LINKED ISSUE #1524: Variable names in element labels for `plot_forest` should be optional
Currently, creating a forest plot with `plot_forest` uses `<var name> <coordinate name>` as the scheme for labels on the y-axis. This creates a slew of repeated variable names in the labels, which I would guess is almost always undesirable (particularly when you have nice element labels as coordinates in the dataset):

Screen Shot 2021-01-27 at 1 02 58 PM

There should be excluded by default, and turned on with a `include_var_names=True` (or similar) argument. 

Alternatively, the first coordinate could include the variable name, and then exclude it for subsequent ones.

fonnesbeck:
Forest plots of multivariate variables currently prepend the variable name to each element label, which adds a lot of clutter to the plot. This removes the variable names from all but the first element.

Can add this feature to other plots as needed, either here or in a separate PR.

Not sure what sort of test would be relevant here, if any.

Closes #1524

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 
 ### Maintenance and fixes
 * Fixed ovelapping titles and repeating warnings on circular traceplot ([1517](https://github.com/arviz-devs/arviz/pull/1517))
+* Removed repetitive variable names from forest plots of multivariate variables ([1527](https://github.com/arviz-devs/arviz/pull/1527))
+
 ### Deprecation
 
 ### Documentation
--- a/arviz/plots/backends/bokeh/forestplot.py
+++ b/arviz/plots/backends/bokeh/forestplot.py
@@ -572,9 +572,16 @@ class VarHandler:
                     skip_dims=skip_dims,
                     reverse_selections=True,
                 )
-                for _, selection, values in datum_iter:
+                datum_list = list(datum_iter)
+                for _, selection, values in datum_list:
                     selection_list.append(selection)
-                    label = make_label(self.var_name, selection, position="beside")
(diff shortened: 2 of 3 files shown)


AlexAndorra (approved):
This looks great, thanks @fonnesbeck ! I don't think we need this for more plots -- or maybe [plot_parallel](https://arviz-devs.github.io/arviz/api/generated/arviz.plot_parallel.html) ?  Curious what the others think. 
The test failure is weird, looks like a problem with Azure

ahartikainen:
How does this look in the case were there are no verbose coords?

fonnesbeck:
@ahartikainen if the variable is not a multivariate random variable, then it uses the variable name. The only change should be in the multivariate case.

fonnesbeck:
@AlexAndorra yes, they look like resource provisioning errors.

fonnesbeck:
Actually, I noticed a bug with multiple non-combined chains: 

Screen Shot 2021-01-28 at 1 50 48 PM

Don't merge until I fix this.

OriolAbril:
Restarting azure and now it seems to run without problem

fonnesbeck:
OK, fixed. A multi-trace multivariate forest plot now looks like this:

Screen Shot 2021-01-28 at 6 58 29 PM

whereas a forest plot of three univariate variables looks like this:

download

aloctavodia (approved):
LGTM!

AlexAndorra (approved):
Thanks @fonnesbeck 🥳

Issue #1514, opened 2021-01-21
Update from_xyz converters to follow schema conventions


OriolAbril:
We recently merged #1063 after deciding on the "ArviZ name" for each of the parameters.

* [x] from_numpyro: @utkarsh-maheshwari updated it with #1525
* [x] from_pyro requires no extra work
* [x] from_emcee requires no extra work 
* [x] from_cmdstan: @utkarsh-maheshwari updated it with #1541
* [x] from_pymc3: @madhucharan working on it
* [x] from_pystan: @madhucharan working on it
* [x] from_cmdstanpy: @utkarsh-maheshwari working on it

## Thoughts on implementation

Some of the libraries already rename some parameters, in many cases it will only mean adding some extra mappings to what is already happening. See for example https://github.com/arviz-devs/arviz/blob/master/arviz/data/io_numpyro.py#L136-L141.

Feel free to work only on one or several converters, but please comment below which are you targetting.

utkarsh-maheshwari:
Hello community, 
I am a beginner in Open Source. I have been following all the issues related to this issue and read the whole discussion. 

I maybe wrong, but shouldn't "potential energy" be matched to "model_logp" and not "lp" because pyMC3 uses "model_logp" to depict potential energy according to this [comment](https://github.com/arviz-devs/arviz/pull/1063#issuecomment-589771200) ?
![image](https://user-images.githubusercontent.com/32706356/105612402-01da2e00-5de2-11eb-9c6d-dd4f70345204.png)

OriolAbril:
Hi, welcome! This screenshot is actually a very good example to illustrate the issue and the motivation behind establishing this sample stats naming convention. 

One of the goals of ArviZ is to be backend agnostic, thus, once converted to InferenceData, there should be no differences other than the attributes (where we store the date, libraries used to get the samples and versions...). However, each of the inference libraries uses its own naming convention, precisely as you have seen in the [comment](https://github.com/arviz-devs/arviz/pull/1063#issuecomment-589771200) you mention above. Currently, some sample stats are already renamed to the ArviZ convention, but not all of them. This issue is to make sure we enforce the sample stats names defined in the schema.

[...]

PR #1541, opened 2021-02-06
Update from_cmdstan converter to follow schema convention

utkarsh-maheshwari:
## Description
 Matched sample stats according to InferenceData schema specification in from_cmdstan converter. Addreses issue #1514

## Checklist

### DIFF
--- a/arviz/data/io_cmdstan.py
+++ b/arviz/data/io_cmdstan.py
@@ -203,22 +203,24 @@ class CmdStanConverter:
     def sample_stats_to_xarray(self):
         """Extract sample_stats from fit."""
         dtypes = {"divergent__": bool, "n_leapfrog__": np.int64, "treedepth__": np.int64}
+        rename_dict = {
+            "divergent": "diverging",
+            "n_leapfrog": "n_steps",
+            "treedepth": "tree_depth",
+            "stepsize": "step_size",
+            "accept_stat": "acceptance_rate",
+        }
 
         sampler_params, sampler_params_warmup = self.sample_stats
 
         for j, s_params in enumerate(sampler_params):
-            rename_dict = {}
             for key in s_params:
-                key_, *end = key.split(".")
-                name = re.sub("__$", "", key_)
-                name = "diverging" if name == "divergent" else name
-                rename_dict[key] = ".".join((name, *end))
-                sampler_params[j][key] = s_params[key].astype(dtypes.get(key_))
-                sampler_params_warmup[j][key] = sampler_params_warmup[j][key].astype(
-                    dtypes.get(key_)
+                name = re.sub("__$", "", key)
+                name = rename_dict.get(name, name)
+                sampler_params[j][name] = s_params[key].astype(dtypes.get(key))
+                sampler_params_warmup[j][name] = sampler_params_warmup[j][key].astype(
+                    dtypes.get(key)
                 )
-            sampler_params[j] = sampler_params[j].rename(columns=rename_dict)
-            sampler_params_warmup[j] = sampler_params_warmup[j].rename(columns=rename_dict)
         data = _unpack_dataframes(sampler_params)
         data_warmup = _unpack_dataframes(sampler_params_warmup)
         return (
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Added confidence interval band to auto-correlation plot ([1535](https://github.com/arviz-devs/arviz/pull/1535)) 
 
 ### Maintenance and fixes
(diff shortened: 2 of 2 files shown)


utkarsh-maheshwari:
Though this work could be done in other ways too, I decided not to edit the code much and do the work in least possible differences.
Is this approach right?

arviz/data/io_cmdstan.py
```diff
@@ -212,6 +212,8 @@ def sample_stats_to_xarray(self):
                 key_, *end = key.split(".")
                 name = re.sub("__$", "", key_)
                 name = "diverging" if name == "divergent" else name
+                name = "n_steps" if name == "n_leapfrog" else name
+                name = "tree_depth" if name == "treedepth" else name
```
OriolAbril, line 216: I think that now that multiple variables are renamed, using a dict as mapping like in `io_numpyro` will be more clear

arviz/data/io_cmdstan.py
```diff
@@ -212,6 +212,8 @@ def sample_stats_to_xarray(self):
                 key_, *end = key.split(".")
```
utkarsh-maheshwari, line 212: Also, according to this [comment](https://github.com/arviz-devs/arviz/pull/1063#issuecomment-589771200), stan variables don't need to be splitted and joined as none of them '.' seperated strings in their names.

arviz/data/io_cmdstan.py
```diff
@@ -205,17 +205,16 @@ def sample_stats_to_xarray(self):
...
+        rename_dict = {
+            "divergent__": "diverging",
+            "n_leapfrog__": "n_steps",
+            "treedepth__": "tree_depth",
+        }
         for j, s_params in enumerate(sampler_params):
-            rename_dict = {}
             for key in s_params:
-                key_, *end = key.split(".")
-                name = re.sub("__$", "", key_)
```
OriolAbril, line 213: Not sure about the split, @ahartikainen will know, but removing the final `__` is still necessary, there are some variables which are not renamed and should have them removed, `lp` for example

OriolAbril (changes requested):

arviz/data/io_cmdstan.py
ahartikainen: Split is needed for nD variables

    theta.1.2

ahartikainen: No wait, I need to reread the code, it might be irrelevant in this step.

utkarsh-maheshwari: > Not sure about the split, @ahartikainen will know, but removing the final `__` is still necessary, there are some variables which are not renamed and should have them removed, `lp` for example

In that case, all the stan variables in the [comment](https://github.com/arviz-devs/arviz/pull/1063#issuecomment-589771200) need to be mapped to unified name. To do that, a separate dictionary for renaming (other then already mentioned `rename_dict`) need to be made.

ahartikainen: Ok, I think that split is not necessary here.

I would use the rename dict for variables which we want to be mapped and then I would leave the rest of the variables as is (just removing `__` like we currently do)

ahartikainen: > I would leave the rest of the variables as is (just removing `__` like we currently do)

This assumes that there might be new diagnostics in future.

utkarsh-maheshwari: Okay. 
Then I am initializing `rename_dict` before the outer for loop.

arviz/data/io_cmdstan.py
```diff
@@ -203,22 +203,24 @@ def posterior_to_xarray(self):
...
-                sampler_params_warmup[j][key] = sampler_params_warmup[j][key].astype(
-                    dtypes.get(key_)
+                name = re.sub("__$", "", key)
+                name = rename_dict.get(name, name)
+                sampler_params[j][name] = s_params[key].astype(dtypes.get(key))
+                sampler_params_warmup[j][name] = sampler_params_warmup[j][key].astype(
+                    dtypes.get(key)
                 )
-            sampler_params[j] = sampler_params[j].rename(columns=rename_dict)
-            sampler_params_warmup[j] = sampler_params_warmup[j].rename(columns=rename_dict)
```
utkarsh-maheshwari, line 221: Now, we don't need these commands to rename the keys as renaming is already been taken care of in the loop.

OriolAbril (approved):
Thanks @utkarsh-maheshwari :smile:

utkarsh-maheshwari:
Done! 
Sorry, forgot to add it earlier.

OriolAbril:
No problem, we have the checklist on the pr template because we all forget if we don't see it there. Thanks for the PR!

PR #1558, opened 2021-02-13
data io - change `from_cmdstanpy` to work with CmdStanPy version 0.9.68

mitzimorris:
## Description

CmdStanPy version 0.9.68 changed several properties and methods on the CmdStanMCMC object;
critically, the protected property `_num_warmup` was replaced with property `num_draws_warmup`
and corresponding `num_draws_sampling`.  Also, the CmdStanMCMC object properties
`stan_vars_cols` and `sampler_vars_cols` provide a mapping from variable names to the output columns.

To fix `from_cmdstanpy`, I added a few helper functions which manipulate lists of variable names.
Going forward, the method `_unpack_fit` can be used instead of `_unpack_frame`.

## Checklist


mitzimorris:
I didn't add any new unit tests; existing tests work.

arviz/data/io_cmdstanpy.py
```diff
@@ -393,6 +438,105 @@ def to_inference_data(self):
             },
         )
 
+    @requires("posterior")
+    def posterior_to_xarray_v68(self):
```
ahartikainen, line 442: Could we change the logic so this is "default" and the other function is "old"?

ahartikainen (commented):
Looks good.

I think the latest release is the default way to handle (and we can drop support for the old code when time goes forward; 1year?)

arviz/data/io_cmdstanpy.py
mitzimorris: yes!

mitzimorris:
hi @ahartikainen, refactored - added a bunch of helper functions that do the data munging, specifically:

- parsing base variable names from the column labels
- munging sampler stats column names, data types

OriolAbril (approved):
It looks great, thanks! 

Regarding developing and maintaining the converters, do you think it would be easier to have a `to_arviz` in cmdstanpy instead? 

It will probably be a bit more complicated in terms of setting up tests and ci in the inference libraries repos, but I feel like all the converters are becoming increasingly complicated due to having to support multiple versions: `from_pystan` is basically 2 converters in one to work with pystan2 and 3, `from_pymc3` also has multiple checks, which get more complicated every time and it already gets to the point of monkeypatching pymc3.

mitzimorris:
> It will probably be a bit more complicated in terms of setting up tests and ci in the inference libraries repos, but I feel like all the converters are becoming increasingly complicated due to having to support multiple versions: from_pystan is basically 2 converters in one to work with pystan2 and 3, from_pymc3 also has multiple checks, which get more complicated every time and it gets to the point of monkeypatching pymc3.

I can see how this is a problem.   are you suggesting that `to_arviz` would return an InferenceData object?

OriolAbril:
> I can see how this is a problem. are you suggesting that to_arviz would return an InferenceData object?

Yes, we can discuss that on the lab meeting this Friday, I have added a point about it. Currently ArviZ has cmdstanpy as a runtime dependency for `from_cmdstanpy` to work, same for pystan, pymc3... we could change the approach so that cmdstanpy has arviz as a runtime dependency for `to_arviz` to work (and we could keep `from_cmdstanpy` in ArviZ as an alias to `to_arviz` in cmdstanpy).

It may also require some changes to ArviZ (making sure all functions used by the converters are not private methods to begin with, maybe documenting how to create converters) but in the long run it this approach will probably serve ArviZ's goals better and ease ArviZ integration with new libraries. In fact, mcx supports InferenceData natively: https://github.com/rlouf/mcx/blob/master/mcx/trace.py using this approach, and it basically uses `dict_to_dataset` and `InferenceData`.

mitzimorris:
PR is now failing because of code coverage - we need tests for CmdStanPy 0.9.65 and 0.9.68.
what is the way to set this up?
other option, merge as is.

mitzimorris:
> Yes, we can discuss that on the lab meeting this Friday, I have added a point about it. Currently ArviZ has cmdstanpy as a runtime dependency for from_cmdstanpy to work, same for pystan, pymc3... we could change the approach so that cmdstanpy has arviz as a runtime dependency for to_arviz to work (and we could keep from_cmdstanpy in ArviZ as an alias to to_arviz in cmdstanpy).

on further consideration, this isn't possible because the dependency is going in the wrong direction - CmdStanPy is upstream.  also, desiderata for CmdStanPy is minimal dependencies on other packages.

OriolAbril:
I think it's no problem that the converters for old cmdstanpy versions are not tested. I am merging like this, it can always be added in a follow up pr.

PR #1559, opened 2021-02-13
Remove varname from legend in plotppc

### LINKED ISSUE #1540: plot_ppc only plots one legend if there are multiple `observed`
**To Reproduce**
```python
import seaborn as sns
import pymc3 as pm
import pandas as pd
import arviz as az

iris = sns.load_dataset("iris")
y_s = pd.Categorical(iris["species"]).codes
x_n = "sepal_length"
x_0 = iris[x_n].values

with pm.Model() as lda:
    μ = pm.Normal("μ", mu=0, sd=10, shape=2)
    σ = pm.HalfNormal("σ", 10)
    setosa = pm.Normal("setosa", mu=μ[0], sd=σ, observed=x_0[:50])
    versicolor = pm.Normal("versicolor", mu=μ[1], sd=σ, observed=x_0[50:])
    bd = pm.Deterministic("bd", (μ[0] + μ[1]) / 2)
    trace_lda = pm.sample(1000, return_inferencedata=True)

with lda:
    ppc = pm.sample_posterior_predictive(trace=trace_lda)

az.plot_ppc(az.from_pymc3(posterior_predictive=ppc, model=lda))
```

image

**Expected behavior**
I think I'd have expected both subplots to have a legend

**Additional context**
```python
> az.__version__, pm.__version__
('0.11.1', '3.11.0')
```

MarcoGorelli:
closes #1540

## Description

Remove varname from legend in plotppc. Motivation is:
- it's already the label of the x-axis
- if the case of multiple plots, then a single legend is still valid

xref https://github.com/arviz-devs/arviz/issues/1540#issuecomment-774667864

> Of course the current behavior is problematic, because we have one legend that is supposed to work for both plots but is referring only to the first. So I suggest to have as default one legend and remove from the legend the name of the variable.

Here's what the output from `examples/matplotlib/mpl_plot_ppc.py` would now look like - note how `obs` only appears as the label of the x-axis, rather than being repeated in the legend

image

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Maintenance and fixes
 * Updated `from_cmdstan` and `from_numpyro` converter to follow schema convention ([1541](https://github.com/arviz-devs/arviz/pull/1541) and [1525](https://github.com/arviz-devs/arviz/pull/1525))
 * Fix calculation of mode as point estimate ([1552](https://github.com/arviz-devs/arviz/pull/1552))
+* Remove variable name from legend in posterior predictive plot ([1559](https://github.com/arviz-devs/arviz/pull/1559))
 
 ### Deprecation
 * Removed Geweke diagnostic ([1545](https://github.com/arviz-devs/arviz/pull/1545))
--- a/arviz/plots/backends/matplotlib/ppcplot.py
+++ b/arviz/plots/backends/matplotlib/ppcplot.py
@@ -124,14 +124,12 @@ def plot_ppc(
             plot_kwargs = {"color": color, "alpha": alpha, "linewidth": 0.5 * linewidth}
             if dtype == "i":
                 plot_kwargs["drawstyle"] = "steps-pre"
-            ax_i.plot(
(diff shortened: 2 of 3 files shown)


OriolAbril (commented):
Matplotlib looks good, thanks for the test! My only comment is boke, does bokeh not show the var names in the legend already?

MarcoGorelli:
Thanks @OriolAbril 

I'm not familiar with bokeh, but running the same code with `backend='bokeh'` gives:

image

so it looks like the Bokeh ppc plots are fine as they are?

OriolAbril:
I think the legend should be present, especially thinking about begginers who use the bokeh backend, I don't think we can assume everyone will know what do each of the colors represent. I have seen my fair share of people confused between the observed kde and the pp mean kde for example. We can open an issue to fix that in some other PR if you prefer too and merge this as is.

MarcoGorelli:
Sure, have opened #1563 for that (I'd prefer to keep it separate as it feels different to the issue which this PR addresses)

PR #1570, opened 2021-02-17
Included estimate of probability within ROPE

### LINKED ISSUE #1567: Include estimate of probability within RoPE when plotting
In cases such as `az.plot_posterior(inference_data, rope=(0.45, 0.55))`, it would be useful to include an estimate of the probability within the rope. (as an option, I suspect)

For Kruschke's version, see: http://doingbayesiandataanalysis.blogspot.com/2013/08/how-much-of-bayesian-posterior.html

we could extent this idea, in a similar way to how `ref_val` works to show probabilities for the Rope and both upper and lower tails.

Rishabh261998:
## Description
Included the estimate of probability within ROPE. To fix #1567 
## Checklist


Rishabh261998:
@OriolAbril could you take a look and give your feedback. Thanks

OriolAbril:
Will take a look whenever posssible, before that though, as the changes affect plotting, can you add some screenshots of the resulting plots?

Rishabh261998:
For:` az.plot_posterior(data, var_names=['mu'], ref_val=0, rope=(-2, 3))`

mpl

For: `az.plot_posterior(data, var_names=['mu'], ref_val=0, rope=(-2, 3), backend="bokeh")`

bokeh

arviz/plots/backends/bokeh/posteriorplot.py
```diff
@@ -187,6 +187,15 @@ def display_rope(max_data):
             line_color="red",
             line_alpha=0.7,
         )
+        less_than_left_probability = (values < vals[0]).mean()
+        less_than_right_probability = (values <= vals[1]).mean()
+        probability_within_rope = less_than_right_probability - less_than_left_probability
```
OriolAbril, line 192:
```suggestion
        probability_within_rope = ((values > vals[0]) & (values <= vals[1])).mean()
```

I think this will perform better in terms of memory usage, values are not generally too large arrays but I think it's better to be cautious. It should return the same value.

arviz/plots/backends/bokeh/posteriorplot.py
```diff
@@ -187,6 +187,15 @@ def display_rope(max_data):
...
+        ax.text(
+            x=(vals[0]+vals[1])/2,
+            y=[max_data * 0.5],
```
OriolAbril, line 195: Can you try other values and see if you can get the rope text to be between the ref_val and hdi text?

OriolAbril (commented):
Thanks for the pr :), I added a couple of cosmetic changes

Rishabh261998:
@OriolAbril is this ok?
For Matplotlib:
mpl
For Bokeh:
bokeh

OriolAbril:
Looks great now, thanks!

Rishabh261998:
Then I will commit the changes and also add it to the CHANGELOG.md

OriolAbril:
If you are up for it I think ideally both plots would look the same, green for rope and orange for ref val. If you decide to work on this make sure to fetch the lastest changes to main branch and rebase

Rishabh261998:
Yes, I will work on this. Thanks @OriolAbril 
btw why are these checks failing?

OriolAbril:
> btw why are these checks failing?

It's the code style check, in this case black here is a direct link to the message https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=3989&view=logs&j=af9a1d3a-ff8f-5483-6823-86518c773a9b&t=6a1cfc46-f4a4-5e3c-e2fb-18ea29f733bf&l=15, you can also get there by clicking details on the job and then the link to Azure.

Rishabh261998:
@OriolAbril does this work?
bokeh

OriolAbril:
Looks good, but it would be great if it was possible to hardcode the colors (in matplotlib the colors used are C0, C1... which are defined by the current style, you can [try out different styles](https://matplotlib.org/stable/tutorials/introductory/customizing.html#using-style-sheets) and see how the colors change). 

Two options to do this come to mind, the first is to add `rope_kwargs`, `ref_val_kwargs`... which is what most plots do (and what I though plot_posterior did :/ ), however, I am not sure how many kwargs should be added, do we need different kwargs for rope text and for rope line?

The 2nd is to use our helper function [`vectorized_to_hex`](https://github.com/arviz-devs/arviz/blob/main/arviz/plots/plot_utils.py#L427) in the bokeh plot. like so:

         ax.line([val, val], [0, 0.8 * max_data], line_color=vectorized_to_hex("C1"), line_alpha=0.65)

Rishabh261998:
@OriolAbril I am familiar with the 2nd method, I have seen an issue regarding that
For the first do you mean to send the colours as arguments to both matplotlib and bokeh `rope` and `ref_val` functions so that the colour used is exactly the same in both the backend?

OriolAbril:
> For the first do you mean to send the colours as arguments to both matplotlib and bokeh rope and ref_val functions so that the colour used is exactly the same in both the backend?

The general approach in plotting functions is to have the base function (in arviz/plots) do some data wrangling and (sometimes also define some defaults as kwarg dictionaries) and then pass these dictionaries to the selected backend function which is the one that does the plotting. See in pairplot for example: 

https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/bokeh/pairplot.py#L123-L136

the defaults are taken from this matplotlib defaults in order to work with mpl styles out of the box, but it is also possible to use kwarg dicts for customization that goes further or diverges from the style.

The 2 approaches I described above are not exclusive, vectorized_to_hex would also be used in the kwargs option, the difference is that the color would not be so hardcoded. I don't remember seeing anyone complaining about the colors in plot_posterior, so maybe the 2nd option is enough to handle this as it already allows some level of customization

Rishabh261998:
@OriolAbril so I am facing this problem:
In the case of bokeh plot we have different arguments for `rope_text` and `rope_line` as you had asked above. The color argument also has different names i.e 'text_color` and `line_color` respectively. So I was thinking that if we could do sth like this, we take a common `kwargs_rope` from the user which has arguments like `text_color`, `line_color`, `line_width` etc.  Then in our `display_rope` function in bokeh we can define separate kwargs for text and line i.e `kwargs_rope_ text` and `kwargs_rope_line` and then do sth like this kwargs_rope_text.setdefault('text_color', kwargs_rope.get('text_color', default value)) similarly for other kwargs. So is this method ok?
Similarly for the case of matplotlib in that case though we have a common argument for `color` for line and text. 
So if we follow this method then we also need to update the documentation and add the 'kwargs_rope'.
Similarly for  `kwargs_ref_val`

OriolAbril:
Given that nobody has ever asked to customize the rope or ref val, I think allowing to change the color only should be enough. I am not sure the feature is worth all the effort to add kwargs or split them.

Having to use styles to change this seems fine to me even if it is a bit cryptic, maybe we could also add a `rope/ref_val_color` arg but more than that is probably too muchwwork for too little to be gained

Rishabh261998:
So I will just modify the code to incorporate only the color arguments(for both `ref` and `rope`)

Rishabh261998:
@OriolAbril could you look at the changes. Thanks

arviz/plots/backends/matplotlib/posteriorplot.py
```diff
@@ -158,15 +165,20 @@ def display_ref_val():
             val,
             format_as_percent(greater_than_ref_probability, 1),
         )
-        ax.axvline(val, ymin=0.05, ymax=0.75, color="C1", lw=linewidth, alpha=0.65)
+        ref_val_props = {
+            "color": vectorized_to_hex("C1")
+            if ref_val_color is None
+            else vectorized_to_hex(ref_val_color)
+        }
+        ax.axvline(val, ymin=0.05, ymax=0.75, lw=linewidth, alpha=0.65, **ref_val_props)
```
OriolAbril, line 173: you can pass `color` as a kwarg directly to `axvline` no need to create the dict, as it is only one kwarg it does not make a difference (also see other comments on removing the if else)

arviz/plots/posteriorplot.py
```diff
@@ -24,6 +24,8 @@ def plot_posterior(
     group="posterior",
     rope=None,
     ref_val=None,
+    rope_color=None,
+    ref_val_color=None,
```
OriolAbril, line 28: I would set the default to `C1` and `C2` respectively, so you can use `vectorized_to_hex` directly down the line. You'll see that there are other string values as defaults. What should not be set as defaults are mutable objects such as dicts or lists which is why you'll not see a dict used as default, always none and the if none initialize the empty dict.

This is a common python gotcha, here is some more info in case you wanted to learn more about that: https://docs.python-guide.org/writing/gotchas/

Rishabh261998:
@OriolAbril made the changes, can you review? Thanks

OriolAbril (approved):

OriolAbril:
Thanks @Rishabh261998!

PR #1583, opened 2021-02-26
add PyStan3 reloo example [docs]

ahartikainen:
## Checklist

cc @OriolAbril


ahartikainen:
There is a httpstan bug that needs to be fixed before this PR can go forward

https://github.com/stan-dev/httpstan/issues/546

ahartikainen:
Also needs a bugfix #1584

arviz/wrappers/wrap_pystan.py
```diff
@@ -55,3 +55,85 @@ def get_inference_data(self, fit):
...
+    description. An example of ``PyStan3SamplingWrapper`` usage can be found
+    in the :ref:`pystan3_refitting` notebook.
+
+    Warnings
+    --------
+    Sampling wrappers are an experimental feature in a very early stage. Please use them
+    with caution.
+    """
+
+    def __init__(
```
OriolAbril, line 73: Not sure renaming model to model code is worth redefining init, in pymc3 for example the "model" is actually a function that returns the model object to then do the `with model:...`. Maybe we can change the name in the base sampler to something more generic?

arviz/wrappers/wrap_pystan.py
```diff
@@ -55,3 +55,85 @@ def get_inference_data(self, fit):
...
+            Variable name containing the pointwise log likelihood data of the excluded
+            data. As PyStan cannot call C++ functions and log_likelihood__i is already
+            calculated *during* the simultion, instead of the value on which to evaluate
+            the likelihood, ``log_likelihood__i`` expects a string so it can extract the
+            corresponding data from the InferenceData object.
+        """
+        raise NotImplementedError("sel_observations must be implemented on a model basis")
+
+    def sample(self, modified_observed_data):
+        """Resample the PyStan model stored in self.model on modified_observed_data."""
```
OriolAbril, line 125: This should be updated to something like "Rebuild and resample the ..." maybe also update the `self.model` if we keep the `model_code` name.

arviz/wrappers/wrap_pystan.py
```diff
@@ -55,3 +55,85 @@ def get_inference_data(self, fit):
     def log_likelihood__i(self, excluded_obs_log_like, idata__i):
         """Retrieve the log likelihood of the excluded observations from ``idata__i``."""
         return idata__i.log_likelihood[excluded_obs_log_like]
+
+
+class PyStan3SamplingWrapper(SamplingWrapper):
```
OriolAbril, line 60: I was just thinking that we could have an intermediate layer `StanSamplingWrapper` to avoid having to repeat the `sel_observations` and `log_likelihood__i` in pystan, pystan3 and cmdstanpy sampling wrappers, as the base logic of generating log likelihoods during sampling is the same for all 3. We can do that in a follow up PR or directly when we implement the cmdstanpy base wrapper.

OriolAbril (commented):
This is great, thanks. I added some comments to discuss a couple open questions, and here come a couple more:

Do you think all the "excluded", "__i", "data_for_fitting" is clear? Would it be better to use train/test?

Should `sel_observations` take 2 arguments instead of one, the first being the positional index and the 2nd a dict that can be used with `sel` to get the excluded/test data?

The only two nits that do require changes are two commented out prints in the notebook that should be removed and adding `PyStan3SamplingWrapper` to the api docs.

OriolAbril (commented):
Looks great, thanks! 

Only two comments, we need to update the `pystan_refitting_xr_lik` item in the sampling wrapper guides toctree and also update the api section on wrappers.

Issue #1581, opened 2021-02-24
Fix mypy errors


OriolAbril:
The latest mypy version now checks more files, see https://github.com/arviz-devs/arviz/pull/1528#issuecomment-783115088. The errors listed in the comment should be fixed. The couple comments after the linked one explain how to do so.

utkarsh-maheshwari:
I read the discussion.
So basically, what I understood is instead of passing `var_names` as a tuple in the functions, we need to convert to `sequence` data type.
Am I correct?

OriolAbril:
The problem is not in the examples but in how the type hints are set. A tuple is a perfectly valid way to pass variable names, so would a numpy array be for example. However, the type hint says the input must be a list, instead of having a type hint saying anything of type `Sequence` is valid.

utkarsh-maheshwari:
Okay. 
I am not very familiar with type-hints. `typing_copilot` is new.
Can I still take the issue? I will learn about it.

OriolAbril:
You can still take the issue, I am not very familiar with type hints either. As we were saying on the issue, one of the goals of working on these issues (instead of just ignoring them because they don't affect the actual library) is precisely for all of us to get a bit more familiar with type hints.

abhisht51:
Hello @OriolAbril, are there more changes to be done to this issue?

OriolAbril:
As per my comment on #1587: https://github.com/arviz-devs/arviz/pull/1586#issuecomment-787580759

> I am realizing that there are many other places where we have inputs type hinted as lists when they should be sequences. I will merge this to go ahead with typing copilot directly with latest mypy but keep the issue open.

So if you know about type hints, it would be great to go over the type hints that are currently in use and make sure they are indeed correct and not too restrictive like this `List` instead of `Sequence` example. I am relabeling the issue however because we are no longer targetting some specific issues and the issue has become quite open ended. Moreover, we are still learning about type hinting, so the guidance we can provide on this is quite limited.

I also want to note that this is a low priority issue

OriolAbril:
We removed the mypy check in CI, so there are no longer mypy errors, but if someone knows about type hints and can help it would be nice to still check the type hints we do have

PR #1528, opened 2021-01-29
Add mypy and typing_copilot checks to the CI pipeline.

obi1kenobi:
Add the tightest-possible passing `mypy.ini` file, autogenerated with `typing_copilot` by running `typing_copilot init` from inside an `arviz` development environment with `arviz` installed within.

Part of #1496. Next steps (in this PR or in separate, future PRs, per the maintainers' choice) would be:
- [x] Add a CI check to ensure `mypy` continues to pass (i.e. the "no type-hint regressions" check)
- [x] Add a CI check that `typing_copilot tighten --error-if-can-tighten` continues to pass (i.e. the "type hinting ratchet is as tight as possible" check)

Technically, `typing_copilot tighten --error-if-can-tighten` will run `mypy` with the current settings first and therefore will duplicate some of the same work from the first check, but I'd recommend running `mypy` directly anyway just to be sure and avoid the risk of `typing_copilot` bugs causing type-hinting quality to decrease unnoticed.

Issue #1578, opened 2021-02-21
Enable pylint use-a-generator check


OriolAbril:
I guess that latest pylint has added a `use-a-generator` check that did not use to exist. I disabled it in #1577 from `.pylintrc` but maybe we should enable the check and either fix the code or disable on a line per line basis. 

Same for `consider-using-generator`

Steps to fix the issue:

1. Remove these two checks from the list of disabled checks in pylintrc: https://github.com/arviz-devs/arviz/blob/main/.pylintrc#L75-L76
2. Run pylint on arviz (or what is the same, run [`scripts/lint.sh`](https://github.com/arviz-devs/arviz/blob/main/scripts/lint.sh))
3. Go to each of the files and lines where pylint shows a warning and try to fix the issue. There is also the possibility of false positives, I expect most occurences to be valid complains, but some may have to still be ignored. This ignores should be done on a line basis, not from `pylintrc`

MgeeeeK:
Hey @OriolAbril , I'd like to work on this issue... thanks, for the steps as well.

OriolAbril:
Great, just assigned you to the issue, as usual, let us know if you need any help

PR #1537, opened 2021-02-06
enable pyjags tests

ahartikainen:
## Checklist

- [ ] Follows [official](https://github.com/arviz-devs/arviz/blob/main/CONTRIBUTING.md#pull-request-checklist) PR format
- [ ] Includes a sample plot to visually illustrate the changes (only for plot-related functions)
- [ ] New features are properly documented (with an example if appropriate)?
- [ ] Includes new or updated tests to cover the new feature
- [ ] Code style  correct (follows pylint and black guidelines)
- [ ] Changes are listed in [changelog](https://github.com/arviz-devs/arviz/blob/main/CHANGELOG.md#v0xx-unreleased)

PR #1586, opened 2021-02-26
[WIP] fix mypy errors

utkarsh-maheshwari:
## Description
Fixed mypy errors mentioned in https://github.com/arviz-devs/arviz/pull/1528#issuecomment-783115088 by adding type hints to `var_names` and `latex_elements`. Addresses #1581

## Checklist

### DIFF
--- a/arviz/plots/traceplot.py
+++ b/arviz/plots/traceplot.py
@@ -1,6 +1,6 @@
 """Plot kde or histograms and values from MCMC samples."""
 import warnings
-from typing import Any, Callable, List, Mapping, Optional, Tuple, Union
+from typing import Any, Callable, List, Mapping, Optional, Tuple, Union, Sequence
 
 from ..data import CoordSpec, InferenceData, convert_to_dataset
 from ..labels import BaseLabeller
@@ -12,7 +12,7 @@ from .plot_utils import KwargSpec, get_plotting_function
 
 def plot_trace(
     data: InferenceData,
-    var_names: Optional[List[str]] = None,
+    var_names: Optional[Sequence[str]] = None,
     filter_vars: Optional[str] = None,
     transform: Optional[Callable] = None,
     coords: Optional[CoordSpec] = None,
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -20,6 +20,7 @@
 import os
 import re
 import sys
+from typing import Dict
 
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
 import arviz
@@ -181,7 +182,7 @@ html_favicon = "_static/favicon.ico"
 
 # -- Options for LaTeX output ---------------------------------------------
 
-latex_elements = {
+latex_elements : Dict[str, str] = {
     # The paper size ('letterpaper' or 'a4paper').
     #
     # 'papersize': 'letterpaper',


utkarsh-maheshwari:
Checks are passed but how should I check if the mypy errors have been resolved?

OriolAbril:
There should only be a `mypy.ini` file that covers the whole package. Both the `mypy.ini` file and pinning mypy `0.790` will be done by #1528 as a temporal fix. Temporal because they are not fixing the issue, they are only ignoring it and sliencing it for a while so we can keep ahead and start using typing_copilot as soon as possible. 

To fix the errors you'll have to edit the actual arviz library, nothing should be modified in the examples folder. The first errors listed in the other PR for example are complaining that `var_names` is a tuple instead of a list. This is a false positive, the mistake is in the type hints that the function (`plot_trace` in this case) has. This is the line that defines the type hint in question:

https://github.com/arviz-devs/arviz/blob/a9e5812dcf0b6c6993c839ab031d2afe82993a67/arviz/plots/traceplot.py#L15

`List` is not the correct type hint, a tuple or a numpy array are perfectly valid options. Thus, the type hint should use `Sequence` instead of list. 

To run mypy I believe you have to run `mypy -p arviz` from the base folder, but if I understand correctly, it will only work if you have the `mypy.ini` defined in #1528.

utkarsh-maheshwari:
Okay. So I have changed the type hint for `var_names`. First six have been resolved now.

The major doubt that I am having is before changing the type hint to `Sequence`, I run the command for 2 different versions of mypy
`typing_copilot init --overwrite`

I can see the 7 errors in case I pin mypy0.812 but not if I pin mypy0.790. So, according to  #1528 (comment) [https://github.com/arviz-devs/arviz/pull/1528#issuecomment-783115088] if we pin mypy<0.8, do we really need to change the type hint?

utkarsh-maheshwari:
Okay. So I have changed the type hint for `var_names`. First six have been resolved now.

The major doubt that I am having is before changing the type hint to `Sequence`, I run the command for 2 different versions of mypy
`typing_copilot init --overwrite`

I can see the 7 errors in case I pin mypy0.812 but not if I pin mypy0.790. So, according to https://github.com/arviz-devs/arviz/pull/1528#issuecomment-783115088 option 1, if we pin mypy<0.8, do we really need to change the type hint?

OriolAbril:
> So, according to #1528 (comment) option 1, if we pin mypy<0.800, do we really need to change the type hint?

Yes, because the ideal solution is doing option 2. Option 1 is a temporal fix that allows us to continue working on #1528 and not have to wait until we have time to fix all the issues (as we are all volunteers, it is not clear when we would have time and if it takes us too long maybe by the time we get to do it instead of 7 there are 30 errors). 

We are doing similar things with #1578 where we disabled some pylint checks to get CI to pass quickly but we did not _fix_ those issues, or with #1537 where we have temporarily disabled pyjags tests because pyjags requires numpy >=1.20 and tensorflow probability requires numpy <=1.19 because our tfp converter is outdated. Fixing the converter is complicated and depends also on changes on tfp side so we feel it's better to wait for now until we can update the code and fix the root issue, if we updated the code now, we may have to rewrite it again in a few months

utkarsh-maheshwari:
Okay I got it now. Very nicely explained. Thank you!

utkarsh-maheshwari:
Do these changes needed to be listed in changelog?

OriolAbril:
I don't think so, thanks! I am realizing that there are many other places where we have inputs type hinted as lists when they should be sequences. I will merge this to go ahead with typing copilot directly with latest mypy but keep the issue open.

PR #1590, opened 2021-03-01
use chain serial rather than logical index with _DefaultTrace.insert

Spaak:
This is part of a fix for an issue in PyMC3: https://github.com/pymc-devs/pymc3/issues/4469 . A trace could have chains that are not necessarily indexed 0, 1, .... However, `_DefaultTrace.insert` *does* take an index 0, 1, ... So, the index argument into `insert` should be zero-based. This PR decouples the `chain` index (which refers to a logical chain index, possibly starting at 1 or whatever) from the index passed to `_DefaultTrace.insert`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Integrate `index_origin` with all the library ([1201](https://github.com/arviz-devs/arviz/pull/1201))
 * Fix pareto k threshold typo in reloo function ([1580](https://github.com/arviz-devs/arviz/pull/1580))
 * Preserve shape from Stan code in `from_cmdstanpy` ([1579](https://github.com/arviz-devs/arviz/pull/1579))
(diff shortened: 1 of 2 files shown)


OriolAbril:
Looks good, thanks. It should probably be also added to https://github.com/pymc-devs/pymc3/pull/4489 better, which is where the converter will live in 4.0+

Spaak:
Thanks, I've pushed the change directly to the `to_inference_data` pymc3 branch as well.

OriolAbril (approved):

OriolAbril:
Can you add it to maintenance and fixes in the changelog? I think it is good to merge after that

Spaak:
Yep, done! (Not sure you'd get a notification after a commit push? Hence this comment :) )

OriolAbril:
Thanks!

PR #1591, opened 2021-03-02
Added Interactive Legend to forestplot Bokeh

### LINKED ISSUE #1572: Add interactive legend to bokeh forestplot
## Tell us about it
plot_density with bokeh backend has a super cool legend that allows to hide lines by clicking on the corresponding element in the legend. plot_forest could do the same.

## Thoughts on implementation
I guess searching for the PR that added the legend for plot_density will be a good starting point. Will try to extend and add more info at some point, but if someone has time to browse the history and track down the relevant pieces of code it would be great

Rishabh261998:
## Description

To fix #1572 

Added the Legend to forest plot in with backend Bokeh. I have also made it interactive but wasn't able to control the different axes with a single legend so had to provide a legend for each axes. See the below image

legend_bokeh

legend_bokeh_interactive

One more problem I am facing is that I am unable to hide the `y_ticks` corresponding to the model when the model plot is hidden. It seems that Bokeh supports interactivity with only glyphs.


Rishabh261998:
@OriolAbril Can you look into this. I was also facing problem in rebasing so I used the git merge command instead of rebasing.

arviz/plots/backends/bokeh/forestplot.py
```diff
@@ -155,11 +159,25 @@ def plot_forest(
...
     if ess:
-        plot_handler.plot_neff(axes[0, idx], markersize)
+        plotted_ess = defaultdict(list)
+        plot_handler.plot_neff(axes[0, idx], markersize, plotted_ess)
+        legend_it_ess = []
+        for (model_name, glyphs) in plotted_ess.items():
+            legend_it_ess.append((model_name, glyphs))
+        legend_ess = Legend(items=legend_it_ess)
+        axes[0, idx].add_layout(legend_ess, "right")
+        axes[0, idx].legend.click_policy = "hide"
```
OriolAbril, line 169: I think all this should only happen if `legend=True`. Same below

arviz/plots/backends/bokeh/forestplot.py
```diff
@@ -607,40 +669,42 @@ def iterator(self):
                 else:
                     row_label = label
                 for values in value_list:
-                    yield y, row_label, label, selection_list[idx], values, self.model_color[
-                        model_name
-                    ]
+                    yield y, row_label, model_name, label, selection_list[
```
OriolAbril, line 672: This is for the code above, not for here, but you need to update the label generation to use the labeller: https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/matplotlib/forestplot.py#L555

OriolAbril (commented):
I had very little time to review, will continue afterwards. 

> Added the Legend to forest plot in with backend Bokeh. I have also made it interactive but wasn't able to control the different axes with a single legend so had to provide a legend for each axes. See the below image

This is amazing, don't worry about the one legend per axes issue. This is perfectly fine. 

What is strange is that the ess/rhat dots should be aligned with the forestplot hdi lines, otherwise they are not very useful. Does this happen always or only with the legend?

> One more problem I am facing is that I am unable to hide the y_ticks corresponding to the model when the model plot is hidden. It seems that Bokeh supports interactivity with only glyphs.

Again, this is no problem at all, I recommend users use the NoModel labeller (or mixtures of it) which should be the default already with `legend=True`. With everything working well, there is only a label per variable, see the last example in the label guide for instance: https://arviz-devs.github.io/arviz/user_guide/label_guide.html#labeller-mixtures. What should not be happening are the 0 and 12 as ticklabels, I will try to take a look afterwards, but wanted to point it out already.

> I was also facing problem in rebasing so I used the git merge command instead of rebasing.

This is to be expected. You are using the same branch you used for the rope PR, which has the changes in the rope PR, but we merged it to master already so you have the same changes in _different_ commits, which makes rebasing tricky. 

We recommend keeping the main branch up to date with the main branch in arviz, and never work on it. And every time you need to make a PR, create a _new_ branch for that feature. After merging, we delete the branch and prune the repo from time to time.

arviz/plots/backends/bokeh/forestplot.py
Rishabh261998: By this, you just mean that I need to generate the row labels as done In matplotib?

Rishabh261998:
Also the ess/rhat dots are not aligned in the current bokeh forestplot without legend, whereas the matplotlib version is perfect. Also the 0 and 12 problem also appears in the bokeh forestplot without legend.
Screenshot from 2021-03-02 19-36-25

Rishabh261998:
And also should I create a new branch or is it fine to work on this branch just for this PR? i.e is there any problem if I use git merge again if I need to sync. Thanks

arviz/plots/backends/bokeh/forestplot.py
OriolAbril: Both row labels and model labels used in legend should be processed by the labeller, you'll see that in matplotlib there are two labeller methods called

OriolAbril:
> @OriolAbril Also the ess/rhat dots are not aligned in the current bokeh forestplot without legend, whereas the matplotlib version is perfect. Also the 0 and 12 problem also appears in the bokeh forestplot without legend.

Do you want to work on these here or open an issue for you or someone else to work on there at some point later in time?

> And also should I create a new branch or is it fine to work on this branch just for this PR? i.e is there any problem if I use git merge again if I need to sync. Thanks

For now it is fine I think, if you ran into problems or the number of commits increased I's probably suggest a rebase, but let's go to than if we get there.

Rishabh261998:
I think for now we should create a new issue and when I have finished with the legend thing in `foresplot` and also the `ppc _plot` then I can take this or someone else who is interested to work can also take it up. Thanks

Rishabh261998:
@OriolAbril in the below plot why are there only orange `ess` dots, plot is from matplotlib backend
Screenshot from 2021-03-04 01-27-58

OriolAbril:
> @OriolAbril in the below plot why are there only orange ess dots, plot is from matplotlib backend

No idea. I am not sure ess and rhat are very useful for multiple models case, but it should not happen, and if it does then we should document that ess/rhat do not work with multiple models.

Rishabh261998:
@OriolAbril fixed the label generation and also fixed the `0, 12` error which was happening before:
Screenshot from 2021-03-04 15-44-50
This is the current plot with bokeh as backend and using `NoModel` labeller. As you can see after updating the label generation in bokeh as it was in `matplotlib` the `ess` dots are only shown for a single model. For this I will create a separate PR and fix this problem for both `mpl` and `bokeh`. Also see the below 2 plots 1 for `mpl` and 1 for `bokeh` with labeller as `DimCoordLabeller`. In this the `ess` dots are plotted for both the model. 
For `Matplolib`:
Screenshot from 2021-03-04 13-41-56
For `Bokeh`:
Screenshot from 2021-03-04 15-45-28

OriolAbril:
> Also see the below 2 plots 1 for mpl and 1 for bokeh with labeller as DimCoordLabeller. In this the ess dots are plotted for both the model.

Oh, I see, good catch. The issue is that the code must be checking a dict or set to see if a dot has been plotted, and with the new labeller capabilities, labels don't need to be unique. I should have updated the code to stop using labels as ids but completely missed that. A new PR to fix this would be great.

arviz/plots/backends/bokeh/forestplot.py
```diff
@@ -191,8 +202,10 @@ def plot_forest(
     ] = -all_plotters[  # pylint: disable=protected-access
         0
     ].group_offset
-    axes[0, 0].y_range._property_values["end"] = y_max  # pylint: disable=protected-access
+    axes[0, 0].y_range._property_values["end"] = y_max  # pylint: disable=protected-
```
OriolAbril, line 205: undo this change to have pyling keep ignoring this warning

OriolAbril (approved):
Thanks, this is great

arviz/plots/backends/bokeh/forestplot.py
Rishabh261998: Did this, still checks are failing:-C

PR #1592, opened 2021-03-02
Using color maps in plot_khat

agustinaarroyuelo:
## Description
`plot_khat` can take a `c` argument to use with a color map

cmaps_khat

## Checklist

### DIFF
--- a/arviz/plots/backends/matplotlib/khatplot.py
+++ b/arviz/plots/backends/matplotlib/khatplot.py
@@ -76,31 +76,35 @@ def plot_khat(
     kwargs = matplotlib_kwarg_dealiaser(kwargs, "scatter")
     kwargs.setdefault("s", markersize)
     kwargs.setdefault("marker", "+")
-    color_mapping = None
-    cmap = None
-    if isinstance(color, str):
-        if color in dims:
-            colors, color_mapping = color_from_dim(khats, color)
-            cmap_name = kwargs.get("cmap", plt.rcParams["image.cmap"])
-            cmap = getattr(cm, cmap_name)
-            rgba_c = cmap(colors)
+
+    c_kwarg = kwargs.get("c", None)
+
+    if c_kwarg is None:
+        color_mapping = None
+        cmap = None
+        if isinstance(color, str):
+            if color in dims:
+                colors, color_mapping = color_from_dim(khats, color)
+                cmap_name = kwargs.get("cmap", plt.rcParams["image.cmap"])
+                cmap = getattr(cm, cmap_name)
+                rgba_c = cmap(colors)
+            else:
+                legend = False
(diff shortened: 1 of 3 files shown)


OriolAbril:
Looking now at the code and the comments in slack, does `az.plot_khat(loo, color=values)` work? I think it should work already with the current rgba_c generation and use a colormap. We may have to handle the presence of the `c` kwarg better though.

agustinaarroyuelo:
You are right `az.plot_khat(loo, color=values)` works. Is this PR enough for handling the `c` kwarg?

OriolAbril:
I would have the `color` argument of `plot_khat` take preference over the `c` passed as kwarg, I think we can do that by adding these lines after the dealiaser, around line 78:

```
c_kwarg = kwargs.pop("c", None)
if c_kwarg is not None:
    warnings.warn("Found kwarg named `c` which will be ignored, use `color` instead")
```
But I would also understand wanting to pass `c` to avoid all the preprocessing done to `color`, ,we could also use the check to instead of printing a warning, use `c` directly and avoid all the `color` preprocessing crazyness.

OriolAbril (approved):

OriolAbril:
Oh, the preprocessing is also used to create a legend based on the colors shown if desired in case the variable that gets color encoded is discrete instead of continuous.

arviz/plots/backends/matplotlib/khatplot.py
```diff
@@ -93,6 +93,7 @@ def plot_khat(
                 rgba_c = to_rgba_array(np.full(n_data_points, color))
         else:
             legend = False
+            color = (color - color.min()) / (color.max() - color.min())
```
OriolAbril, line 96: This has to come after the except. At this point `color` can be a list of strings or a nx3 array of rgb colors. I would also suggest using matplotlib normalizer function in case someone wanted to use a different one

```
cmap_name = kwargs.get("cmap", plt.rcParams["image.cmap"])
cmap = getattr(cm, cmap_name)
norm_fun = kwargs.get("norm", mpl.colors.Normalize(color.min(), color.max()))
rgba_c = cmap(norm_fun(color))
```

arviz/plots/backends/matplotlib/khatplot.py
```diff
@@ -105,6 +106,14 @@ def plot_khat(
...
         rgba_c = vectorized_to_hex(rgba_c)
         kwargs["c"] = rgba_c
+    else:
+        if isinstance(c_kwarg, str):
+            if c_kwarg in dims:
+                colors, color_mapping = color_from_dim(khats, c_kwarg)
+            else:
+                legend = False
+        else:
+            legend = False
```
OriolAbril: I would not add that. I think `c` should always behave _exactly_ like the argument in `matplotlb.scatter` if they want any of the other features, users should use `color`

PR #1593, opened 2021-03-02
Fix ax argumentin plot_elpd

agustinaarroyuelo:
## Description

`plot_elpd` can now take the `ax` argument. Naive example:

```
d1 = az.load_arviz_data("centered_eight")
d2 = az.load_arviz_data("non_centered_eight")

fig, ax = plt.subplots(1, 2, figsize=(10, 4))

az.plot_elpd({"Centered eight": d1, "Non centered eight": d2}, xlabels=True, ax=ax[0])
az.plot_elpd({"Centered eight": d1, "Non centered eight": d2}, xlabels=True, ax=ax[1])

```

elpd

## Checklist

### DIFF
--- a/arviz/plots/backends/matplotlib/elpdplot.py
+++ b/arviz/plots/backends/matplotlib/elpdplot.py
@@ -140,6 +140,8 @@ def plot_elpd(
                 sharey="row",
                 sharex="all",
             )
+        else:
+            fig = ax.get_figure()
 
         for i in range(0, numvars - 1):
             var1 = pointwise_data[i]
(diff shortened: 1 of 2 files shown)


OriolAbril (commented):
Good catch, thanks! the same thing is also needed in line 131 in the case of >2 models being compared, where the ax is an array like in pairplot

arviz/plots/backends/matplotlib/elpdplot.py
```diff
@@ -138,6 +140,8 @@ def plot_elpd(
                 sharey="row",
                 sharex="all",
             )
+        else:
+            fig = ax.get_figure()
```
OriolAbril, line 144: I think it should be `ax.ravel()[0]` or something like this, because in these cases `ax` is an array of axes, they should all have the same figure as reference but the array does not know about the figure.

PR #1598, opened 2021-03-04
Fix cmdstanpy io

### LINKED ISSUE #1594: from_cmdstanpy uses posterior in place of prior when both provided
**Describe the bug**
I'm seeing some odd behavior on a branch of arviz when using `from_cmdstanpy`. Basically, if I fit a model twice using cmdstanpy & construct the az.InferenceData object using both the prior & the posterior samples, I see that the posterior draws are in both the prior & posterior groups. 

In other words, say I construct my arviz InferenceData object as follows:

```py
idata = az.from_cmdstanpy(posterior_fit, 
    prior = prior_fit, 
    coords = dict(school=["A", "B", "C", "D", "E", "F", "G", "H"]), 
    dims = dict(
        theta=["school"],
        y=["school"],
        log_lik=["school"],
        y_hat=["school"],
        theta_tilde=["school"],
    ),
    posterior_predictive = ["y_hat"],
    prior_predictive = ['y_hat'],
    log_likelihood = ["log_lik"])
```

The draws for the prior & posterior are identical.

```py
all(idata.prior == idata.posterior)
> True
```

Notably, this is not the case for the predicted quantities:

```py
all(idata.prior_predictive == idata.posterior_predictive)
> False
```

**To Reproduce**
I have set up a reproducible example here: https://gist.github.com/jburos/b8839b7919455d6a0019d7ce33394c2f

**Expected behavior**
The prior draws should be saved in the prior group.

**Additional context**

I had installed `arviz` from the `cmdstanpy` branch, due to previous bugs. I was on commit d95e9763b19611fcdeefa7b9aecb2746d5534fc9, according to my Pipfile.lock.

My system:

```py
In [3]: sys.version
Out[3]: '3.8.7 (default, Feb  3 2021, 06:31:03) \n[Clang 12.0.0 (clang-1200.0.32.29)]'
In [5]: arviz.__version__
Out[5]: '0.11.2'

### PR DESCRIPTION
**Pull request by ahartikainen**
Fix missing posterior calls and fix default log_likelihood for older versions. closes #1594

### DIFF
--- a/arviz/data/io_cmdstanpy.py
+++ b/arviz/data/io_cmdstanpy.py
@@ -491,8 +491,7 @@ class CmdStanPyConverter:
             ),
         )
 
-    @requires("posterior")
-    def sample_stats_to_xarray_pre_v_0_9_68(self, fit):
+    def sample_stats_to_xarray_pre_v_0_9_68(self, fit)
         """Extract sample_stats from fit."""
         dtypes = {"divergent__": bool, "n_leapfrog__": np.int64, "treedepth__": np.int64}
         columns = fit.column_names
--- a/arviz/data/io_cmdstanpy.py
+++ b/arviz/data/io_cmdstanpy.py
@@ -54,8 +54,10 @@ class CmdStanPyConverter:
             if self.log_likelihood is None and "log_lik" in self.posterior.stan_vars_cols:
                 self.log_likelihood = ["log_lik"]
         else:
-            if self.log_likelihood is None and any(
-                name.split("[")[0] == "log_lik" for name in self.posterior.column_names
+            if (
+                self.log_likelihood is None
+                and self.posterior is not None
+                and any(name.split("[")[0] == "log_lik" for name in self.posterior.column_names)
             ):
                 self.log_likelihood = ["log_lik"]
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by ahartikainen** [comment]
If tests pass and everything looks good, we should merge this and start to think when we can create patch release (`from_cmdstanpy` is broken currently for older cmdstanpy folders).

cc @OriolAbril

File: arviz/data/io_cmdstanpy.py [hunk 1]
```diff
@@ -483,8 +491,7 @@ def posterior_to_xarray_pre_v_0_9_68(self):
             ),
         )
 
-    @requires("posterior")
-    def sample_stats_to_xarray_pre_v_0_9_68(self, fit):
+    def sample_stats_to_xarray_pre_v_0_9_68(self, fit)
```
OriolAbril, line 494:
```suggestion
    def sample_stats_to_xarray_pre_v_0_9_68(self, fit):
```

OriolAbril:
> and start to think when we can create patch release 

We can do a minor release probably better, I'd like to have tests for the labels before releasing though, also delete the `NoRepeatLabeller` and add a `NoVarNameLabeller` instead

ahartikainen:
This PR is ready. Tests are passed (pymc3 test is failing)

PR #1601, opened 2021-03-05
Fix xlabels and ax argument in plot_elpd

agustinaarroyuelo:
## Description
Fix xlabels and ax argument in plot_elpd for numvars > 2. 

elpd+2numvars

## Checklist

### DIFF
--- a/arviz/plots/backends/matplotlib/elpdplot.py
+++ b/arviz/plots/backends/matplotlib/elpdplot.py
@@ -141,7 +141,7 @@ def plot_elpd(
                 sharex="all",
             )
         else:
-            fig = ax.get_figure()
+            fig = ax.ravel()[0].get_figure()
 
         for i in range(0, numvars - 1):
             var1 = pointwise_data[i]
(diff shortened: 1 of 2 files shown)


OriolAbril (approved):

arviz/plots/backends/matplotlib/elpdplot.py
```diff
@@ -179,7 +179,8 @@ def plot_elpd(
                     "{} - {}".format(models[i], models[j + 1]), fontsize=titlesize, wrap=True
                 )
         if xlabels:
-            set_xticklabels(ax[-1, -1], coord_labels)
+            for i in range(len(ax)):
+                set_xticklabels(ax[-1, :][i], coord_labels)
```
OriolAbril, line 183: I am not completely sure about what is causing the test error. Matplotlib locators do take steps as an argument, and there has been no breaking change on that. The only thing that would make sense would be for the loop to be over the last column instead of being over the bottom row so that `set_xticklabels` is trying to format an axis that has been set to `off` and thus has no locator. But then you would not have been able to generate the example image you shared, so I am completely at a loss here, no idea what is happening.

OriolAbril: Maybe we don't need to loop over the last row, the x axis is shared and should have the same labels for all plots automatically I think :thinking:

OriolAbril (approved):

PR #1608, opened 2021-03-09
Update setup.py to require python >= 3.7

ahnitz:
Arxviz due to its requirement on xarray>=0.16.1 (and possibly for internal reasons as well) in practice has a hard requirement on the python version of >=3.6. However, it is not declared at the moment in the setup.py so the package resolution doesn't notice this through pypi until attempting to satisfy the xarray requirement.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,6 +58,7 @@ setup(
     long_description=get_long_description(),
     long_description_content_type="text/markdown",
     include_package_data=True,
+    python_requires='>=3.6',
     classifiers=[
         "Development Status :: 4 - Beta",
         "Framework :: Matplotlib",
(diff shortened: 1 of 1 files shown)


ahnitz:
@OriolAbril I believe python3.6 should in fact be OK (unless other PRs have already implicitly removed 36 support)

OriolAbril:
Oh, I missed the equal, we should drop support for python 3.6 following [nep 29](https://numpy.org/neps/nep-0029-deprecation_policy.html), most of the packages in the pydata ecosystem have already done so: https://github.com/numpy/numpy/blob/main/setup.py#L407, https://github.com/scipy/scipy/blob/master/setup.py#L525, https://github.com/pandas-dev/pandas/blob/master/setup.cfg#L37, https://github.com/pydata/xarray/blob/master/setup.cfg#L75, https://github.com/pymc-devs/pymc3/blob/master/setup.py#L85... 

Can you update the PR?

PR #1611, opened 2021-03-11
add some extra rcparams

OriolAbril:
## Description
Adds two new rcParams, data.log_likelihood and stats.ic_compare_method. Also fixes a couple of
mishaps in index_origin usage.

## Checklist

      PR format?
      section of the changelog?


ahartikainen:
Logic is now wrong.

I have a model with a variable `log_lik` and also the correct log likelihood as `log_lik2`.

from_xyz(log_likelihood="log_lik2") --> log_lik goes to log_likelihood.

OriolAbril:
True, it only has to be triggered if log_likelihood is exactly true, not if it evaluates to true, I always mess this up

ahartikainen (approved):

ahartikainen (approved):

ahartikainen:
I wonder if there was something in the main that could fix that weird error.

ahartikainen:
Yeah, when there is true (?) and no log_lik, it will of course fail, so we need to put it back to None.

OriolAbril:
I was thinking that it may be easier to set `self.log_likelihood = None` and then start evaluating conditions on rcparams and `log_likelihood` argument to see if it needs to be updated and how. At least I don't see the current logic very clear which ends up in messing up. I'll give it a go later this evening

OriolAbril:
nevermind, it's basically the same, it would be something like:

```python
self.log_likelihood = None
log_likelihood = rcParams if ...  # instead of doing that on self.log_likelihood
# [...]
# keep the same conditions but using log_likelihood instead of self.log_likelihood
# skip the if isinstance(..., bool)
```

which is not any improvement. Let's merge like this

Issue #1291, opened 2020-07-15
Add linear regression code to examples repo


OriolAbril:
Add the code used to generate `regression1d` and `regression10d` examples available with `load_arviz_data` to https://github.com/arviz-devs/arviz_example_data

aloctavodia:
Closing as those examples have been superseded

PR #1615, opened 2021-03-15
Added example to plot_hdi using Inference data

### LINKED ISSUE #1479: Add example to az.plot_hdi that shows how to plot hdi from InferenceData posterior or posterior predictive
## Tell us about it
The `az.plot_hdi` docs only show how to plot the hdi from synthetic data. What would be nice is if it also shows users how to plot the hdi from an `az.InferenceData` as thats the standard data representation most people will be working with.

image

## Thoughts on implementation
Use one of the precomputed datasets that has a posterior predictive group and create a plot.

Here's an example of loading a dataset with posterior predictive group
https://arviz-devs.github.io/arviz/api/generated/arviz.plot_ppc.html

Here's an (unrelated) example where PyMC3 generates a trace and posterior predictive group and its used with `az.plot_hdi`

```
with pm.Model() as model_linear:
    β = pm.Normal('β', sd=10, shape=2)

    μ = pm.Deterministic('μ', pm.math.dot(babies[["Intercept", "Month"]], β))
    ϵ = pm.HalfNormal('ϵ', sd=10)

    length = pm.Normal('length', mu=μ, sd=ϵ, observed=babies.Length)

    linear_trace = pm.sample(2000, tune=4000)
    linear_ppc = pm.sample_posterior_predictive(inf_data_linear)
    inf_data_linear = az.from_pymc3(trace=linear_trace, posterior_predictive= linear_ppc)

fig, ax = plt.subplots()

ax.set_ylabel("Length")
ax.set_xlabel("Month");

μ_m = inf_data_linear.posterior["μ"].values.reshape(-1, babies.Length.shape[0]).mean(axis=0)

ax.plot(babies.Month, μ_m, c='C4')
az.plot_hdi(babies.Month, inf_data_linear.posterior_predictive["length"], hdi_prob=.94, ax=ax)

ax.plot(babies.Month, babies.Length, 'C0.', alpha=0.1)

plt.savefig('img/Baby_Length_Linear_Fit.png', dpi=300)

```
image
image

Rishabh261998:
## Description

To fix #1479 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 * Added "Label guide" page and API section for `arviz.labels` module ([1201](https://github.com/arviz-devs/arviz/pull/1201) and [1635](https://github.com/arviz-devs/arviz/pull/1635))
 * Add "Installation guide" page to the documentation ([1551](https://github.com/arviz-devs/arviz/pull/1551))
 * Improve documentation on experimental `SamplingWrapper` classes ([1582](https://github.com/arviz-devs/arviz/pull/1582))
+* Added example to `plot_hdi` using Inference Data ([1615](https://github.com/arviz-devs/arviz/pull/1615))
 
 ## v0.11.2 (2021 Feb 21)
 ### New features


arviz/plots/hdiplot.py
```diff
@@ -109,6 +109,17 @@ def plot_hdi(
...
+    ``plot_hdi`` can also be used with Inference Data objects. Here we use the posterior predictive
+    to plot the HDI interval.
+
+    .. plot::
+        :context: close-figs
+
+        >>> data = az.load_arviz_data('radon')
+        >>> x_data = np.array(data.posterior_predictive.obs_id)
+        >>> y_data = np.array(data.posterior_predictive.y)
+        >>> az.plot_hdi(x_data, y_data)
```
OriolAbril, line 121: not sure if you have looked at how the plot looks like, but `obs_id` are integer indexes that don't really have any meaning, they should not be used as x values. They are also not continuous, so plotting a continuous hdi band over all observations is not a good idea either. 

I think this should be done with the regression1d example better, but it looks like the InferenceData does not have the x values used as predictiors. Going down this path would require solving https://github.com/arviz-devs/arviz/issues/1291 and extending the code to get the x data into the inferencedata, which will probably take some time.

As a more immediate solution, I would use simulated data (maybe even the same as above) to create a new inferencedata instead of loading a preexisting one and then use this one to plot the hdi

Rishabh261998: Yeah, that makes sense, I too was confused about this bit. I was also thinking of creating a new Inference Object and using that to plot the HDI but as the issue suggested to use pre-computed data so went on with that approach to see if it was possible. So as of now, it is not possible as there are no values in the posterior predictive or the posterior. Will make the necessary changes. Thanks

OriolAbril: Minor note, the x values would be in the `constant_data` group

Rishabh261998: Done!, could you check @OriolAbril? Thanks

arviz/plots/hdiplot.py
```diff
@@ -109,6 +109,21 @@ def plot_hdi(
...
+    to plot the HDI interval.
+
+    .. plot::
+        :context: close-figs
+
+        >>> X = np.random.normal(0,1,100)
+        >>> Y = np.random.normal(2 + X * 0.5, 0.5, (10,100))
+        >>> X = {"x":X}
+        >>> Y = {"y": Y}
+        >>> dataset = az.from_dict(posterior=Y, constant_data = X)
```
OriolAbril, line 122:
```suggestion
        >>> idata = az.from_dict(posterior_predictive={"y" : Y}, constant_data = {"x" : X})
```

I'd recommend using this names instead. `y` is generally used for observed data, and its values will generally be stored in the posterior predictive group instead of the posterior, and I would avoid `dataset` as a name because the object it represents is an `InferenceData` as opposed to an `xr.Dataset` which could be confusing, due to this double meaning of dataset.

OriolAbril (approved):

Issue #1614, opened 2021-03-15
Error comparing PyMC3 models with multiple observed variables


rpgoldman:
**Describe the bug**
I have a set of PyMC3 models with multiple observed variables that I would like to compare with ArviZ.  When I try to invoke the comparison as follows:

```
arviz.compare({"simple_model": simple_model_idata, "complex_model": complex_model_idata})
```

I get this error:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-14-ae768c0c0627> in <module>
----> 1 arviz.compare({"simple_model": simple_model_idata, "complex_model": complex_model_idata})

~/tacc-work/jupyter_packages/envs/xplan-experiment-analysis/lib/python3.9/site-packages/arviz/stats/stats.py in compare(dataset_dict, ic, method, b_samples, alpha, seed, scale)
    210     for name, dataset in dataset_dict.items():
    211         names.append(name)
--> 212         ics = ics.append([ic_func(dataset, pointwise=True, scale=scale)])
    213     ics.index = names
    214     ics.sort_values(by=ic, inplace=True, ascending=ascending)

~/tacc-work/jupyter_packages/envs/xplan-experiment-analysis/lib/python3.9/site-packages/arviz/stats/stats.py in loo(data, pointwise, var_name, reff, scale)
    616     """
    617     inference_data = convert_to_inference_data(data)
--> 618     log_likelihood = _get_log_likelihood(inference_data, var_name=var_name)
    619     pointwise = rcParams["stats.ic_pointwise"] if pointwise is None else pointwise
    620 

~/tacc-work/jupyter_packages/envs/xplan-experiment-analysis/lib/python3.9/site-packages/arviz/stats/stats_utils.py in get_log_likelihood(idata, var_name)
    424         var_names = list(idata.log_likelihood.data_vars)
    425         if len(var_names) > 1:
--> 426             raise TypeError(
    427                 f"Found several log likelihood arrays {var_names}, var_name cannot be None"
    428             )

TypeError: Found several log likelihood arrays ['AND obs', 'NAND obs', 'NOR obs', 'OR obs', 'XNOR obs', 'XOR obs'], var_name cannot be None
```

This is unfortunate for at least two reasons:

1. I'd like to be able to compare these models.
[...]

PR #1616, opened 2021-03-15
Modify compare() docstring, error-check, pass var_name.

rpgoldman:
## Description

Check to see if the log_likelihood groups have more than one data
variable, and if so, require the var_name parameter. This avoids
having a less understandable error message crop up later.

Introduce `var_name` parameter to `compare()`, and pass it through to subsidiary IC function. @OriolAbril reports it was simply an oversight that this wasn't done before.

Addresses issue #1614 

## Checklist


arviz/stats/stats.py
```diff
@@ -206,7 +214,16 @@ def compare(
...
     for name, dataset in dataset_dict.items():
         names.append(name)
-        ics = ics.append([ic_func(dataset, pointwise=True, scale=scale)])
+        if len(dataset.log_likelihood.data_vars) > 1 and var_name is None:
+            raise ValueError(
+                (
+                    f"Dataset {name} has multiple variables in its log_likelihood.\n"
+                    "In such cases, the var_name parameter is mandatory."
+                )
+            )
```
OriolAbril, line 223: This is basically the same error in https://github.com/arviz-devs/arviz/blob/main/arviz/stats/stats_utils.py#L426, I think it would be better to modify the error in `get_log_likelihood` so compare, loo and waic all show the same error

arviz/stats/stats.py
```diff
@@ -45,7 +43,14 @@
 
 
 def compare(
-    dataset_dict, ic=None, method="stacking", b_samples=1000, alpha=1, seed=None, scale=None
+    dataset_dict: Optional[Dict[str, InferenceData]],
```
OriolAbril, line 46: I don't think this is optional, and I think we should use `Mapping` instead of dict for inputs, similarly to List vs Sequence

rpgoldman: That looks right to me. I will change it. I can't recall why I thought `Optional` was right.

arviz/stats/stats.py
rpgoldman: There is an error check there, but it's not helpful because it has lost the identity of the `InferenceData` object. Having the error here means that we can tell the user *which* of the entries in the `dataset_dict` has multiple observed variables (with `name`, as a I do here).  The individual calls to `get_log_likelihood` don't have access to this information.  If we just use the subsidiary errors, then the poor user has to restart and run in the debugger in order to figure out which of the set of `InferenceData` objects has the problem.  This way they can identify the location of the error immediately.

Possibly we could do what I want by catching and re-signaling *all* errors in the `ic_func`, and raise new ones that add `Error in computing the {ic} for {name}` to the original error string.  Would you prefer that?

Also, slightly OT, I note that the subsidiary functions raise a `TypeError` for this situation, where I had thought of it as a `ValueError`.

OriolAbril: I have a very bad grasp on the meaning of errors, don't hesitate to change it and use the right one.

I was going to write that I'd rather do "catch->re-raise" than defining a completely new error here. But while thinking about this I realized that passing inferencedatas to compare does not make much sense and limits a lot the possibilities. We should be passing elpddata to compare. For now it's mostly a matter of efficiency, if you want to use compare and plot_elpd for example, you have to compute loo/waic twice, there is no way to reuse the result. However, going forward, the current approach is actually making it impossible to use `compare` with reloo or with IWMM. 

I think we should deprecate passing inferencedata and take only elpddata. The first step would be taking both (like plot_elpd does) and print a warning if inferencedata is passed, then eventually remove it.

This does not need to happen here though, we can open an issue about this and homogenize compare, plot_elpd, plot_khat..., for now catching+re-raising is enough. What do you prefer?

rpgoldman:
Looking at the test failures, it seems that sometimes ArviZ *computes* the `log_likelihood` group for input `InferenceData` objects, instead of just extracting it. That was not what I expected. It suggests that this should be redone as I suggested above: catch the errors from the `ic_func` and annotate them, instead of doing the error-check in `compare()`

OriolAbril:
> Looking at the test failures, it seems that sometimes ArviZ computes the log_likelihood group for input InferenceData objects, instead of just extracting it. That was not what I expected. It suggests that this should be redone as I suggested above: catch the errors from the ic_func and annotate them, instead of doing the error-check in compare()

Maybe the test helpers have not been updated and still have `log_likelihood` as a variable in `sample_stats`. The `get_log_likelihood` prints a warning if this happens but it still works for backward compatibility reasons.

rpgoldman:
I tried the error-catching and re-raising approach and it worked locally for me, so I am pushing it for inspection and testing.

I still need to add some new tests to verify that the errors are signaled as I expect.

arviz/stats/stats.py
```diff
@@ -142,6 +150,11 @@ def compare(
     loo : Compute the Pareto Smoothed importance sampling Leave One Out cross-validation.
     waic : Compute the widely applicable information criterion.
 
+    Notes
+    -----
+    If the `log_likelihood` group is not present in the input datasets, ArviZ will attempt
+    to compute it.
```
OriolAbril, line 156: I don't think this is right nor possible even. If compare is working without the log likelihood group present it *must* be because log likelihood data is stored as `log_likelihood` variable in `sample_stats` group. This was the original approach and still works for backward compatibility with old netcdf files.

OriolAbril (changes requested):

rpgoldman:
@OriolAbril I think this now does things the way you wanted: I grab up the errors and signal a new error `from` the old one, with additional information identifying which model caused the problem. I'll look at your cross-reference now.

OriolAbril (approved):

arviz/stats/stats.py
rpgoldman: I would also prefer we push the elpddata issue to a different PR. This one has dragged on long enough!

I need to figure out what multi-observed inference data to use for tests, and then I think this can be finished.

arviz/data/io_pymc3.py
```diff
@@ -89,7 +89,8 @@ def __init__(
         # this permits us to get the model from command-line argument or from with model:
         try:
             self.model = self.pymc3.modelcontext(model or self.model)
-        except TypeError:
+        except TypeError as e:
+            _log.error("Got error %s trying to find log_likelihood in translation.", e)
```
OriolAbril, line 93: Kind of a side note, we may want to port that to https://github.com/pymc-devs/pymc3/blob/v4/pymc3/backends/arviz.py, but I would wait for a while before that because log likelihood computation will probably be rewritten completely.

arviz/rcparams.py
```diff
@@ -9,11 +9,22 @@
 from collections.abc import MutableMapping
 from pathlib import Path
 from typing import Any, Dict
+from typing_extensions import Literal
+
+NO_GET_ARGS: bool = False
+try:
+    from typing_extensions import get_args
+except ImportError:
+    NO_GET_ARGS = True
```
OriolAbril, line 18: It looks like we can simplify all this if we require py>=3.7, which we should already do by next release as we agreed to follow numpy guidance on this. Note that numpy, scipy, pandas, matplotlib... latest versions all require >=3.7.

As how to proceed, it is probably better to merge this to avoid conflicts and having this hanging but after open an issue to fix this once we drop 3.6 (there are already a couple PRs on that).

references:
https://github.com/python/typing/blob/master/typing_extensions/src_py3/typing_extensions.py#L17
https://github.com/python/typing/blob/master/typing_extensions/src_py3/typing_extensions.py#L157

.pylintrc
```diff
@@ -265,7 +266,7 @@ good-names=b,
            ok,
            sd,
            tr,
-	         eta,
+                 eta,
```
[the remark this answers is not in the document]
rpgoldman: Fixed.

arviz/rcparams.py
rpgoldman: Added #1638 for this.

arviz/data/io_pymc3.py
rpgoldman: I'm going to mark this as resolved, since it's for future reference.

arviz/stats/stats.py
```diff
@@ -1280,7 +1311,9 @@ def summary(
     n_vars = np.sum([joined[var].size // n_metrics for var in joined.data_vars])
 
     if fmt.lower() == "wide":
-        summary_df = pd.DataFrame(np.full((n_vars, n_metrics), np.nan), columns=metric_names)
+        summary_df = pd.DataFrame(
+            (np.full(cast(Tuple[int, int], (n_vars, n_metrics)), np.nan)), columns=metric_names
```
OriolAbril, line 1315: why do we have to cast this to a tuple of ints? shouldn't it be one by construction? `n_metrics` is a length and `n_vars` is a sum over sizes integer divided by `n_metrics` :confused:

arviz/tests/base_tests/test_stats.py
```diff
@@ -40,6 +40,23 @@ def non_centered_eight():
     return non_centered_eight
 
 
+@pytest.fixture
```
OriolAbril, line 43: I think we should make it have scope "module", there is no need to rerun the fixture function for every parameter in the test right?

arviz/tests/base_tests/test_stats.py
```diff
@@ -192,6 +209,20 @@ def test_compare_different_size(centered_eight, non_centered_eight):
...
+@pytest.mark.parametrize("ic", ["loo", "waic"])
+def test_compare_multiple_obs(multivariable_log_likelihood, centered_eight, non_centered_eight, ic):
+    compare_dict = {
+        "centered_eight": centered_eight,
+        "non_centered_eight": non_centered_eight,
+        "problematic": multivariable_log_likelihood,
+    }
+    with pytest.raises(TypeError, match="several log likelihood arrays"):
+        get_log_likelihood(compare_dict["problematic"])
+    with pytest.raises(TypeError, match="model problematic"):
```
OriolAbril, line 221: we can use a regular expression to check that the error message includes both `problematic` and the ic name, I think `f"{ic}\.+model problematic"` will do this.

OriolAbril (commented):
I added a couple minor comments, looks good

arviz/tests/base_tests/test_stats.py
rpgoldman: OK, I will change that.

arviz/stats/stats.py
rpgoldman: Without that cast, I get this error:

```
arviz/stats/stats.py:1314: error: Argument 1 to "full" has incompatible type "Tuple[Union[number[Any], ndarray], int]"; expected "Union[int, Sequence[int]]"
```

mypy believes that `n_vars` could be an `ndarray`, because I believe it does not think deeply enough about what `np.sum()` could return.

I can move the `cast` so that it's just `cast(int, n_vars)`, but I can't get rid of it.  See my latest push.

arviz/tests/base_tests/test_stats.py
rpgoldman: Done!

rpgoldman:
@OriolAbril If my changes meet with your approval, please squash and merge.

PR #1626, opened 2021-03-21
Remove PyMC4 from doc index and add badges

canyon289:
## Description
Pretty sure we're not going to be supporting PyMC4 anytime soon. Are we still going to support edward or edward2?

**Edit**
Also add pypi badge like we are doing for README.md


OriolAbril (approved):
> Are we still going to support edward or edward2?

I don't think it makes any sense to support edward at this point, at most edward2.

OriolAbril (approved):

PR #1629, opened 2021-03-24
enabled type checking for io_pymc3 input

### LINKED ISSUE #1617: Check for InferenceData input on from_pymc3
`pymc3.sample` has a keyword argument that makes the function return an InferenceData object, `from_pymc3` is called under the hood by pymc3. This can generate confusion when doing also posterior predictive checks or similar postprocessing, and it is common to end up calling `from_pymc3` manually again but using the `InferenceData` as `trace` argument, which does not work nor should it work.

We should check for InferenceData input on io_pymc3 to make sure the input is not an inferencedata and if it is raise an error pointing people out to https://arviz-devs.github.io/arviz/api/generated/arviz.InferenceData.extend.html#arviz.InferenceData.extend.

MgeeeeK:
## Description

Enabled type checking for `io_pymc3` input... Now it raises an error whenever input type is of `InferenceData`.

Fix #1617 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@
 * Fix `ess/rhat` plots in `plot_forest` ([1606](https://github.com/arviz-devs/arviz/pull/1606))
 * Fix `from_numpyro` crash when importing model with `thinning=x` for `x > 1` ([1619](https://github.com/arviz-devs/arviz/pull/1619))
 * Upload updated mypy.ini in ci if mypy copilot fails ([1624](https://github.com/arviz-devs/arviz/pull/1624))
-* Enforced using `pymc3.backends.base.MultiTrace` as the `trace` value for `io_pymc3`  ([1629](https://github.com/arviz-devs/arviz/pull/1629))
+* Added type checking to raise an error whenever `InferenceData` object is passed using `io_pymc3`'s `trace` argument ([1629](https://github.com/arviz-devs/arviz/pull/1629))
 
 ### Deprecation
 * Deprecated `index_origin` and `order` arguments in `az.summary` ([1201](https://github.com/arviz-devs/arviz/pull/1201))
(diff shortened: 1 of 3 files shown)


OriolAbril (commented):
looks good, can you also add a test for this?

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 * Fix `ess/rhat` plots in `plot_forest` ([1606](https://github.com/arviz-devs/arviz/pull/1606))
 * Fix `from_numpyro` crash when importing model with `thinning=x` for `x > 1` ([1619](https://github.com/arviz-devs/arviz/pull/1619))
 * Upload updated mypy.ini in ci if mypy copilot fails ([1624](https://github.com/arviz-devs/arviz/pull/1624))
+* Enforced using `pymc3.backends.base.MultiTrace` as the `trace` value for `io_pymc3`  ([1629](https://github.com/arviz-devs/arviz/pull/1629))
```
OriolAbril, line 28: I would say something along the lines of raising an informative error if inferencedata is passed

arviz/tests/external_tests/test_data_pymc.py
```diff
@@ -139,6 +139,16 @@ def test_from_pymc_predictions(self, data, eight_schools_params):
             assert ivalues.shape[0] == 1  # one chain in predictions
             assert np.all(np.isclose(ivalues[0], values))
 
+    def test_from_pymc_trace_inference_data(self):
+        # check if error is raised successfully after passing InferenceData as trace
```
OriolAbril, line 143: This should be a docstring using ```"""``` instead of a comment

arviz/tests/external_tests/test_data_pymc.py
```diff
@@ -139,6 +139,16 @@ def test_from_pymc_predictions(self, data, eight_schools_params):
...
+        with pm.Model():
+            p = pm.Uniform("p", 0, 1)
+            pm.Binomial("w", p=p, n=2, observed=1)
+            trace = pm.sample(100, chains=2, return_inferencedata=True)
+            assert isinstance(trace, InferenceData)
```
OriolAbril, line 148: I think we don't need to sample from a model, we can use an example inferencedata or create it with `from_dict` and then do `az.from_pymc3(idata, model=pm.Model())` directly

OriolAbril (approved):

PR #1632, opened 2021-03-25
Infer dtypes for all Stan interfaces

### LINKED ISSUE #1630: dtypes with cmdstanpy
We should take a look at how we handle dtypes in `from_cmdstanpy` (or directly on cmdstanpy). See https://github.com/teddygroves/figure_skating/pull/1 for an example. 

Note this is more of an issue _candidate_ for now. I will try to play around a bit and see where things are going wrong, but I am posting anyways because 1) someone may already know the answer and I will not need to look into this and 2) maybe someone beats me into looking, I can't promise to look into this soon and we may want to fix this before the next release.

ahartikainen:
## Description
closes #1630

## Checklist

Infer dtypes from model code (all interfaces) or use user given dictionary (cmdstanpy, cmdstan).

### DIFF
--- a/arviz/data/io_cmdstanpy.py
+++ b/arviz/data/io_cmdstanpy.py
@@ -117,6 +117,7 @@ class CmdStanPyConverter:
             self.posterior,
             items,
             self.save_warmup,
+            self.dtypes,
         )
 
         # copy dims and coords  - Mitzi question:  why???
@@ -229,6 +230,7 @@ class CmdStanPyConverter:
                 fit.column_names,
                 valid_cols,
                 self.save_warmup,
+                self.dtypes,
             )
 
         return (
@@ -269,6 +271,7 @@ class CmdStanPyConverter:
                 columns,
                 valid_cols,
                 self.save_warmup,
+                self.dtypes,
             )
 
         return (
@@ -309,6 +312,7 @@ class CmdStanPyConverter:
                 columns,
                 valid_cols,
                 self.save_warmup,
+                self.dtypes,
             )
         if isinstance(self.log_likelihood, dict):
             data = {obs_name: data[lik_name] for obs_name, lik_name in self.log_likelihood.items()}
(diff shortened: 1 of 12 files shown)


OriolAbril (commented):
Looks good, are you planning on adding tests? Maybe a test directly on `infer_stan_dtypes` is enough. We can always create an issue to see if it gets done

ahartikainen:
Yes. (I have not tested these yet)

arviz/data/io_cmdstanpy.py
```diff
@@ -755,6 +794,9 @@ def from_cmdstanpy(
     save_warmup : bool
         Save warmup iterations into InferenceData object, if found in the input files.
         If not defined, use default defined by the rcParams.
+    dtypes: dict or str
```
OriolAbril, line 797: not sure how to combine with this, but I think in general it will be more convenient to pass a `CmdStanModel` instance to the converter and have it take the model code from there, instead of either dict or string. Maybe there is even more useful info to be retrieved from the model object? @mitzimorris?

OriolAbril: I am quite sure it's in the model, but not in the CmdStanMCMC object that is the only one currently being passed to the converter

ahartikainen: Program code is read once, but it is not saved.

https://github.com/stan-dev/cmdstanpy/blob/ecd844e9945ecf153be21ab69112d71862d2e0be/cmdstanpy/model.py#L130

I don't think there is anything needed from the CmdStanModel object (at least for now).

ahartikainen: Ok, there is `.code` method, but I think if users need to give model object, then calling `.code` might do the same thing.


Should we add option to accept CmdStanModel?

ahartikainen:
I also added dtypes option for from_pystan, so if users have complicated stan file, they still have possibility to define correct dtypes.

arviz/data/io_cmdstanpy.py
ahartikainen: Updated

OriolAbril (approved):
Not sure I did a very good job with the changelog, either way, feel free to merge whenever

PR #1637, opened 2021-03-26
More informative error message for error reading InferenceData from NetCDF

rpgoldman:
## Description

Previously, when trying to read an `InferenceData` object using `from_netcdf` on a filesystem that did not support HDF5 file locking (full disclosure: I have no idea what HDF5 file locking is), one would get a cryptic `OSError: -101` which did not provide any assistance in fixing the problem.

I tracked this down, and found a solution on StackOverflow.  Because it's difficult to figure out (happens due to a dependency that is well down in the call tree), I tweaked `from_netcdf` to trap this error, and add an explanation.

## Checklist

## Remarks

WRT the code style, running pylint locally I get it complaining that the use of `e` as a variable name is bad, but this is quite normal usage, stylistically, for variables that will be bound to `Exception`'s.


arviz/data/inference_data.py
```diff
@@ -328,16 +328,30 @@ def from_netcdf(filename: str) -> "InferenceData":
...
-                    groups[group] = data
-        return InferenceData(**groups)
+            for group in data_groups:
+                with xr.open_dataset(filename, group=group) as data:
+                    if rcParams["data.load"] == "eager":
+                        groups[group] = data.load()
+                    else:
+                        groups[group] = data
+            return InferenceData(**groups)
+        except OSError as e:  # pylint: disable=invalid-name
```
OriolAbril, line 342: To use `e` (instead of `err` or something a bit longer that pylint considers optimal) as variable to raise errors from other errors, it may be worth it to add it to `.pylintrc` instead of having to disable the error every time

OriolAbril (commented):
LGTM, only needs to be added to the changelog

rpgoldman:
@OriolAbril @ColCarroll This is ready to go, if you are OK with it, please squash and merge.

OriolAbril (approved):

Issue #1639, opened 2021-03-28
have get_bins distinguish between float and int data


OriolAbril:
[get_bins](https://github.com/arviz-devs/arviz/blob/main/arviz/stats/density_utils.py#L923) currently assumes that `values` passed to it are always integer values, it therefore enforces a minimum bin width of 1. There are some cases however where we may want to use a histogram for continuous data and we do have a `kind` argument in order to override the default continuous->kde, discrete->hist, so it should work.

We should look at the dtype of the input values and act accordingly. if the dtype is a discrete type, we should keep the exact same code and return the same number of bins, otherwise, we can either keep a similar version of the code without the specifics of discrete variables or call https://numpy.org/doc/stable/reference/generated/numpy.histogram_bin_edges.html with `bins="auto"`

ahartikainen:
We probably want higher bin count than auto gives.

rcparams could work? Then a few examples with rccontext?

OriolAbril:
I think our custom way to get the bins that we currently use for int data will do that (provided we remove the bin_width>=1 constraint)

PR #1641, opened 2021-03-29
Added floating margins for get_bins

rohits5496:
## Description

Added ability to calculate float margins with get_bins. This addresses issue #1639 .

The code distinguishes between dtype of the values array. If int32, previously defined code is used. If not int32, numpy [histogram_bin_edges ](https://numpy.org/doc/stable/reference/generated/numpy.histogram_bin_edges.html) function is used with bins='auto' and default arguments. 

Open to feedback and revisions to suit codebase.

## Checklist

### DIFF
--- a/arviz/stats/density_utils.py
+++ b/arviz/stats/density_utils.py
@@ -947,8 +947,14 @@ def get_bins(values):
     than the standard deviation. However, the IQR depends on fewer points than the standard
     deviation, so it is less accurate, especially for long tailed distributions.
     """
-    x_min = values.min().astype(int)
-    x_max = values.max().astype(int)
+    dtype = values.dtype.kind
+
+    if dtype == "i":
+        x_min = values.min().astype(int)
+        x_max = values.max().astype(int)
+    else:
+        x_min = values.min().astype(float)
+        x_max = values.max().astype(float)
 
     # Sturges histogram bin estimator
     bins_sturges = (x_max - x_min) / (np.log2(values.size) + 1)
@@ -957,9 +963,16 @@ def get_bins(values):
     iqr = np.subtract(*np.percentile(values, [75, 25]))  # pylint: disable=assignment-from-no-return
     bins_fd = 2 * iqr * values.size ** (-1 / 3)
 
-    width = np.round(np.max([1, bins_sturges, bins_fd])).astype(int)
+    if dtype == "i":
+        width = np.round(np.max([1, bins_sturges, bins_fd])).astype(int)
+        bins = np.arange(x_min, x_max + width + 1, width)
+    else:
+        width = np.max([bins_sturges, bins_fd])
+        if np.isclose(x_min, x_max):
+            width = 1e-3
+        bins = np.arange(x_min, x_max + width, width)
 
-    return np.arange(x_min, x_max + width + 1, width)
+    return bins
 
 
 def _sturges_formula(dataset, mult=1):
--- a/arviz/stats/density_utils.py
+++ b/arviz/stats/density_utils.py
@@ -8,7 +8,6 @@ from scipy.optimize import brentq
 from scipy.signal import convolve, convolve2d, gaussian  # pylint: disable=no-name-in-module
 from scipy.sparse import coo_matrix
 from scipy.special import ive  # pylint: disable=no-name-in-module
-from numpy import histogram_bin_edges
 
 from ..utils import _cov, _dot, _stack, conditional_jit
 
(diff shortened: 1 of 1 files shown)


arviz/stats/density_utils.py
```diff
@@ -946,20 +947,37 @@ def get_bins(values):
...
+    numpy.histogram_bin_edges are default.
+
+
     """
-    x_min = values.min().astype(int)
-    x_max = values.max().astype(int)
 
-    # Sturges histogram bin estimator
-    bins_sturges = (x_max - x_min) / (np.log2(values.size) + 1)
+    if values.dtype == "int32":
```
OriolAbril, line 962: we should use `"i"` and/or `"f"`, more often than not data will be `int64` instead of `int32`, but the code should work in all cases. See for example https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/matplotlib/ppcplot.py#L130

arviz/stats/density_utils.py
```diff
@@ -946,20 +947,37 @@ def get_bins(values):
...
+        bins_fd = 2 * iqr * values.size ** (-1 / 3)
 
-    width = np.round(np.max([1, bins_sturges, bins_fd])).astype(int)
+        width = np.round(np.max([1, bins_sturges, bins_fd])).astype(int)
 
-    return np.arange(x_min, x_max + width + 1, width)
+        return np.arange(x_min, x_max + width + 1, width)
+
+    else:
+        return histogram_bin_edges(values, bins="auto")
```
OriolAbril, line 980: given that we are using the same approach as `auto` in numpy, I think it would be better to use our `bin_sturges` and `bins_fd`, therefore, if we decided to update the default behaviour, both cases would be updated. Only the `astype` in the min, max and width and the considering 1 in the call to max differ between int and float, the rest are the same.

arviz/stats/density_utils.py
rohits5496: Ofcourse, thanks for the example, using same technique now

arviz/stats/density_utils.py
rohits5496: Using the same method as defined in the code where `bin_sturges` and `bin_fd` are calculated with x_min and x_max astype as float. Note that the max possible value for last bin is `x_max + width`. The results are different than `histogram_bin_edges ` function since it places a hard limit on max value of last bin.

https://github.com/rohits5496/arviz/blob/aca1c21d68a8364617fc9c88badf57abe93974ab/arviz/stats/density_utils.py#L977-L979

For example: auto mode on histogram_bin_edges

1. Integer values
Given array : [0,1,1,2,3,5]
get_bins() : [0, 2, 4, 6]
histogram_bin_edges() : [0.  , 1.25, 2.5 , 3.75, 5.  ]

2. Float values
Given array : [0.01, 1,1,2,3,5]
get_bins() : [0.01      , 1.93612423, 3.86224846, 5.78837269]
histogram_bin_edges() : [0.01  , 1.2575, 2.505 , 3.7525, 5.    ]

ahartikainen:
The following is not needed to be solved in this PR, but what will happen when min and max are the same value (this will happen when there is a parameter which has stuck into a single value)?

Maybe we need to create an issue for this? Or did we have already fail-safe solution?

arviz/stats/density_utils.py
```diff
@@ -946,9 +946,23 @@ def get_bins(values):
...
     than the standard deviation. However, the IQR depends on fewer points than the standard
     deviation, so it is less accurate, especially for long tailed distributions.
+
+    -------
+    Adding functionality for non integer data
+
+    Segment only works if input dtype is not int32. Width of bins is computed with
+    numpy.histogram_bin_edges with bins='auto'. Auto mode uses the maximum of Sturges and
+    Freedman-Diaconis estimators to provide bin margins. Rest of the parameters to
+    numpy.histogram_bin_edges are default.
```
OriolAbril, line 956: this should be removed now

OriolAbril:
> Can either of you please highlight the issue in some more detail? If the max and min values are the same, what should be the ideal result ?

For histograms, if dtype is of integer kind, I think we don't need any changes, the function will generate a single bin of width 1. If dtype if of kind float instead, I think we should retrun a bin of width 1e-2 or something like this.

For KDE, I am not completely sure what will fix the issue, in that case the function returns both a grid of x values and the corresponding y values for the kde. My guess is that making the x grid an array of length 2 (similarly to what we did with the histogram case) and then making y an array of nans will fix this issue. But we'll probably need to be careful when computing the bandwidth and only try to compute it if min and max are not equal.

rohits5496:
> For histograms, if dtype is of integer kind, I think we don't need any changes, the function will generate a single bin of width 1. If dtype if of kind float instead, I think we should retrun a bin of width 1e-2 or something like this.
> 
> For KDE, I am not completely sure what will fix the issue, in that case the function returns both a grid of x values and the corresponding y values for the kde. My guess is that making the x grid an array of length 2 (similarly to what we did with the histogram case) and then making y an array of nans will fix this issue. But we'll probably need to be careful when computing the bandwidth and only try to compute it if min and max are not equal.

Fixed it for the histogram issue
https://github.com/rohits5496/arviz/blob/698e99a9deb0e9cfccd82cffb9ff3e25d19618bf/arviz/stats/density_utils.py#L971-L972

After going over the code, I feel setting up a separate PR for kde would be better since the issue is a lot more involved and might touch different functions. 

Apologies for the multiple commits.

OriolAbril:
Sorry about the delay @rohits5496, I have rebased the PR to fix the diff, I'll merge once tests pass, thank you for the PR

Issue #1645, opened 2021-03-30
Add a coords/dims guide


OriolAbril:
This guide should cover the following topics. 1) coords and dims definition in ArviZ (complementary to the conversion guides) 2) basic xarray selection and how this relates to the coords argument (also could be a good idea to cover var_names and filter_vars), 3) diving deeper into xarray. ArviZ itself uses a relatively small subset of xarray, even from a pure data storage prespective. In ArviZ each dimension has a coord with the same name by default, in xarray there can be many combinations, some of which I barely understand. One of the beauties of xarray is that as long as chain and draw dimensions are not removed or renamed, users are free to use any xarray feature and everything should still work. We should cover a bit this context info and then show some examples of bringing xarray features like multidim coordinates into ArviZ. 

## Thoughts on implementation
We have examples on several issues about mask, pointwise selection, multidim coords...

almostmeenal:
> basic xarray selection and how this relates to the coords argument (also could be a good idea to cover var_names and filter_vars),

I might take some time, but I wanna play around more with xarrays so I guess if I work on this, it'll probably give me a tangible roadmap to learn a bit and then write it down. I understand this isn't marked beginner friendly or something, so I get it if it's not an appropriate issue for me, but I'd like to work on it.

OriolAbril:
Thanks! I actually forgot to link some issues in the original description with questions about slicing and coords  usage that can be helpful.

* https://github.com/arviz-devs/arviz/issues/1628 (for "section" 3)
* https://github.com/arviz-devs/arviz/issues/1627 (for "section" 2 advanced)

almostmeenal:
> Thanks! I actually forgot to link some issues in the original description with questions about slicing and coords usage that can be helpful.
> 
> * #1628 (for "section" 3)
> * #1627 (for "section" 2 advanced)

I'll have a look at these and get to it once I'm done making the inferencedata documentation PR. thanks!!

PR #1647, opened 2021-03-31
Renamed `sample` dim to `__sample__` when stacking `chain` and `draw`

### LINKED ISSUE #1613: az.loo fails when "sample" already exists as a dimension
**Describe the bug**

`az.loo` fails if there is a dimension named `sample` in the `log_likelihood` data variable. It looks like the `log_likelihood.stack(sample=("chain", "draw"))` line is the culprit.

**To Reproduce**

```python
import arviz as az
data.sample_stats = data.sample_stats.rename({"school": "sample"})
az.loo(data)
```

<details>
  <summary>Error</summary>
  
  ```
Traceback (most recent call last):
  File "tmp/loo_issue.py", line 5, in <module>
    az.loo(data)
  File "/Users/gibs/miniconda3/envs/birdman/lib/python3.7/site-packages/arviz/stats/stats.py", line 618, in loo
    log_likelihood = log_likelihood.stack(sample=("chain", "draw"))
  File "/Users/gibs/miniconda3/envs/birdman/lib/python3.7/site-packages/xarray/core/dataarray.py", line 2035, in stack
    ds = self._to_temp_dataset().stack(dimensions, **dimensions_kwargs)
  File "/Users/gibs/miniconda3/envs/birdman/lib/python3.7/site-packages/xarray/core/dataset.py", line 3578, in stack
    result = result._stack_once(dims, new_dim)
  File "/Users/gibs/miniconda3/envs/birdman/lib/python3.7/site-packages/xarray/core/dataset.py", line 3524, in _stack_once
    stacked_var = exp_var.stack(**{new_dim: dims})
  File "/Users/gibs/miniconda3/envs/birdman/lib/python3.7/site-packages/xarray/core/variable.py", line 1554, in stack
    result = result._stack_once(dims, new_dim)
  File "/Users/gibs/miniconda3/envs/birdman/lib/python3.7/site-packages/xarray/core/variable.py", line 1507, in _stack_once
    "cannot create a new dimension with the same "
ValueError: cannot create a new dimension with the same name as an existing dimension

  ```
</details>

**Expected behavior**

I expected that the computation would proceed as in the [documentation page](https://arviz-devs.github.io/arviz/api/generated/arviz.loo.html):

```
Computed from 2000 by 8 log-likelihood matrix

         Estimate       SE
elpd_loo   -30.81     1.43
p_loo        0.95        -
```

**Additional context**

This was with `arviz` commit 5957527 installed through `pip install git+git://github.com/arviz-devs/arviz.git` and `xarray` version `0.17.0`.

---

I realize that I can just change the name of my dimension from `sample` to something else to address this. However, I think it would be nice to not have to rename things, especially as the temporary dataset is not returned at all.

MgeeeeK:
## Description

Renamed `sample` dim to `__sample__` when stacking in `loo` and `loo_pit` to avoid dimension collision. Related methods that require stacking `chain` and `draw` will now expect a  `__sample__` dim.
Closes #1613 

## Checklist

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -752,8 +752,13 @@ def psislw(log_weights, reff=1.0):
     """
     Pareto smoothed importance sampling (PSIS).
 
-    Note that psislw will only act on the last dimension if the log_weights input is an
-    xarray.dataarray without a `__sample__` dimension or if it is a numpy array.
+    Notes
+    -----
+    If the ``log_weights`` input is an :class:`~xarray.DataArray` with a dimension
+    named ``__sample__`` (recommended) ``psislw`` will interpret this dimension as samples,
+    and all other dimensions as dimensions of the observed data, looping over them to
+    calculate the psislw of each observation. If no ``__sample__`` dimension is present or
+    the input is a numpy array, the last dimension will be interpreted as ``__sample__``.
 
     Parameters
     ----------
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -648,7 +648,7 @@ def loo(data, pointwise=None, var_name=None, reff=None, scale=None):
     log_likelihood = _get_log_likelihood(inference_data, var_name=var_name)
     pointwise = rcParams["stats.ic_pointwise"] if pointwise is None else pointwise
 
-    log_likelihood = log_likelihood.stack(sample=("chain", "draw"))
+    log_likelihood = log_likelihood.stack(__sample__=("chain", "draw"))
     shape = log_likelihood.shape
     n_samples = shape[-1]
     n_data_points = np.product(shape[:-1])
@@ -692,7 +692,7 @@ def loo(data, pointwise=None, var_name=None, reff=None, scale=None):
         warn_mg = True
 
     ufunc_kwargs = {"n_dims": 1, "ravel": False}
-    kwargs = {"input_core_dims": [["sample"]]}
+    kwargs = {"input_core_dims": [["__sample__"]]}
     loo_lppd_i = scale_value * _wrap_xarray_ufunc(
         _logsumexp, log_weights, ufunc_kwargs=ufunc_kwargs, **kwargs
     )
@@ -782,9 +782,11 @@ def psislw(log_weights, reff=1.0):
            ...: az.psislw(-log_likelihood, reff=0.8)
 
     """
-    if hasattr(log_weights, "sample"):
-        n_samples = len(log_weights.sample)
-        shape = [size for size, dim in zip(log_weights.shape, log_weights.dims) if dim != "sample"]
+    if hasattr(log_weights, "__sample__"):
+        n_samples = len(log_weights.__sample__)
+        shape = [
+            size for size, dim in zip(log_weights.shape, log_weights.dims) if dim != "__sample__"
+        ]
     else:
         n_samples = log_weights.shape[-1]
         shape = log_weights.shape[:-1]
@@ -799,7 +801,7 @@ def psislw(log_weights, reff=1.0):
     # define kwargs
     func_kwargs = {"cutoff_ind": cutoff_ind, "cutoffmin": cutoffmin, "k_min": k_min, "out": out}
     ufunc_kwargs = {"n_dims": 1, "n_output": 2, "ravel": False, "check_shape": False}
-    kwargs = {"input_core_dims": [["sample"]], "output_core_dims": [["sample"], []]}
+    kwargs = {"input_core_dims": [["__sample__"]], "output_core_dims": [["__sample__"], []]}
     log_weights, pareto_shape = _wrap_xarray_ufunc(
         _psislw,
         log_weights,
@@ -808,7 +810,7 @@ def psislw(log_weights, reff=1.0):
         **kwargs,
     )
     if isinstance(log_weights, xr.DataArray):
-        log_weights = log_weights.rename("log_weights").rename(sample="sample")
+        log_weights = log_weights.rename("log_weights")
     if isinstance(pareto_shape, xr.DataArray):
         pareto_shape = pareto_shape.rename("pareto_shape")
     return log_weights, pareto_shape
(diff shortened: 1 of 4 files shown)


OriolAbril (commented):
looks good, can you also add a note in the docstring of psislw about this convention?

MgeeeeK:
> can you also add a note in the docstring of psislw about this convention?

Okay :+1:

Should I also reflect these changes in the examples? Though, I think the note will do the job but still what do you say...

OriolAbril:
Yes,  updating the example too, good catch, I had missed that

arviz/stats/stats.py
```diff
@@ -719,6 +719,9 @@ def psislw(log_weights, reff=1.0):
     """
     Pareto smoothed importance sampling (PSIS).
 
+    Note that `__sample__` name is reserved only for the dimension storing
```
MgeeeeK, line 722: @OriolAbril can you verify if the message is correct... I'm having mixed feelings about it.

ahartikainen:
```suggestion
    Note that `__sample__` name is reserved.
```

Maybe just say this?

OriolAbril: It is not a reserved name here though. Here inputs _should_ have this dimension as it's the one that will be used. `psislw` can take either a numpy array or an xarray dataarray. 

* if the input if an xarray dataarray and has a `__sample__` dimension, `psislw` will **always** work, independently of the position of the `__sample__` dim, it can be the last, the first, any dimension in the middle.
* if the input is an xarray dataarray withoug a `__sample__` dimension, it is treated as a numpy array, see below.
* if the input is a numpy array, psislw will act on its _last_ dimension, if the input array does not fulfill the convention, psislw won't return the right results. It should still work, but it will act on the wrong dimension.

It's the inputs to the functions that call `psislw` (`loo` and `loo_pit`). 

As a summary, what we should document somewhere, probably in https://github.com/arviz-devs/arviz/issues/1645 or some of the arviz+xarray docs is that ArviZ has 3 reserved dimension names: `chain, draw and __sample__`

OriolAbril (changes requested):

arviz/stats/stats.py
MgeeeeK: > if the input array does not fulfill the convention, psislw won't return the right results

I was trying to covey this only, user may assign this name to some other dimension which will lead to broadcast error, the message was far from clear though. Since, `psislw` will be used as a private function only i don't think this message is needed.

> add note in the docstring of psislw about this convention

Should have asked this earlier, what convention were you talking about here?

OriolAbril: > which will lead to broadcast error,

I expect everything to work actually, but it will return the wrong result, which is even more dangerous.

> what convention were you talking about here?

That `psislw` is applied to the `__sample__` dimension. Take the example and pass both the current log_weights and the transposed version of log weights and you'll see that both work, in xarray one does not care about the order of the dimensions (case 1). Then do the same but when stacking, name the dimension `dim` instead of `__sample__`, now the results differ between original and transposed arrays (case 2). And finally do the same but instead of passing the xarray dataarray to it, use `.values` to pass numpy arrays, here again the result will differ between original and transposed array (case 3)

> Since, psislw will be used as a private function only i don't think this message is needed.

It is not a function the average used would use, but it is a public function, advanced users may use it as they please.

MgeeeeK: Sounds good, I'll update the docstring

MgeeeeK:
For some reason I'm getting this error in the build doc pipeline:

```bash
Exception occurred:
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/docutils/parsers/rst/__init__.py", line 196, in parse
    if len(line) > self.document.settings.line_length_limit:
AttributeError: 'Values' object has no attribute 'line_length_limit'
 ```

File: arviz/stats/stats.py (hunk 1 above)
Comment (MgeeeeK) [reply arviz/stats/stats.py]: I've updated the note now

**Comment by ahartikainen** [comment]
> For some reason I'm getting this error in the build doc pipeline:
> 
> ```shell
> Exception occurred:
>   File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/docutils/parsers/rst/__init__.py", line 196, in parse
>     if len(line) > self.document.settings.line_length_limit:
> AttributeError: 'Values' object has no attribute 'line_length_limit'
> ```

Probably docutils has been updated (docutils==0.17, 2021-04-03 https://pypi.org/project/docutils/#history )

Working workflow has docutils==0.16 https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=4267&view=logs&j=5e60a62b-366f-55f7-ec2d-5209191ccb17&t=3b19a6af-f535-5ba4-1e9a-379fb7d07d81

I will create a quick fix PR and then we can create a correct fix later.

File: arviz/stats/stats.py [hunk 2]
```diff
@@ -719,6 +719,9 @@ def psislw(log_weights, reff=1.0):
     """
     Pareto smoothed importance sampling (PSIS).
 
+    Note that psislw will only act on the last dimension if the log_weights input is an
+    xarray.dataarray without a `__sample__` dimension or if it is a numpy array.
```
OriolAbril, line 723:
```suggestion
    Notes
    -----
    If the ``log_weights` input is an :class:`~xarray.DataArray` with a dimension named ``__sample__`` (recommended)
    ``psislw`` will interpret this dimension as samples, and all other dimensions as dimensions of the observed data,
    looping over them to calculate the psislw of each observation. If no ``__sample__`` dimension is present or
    the input is a numpy array, the last dimension will be interpreted as ``__sample__``.
```

OriolAbril (commented):
I think you can already update and rebase to make sure docs build is fixed

arviz/stats/stats.py
MgeeeeK: Thanks, this is definitely more appropriate.

MgeeeeK:
Hey @OriolAbril I've update the docstring and rebased the branch, I think its ready now.

OriolAbril (approved):

PR #1655, opened 2021-04-08
[DOC][WIP] Added user guide for dask support inside ArviZ 

### LINKED ISSUE #1631: [DOC] Add docs to showcase existing `Dask` support.
## Tell us about it

As @OriolAbril has pointed out in the gitter room, Arviz lacks on the documentation side for showing the current `Dask` capabilities in the stats module which was recently added in #1229 . 

[Computation](https://arviz-devs.github.io/arviz/user_guide/computation.html) page mentions about this as well.

## Thoughts on implementation

Just like [Numba - an overview](https://arviz-devs.github.io/arviz/user_guide/Numba.html), we can create a notebook where we can show the usage and effectiveness of `Dask` in an example workflow.

MgeeeeK:
This pull request adds a user guide for showing current dask capabilities in ArviZ. 
Closes #1631 

Link to the [notebook](https://nbviewer.jupyter.org/github/MgeeeeK/arviz/blob/dask_doc/doc/source/user_guide/Dask.ipynb)

Edits to be made:
- [x] Note on out of memory computation
- [x] Provide a link to [best practices](https://docs.dask.org/en/latest/array-best-practices.html) doc.
- [x] Section on how to get dask backed `InferenceData` objects.
- [x] For both non-dask and dask enabled method, initialize the data at the start instead of loading it everytime.
- [x] Small note on `Client` and `ResourceProfiler`
- [x] High level explanation on `az.Dask` and link to it's api docs
- [x] Check if `dask="allowed"` works or not


MgeeeeK:
hey! @OriolAbril , I'm sure that lot of places will require changes so keeping this pr as WIP.

You can view the notebook [here](https://nbviewer.jupyter.org/github/MgeeeeK/arviz/blob/dask_doc/doc/source/user_guide/Dask.ipynb).

I'm very late due to coursework, exams and internship :/... but hopefully will come up with a draft for feedbacks before weekend.

OriolAbril:
Thanks, this is a great start!

Some general level comments. The intro on what is Dask and when to use it is great. I would add a note on out or memory computation, as this is in my opinion the most relevant feature in ArviZ context. Probably also link to both why dask like you already do and to https://docs.dask.org/en/latest/array-best-practices.html. 

On the general structure of the doc, I would start with a section on how to get dask backed inferencedata objects (this is currently only possible by getting dask data somehow and then converting it to inferencedata, so your approach of dask array + from_dict is great) that we would progressively extend as more options become available. Then use always that same data for computation so that memit and time only count the computation time.

Add a note on `Client` and `ResourceProfiler` indicating that those are optional. Also, before starting to use dask in `ess`, `rhat`... add some high level explanation on `az.Dask` class and link to it's api docs. As notebooks are MyST format, you can do that with ``{class}`~arviz.Dask` `` 

As I was saying above, in my opinion the main use of Dask is allowing us to work with data that doesn't fit into RAM, so I would focus mostly on that, and mention any time improvement as a welcome side effect, but not our goal. To really speed things up we should improve our use of Numba. As per the best practices page I linked above:

> If your data fits comfortably in RAM and you are not performance bound, then using NumPy might be the right choice. Dask adds another layer of complexity which may get in the way.
>
> If you are just looking for speedups rather than scalability then you may want to consider a project like Numba

And lastly, mostly out of curiosity, can you try what happens when using `dask="allowed"`? I have a feeling that `rhat` may work, not much hope for the others but who knows. This is completely secondary, there is no rush nor need to do this, feel free to ignore that if you are time constrained

MgeeeeK:
Thanks, I've listed out the suggestions, will start working on them.

Also, do I need to run the sphinx builder to update the changes on the documentation side?

OriolAbril:
> Also, do I need to run the sphinx builder to update the changes on the documentation side?

Not sure I understand, what you'll have to do eventually is edit https://github.com/arviz-devs/arviz/blob/main/doc/source/user_guide/computation.md so the dask guide is included in the right place of the docs.

> Edit: If you would've noticed, there is a dask's task scheduling graph as well. I have some mixed feelings on that as its size is around 4MB... what do you say? should I remove it completely?

Can you build docs locally? If so, I would recommend to build them locally and open them in the browser to see if the page is slow to load due to the bokeh plots, if it isn't I think we can leave that there to show how dask uses the resources available. 

As an example of slow borderline too slow to be acceptable, take https://arviz-devs.github.io/arviz/examples/plot_khat.html. You'll see that when you open the page it freezes and have to wait a couple seconds before being able to click on the bokeh tab or the link to the api docs.

MgeeeeK:
@OriolAbril I've added the changes, can you review it once again... I also tried to run each method using `dask="allowed"` but none of them ran properly (it was really slow and i think computation was single threaded), will look into this later.

Also, I'll resume working on this pr after finishing my draft proposal :(

OriolAbril:
> I've added the changes, can you review it once again

Thanks, happy to review, will probably do so in a few days.

> I also tried to run each method using dask="allowed" but none of them ran properly (it was really slow and i think computation was single threaded), will look into this later.

This is probably due to how `make_ufunc` works, it probably prevents parallelizing, not sure if there is a good way to work on that without duplicating much of our code and also avoiding to make numba a dependency (not optional anymore)

MgeeeeK:
Hey @OriolAbril , lemme know what all areas require change in the notebook... 

Also, (this is more about the project), since I was not able to get feedback on my proposal due to end moment submission and didn't discuss the project properly, my estimation for each milestone can be little off... Idk if it's allowed but if you have any suggestions or questions related to the proposal (cos some of the areas are too abstract) please let me know about those as well.

OriolAbril:
> Also, (this is more about the project), since I was not able to get feedback on my proposal due to end moment submission and didn't discuss the project properly, my estimation for each milestone can be little off... Idk if it's allowed but if you have any suggestions or questions related to the proposal (cos some of the areas are too abstract) please let me know about those as well.

We can discuss more on that once the community bonding period starts, no need to worry too much about this for now.

mortonjt:
Thank you @MgeeeeK , @OriolAbril for putting these docs together - this is incredibly useful.

One thing that could be helpful for users if there was a list of functions that accepts `dask_kwargs`.  For instance, `ess` and `rhat` both accept `dask_kwargs`, but `loo` doesn't. 

Also, from the notebooks, it looks like dask can only be used by initializing dask data arrays and passing them into arviz via `from_dict`, but I don't think this is the case, since I'm able to load InferenceData objects from disk and pass in the `dask_kwargs` without any issues.

MgeeeeK:
Hey @mortonjt thanks for all the suggestions, your feedback will definitely help as this still require some work.

---
> so that you don't have to pass in dask_kwargs into every function call?

yeah, its possible you can see for `rhat` we used this approach. I was trying to show here that if some function wants some different argument other than those that are stored globally then we can pass them to the function directly. 

example:

```python
az.Dask.enable_dask(dask_kwargs={"output_dtypes": [float], "dask": "parallelized"}) 
az.rhat(idata_dask)
az.ess(idata_dask, dask_kwargs={"output_dtypes": [int]}) # will overwrite the global parameters 
```

lmk if this was not what you were trying to say.

---
> Also, from the notebooks, it looks like dask can only be used by initializing dask data arrays and passing them into arviz via from_dict, but I don't think this is the case, since I'm able to load InferenceData objects from disk and pass in the dask_kwargs without any issues.

Got it, will discuss more about this with @OriolAbril , this notebook surely needs more workflows.

OriolAbril:
> in the very beginning, so that you don't have to pass in dask_kwargs into every function call?

This may not be very relevant yet that there are still only a few functions working with Dask so `dask_kwargs={"output_dtypes": [float], "dask": "parallelized"}` is a good default for basically everything. But that won't be the case as @MgeeeeK points out. It may still be worth it to set these as defaults so ess/rhat work without extra arguments, but loo, or hdi or even summary will need different values, some functions could even take `dask="allowed"` for extra performance, which is why I think its valuable to have the examples of function specific kwargs, not only `Dask` class level ones. We may need some extra explanation on this on the guide though so it is clear.

> One thing that could be helpful for users if there was a list of functions that accepts dask_kwargs. For instance, ess and rhat both accept dask_kwargs, but loo doesn't.

I agree this would be good, but we have to be careful because it's easy for this list to become outdated. 

> Also, from the notebooks, it looks like dask can only be used by initializing dask data arrays and passing them into arviz via from_dict, but I don't think this is the case, since I'm able to load InferenceData objects from disk and pass in the dask_kwargs without any issues.

It is not the only option, and the creating dask backed inferencedata objects should explain that, but I think it's the best option for the tutorial. But we could have some non executable examples for illustrative purposes.

Regarding your case @mortonjt , could you share how did you load the data? Did you chunk it after loading? I think that for numpy backed data, dask=parallelized works but doesn't really do anything.

mortonjt:
Regarding the loading, I was using `from_netcdf`, I wasn't using chunking,
so I actually didn't notice a performance boost after investigation. If
`from_netcdf` had support for chunking, that could be useful as well :)

mortonjt:
Hi @OriolAbril I've figured out how enabling chunking with from_netcdf (it just requires passing in separate kwargs into `xr.open_dataset`).  Opening a separate pull request

OriolAbril:
> Hi @OriolAbril I've figured out how enabling chunking with from_netcdf (it just requires passing in separate kwargs into xr.open_dataset). Opening a separate pull request

Sounds great, thanks for contributing. We should be a bit careful with the api I think though. We can continue the discussion on the PR, but I have some questions we should have answers for before merging. For example, not all groups will have the same dimensions, what happens if the chunk argument passed to xr.open_dataset has non existing dimensions? or if we want to chunk the posterior predictive and log likelihood in blocks of 10 draws but we don't want to chunk posterior nor sample stats ones?

PR #1660, opened 2021-04-13
modify io_pymc for v3&4 compatibility

### LINKED ISSUE #1278: Remove technical debt in `from_pymc3` and move to pymc3 repo
PyMC3 already depends on ArviZ and is moving towards a seamless integration between the two packages (i.e. autocoords and dims, returning `InferenceData` from `pm.sample`, having `from_pymc3` retrieve and use the model context...). Similarly to PyMC4, we thought it would make development of both packages easier to move `from_pymc3` to PyMC3 codebase at some point. This would allow to simplify the code of `from_pymc3` and to update it straight away changes in PyMC3 require it thus eliminating the need for synchronized releases of ArviZ and PyMC3 or the [current monkeypatching](https://github.com/arviz-devs/arviz/blob/master/arviz/data/io_pymc3.py#L33) of PyMC3 done in `io_pymc3`. 

This issue should be a place for roadmap and discussion on this. cc @AlexAndorra @michaelosthege @rpgoldman 

I think the first step would be identifying things that could be problematic when going ahead with the merger. Things that come to mind are maybe #1063 or making sure all functions used are public.

OriolAbril:
## Description

With this `from_pymc3` will only be available if pymc3 v4 is not installed. If it is, it will be a placeholder with an informative not implemented error. 

I think that with that we should freeze `from_pymc3` here,
and eventually remove all pymc3 io as it will always return inferencedata.

will eventually close #1278 

## Checklist

      PR format?
- [ ] Does the PR include new or updated tests to cover the new feature (using [pytest fixture pattern](https://docs.pytest.org/en/latest/fixture.html#fixture))?
- [ ] Is the new feature listed in the [changelog](https://github.com/arviz-devs/arviz/blob/main/CHANGELOG.md)?


twiecki:
Any progress on this? It's holding up https://github.com/pymc-devs/pymc3/pull/4726.

OriolAbril:
only mypy needs to be fixed or ignored, not really how how to do either. Regarding 4762, will merging this be enough or will it also need a release?

twiecki:
Also a release.

AlexAndorra (approved):
Well done figuring out the mypy stuff 😉 
LGTM, thanks @OriolAbril. Should I merge?

Issue #1651, opened 2021-04-05
contour plot


Hossein-1990:
Hi everybody
I’m going to plot a 95% probability contours using Arviz. I know that plot_pair is for constructing contours, but there isn't any option to adjust the HDI interval. Is there any other function to approach this problem?

Thanks

ahartikainen:
Our plot_pair basically calculates density with `_fast_kde_2d` function and then pushes that density on a grid to matplotlib contour function.

https://github.com/arviz-devs/arviz/blob/d10e5b3f1d71152f06e6d138807118c36cbd122b/arviz/stats/density_utils.py#L852

I'm not sure if there is a way to set specific values for contours, but if there is, then it should be possible through `kde_kwargs` -> `contour_kwargs`

    az.plot_pair(..., kde_kwargs={"contour_kwargs":  {"levels": [hdi_025, hdi_975]}})

But then, what are the correct hdi_025 and hdi_975 values.

@OriolAbril @aloctavodia 2D density -> user manually calls the `_fast_kde_2d` and then finds the correct density levels from that grid? What values should user look for?

OriolAbril:
I guess it is possible provided you trust the kde and don't mind computing things only within the grid it's defined in. As Ari suggested, you can choose which levels to plot when calling matplotlib contour function, but I don't think this is of any use in practice. There are too many things to check and that will probably depend on each specific pair of variables (kwargs are common for all of them). 

Here is a pseudo list of steps I would follow:
1. compute the 2d kde with `_fast_kde_2d` _and ensure it is normalized_ (I don't remember if we do normalize it, I don't think this is visible in the countour plots)
2. define a function to find the levels you want, that will probably have to be a numerical root finding algorithm. You choose a kde level, find the area in the 2d space with kde values larger than that, and integrate the kde in that area only, compare to your probability and find a level close enough. This will most probably give different levels for each pair of variables
   - You can also define a fine enough grid between 0 and max(kde_2d) and compute the integral for each point in the grid, then choose the level closer to your desired value. If done carefully, I believe you'll be able to evaluate some simple quadrature in a vectorized way along all the grid at once. 
3. use those levels as levels in the matplotlib contourplot.

ahartikainen:
One more thing to consider, you can probably calculate the area with the calculated contour (see bokeh code for calculating the contour locations with mpl functions manually).

Hossein-1990:
[...]

PR #1665, opened 2021-04-15
[WIP] Add contours of highest posterior density for 2D KDE plot

wm1995:
## Description
This pull request is designed to add the option to plot contours of highest posterior density, as discussed in #1651 with @aloctavodia. I think I've got the basic code working, but I'd deeply appreciate any comments. I'll start working my way through the checklist for a complete pull request now.

_Update (22/04/21):_ I think I'm nearly there with the pull request. I've attached a couple of sample plots

HDI contour plot with 1, 2 and 3 sigma contours

This is an example KDE plot of samples from a 2D multivariate normal distribution, with the 1-, 2-, and 3-sigma contours shown - the code is [here](https://github.com/arviz-devs/arviz/pull/1665/files#diff-13681bd267b761a9e3d69836047d61573d332064edccfe56d36e66c730d7a6cd).

HDI contour pair plot with 30%, 60% and 90% HDI contours

This is an example pair plot of the `centered_eight` example with 30%, 60% and 90% HDI contours - the code is [here](https://github.com/arviz-devs/arviz/pull/1665/files#diff-dd0be988feff035d525b010611d727db2116f33b9a1b9692d88c5f516bf983b3).

~~I tried to test the new feature with the Bokeh backend as well but because of #1673 the test fails.~~

_Update (23/04/21):_ I've added some example Bokeh plots, they're the same as the matplotlib ones just with the different backend
HDI
bokeh_pair
(These are just the HDI examples, I added a couple more general contour ones as well to mirror those available for matplotlib)

_Update (24/04/21):_ Noticed an error in the Bokeh pair plot with the HDI, so I've quickly fixed that and updated the picture above to reflect it. 

## Checklist

- [x] Ensure `pair_plot` compatibility with an example
- [x] Add Bokeh tests and examples


arviz/plots/kdeplot.py
```diff
@@ -20,6 +20,7 @@ def plot_kde(
     quantiles=None,
     rotated=False,
     contour=True,
+    hpd_levels=None,
```
OriolAbril, line 23: All `hpd` should be changed to `hdi`, or maybe even to `levels` directly. Is there any guarantee that those areas are the minimal area with probability X?

arviz/plots/kdeplot.py
```diff
@@ -261,6 +264,33 @@ def plot_kde(
...
+            if min(hpd_levels) < 0 or max(hpd_levels) > 1:
+                raise ValueError(
+                    "Highest posterior density confidence levels must be between 0 and 1"
+                )
+
+            # Need to include 0 & 1 if not included so that contours plot as expected
+            if 0 not in hpd_levels:
+                hpd_levels.append(0)
+            if 1 not in hpd_levels:
+                hpd_levels.append(1)
```
OriolAbril, line 278: I don't think this should be necessary.

arviz/stats/density_utils.py
```diff
@@ -1036,3 +1036,40 @@ def histogram(data, bins, range_hist=None):
...
+    # Calculate cumulative density
+    cum_density = sorted_density.cumsum()
+
+    # Find index of contour levels
+    contour_inds = np.empty_like(hpd_levels, dtype=int)
+    for idx, hpd_level in enumerate(hpd_levels):
+        if hpd_level == 0:
+            contour_inds[idx] = cum_density.shape[0] - 1
+        else:
+            contour_inds[idx] = np.argmax(cum_density >= (1 - hpd_level) * norm)
```
OriolAbril, line 1072: I kind of prefer the mask approach used in corner.py because I find it easier to understand, I select all the values with cumsum lower than the level then get the last one as level. With argmax I never remember what happens with ties, if the first occurrence or the last one is selected. 

Side notes, if we use this approach instead of try except, we shouldn't check for `==` but use `np.isclose`. Also, can't `cum_density.shape[0] - 1` be directly `-1`?

arviz/stats/density_utils.py
```diff
@@ -1036,3 +1036,40 @@ def histogram(data, bins, range_hist=None):
...
+
+
+def _find_hpd_contours(density, hpd_levels):
+    """
+    Find contours enclosing regions of highest posterior density.
+
+    Parameters
+    ----------
+    density : array-like
+        A gridded 2D KDE.
```
OriolAbril, line 1048: I would be explicit in that we need a grid where each value has the same area underneath. The area doesn't need to be a square, but it needs to be the same for all values.

arviz/plots/kdeplot.py
wm1995: I've changed all `hpd_levels` to `levels` in `plot_kde`, and `hdi_levels` in `_find_hdi_contours` (which I've also renamed). I was wondering whether `hdi_probs` might be more appropriate than either `levels` or `hdi_levels`, given that it'd be similar to the argument for `arviz.hdi`?

I think the areas given are as close to minimal area as we can get with a discretised KDE?

arviz/plots/kdeplot.py
wm1995: The reason I included 0 and 1 was to ensure that all the contours are filled in correctly (`plt.contourf` will not fill the highest contour otherwise, and using the `fill_last` argument to ensure the lowest contour is filled introduces an extraneous contour on the plot that I think outlines all non-zero values)

I've rewritten this section so that it's hopefully clearer that we're including 0 and `density.max()`, but please let me know if it's not clear and I can rewrite it again (or put some comments explaining the reasoning)

arviz/stats/density_utils.py
wm1995: I've changed the approach to reflect the one used in `corner.py`

arviz/stats/density_utils.py
wm1995:
```suggestion
        A 2D KDE on a grid with cells of equal area.
```
Would this suffice?

arviz/plots/kdeplot.py
aloctavodia: I agree on using hdi_probs for consistency

arviz/plots/kdeplot.py
aloctavodia: I think this is clear

arviz/plots/kdeplot.py
```diff
@@ -74,6 +75,8 @@ def plot_kde(
     contour : bool
         If True plot the 2D KDE using contours, otherwise plot a smooth 2D KDE.
         Defaults to True.
+    levels : list
+        Confidence levels for highest density interval contours of a 2D KDE.
```
aloctavodia, line 79: may be we can write  Confidence levels for highest density (2-dimensional) interval contours of a 2D KDE.

wm1995:
```suggestion
        Confidence levels for highest density (2-dimensional) interval contours of a 2D KDE.
```
Like this? Saying 2D KDE at the end might be redundant, so I can change that to just KDE?

arviz/stats/density_utils.py
OriolAbril: yes

arviz/plots/kdeplot.py
OriolAbril: This looks good, I'd keep the 2d kde as it's not too cluttery, but either alterntive is fine by me

arviz/tests/base_tests/test_utils.py
```diff
@@ -291,3 +291,48 @@ def test_normalize_angle(mean):
...
+    "cov",
+    [
+        np.diag([1, 1]),
+        np.diag([0.5, 0.5]),
+        np.diag([0.25, 1]),
+        np.array([[0.4, 0.2], [0.2, 0.8]]),
+    ],
+)
+@pytest.mark.parametrize("contour_sigma", [np.array([1, 2, 3])])
+def test_find_hdi_contours(mean, cov, contour_sigma):
```
wm1995, line 307: I'm a bit worried that this unit test is overly complicated, but I wasn't sure how best to approach it.

arviz/plots/kdeplot.py
aloctavodia: let's keep the 2d, a little redundancy is better in this case I think

arviz/tests/base_tests/test_utils.py
aloctavodia: I think is ok.

arviz/plots/kdeplot.py
wm1995: I've changed it to `hdi_probs` in cf10583

wm1995:
I think I've made all the requested changes and followed the proper format - I'll mark the code as ready for review again

wm1995:
I'm going to add some tests for the Bokeh backend, given that #1673 has been resolved

arviz/plots/kdeplot.py
```diff
@@ -74,6 +75,8 @@ def plot_kde(
     contour : bool
         If True plot the 2D KDE using contours, otherwise plot a smooth 2D KDE.
         Defaults to True.
+    hdi_probs : list
+        Confidence levels for highest density (2-dimensional) interval contours of a 2D KDE.
```
OriolAbril, line 79: Confidence feels strange, in this context, maybe something like "Plots highest density credibility regions for the provided probabilities. Defaults to matplotlib chosen levels with no fixed probability associated"?

OriolAbril (commented):
Looks good so far, adding bokeh and bokeh example counterparts would be great too

arviz/plots/kdeplot.py
wm1995: I've updated this in 6f2d912

wm1995:
Right, I've tested that it works with Bokeh and added a few Bokeh examples to complement the matplotlib ones, so hopefully it's all good to go now!

arviz/plots/kdeplot.py
```diff
@@ -261,6 +272,27 @@ def plot_kde(
...
+            contour_levels = _find_hdi_contours(density, hdi_probs)
+            contour_levels.sort()
+
+            contour_level_list = [0] + list(contour_levels) + [density.max()]
+
+            # Add keyword arguments to contour, contourf
+            if contour_kwargs is None:
+                contour_kwargs = {"levels": contour_level_list}
+            else:
+                contour_kwargs.setdefault("levels", contour_level_list)
```
OriolAbril, line 290: Shouldn't `hdi_probs` have preference over the `levels` in `kwargs`?

examples/bokeh/bokeh_plot_kde_2d.py
```diff
@@ -1,6 +1,6 @@
 """
-2d KDE Bokeh
-============
+2d KDE
```
OriolAbril, line 2:
```suggestion
2d KDE (default style)
```

See also next comment. We have a sidebar that helps navigate the examples: https://arviz-devs.github.io/arviz/examples/index.html. Now with the same title it's a bit confusing. Same should be done for matplotlib titles to match the ones here.

OriolAbril (approved):
This is great and very thorough, thanks!

I left two final comments as kind of cherry on top

examples/bokeh/bokeh_plot_kde_2d.py
wm1995: I've made this change in 3b56be1

arviz/plots/kdeplot.py
wm1995: I've set it up so that it warns the user if both are provided, and then uses hdi_probs preferentially; the update is in 5595173, with unit tests in cf9be3f.

wm1995:
I don't know whether it would be more appropriate to file a bug report or just add a commit here to fix it but the CI checks are failing because the newly-updated Pylint doesn't like lines 404:405 of `arviz/stats/density_utils.py` (they're not lines I've touched, but it throws an R1731 consider-using-max-builtin). I'll put in a commit here to fix it, but I think it'll cause checks on other builds to fail too until fixed, so perhaps a bug report would be a better course of action?

OriolAbril (approved):

wm1995:
Is there anything else you need me to do for the PR?

OriolAbril:
It looks like you now have to merge/rebase to fix the conflicts with `main`, other than that I think everything is ready. I'll wait for someone else to review and merge.

wm1995:
I've resolved the conflict, hopefully the checks will pass without the need for further intervention from me. Let me know if there's anything more you need!

OriolAbril:
Thanks @wm1995 !

wm1995:
Thank you!

PR #1666, opened 2021-04-16
fixed implementation of R2 score. R2 now measures variance over the correct axis. also added an input to return the full distribution 

nrakocz:
**Describe the bug**
there is a bug in implementing the Bayesian R-square as described by Aki Vehtari.

In his description:

R^2 = \frac{\mathrm{Var}_{\mu}}{\mathrm{Var}_{\mu}+\mathrm{Var}_\mathrm{res}}

For each posterior sample
here's a link: https://avehtari.github.io/ROS-Examples/Rsquared/rsquared.html

**The Arviz implementation:**
var_y_est = _numba_var(svar, np.var, y_pred.mean(0))
var_e = _numba_var(svar, np.var, (y_true - y_pred), axis=0)
r_squared = var_y_est / (var_y_est + var_e)

dimensions using example in source code
data = az.load_arviz_data('regression1d')
y_true = data.observed_data["y"].values
y_pred = data.posterior_predictive.stack(sample=("chain", "draw"))["y"].values.T
az.r2_score(y_true, y_pred)

dimensions:
y_true.shape
(100,)
y_pred.shape
(2000,100)

**The problem**
Notice that the original description generates n R-square samples, one for each posterior sample, while the Arviz implementation generates N samples, one for each data point.

**my implementation:**
var_y_est = _numba_var(svar, np.var, y_pred,axis=1)
var_e = _numba_var(svar, np.var, (y_true - y_pred), axis=1)
r_squared = var_y_est / (var_y_est + var_e)


arviz/stats/stats.py
```diff
@@ -963,22 +963,24 @@ def _gpinv(probs, kappa, sigma):
...
 
     Parameters
     ----------
-    y_true: array-like of shape = (n_samples) or (n_samples, n_outputs)
+    y_true: array-like of shape = (n_outputs,)
         Ground truth (correct) target values.
-    y_pred: array-like of shape = (n_samples) or (n_samples, n_outputs)
+    y_pred: array-like of shape = (n_posterior_samples, n_outputs)
         Estimated target values.
+    dist(defualt=False): Flag indicating whether to return all posterior samples of the Bayesian R² or only summary stats.
```
aloctavodia, line 975: Probably a better name for the argument would be "summary" or "samples". Also the following wording is more consistent with our doc style

summary: If False return all posterior samples of the Bayesian R² if True only the mean and standard deviation. Defaults to False.

arviz/stats/stats.py
```diff
@@ -993,14 +995,16 @@ def r2_score(y_true, y_pred):
...
+#     if y_pred.ndim == 1:
+#         var_y_est = _numba_var(svar, np.var, y_pred)
+#         var_e = _numba_var(svar, np.var, (y_true - y_pred))
+#     else:
+    var_y_est = _numba_var(svar, np.var, y_pred,axis=1)
+    var_e = _numba_var(svar, np.var, (y_true - y_pred), axis=1)
     r_squared = var_y_est / (var_y_est + var_e)
-
+    
+    if(dist):
```
aloctavodia, line 1006: if dist:

aloctavodia:
Hi @nrakocz do you need help?

ahartikainen:
I would still create a new function that returns the array and r2_score would then call this new function.

This way the function output won't change based on the kw flag.

nrakocz:
I've implemented the function that does that.
Need to change the naming in accordance to naming convention.

Was a little busy.  I'll do it this weekend.

Best regards,
Nadav

nrakocz (commented):
I've addressed the given comments.

I also think we should change the way "y_pred.ndim==1" is handled since this is not a desired input.

ahartikainen:
Hi, I still think a different function should return all the R2 values and then the current score function would call it and return the summary. This way our functions won't change the output based on keywords.

Issue #1673, opened 2021-04-22
Setting levels for 2D KDE contour plot returns errors for Bokeh backend


wm1995:
**Describe the bug**
Setting the `levels` parameter for either `contour_kwargs` or `contourf_kwargs` in `plot_kde` returns errors for the Bokeh backend.

**To Reproduce**
```python
import numpy as np
import arviz as az

rng = np.random.default_rng()
data = rng.multivariate_normal([2, 2], [[1, 0.4], [0.4, 0.8]], 1000000)

levels = [0, 0.01, 0.07, 0.13, 0.2]
ax = az.plot_kde(
    data[:, 0],
    data[:, 1],
    contourf_kwargs={"levels": levels},
    contour_kwargs={"levels": levels},
    backend="bokeh",
)
```
returns the error
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-7-6f97bf89ac37> in <module>
      6 
      7 levels = [0, 0.01, 0.07, 0.13, 0.2]
----> 8 ax = az.plot_kde(
      9     data[:, 0],
     10     data[:, 1],

/opt/arviz/arviz/plots/kdeplot.py in plot_kde(values, values2, cumulative, rug, label, bw, adaptive, circular, quantiles, rotated, contour, hdi_probs, fill_last, figsize, textsize, plot_kwargs, fill_kwargs, rug_kwargs, contour_kwargs, contourf_kwargs, pcolormesh_kwargs, is_circular, ax, legend, backend, backend_kwargs, show, return_glyph, **kwargs)
    341     # TODO: Add backend kwargs
    342     plot = get_plotting_function("plot_kde", "kdeplot", backend)
--> 343     ax = plot(**kde_plot_args)
    344 
    345     return ax

/opt/arviz/arviz/plots/backends/bokeh/kdeplot.py in plot_kde(density, lower, upper, density_q, xmin, xmax, ymin, ymax, gridsize, values, values2, rug, label, quantiles, rotated, contour, fill_last, figsize, textsize, plot_kwargs, fill_kwargs, rug_kwargs, contour_kwargs, contourf_kwargs, pcolormesh_kwargs, is_circular, ax, legend, backend_kwargs, show, return_glyph)
    185             else:
    186                 levels_scaled_nonclip = _scale_axis(np.asarray(levels), scaled_density_args)
--> 187                 levels_scaled = np.clip(levels_scaled_nonclip, 0, 1)
    188 
[...]

PR #1674, opened 2021-04-22
fix bokeh contour colors and levels

ahartikainen:
Small fixes for bokeh side.


wm1995:
Just continuing the discussion we were having in #1673, your pull request covers everything that mine covered and more; the only thing mine had that yours doesn't are a couple of unit tests to verify everything worked without throwing errors.

ahartikainen:
I think we can merge this is and can you add the unit tests in your current PR for kde levels?

wm1995:
Sure, will do

Issue #1488, opened 2021-01-08
Write plotting with ArviZ guides


OriolAbril:
## Tell us about it

Write a guide (one per backend) about plotting with ArviZ. It should not be about what to plot (what to plot would be in the educational resources) and focus on _how_. It should at least cover the following:

* [x] ~Sorting subplots. This [PyMC3 discourse topic](https://discourse.pymc.io/t/control-the-order-of-coordinates-in-arviz-inferencedata/4812?u=oriolabril) has many options~ This is independent of the backend, so it doesn't make much sense here. It is also independent of plotting (and useful for `az.summary` too for example). And probably the most important point, it is already covered in https://python.arviz.org/en/latest/user_guide/label_guide.html
* [ ] Defining axes manually. 
* [ ] Combining plots (both combining multiple arviz plots, arviz plots with "manual" plots and overlaying "manual" plots on top of ArviZ ones), synomym to use of `ax` argument.
* [ ] Customizing the look of the plots, titles, ticks... This section should be very simple and short and only show how working on arviz plots is basically the same as working on "manual" plots, the bulk of the section should be links to matplotlib/bokey guides.
* [ ] Use of `show`, `backend_kwargs` arguments. 
  * (matplotlib only) For `show` related to https://github.com/arviz-devs/arviz/pull/1768 and https://github.com/arviz-devs/arviz/issues/1767. It is a convenience function that calls `plt.show()` behind the scenes it has no effect on how the figure and axes are instatiated (when necessary) which is always with `plt.subplots(...)`.
* [ ] (bokeh only) bokeh tricks when in a jupyter-like environment
* [ ] (bokeh only) server mode?
* [x] ~Maybe also how to use coords? All plots work the same. See #1436~ Similar comments to the sorting point, this is independent of the backend and of plotting, and should be covered in a different guide, see https://github.com/arviz-devs/arviz/issues/1645 for tracking of the coords and dims guide. This two plotting guides can be seen as integration with matplotlib and bokeh, the coords and dims is more of a integration with xarray one.

symeneses:
I would like to work on this issues 🙋🏽‍♀️

PR #1676, opened 2021-04-23
Expose skip_dims argument as combine_dims

### LINKED ISSUE #1644: Expose skip_dims/flatten argument
All density plots (except plot_ppc) reduce the chain and draw dimensions, some have a combined argument to avoid reducing the chain dimension,  but in some cases this is not enough, and similarly to plot_ppc behaviour, we would like to choose at will which dimensions to reduce. I think we should support that. 

## Thoughts on implementation
Exposing skip_dims argument from xarray_var_iter should do the trick. This is basically what plot_ppc does with its flatten argument

ericch99:
## Description

This pull request is meant to close #1644. We have "exposed" the `skip_dims` argument in `xarray_var_iter` by adding a new argument `combine_dims` to all top-level plotting functions except for `plot_forest`, which we may need some more guidance on. Relevant tests and examples have also been added.

## Checklist

- [X] Includes sample plots to visually illustrate the changes (only for plot-related functions) -- any file in `arviz/examples` with "combinedims" in the script name
- [X] Includes new or updated tests to cover the new feature -- any test 
- [X] Code style correct (follows pylint and black guidelines)


arviz/plots/densityplot.py
```diff
@@ -17,6 +17,7 @@ def plot_density(
     group="posterior",
     data_labels=None,
     var_names=None,
+    combine_dims=[],
```
OriolAbril, line 20: The default value should be `None`. Using mutable objects like lists as default arguments can have unexpected effects, see for example https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

arviz/plots/densityplot.py
```diff
@@ -57,6 +58,8 @@ def plot_density(
         List of variables to plot.  If multiple datasets are supplied and var_names is not None,
         will print the same set of variables for each dataset.  Defaults to None, which results in
         all the variables being plotted.
+    combine_dims : Optional[List[str]]
```
OriolAbril, line 61: This style is mypy like, and should be used only in type hints (that is in the code itself, not in docstrings). We initially had not set a convention on this so as you have seen each docstring has it's own style. We are progressively moving to type hints where they should be and numpydoc format for docstrings. In this particular case, I would use `set of str, optional` or `set_like of str, optional`

arviz/tests/base_tests/test_plots_bokeh.py
```diff
@@ -148,6 +148,69 @@ def test_plot_density_bad_kwargs(models):
     with pytest.raises(ValueError):
         plot_density(obj, hdi_prob=2, backend="bokeh", show=False)
 
+def combinedims_test_plot_density_float(models, kwargs):
+    obj = [getattr(models, model_fit) for model_fit in ["model_1", "model_2"]]
+    axes = plot_density(obj, combine_dims=['school'], backend="bokeh", show=False, **kwargs)
+    assert axes.shape[0] >= 6
+    assert axes.shape[0] >= 3
```
OriolAbril, line 155: The 2nd assert is unnecessary. I don't understand however why are there more than 6 subplots, I think there should only be 4, one per variable: mu, tau, theta and theta_tilde. I also don't understand how is this test passing when `kwargs` is not defined anywhere :confused:

OriolAbril: Ohhhh, I see now. Pytest only executes the functions that start with `test_`, as the new tests start with `combineddims` they are not executed.

examples/matplotlib/mpl_plot_distcomp_combinedims.py
```diff
@@ -0,0 +1,5 @@
+import matplotlib.pyplot as plt
+import arviz as az
+
+data = az.load_arviz_data('rugby')
+az.plot_dist_comparison(data, var_names=["defs"], coords={"team" : ["Italy"]}, combine_dims=[''])
```
OriolAbril, line 5: Here combine_dims is not doing anything. As we are subsetting data to use only the italy coordinate value, there are no more dims that could be combined.

OriolAbril (commented):
The first step is to get the tests to run. After that I'll try to go over them and see which ones can be removed or merged with already existent tests. bad arguments tests for example don't need to be duplicated with and without combine_dims, maybe some of the tests where we don't assert a specific shape for the output axis can be merged into the general tests with the parametrized kwargs. As they all use the same undelying function, testing that all work and testing that one works correctly is probably enough.

For examples, this must be added to https://github.com/arviz-devs/arviz/issues/1488 more than anything. Maybe add 1-2 examples in the example gallery and n example usage in the docstring of the more common functions like plot_posterior, but we shoud try to avoid an example for each function and centralize the use of common args in the plotting guide to keep our documentation clear and concise.

Thank you both for the contribution!

arviz/plots/densityplot.py
ericch99: I believe this is resolved -- I changed the default to be `None`, but I believe `skip_dims` still requires a set-like object, so I added in a line that passes in an empty list instead, if `skip_dims = None`.

arviz/plots/densityplot.py
ericch99: This should be resolved! Please let us know if we happened to miss something. Thanks for the suggestion!

arviz/tests/base_tests/test_plots_bokeh.py
ericch99: I've changed all `combine_dims`-related test cases to follow the correct format. I have also updated the `assert` conditions to be `assert axes.shape[0] == 4` and added the pytest decorator to "pass in" any other arguments we may need, so `kwargs` should not be undefined now. The tests do all pass, but may need to be inspected for redundancy, per your previous comment. 

Notably, I have taken out all the new Bokeh tests for simplicity; once we receive feedback, I'll add them in, since they should 
1) follow a similar structure as the existing tests
2) test the same functionality as the Matplotlib tests.

Please let us know if we're missing something key, and thanks for the detailed feedback!

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -571,6 +575,53 @@ def test_plot_pair_overlaid(models, kwargs):
     assert ax is ax2
     assert ax.shape
 
+@pytest.mark.parametrize(
+    "kwargs", [{"var_names": ["mu", "theta"]}]
+)
+def test_plot_pair_combinedims(models, kwargs):
+    ax = plot_pair(models.model_1, combine_dims=["school"], **kwargs)
+    assert np.all(ax)
```
OriolAbril, line 583: This test can be added as an extra case in `test_plot_pair`, it will test the same it is doing here. It may be even better to add the `combine_dims=["school"]` to one of the many already present cases.

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -571,6 +575,53 @@ def test_plot_pair_overlaid(models, kwargs):
...
+    _, ax = fig_ax
+    ax = plot_pair(discrete_model, combine_dims=["school"], ax=ax, **kwargs)
+    assert ax
+
+
+def test_plot_pair_bad_combinedims(models):
+    with pytest.raises(ValueError):
+        plot_pair(models.model_1, combine_dims=["school"], kind="bad_kind")
+    with pytest.raises(Exception):
+        plot_pair(models.model_1, var_names=["mu"], combine_dims=["school"])
```
OriolAbril, line 599: I don't think it's necessary to check for errors adding a `combined_dims`, failing or not based on kind or var_names doesn't depend on `cobine_dims` so I think this is redundant with the original test. But I'lll deffer to our test expert @canyon289

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -571,6 +575,53 @@ def test_plot_pair_overlaid(models, kwargs):
...
+    assert np.all(ax)
+
+
+@pytest.mark.parametrize(
+    "kwargs", [{"kind": "scatter"}, {"kind": "kde"}, {"kind": "hexbin", "colorbar": True}]
+)
+def test_plot_pair_2var_combinedims(discrete_model, fig_ax, kwargs):
+    _, ax = fig_ax
+    ax = plot_pair(discrete_model, combine_dims=["school"], ax=ax, **kwargs)
+    assert ax
```
OriolAbril, line 592: I am not sure what this is testing nor why is the fixture axis used. Do we want to make sure a single axes is used?

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -571,6 +575,53 @@ def test_plot_pair_overlaid(models, kwargs):
...
+
+def test_plot_pair_bad_combinedims(models):
+    with pytest.raises(ValueError):
+        plot_pair(models.model_1, combine_dims=["school"], kind="bad_kind")
+    with pytest.raises(Exception):
+        plot_pair(models.model_1, var_names=["mu"], combine_dims=["school"])
+
+
+@pytest.mark.parametrize("has_sample_stats", [True, False])
+def test_plot_pair_divergences_warning_combinedims(has_sample_stats):
```
OriolAbril, line 603: Same comment for this test, I think its redundant, plus, why is it using `combine_dims=["tau"]`?

OriolAbril (commented):
Will try to continue reviewing afterwards if I have time, I hope the comments are already useful

OriolAbril:
As a way of showing how this feature will have immediate use and application, [here](https://discourse.mc-stan.org/t/ppcs-in-arviz/22176) is one question where we could use this to mimic `ppc_stat_2d` in bayesplot with our `pair_plot`, so thanks again for the PR! :rocket:

OriolAbril:
Thanks @ericch99 and @Tsai-Ni for getting this started

PR #1678, opened 2021-04-27
Removed extra line

roshnaeem:
There was an extra line at the start of `index.md`. It seemed to be there by mistake.

### DIFF
--- a/doc/source/getting_started/index.md
+++ b/doc/source/getting_started/index.md
@@ -1,4 +1,3 @@
-(getting_started)=
 
 # Getting Started
 


OriolAbril (commented):
This line is important actually, it's defining an anchor using [myst](https://myst-parser.readthedocs.io/en/latest/using/syntax.html#extra-markdown-syntax) syntax so we can link to the getting started guide using this anchor instead of needing to use the whole path.

roshnaeem:
> This line is important actually, it's defining an anchor using [myst](https://myst-parser.readthedocs.io/en/latest/using/syntax.html#extra-markdown-syntax) syntax so we can link to the getting started guide using this anchor instead of needing to use the whole path.

Can we add a comment explaining this line, so it won't bug newcomers?

OriolAbril:
I think we should have a proper section on writing documentation in the developer guide, not sure a comment is the way to go. The anchors are parsed and therefore now shown in the documentation (i.e. https://arviz-devs.github.io/arviz/getting_started/index.html) which is what newcomers should read. 

Having this section in the developer guide would guide people who are interested in working on docs about rst, myst, general doc organization... Otherwise why should the comment be on the getting started page but not in the [sampling wrappers section](https://github.com/arviz-devs/arviz/blob/main/doc/source/user_guide/sampling_wrappers.md)? or the [sampling wrapper guides](https://github.com/arviz-devs/arviz/blob/main/doc/source/user_guide/pystan_refitting.ipynb)? the or even in the installation guide (which uses rst anchors instead of myst)? Adding comments everywhere doesn't feel right to me

I know everything is quite confusing right now, and trying to follow the current docs doesn't help much because they mix rst and myst (and notebook with myst), multiple docstring formats, multiple directives to achieve the same goal, even multiple formatting for the same directive, so don't hesitate to ask and we'll do our best to clarify.

roshnaeem:
Thank you so much for your response. Yes, you are right. Let me go through all the docs and I will get back back to you.

PR #1681, opened 2021-04-28
Use only `is_circular` argument in `plot_dist`

agustinaarroyuelo:
## Description
`circular` was used to indicate `"taylor"` bandwidht for the `kde` function and it co existed with the `is_circular` argument. In this PR I removed `circular` in favor of `is_circular`.

## Checklist


arviz/stats/density_utils.py
```diff
@@ -522,6 +522,8 @@ def kde(x, circular=False, **kwargs):
     arviz.stats.density_utils.kde: Arviz KDE estimator
     """
     if circular:
+        if circular == "degrees":
```
agustinaarroyuelo, line 525: Given that `_kde_circular` assumes inputs are in radians I added this internal conversion. We can discuss where else we can place it or if it shoudl exist at all.

OriolAbril (approved):
Looks good, I think the conversion is fine where it is

Issue #1680, opened 2021-04-27
plot_dist_comparison has no warning when max_subplots goes over the max


OriolAbril:

alexisperakis:
Hi, first-time contributor here, can I be assigned to this issue? Is there any further info?
Thanks

ColCarroll:
Welcome! Happy to assign to you.

You can check out [here](https://github.com/arviz-devs/arviz/blob/main/arviz/plots/traceplot.py#L200-L212) in traceplot, there are 12 lines that get the "plotters", and check that against the maximum number of subplots. I think this would be fixed by:

1. Putting some very similar code [here](https://github.com/arviz-devs/arviz/blob/main/arviz/plots/distcomparisonplot.py#L132)
2. Do something similar to [this test](https://github.com/arviz-devs/arviz/blob/main/arviz/tests/base_tests/test_plots_matplotlib.py#L231), but with `plot_dist_comparison`: you'll want to ask for a plot that would generate more than 6 subplots, and verify you still only get 6.

Let me know if you need more pointers!

alexisperakis:
Hi, I would like to ask you a question.
We want to check if `dc_plotters` goes over the limit.
If I understood correctly, is the maximum number of plots that arviz allows to be displayed the variable `len_plots`? If this is the case, I need to compare it with this variable

OriolAbril:
The behaviour is already correct, we want to cut `dc_plotters` to have at most `len_plots`. The issue is that when `dc_plotters` is actually truncated, a warning must be shown, like it's done in the code example linked above by Colin. If `dc_plotters` is short enough that it doesn't need to be truncated, then there should be no warning.

alexisperakis:
> The behaviour is already correct, we want to cut `dc_plotters` to have at most `len_plots`. The issue is that when `dc_plotters` is actually truncated, a warning must be shown, like it's done in the code example linked above by Colin. If `dc_plotters` is short enough that it doesn't need to be truncated, then there should be no warning.

Thank you

PR #1688, opened 2021-05-06
Development/plot dist comparison warning

alexisperakis:
## Checklist

Issue: #1680 
* if statement added [code](https://github.com/alexisperakis/arviz/blob/development/plot_dist_comparison_warning/arviz/plots/distcomparisonplot.py#L137%20-%20#L144) to warn users if the subplots are over the limit. Notice that in the comparison, the length of the list should be increased by one:`len(dc_plotters) + 1` 
* There is a  [test](https://github.com/alexisperakis/arviz/blob/development/plot_dist_comparison_warning/arviz/tests/base_tests/test_plots_matplotlib.py#L238-#L242) case at the corresponding test file, with `max_sublots` set to 6.  

Let me know if there is any issue regarding my work. Thank you for the help.


arviz/plots/distcomparisonplot.py
```diff
@@ -130,9 +131,17 @@ def plot_dist_comparison(
     len_plots = rcParams["plot.max_subplots"] // (len(groups) + 1)
     len_plots = len_plots if len_plots else 1
     dc_plotters = [
-        list(xarray_var_iter(data, var_names=var, combined=True))[:len_plots]
+        list(xarray_var_iter(data, var_names=var, combined=True)) [:len_plots]
```
OriolAbril, line 134: It looks like it is being truncated twice now, so I think the warning will always be triggered. Can you check locally in a few cases in addition to the test?

OriolAbril (commented):
The `.DS_Store` files were added to the PR, they should be removed and probably have `.gitignore` ignore them from now on.

arviz/plots/distcomparisonplot.py
alexisperakis: Yes, I will look into it

alexisperakis: Should I alter the .gitignore?

OriolAbril: Yes, adding things to the gitignore is fine, removing them is not

arviz/plots/distcomparisonplot.py
```diff
@@ -133,6 +134,14 @@ def plot_dist_comparison(
...
         for data, var in zip(datasets, var_names)
     ]
+    if len(dc_plotters) + 1 > len_plots:
+        warnings.warn(
+            "rcParams['plot.max_subplots'] ({max_plots}) is smaller than the number "
+            "of variables to plot ({len_plotters}), generating only {max_plots} "
+            "plots".format(max_plots=len_plots, len_plotters=len(dc_plotters)),
+            UserWarning,
+        )
+        dc_plotters = dc_plotters[:len_plots]
```
OriolAbril, line 144: the `xarray_var_iter` list is still getting truncated when defining `dc_plotters`, and it should be truncated at that level but afterwards while printing the warning. 

`dc_plotters` will have length of 2 or 3 always, as it is a list of lists, it's those sublists the ones that need to be truncated.

OriolAbril (commented):
there is still one ds store left

arviz/plots/distcomparisonplot.py
alexisperakis: I used a for loop to truncate those sublists, however I am not sure about the tests that I used. `test_plot_dist_comparison(models, kwargs)` and `test_plot_dist_comparison_different_vars()`
Any suggestions?

alexisperakis: Okay so I change the code, summing up the length of each list in one variable and then compare it with `len_plots`.
I also used a for loop to truncate the sublists. However, I am not sure about the tests.

ColCarroll: Yeah, this is confusing since `len_plots` is different from `rcParams["plot.max_subplots"]`, and it isn't obvious to me that 
1. you can plot exactly that number of subplots
2. you won't plot more

I think 
```python
total_plots = sum(sum(1 for _ in xarray_var_iter(data, var_names=var, combined=True)) for data, var in zip(datasets, var_names))
```
gives you the number of "desired plots". I would check if that number is bigger than `rcParams["plot.max_subplots"]`, and if so, issue a warning, and say you'll plot however many things are in `dc_plotters` (which i guess is `sum(len(j) for j in dc_plotters)`). The calculation on line 134 already verifies that you shouldn't have to truncate further (I think?), so you'd just have to issue a warning.

ColCarroll: Specifically, I think it is an accident that the test passes, and line 137 isn't the condition we care about at all!

alexisperakis: Yes I also thought that at line 134 ` dc_plotters = [ list(xarray_var_iter(data, var_names=var, combined=True))[:len_plots]`dc_plotters is already truncated, so I think that it should stay as is. 
However, do you think I should compare `total_plots` with `rcParams["plot.max_subplots"]` and not with `len_plots` which divides `rcParams["plot.max_subplots"]` with `(len(groups) + 1)`, because it was [mentioned](https://github.com/arviz-devs/arviz/issues/1680#issuecomment-832994459) that we want to cut `dc_plotters` to have at most `len_plots`?

arviz/plots/distcomparisonplot.py
```diff
@@ -134,6 +135,22 @@ def plot_dist_comparison(
...
+
+    if total_plots > rcParams["plot.max_subplots"]:
+        warnings.warn(
+            "rcParams['plot.max_subplots'] ({max_plots}) is smaller than the number "
+            "of variables to plot ({len_plotters}), generating only {max_plots} "
+            "plots".format(max_plots=maxplots, len_plotters=total_plots),
+            UserWarning,
+        )
+        for i in range(0, len(dc_plotters)):
+             dc_plotters[i] = dc_plotters[i][:len_plots]
```
OriolAbril, line 152: This is not necessary because the inner lists are truncated at creation, this would have no effect.

arviz/plots/distcomparisonplot.py
```diff
@@ -134,6 +135,22 @@ def plot_dist_comparison(
         for data, var in zip(datasets, var_names)
     ]
 
+    total_plots = sum(
+        sum(1 for _ in xarray_var_iter(data, var_names=var, combined=True))
+        for data, var in zip(datasets, var_names)
+    )
+    maxplots = sum(len(splot) for splot in dc_plotters)
```
OriolAbril, line 142: `maxplots` as defined here is not necessarily equal to the rcParam, here plots are done in blocks of 3-4 subplots, so this will be equal _or lower_ than the rcParam. To avoid confusion, I think the warning should use the rcParam first, then in the second occurrence it can be `maxplots`. It is not necessary nor correct to loop over dc_plotters though, it should be:

    maxplots = len(dc_plotters[0]) *  (len(groups) + 1)

arviz/plots/distcomparisonplot.py
```diff
@@ -134,6 +135,22 @@ def plot_dist_comparison(
         for data, var in zip(datasets, var_names)
     ]
 
+    total_plots = sum(
+        sum(1 for _ in xarray_var_iter(data, var_names=var, combined=True))
+        for data, var in zip(datasets, var_names)
+    )
```
OriolAbril, line 141: Here similar comment as for next line, the nested loop is not needed and is actually incorrect. It should be something like:

    total_plots = sum(1 for _ in xarray_sel_iter(datasets[0], var_names=var_names[0], combined=True))

Also as we only want to count, we can use `sel` instead of using `var`, we don't need to subset the data to get the values in each position to count how many positions there are.

OriolAbril (commented):
Added some comments, I think it will also be helpful to explain a bit better what the plotting function does to avoid further confusion.

`plot_dist_comparison` allows comparing distributions and can either compare 2 distributions between them or 3 of them at the same time. This is achieved by having a subplot per distribution and an extra plot where all 2 or 3 are overlayed together. Therefore, subplots are defined on a variable basis first, but each variable has 3-4 subplots, so subplots are added by blocks, not independently. In the case of comparing 2 dists, the look of the plot is the following:

```
====
(var1, dist1)  |  (var1, dist2)
----
  (var1, dist1+dist2)
====
(var2, dist1)  |  (var2, dist2)
----
  (var2, dist1+dist2)
====
...
```

The example in the docs has a single variable: https://arviz-devs.github.io/arviz/api/generated/arviz.plot_dist_comparison.html. 

`dc_plotters` is a nested list of length # of distributions to compare, each of the lists within has length # of variables to plot. As it makes no sense to _compare_ distributions yet plot one of them and not the one it should be compared with, we can't cut the number of plots by cutting `dc_plotters` we need to cut the number of variables (like we are currently doing). 

`len_plots` takes this block structure into account to translate the number of subplots to number of varibles so we can use that value to cut the inner list of variables. Most of the new changes do not consider the extra overlayed plot and therefore would return incorrect results.

arviz/plots/distcomparisonplot.py
alexisperakis: Okay thank you for this explanation it was really useful and sorry for all the trouble caused, it's my first time contributing.  
So what I have understood and already done (no commit yet) is that I have compared the `total_plots` which are the desired plots by the user with the `rcParams["plot.max_subplots"]`. If `total_plots` are greater than this then the warning is popping up and it does not truncate the plots because they are already truncated at the creation. I also used the `maxplots` only at the description of the warning. I've also included a test case.

OriolAbril: Sounds great, and don't worry about the discussion and back-and-forth comments, I think they were useful to all of us.

arviz/plots/distcomparisonplot.py
```diff
@@ -134,6 +135,19 @@ def plot_dist_comparison(
         for data, var in zip(datasets, var_names)
     ]
 
+    total_plots = sum(
+        1 for _ in xarray_sel_iter(datasets[0], var_names=var_names[0], combined=True)
+    )
```
OriolAbril, line 140: I forgot to multiply in the sample code. This counts the number of variables which needs to be multiplied by `(len(groups) + 1)` to be equal to the total number of subplots that would be needed if we weren't truncating the lists of plotters (also read the next comment about the warning message).

arviz/plots/distcomparisonplot.py
```diff
@@ -134,6 +135,19 @@ def plot_dist_comparison(
...
+    maxplots = len(dc_plotters[0]) * (len(groups) + 1)
+
+    if total_plots > rcParams["plot.max_subplots"]:
+        warnings.warn(
+            "rcParams['plot.max_subplots'] ({max_plots}) is smaller than the number "
+            "of variables to plot ({len_plotters}), generating only {max_plots} "
+            "plots".format(max_plots=maxplots, len_plotters=total_plots),
```
OriolAbril, line 147: This needs to be updated, not sure what is the best way to do so, but somehow. Here are the things that need to somehow be changed:

* `smaller than the number of variables to plot` does not work here. In most plots the number of subplots is the same as the number of variables, but here each variable has 3-4 subplots. Maybe using subplots instead of variable is enough?
* `rcParams['plot.max_subplots'] ({max_plots})` precisely because of this 3-4 factor, `max_plots` is not necessarily equal to the rcParam. I think it will be very confusing to users to set the rcParam to 7, then get a warning saying we'll only plot 6 subplots because you have set the rcParam to 6.

OriolAbril (approved):
I think this is good to merge, it's only missing being added to the changelog

OriolAbril (approved):

Issue #1461, opened 2020-12-09
Posterior predictive check (plot_ppc) with dataframe encoding. 


sperezconesa:
Hello everyone!

I have a model which has two cases enconded in df['membrane_enc'], so this should give two indepently modelled likelyhoods and posteriors. I train the model in this fashion:
```python
shape = df['membrane_enc'].unique().shape
with pm.Model() as my_model:
    tau = pm.HalfNormal('tau', sigma=1.,shape =shape)
    tau_0 = pm.TruncatedNormal('tau_0', mu = min_val, sigma = 1, lower = 0, upper = min_val, shape =shape)
    like = DensityDist('like', logp_exp_lag(tau_0[df['membrane_enc']],tau[df['membrane_enc']]), observed=df['tpore'], random=custom_random, shape = (2,180))
with my_model:
    my_model_trace = pm.sample(4000, tune=3000,  random_seed=RANDOM_SEED)
    ppc = pm.sample_posterior_predictive(my_model_trace, var_names = ['like'],  random_seed=RANDOM_SEED)
    my_model_trace = az.from_pymc3(trace=my_model_trace, model=my_model, posterior_predictive=ppc)
    my_model_trace.to_netcdf(model_path)
```
The model converges nicely and gives reasonable posteriors for the two encodings. The problem comes when I try to plot the posterior predictive. I cannot find the way to plot the two distinct ppc's in two graphs:
```python
az.plot_ppc(my_model_trace, random_seed=RANDOM_SEED, var_names = ['like'], flatten = [])
```
This code fails and I am not sure which dimension I should flatten since there is only one in `my_model_trace`, `like_dim_0` and it doesn't work. All this code works nicely if you only work with one dimension. 

Can someone please help me make this plotting work?

By the way, I am using arviz 0.10.0.

Thank you very much in advance!
Best,
Sergio

OriolAbril:
Hi! 

Can you share the netcdf file or some code to generate fake data with the same/similar dims? I think I have a vague idea of the issue but I'd like to be sure before diving in.

sperezconesa:
Hello,
I think I narrowed down where the problem is. The problem is that in my_model_trace.observed_data and my_model_trace.posterior_predictive the data is 1d. plot_ppc does not know how to split it into chunks according to a label array. I made this quick hack but it is far from ideal:
```python
    fig, ax = subplots(n_rows, n_cols, figsize=(sizex,5.5*n_rows), sharex=False, sharey=False)
[...]

Issue #641, opened 2019-04-12
add example about transforming and/or creating variables


aloctavodia:
We should have at least one example about modifying, transforming and/or creating a variable before plotting, something like 

`data.posterior['log_temperature'] = np.log(data.posterior.temperature)`

`data.posterior['diff_temperature'] = data.posterior.temperature1 - data.posterior.temperature2`

canyon289:
@aloctavodia I'll add an example in the xarray section of tutorial in the meanwhile

corriebar:
Hey, I recently went through a few xarray tutorials to understand how to best modify and transform my posterior, so I would love to take this issue on to make it easier for the next one.
I can write something transformations such as @aloctavodia mentioned above, but also how to stack dimensions, group by dimension + aggregate, e.g. mean/median per group.

Where would be a good start for this? Just extend the section about inference data? Or the cookbook? Or better a new tutorial/notebook?

ahartikainen:
Hi, sound great

Maybe a new notebook? e.g. "Working with the InferenceData" or something similar?

corriebar:
Okay good, will do :)

OriolAbril:
From a conversation in slack and wanted to suggest using `idata.posterior = idata.posterior.assign(...)` also, in addition to `idata.posterior["new_var"] = ...`, as assign allows to add multiple variables at the same time. Moreover, it could also be interesting to add some examples on using [`idata.map`](https://arviz-devs.github.io/arviz/generated/arviz.InferenceData.map.html#arviz.InferenceData.map) to transform variables in several groups at the same time and add them to the corresponding group.

OriolAbril:
Closing in favour of #1486

Issue #1693, opened 2021-05-13
Ensure dimension order can be anything


sethaxen:
Currently all examples I have seen order the dimensions for samples as `(chain, draw, shape...)`, where `shape` is the shape of a single draw of the variable. This particular ordering shouldn't be necessary, and allowing alternative orderings might enable samples from PPLs to be used without copying. e.g. Turing stores variables as `(draw, shape..., chain)`.

This might already be supported. If so, it should be tested.

junpenglao:
+1. TFP mcmc currently output variables as `(draw, chain, shape, ...)`
I think an option to pass an indication or mapping of where the `draw` and `chain` is would be sufficient

OriolAbril:
There are two things to take into account for this to work, luckily with xarray the second is already taken care of. 

The first thing is conversion from ppl data to inferencedata. ppls use a given ordering but so far none uses named dimensions, so we need to know the ordering for things to work. Currently all converters end up calling `dict_to_dataset` which assumes the input are dicts with var names as keys and arrays as values. Those arrays currently need to be `chain, draw, *shape` which has the great advantage of being able to do things like `az.from_dict({"a": np.random.normal(size=(4, 100, 8))}, dims={"a": ["team"]})` and it works because it knows that if chain and draw are not in dims they need to be prepended. 

I think it would also be great to allow something like

```
az.from_dict({"a": np.random.normal(size=(100, 8, 4))}, dims={"a": ["draw", "team", "chain"]})
```

And have it work, which will need some work in the base converters but should be possible. Once this is done, we can also look into updating all the converters that currently use swapdims, as it may (or may not) be easier to handle things this way.

The second thing is that once data is already as inferencedata, ArviZ needs to **always** used named dims, so it doesn't actually care about the order of the dimensions anymore, only about their names (like xarray does). I am 90% sure that all ArviZ code does that already but we need to check. Here are some examples showing how this already works:

```python
import arviz as az
idata = az.load_arviz_data("rugby")
ds = idata.posterior[["atts", "defs"]]
ds = ds.transpose("draw", "team", "chain")
az.summary(ds)
```

```
          mean     sd  hdi_3%  hdi_97%  ...  mcse_sd  ess_bulk  ess_tail  r_hat
atts[0]  0.171  0.042   0.095    0.251  ...    0.001    2574.0    1409.0    1.0
atts[1] -0.083  0.044  -0.164   -0.001  ...    0.001    2330.0    1680.0    1.0
[...]

PR #1697, opened 2021-05-15
Working with InferenceData page

mjhajharia:
Addresses #1486

Notebook modified: [Working with InferenceData](https://arviz-devs.github.io/arviz/getting_started/XarrayforArviZ.html)

Tasks:
- [x]  modifying, transforming and/or creating a variable within an InferenceData group or groups. See #641 for code examples that can be used for this
- [x]  How to combine/merge/extend multiple InferenceData objects
- [ ]  Coordinate modification. Maybe something like #1461 (includes example code and explanation)
- [x]  Add links to/from [InferenceData schema](https://arviz-devs.github.io/arviz/schema/schema.html) and [xarray](https://arviz-devs.github.io/arviz/getting_started/XarrayforArviZ.html) page

Open to suggestions/improvements


doc/source/getting_started/WorkingWithInferenceData.ipynb
```diff
@@ -13,11 +13,12 @@
   },
```
OriolAbril, line 13: <p>I like the idea of using the variable just created for a 2nd computation. The cumulative mean however I think is to complicated for this tutorial is a bit too complicated for this tutorial. It is also using <code>len</code>  and positional indexing to loop over chains which we should try to avoid. Maybe there is something else we could do instead?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/arviz-devs/arviz/pull/1697/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

OriolAbril, line 13: <p>How about computing pairwise differences instead? Seeing this gave me the idea and shows cool xarray broadcasting and aligning:</p><p> </p><pre class="ql-syntax" spellcheck="false">idata.posterior.theta - idata.posterior.theta.rename(school="school_bis")
</pre><p> </p><p>Then, selecting <code>.sel(school="Choate", school_bis="Deerfield")</code>  gives the difference in effect between Choate and Deerfield, and likewise for all pairwise combinations.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/arviz-devs/arviz/pull/1697/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

almostmeenal: hi! this was brilliant, v cool, i added it, thanks for suggesting!!

almostmeenal: I used xarray rolling mean instead, have a look and let me know if that works? also, just curious, are len and positional indexing considered bad practices?

almostmeenal:
@OriolAbril any ideas for combining/extending `InferenceData` objects? I was thinking maybe converting a numpy array/pandas dataframe to create a new `InferenceData` object which can be merged with the default one would be a cool idea? It would also help show the ways to create `InferenceData`, It is something I found confusing initially and so people would probably like to know that Arviz `InferenceData` can be easily generated (for those not familiar with xarrays). Another option is we sample `posterior` with pymc3 and then extend the trace variable by adding `posterior_predictive`? Let me know which sounds like a better idea and I'll get to it

canyon289:
@almostmeenal in regards to combining have you seen these methods like adding groups and such? Any thoughts?

https://arviz-devs.github.io/arviz/api/inference_data.html#inferencedata-contents

almostmeenal:
> @almostmeenal in regards to combining have you seen these methods like adding groups and such? Any thoughts?
> 
> https://arviz-devs.github.io/arviz/api/inference_data.html#inferencedata-contents

I was thinking `extend` would be a simpler thing, for basics, maybe we can use extend and `concat`, and add a link to this page for other operations? also, i didn't exactly understand the difference between add_groups and extend, are groups like a dict of multiple `InferenceData` objects? there are no examples in the documentation there, so somewhat unclear on that one. Another point, we should probably atleast *try to* have a small example for every function. Maybe I'm just dumb, but I get confused with what the arguments actually mean unless I see an example? since xarray doesn't have a lot of "blogs" or tutorials as well, Arviz examples become more important for people to know what to do with `InferenceData`

doc/source/getting_started/WorkingWithInferenceData.ipynb
OriolAbril: I would say they are borderline bad practices because xarray code should ideally not depend on the order of the dimensions, only on their name. Therefore, doing `len(dataarray)` won't do this, the ideal approach would be `len(dataarray.dim_name)` so we are explicit on which is the dim whose length we want instead of grabbing the first dimension. 

Currently ArviZ enforces the `chain, draw, *shape` convention by reshaping and adding indices but this is an _input_ convention and code should not take as a property that once converted to xarray this order will still hold, because it will eventually break due to transpositions or https://github.com/arviz-devs/arviz/issues/1693. 

The convention should be understood as "if ArviZ gets unlabeled arrays as input it will default to interpreting their dimensions as `chain, draw, *shape`" not as something applied to the data once converted to labeled arrays

almostmeenal:
> I lean towards a completely synthetic and fake example/set of examples. i.e. take the centered_eight idata, make a copy (optionally subset it) and then say that the copied idata is actually data from a 2nd run. i.e. the first mcmc run doesn't give high enough ess, so you take the same model and run 2 extra chains so that combining the original 4 plus this 2 gives enough ess. Similar thing can be done with draws instead of chains.

I'm not sure if I understand this, but wouldn't we actually need to sample again to have 2 additional chains, and what if it creates confusion when we copy the same data and say that its from a different run?

almostmeenal:
> Creation examples should go in the tutorial dedicated to creating idata. I think we have both ndarray and dataframe conversion, but it could be interesting to add a couple extra cases.

okayy, that makes sense, lets leave that aside for a moment, will make a separate PR for creating inferencedata page?

almostmeenal:
> This has to be documented somewhere eventually, but now that we have moved `from_pymc3` to the pymc3 codebase, it should probably be in pymc docs. The task of having a posterior and using `.extend` to add posterior predictive is a great choice to show here, but I would use synthetic datasets. Other than the "creating inferencedata" page and the sampling wrappers, the rest of our docs should be ppl agnostic.

I agree with that, can we do something like - create two subsets of centered_eight and use extend, concat on those two subsets

almostmeenal:
> `extend` and `concat` (when not specifying dims! the examples above are the specifying dims case) are _extremely_ similar. The main difference is that `concat` has several copy/inplace options and _requires_ both idata to not have any common groups whereas `extend` takes a 2nd idata and adds all the groups the original one is missing to it. `add_groups` is a bit like `extend` but instead of taking the groups to be added from a 2nd idata it takes a dict of dicts or a dict of datasets.

then maybe a markdown cell explaining this difference would be useful?

almostmeenal:
> Yes, we definitely should, I think we have that written in the contributing guide but it is not enforced in many cases. We also need better ways of linking docstring examples and tutorial/guide examples using those same functions, hopefully gsod will help on that.

yes, okay then i'll try to only add existing links and not end up stepping on Rosheen's toes ahahaha

OriolAbril:
> I'm not sure if I understand this, but wouldn't we actually need to sample again to have 2 additional chains, and what if it creates confusion when we copy the same data and say that its from a different run?

I think we can do something _like_ (don't expect the code to run as is, it probably needs some extra work):

```
idata = az.load_...
idata_rerun = idata.sel(chain=[0, 1]).copy()
idata_rerun = idata_rerun.assign_coords(chain=[4, 5])
idata_complete = az.concat(idata, idata_rerun, dim="chain")
```

I envision this document as a tutorial (using [diataxis nomenclature](https://diataxis.fr/tutorials/)) so the focus should be on learning by doing via attainable tasks. While these tasks should be meaningful, they are chosen mostly because of their pedagogical value, not because of their intrinsec or real world value (how-to guides focus on real world value not on learning). We should also try to avoid giving too much explanations or context. 

It may be better to create a couple of new examples dedicated to the tutorial instead of doing copy trick or using synthetic data, but I think we should avoid having too much context, sampling... at all costs. The main audience for this page should be people who are new to InferenceData and want to learn how to use it, so they will follow the steps in the tutorial and achieve enough inferencedata ability to start using it on their own. 

Note that most of this is future planning, the doc is currently mostly a cookbook of common operations, but I think that by adding more cases and examples we'll eventually see how to organize everything in a more coherent way.

> then maybe a markdown cell explaining this difference would be useful?

I think each function should be clearly explained in their docstring, as well as linking to each other with the `See also` directive, but it could also have a couple lines here I guess, not sure it is the best place for that though but we can always move it later.

> yes, okay then i'll try to only add existing links and not end up stepping on Rosheen's toes ahahaha

Feel free to link to your heart's content, we should probably try a couple approaches before we set and write down the arviz "official" way to do so in the developer guide.

almostmeenal:
> I think we can do something _like_ (don't expect the code to run as is, it probably needs some extra work):

okayyy i'll try to get that done. about the rest of the comments, makes sense to me, i guess everything is really dynamic right now so i'll just add more content which you can review/redirect

OriolAbril:
yeah, docs are generally quite dynamic, and now ours are undergoing a period of particularly fast development. Moreover, even when aiming to use diataxis, we should not get stuck in the "not doing anything until we can do _everyhing_ right" but instead move forward step by step. https://diataxis.fr/how-to-use-diataxis/#allow-your-work-to-develop-organically has nice explanations and analogies about this.

doc/source/getting_started/WorkingWithInferenceData.ipynb
```diff
@@ -13,11 +13,13 @@
...
```
OriolAbril, line 13: <p>The explanation should not be as a comment but in a cell above, also with a title so it shows up in the right toc, same for all code samples.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/arviz-devs/arviz/pull/1697/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

OriolAbril, line 13: <p>Instead of adding this here afterwards, I would add a link with intersphinx in the initial explanation, something like "we will compute the rolling mean with <code> {meth}xarray.DataArray.rolling</code>...".  More on intersphinx  within ArviZ at https://arviz-devs.github.io/arviz/contributing/developer_guide.html#hyperlinks</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/arviz-devs/arviz/pull/1697/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

OriolAbril, line 13: <p>We could use a note directive here so it gets better formatting.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/arviz-devs/arviz/pull/1697/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

OriolAbril, line 13: <p>computation happens in the previous cell with the difference, here we are only selecting the difference we are interested in, I would not use the verb "compute"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/arviz-devs/arviz/pull/1697/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

OriolAbril, line 13: <p>related to these titles and sections I menioned above, I would have the concat and the extend examples in different sections, or maybe in the same section but different subsection. I would not compare them here, only show concat to extend with new chains, extend to add posterior_predictive to an already existing inferencedata with posterior samples (so different example), nothing else. Users at this level don't need to know more, more advanced guides or the docstring could be a place to discuss the differences.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/arviz-devs/arviz/pull/1697/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

almostmeenal:
@OriolAbril did the needful changes, hopefully its okay now

doc/source/getting_started/WorkingWithInferenceData.ipynb
OriolAbril, line 13: <p>This needs a title, it's no longer "Compute posterior mean values along draw and chains dimensions"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/arviz-devs/arviz/pull/1697/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

OriolAbril, line 13: <p>This is rST syntax for directives, in notebooks myst has to be use: https://myst-parser.readthedocs.io/en/latest/using/syntax.html#syntax-directives. In this case it should be:</p><p> </p><pre class="ql-syntax" spellcheck="false"><code></code>`{note}
text of note
<code></code>`
</pre><p> </p><p>I can' seem to get this to format properly though</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/arviz-devs/arviz/pull/1697/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

OriolAbril:
We recently merged some fixes to text and wording on that notebook, see https://github.com/arviz-devs/arviz/pull/1784, so you'll need to rebase. Let us know if you have any issue

OriolAbril:
Did some updates, I think we can merge like this and continue with more updates in other PRs

PR #1699, opened 2021-05-20
[WIP] improvement suggestions for the label guide documentation (until sorting dimensions chapter)

Eva-Lotte:

### DIFF
--- a/doc/source/user_guide/label_guide.rst
+++ b/doc/source/user_guide/label_guide.rst
@@ -20,27 +20,28 @@ Example: Default labelling
      ...: schools = az.load_arviz_data("centered_eight")
      ...: az.summary(schools)
 
-ArviZ supports label based indexing.
+ArviZ supports label based indexing powered by xarray.
 Through label based indexing you can use labels to plot a subset of selected variables.
 
 Example: Label based indexing
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-For a case where the coordinate values shown for the theta variable correspond to the `school` dimension
-you can plot only the ``tau`` variable to better inspect its 1.03 :func:`~arviz.rhat` value,
-only the ``theta`` variable to the ``Choate`` and ``St. Paul's`` coordinates, which have higher means, you can ploy only the ``theta`` variable by running the following command:
+For a case where the coordinate values shown for the ``theta`` variable coordinate to the ``school`` dimension
+you can indicate ArviZ to plot ``tau`` by including it in the ``var_names`` argument to inspect its 1.03 :func:`~arviz.rhat` value.
+To inspect the ``theta`` values for the ``Choate`` and ``St. Paul's`` coordinates, you can include ``theta`` in ``var_names`` and use the ``coords`` argument to select only these two coordinate values.
+You can generate this plot with the following command:
 
 .. ipython:: python
 
     @savefig label_guide_plot_trace.png
     az.plot_trace(schools, var_names=["tau", "theta"], coords={"school": ["Choate", "St. Paul's"]}, compact=False);
 
-With this you can now identify issues for the low ``tau`` values.
+With this you can now identify issues for low ``tau`` values.
 
 Example: Using the labeller argument
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To create a report on Deerfield, Hotchkiss and Lawrenceville schools for the probability of ``theta > 5``, use the labeller argument to customize labels. 
+To create a report on Deerfield, Hotchkiss and Lawrenceville schools for the probability of ``theta > 5`` and use the labeller argument to customize labels. 
 Unlike the default labels that show ``theta``, not $\theta$ (generated from ``$\theta$`` using $\LaTeX$), the labeller argument presents the label with proper math notation.
 
 You can use :class:`~arviz.labels.MapLabeller` to rename the variable ``theta`` to ``$\theta$``, as shown in the following example:
@@ -54,11 +55,9 @@ You can use :class:`~arviz.labels.MapLabeller` to rename the variable ``theta`` 
     @savefig label_guide_plot_posterior.png
     In [1]: az.plot_posterior(schools, var_names="theta", coords=coords, labeller=labeller, ref_val=5);
 
-Additional Resources
-~~~~~~~~~~~~~~~~~~~~
+.. seealso::
 
 - For a list of labellers available in ArviZ, see the :ref:`the API reference page <labeller_api>`.
-- For common customization classes, see the ``arviz.labels`` module.
 
 Sorting labels
 --------------
@@ -66,14 +65,14 @@ Sorting labels
 ArviZ allows labels to be sorted in two ways:
 
 - Using the arguments passed to ArviZ plotting functions 
-- Sorting the underlying xarray Dataset
+- Sorting the underlying :class:`xarray.Dataset`
 
 The first option is more suitable for single time ordering whereas the second option is more suitable for sorting plots consistently.
 
 .. note::
 
   Both ways are limited.
-  Multidimensional variables can not be separated from each other. 
+  Multidimensional variables can not be separated. 
   For example, it is possible to sort ``theta, mu,`` or ``tau`` in any order, and within ``theta`` to sort the schools in any order, but it is not possible to sort half of the schools, then ``mu`` and ``tau`` and then the rest of the schools.
 
 
@@ -104,7 +103,7 @@ Sorting variable names
 Sorting coordinate values
 .........................
 
-To sort coordinate values by mean you have to locate the means of each school and then use the DataArray result to sort the coordinate values. 
+To sort coordinate values you have to define the order, store it, and use the result to sort the coordinate values. 
 
 Example: Sorting the schools by mean
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -121,8 +120,8 @@ There are two ways of sorting:
 
 .. tabbed:: ArviZ args
 
-Sort the coordinate values to pass them as a `coords` argument and choose the order of the rows.
-To manually sort the schools, `sorted_schools`, define sorted_schools as a list.
+  Sort the coordinate values to pass them as a `coords` argument and choose the order of the rows.
+  To manually sort the schools, `sorted_schools`, define sorted_schools as a list.
 
   .. ipython::
 
@@ -131,7 +130,7 @@ To manually sort the schools, `sorted_schools`, define sorted_schools as a list.
 
 .. tabbed:: xarray
 
-You can use the :meth:`~xarray.Dataset.sortby` method to order our coordinate values directly at the source.
+  You can use the :meth:`~xarray.Dataset.sortby` method to order our coordinate values directly at the source.
 
   .. ipython::
 
--- a/doc/source/user_guide/label_guide.rst
+++ b/doc/source/user_guide/label_guide.rst
@@ -41,8 +41,8 @@ With this you can now identify issues for low ``tau`` values.
 Example: Using the labeller argument
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To create a report on Deerfield, Hotchkiss and Lawrenceville schools for the probability of ``theta > 5`` and use the labeller argument to customize labels. 
-Unlike the default labels that show ``theta``, not $\theta$ (generated from ``$\theta$`` using $\LaTeX$), the labeller argument presents the label with proper math notation.
+You can use the labeller argument to customize labels. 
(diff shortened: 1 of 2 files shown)


doc/source/user_guide/label_guide.rst
```diff
@@ -4,42 +4,46 @@
...
      ...: schools = az.load_arviz_data("centered_eight")
      ...: az.summary(schools)
 
-Thanks to being powered by xarray, ArviZ supports label based indexing.
-We can therefore use the labels we have seen in the summary to plot only a subset of the variables,
-the one we are interested in.
-Provided we know that the coordinate values shown for theta correspond to the `school` dimension,
-we can plot only ``tau`` to better inspect it's 1.03 :func:`~arviz.rhat` and
-``theta`` for ``Choate`` and ``St. Paul's``, the ones with higher means:
+ArviZ supports label based indexing.
```
OriolAbril, line 23: I would like to still be explicit about xarray. Many users only use ArviZ functions, plots and diagnostics and are not aware that all of that is powered by xarray under the hood. This is particularly relevant for researchers who then don't cite xarray when publishing even though they did cite ArviZ. It is not "our" problem per se, but I think we should try to nudge users towards the right citations and it is also bad for us indirectly because xarray is generally undercited which dificults it getting grants and slows down its development.

doc/source/user_guide/label_guide.rst
```diff
@@ -4,42 +4,46 @@
...
+Through label based indexing you can use labels to plot a subset of selected variables.
+
+Example: Label based indexing
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For a case where the coordinate values shown for the theta variable correspond to the `school` dimension
+you can plot only the ``tau`` variable to better inspect its 1.03 :func:`~arviz.rhat` value,
+only the ``theta`` variable to the ``Choate`` and ``St. Paul's`` coordinates, which have higher means, you can ploy only the ``theta`` variable by running the following command:
```
OriolAbril, line 31: My original description wasn't very good, but it tried to say that we are plotting all the things that were described, we are not plotting only the `theta` variable. How about something like:

```
Given that summary by default shows the coordinate values and not their indices or some aliases, 
you can use what you see to subset the data. 
The summary shows that ``tau`` has 1.03 :func:`~arviz.rhat` value which deserves closer inspection. 
You can indicate ArviZ to plot ``tau`` by including it in the ``var_names`` argument. 
Moreover, you may also want to inspect closely the ``theta`` values for the ``Choate`` 
and ``St. Paul's`` coordinates because of their higher means. 
You can do so by including ``theta`` too in ``var_names`` and 
using the ``coords`` argument to select only these two coordinate values for the ``school`` dimension. 
You can generate this plot with the following command:
```

doc/source/user_guide/label_guide.rst
```diff
@@ -4,42 +4,46 @@
...
 .. ipython:: python
 
     @savefig label_guide_plot_trace.png
     az.plot_trace(schools, var_names=["tau", "theta"], coords={"school": ["Choate", "St. Paul's"]}, compact=False);
 
-So far so good, we can identify some issues for low ``tau`` values which is great start.
-But say we want to make a report on Deerfield, Hotchkiss and Lawrenceville schools to
-see the probability of ``theta > 5`` and we have to present it somewhere with math notation.
-Our default labels show ``theta``, not $\theta$ (generated from ``$\theta$`` using $\LaTeX$).
+With this you can now identify issues for the low ``tau`` values.
```
OriolAbril, line 38:
```suggestion
With this you can now identify issues for low ``tau`` values.
```

doc/source/user_guide/label_guide.rst
```diff
@@ -4,42 +4,46 @@
...
+
+Example: Using the labeller argument
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Fear not, we can use the labeller argument to customize the labels.
-The ``arviz.labels`` module contains some classes that cover some common customization classes.
+To create a report on Deerfield, Hotchkiss and Lawrenceville schools for the probability of ``theta > 5``, use the labeller argument to customize labels.
```
OriolAbril, line 43:
```suggestion
You'll now create a report on Deerfield, Hotchkiss and Lawrenceville schools for the probability of ``theta > 5`` and use the labeller argument to customize labels.
```

doc/source/user_guide/label_guide.rst
```diff
@@ -50,24 +54,28 @@ tell it to rename the variable name ``theta`` to ``$\theta$``, like so:
...
     In [1]: az.plot_posterior(schools, var_names="theta", coords=coords, labeller=labeller, ref_val=5);
 
-You can see the labellers available in ArviZ at :ref:`their API reference page <labeller_api>`.
-Their names aim to be descriptive and they all have examples in their docstring.
-For further customization continue reading this guide.
+Additional Resources
+~~~~~~~~~~~~~~~~~~~~
+
+- For a list of labellers available in ArviZ, see the :ref:`the API reference page <labeller_api>`.
+- For common customization classes, see the ``arviz.labels`` module.
```
OriolAbril, line 61: The api reference has the docs for the `arviz.labels` module, the original last line tried to give a choice to readers. At this point they have seen how to use the `labeller` argument. Therefore, if they only want to do some more or less simple changes to the labels, they don't need to continue reading this guide, but they will probably need to read the api reference to find the class that suits their needs (which may be IdxLabeller instead of MapLabeller). Does this make sense? Maybe deleting this 2nd bullet point is enough already?

Also, how about using the `seealso` directive? https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-seealso

doc/source/user_guide/label_guide.rst
```diff
@@ -50,24 +54,28 @@ tell it to rename the variable name ``theta`` to ``$\theta$``, like so:
...
-Labels in ArviZ can generally be sorted in two ways,
-using the arguments passed to ArviZ plotting functions or
-sorting the underlying xarray Dataset.
-The first one is more convenient for single time ordering
-whereas the second is better if you want plots consistently sorted that way and
-is also more flexible, using ArviZ args is more limited.
+ArviZ allows labels to be sorted in two ways:
+
+- Using the arguments passed to ArviZ plotting functions 
+- Sorting the underlying xarray Dataset
```
OriolAbril, line 69:
```suggestion
- Sorting the underlying :class:`xarray.Dataset`
```

That was (yet another) oversight on my part when writing the guide.

doc/source/user_guide/label_guide.rst
```diff
@@ -50,24 +54,28 @@ tell it to rename the variable name ``theta`` to ``$\theta$``, like so:
...
+
+The first option is more suitable for single time ordering whereas the second option is more suitable for sorting plots consistently.
+
+.. note::
+
+  Both ways are limited.
+  Multidimensional variables can not be separated from each other.
```
OriolAbril, line 76:
```suggestion
  Multidimensional variables can not be separated. 
```

doc/source/user_guide/label_guide.rst
```diff
@@ -97,22 +104,25 @@ Sorting variable names
 Sorting coordinate values
 .........................
 
-We may also want to sort the schools by their mean.
-To do so we first have to get the means of each school:
+To sort coordinate values by mean you have to locate the means of each school and then use the DataArray result to sort the coordinate values.
```
OriolAbril, line 107:
```suggestion
To sort coordinate values you have to first define the order and store it, then use that result to sort the coordinate values. 
```

doc/source/user_guide/label_guide.rst
```diff
@@ -97,22 +104,25 @@ Sorting variable names
...
+2) You can use the DataArray result to sort the coordinate values for ``theta``.
+There are two ways of sorting:
 
 .. tabbed:: ArviZ args
 
-  Here the first step is to sort the coordinate values so we can pass them as `coords` argument and
-  choose the order of the rows.
-  If we want to manually sort the schools, `sorted_schools` can be defined straight away as a list
+Sort the coordinate values to pass them as a `coords` argument and choose the order of the rows.
+To manually sort the schools, `sorted_schools`, define sorted_schools as a list.
```
OriolAbril, line 125:
```suggestion
  Sort the coordinate values to pass them as a `coords` argument and choose the order of the rows.
  To manually sort the schools, `sorted_schools`, define sorted_schools as a list.
```

This needs to be indented for rST to parse the tabs correctly

doc/source/user_guide/label_guide.rst
```diff
@@ -121,7 +131,7 @@ Again we have two alternatives:
 
 .. tabbed:: xarray
 
-  We can use the :meth:`~xarray.Dataset.sortby` method to order our coordinate values straight at the source
+You can use the :meth:`~xarray.Dataset.sortby` method to order our coordinate values directly at the source.
```
OriolAbril, line 134:
```suggestion
  You can use the :meth:`~xarray.Dataset.sortby` method to order our coordinate values directly at the source.
```

OriolAbril (commented):
Thanks :heart: these really improve the clarity of the guide.

doc/source/user_guide/label_guide.rst
Eva-Lotte: Thank you for the hint with the `seealso` directive! This is what I was looking for.

Eva-Lotte:
Thank you very much for your review!

doc/source/user_guide/label_guide.rst
```diff
@@ -20,27 +20,28 @@ Example: Default labelling
...
     az.plot_trace(schools, var_names=["tau", "theta"], coords={"school": ["Choate", "St. Paul's"]}, compact=False);
 
-With this you can now identify issues for the low ``tau`` values.
+With this you can now identify issues for low ``tau`` values.
 
 Example: Using the labeller argument
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To create a report on Deerfield, Hotchkiss and Lawrenceville schools for the probability of ``theta > 5``, use the labeller argument to customize labels. 
+To create a report on Deerfield, Hotchkiss and Lawrenceville schools for the probability of ``theta > 5`` and use the labeller argument to customize labels.
```
OriolAbril, line 44: This sentence now feels incomplete. It starts with "to create" and a description on what to create but nothing on how to create it.

A bit more context in case it helps. Here there are two goals: creating a report of theta > 5 and using the labeller to customize the labels. 

One can be done without the other, but instead of showing how to use the labeller (which I think should be done in the example in he docstring that is still wip), I wanted to show how to use the labeller to solve a more specific and real task. When exploring the mode ourselves, we won't generally care much about the labels, having the same thing as the code is fine. However, if generating a report to be published or presented, we'll probably want to take better care of the presentation, and match the labels to the labels in the equations of the paper instead of variables in the code. Hence the MapLabeller

doc/source/user_guide/label_guide.rst
```diff
@@ -97,22 +103,25 @@ Sorting variable names
...
+2) You can use the DataArray result to sort the coordinate values for ``theta``.
+There are two ways of sorting:
 
 .. tabbed:: ArviZ args
 
-  Here the first step is to sort the coordinate values so we can pass them as `coords` argument and
-  choose the order of the rows.
-  If we want to manually sort the schools, `sorted_schools` can be defined straight away as a list
+  Sort the coordinate values to pass them as a `coords` argument and choose the order of the rows.
+  To manually sort the schools, `sorted_schools`, define sorted_schools as a list.
```
OriolAbril, line 124: To keep the example only about sorting by mean, I propose to move this above (I am commenting later in time on the above section in hopes that github will preserve the chronology).

```suggestion
```

doc/source/user_guide/label_guide.rst
```diff
@@ -97,22 +103,25 @@ Sorting variable names
 Sorting coordinate values
 .........................
 
-We may also want to sort the schools by their mean.
-To do so we first have to get the means of each school:
+To sort coordinate values you have to define the order, store it, and use the result to sort the coordinate values.
```
OriolAbril, line 106:
```suggestion
To sort coordinate values you have to define the order, store it, and use the result to sort the coordinate values.
The order can be defined by performing some operations on our xarray objects (like it is shown in the example below)
or by manually creating a list with the desired order.
```

doc/source/user_guide/label_guide.rst
Eva-Lotte: Thank you for providing a bit of context to this matter. The way I see it, since this piece of text is inside of the example, we don't even need the first part of the sentence.

OriolAbril (approved):
Looks great, only thing missing is adding to the [changelog](https://github.com/arviz-devs/arviz/blob/main/CHANGELOG.md#documentation) and merging.

Eva-Lotte:
Thank you, Oriol! I added the changes to the changelog! I can't seem to merge the PR, probably because of access rights?

OriolAbril:
Yeah, only core contributors have merge privilieges, I wasn't very clear in my previous post. Thanks for the PR!

PR #1707, opened 2021-05-27
Fixed Posterior plot errors with boolean array.

### LINKED ISSUE #1694: Posterior plot errors with boolean array
When passed a boolean array, `plot_posterior` fails with the following error because `np.histogram` assumes that subtraction is defined on the array:
```python
TypeError('numpy boolean subtract, the `-` operator, is not supported, use the bitwise_xor, the `^` operator, or the logical_xor function instead.')
```

This example reproduces the error:

```python
import numpy as np
import arviz as az

data = np.random.choice(a=[False, True], size=(4, 100))
az.plot_posterior(data)
```

utkarsh-maheshwari:
## Description
Plotted bar instead of hist for a boolean value.
Also, should HDI also be visible? I have assumed no. 

Fixes #1694

## Checklist


OriolAbril (commented):
I think we could use `if dtype kind == "f" -> elif dtype kind == "i" else (dtype == "b")` instead of converting to int and storing `is_bool`. We may also want to update a bit more the apperance of the plot? @sethaxen? i.e. showing the y axis in the boolean case, or hiding the mean and annotating both bars with `count (xx%)` on top of them, things like that

utkarsh-maheshwari:
Yes, it looks incomplete without the y axis. Don't we need it in the case of integers? 

Achieved this. 
image

through this:
```
import numpy as np
import arviz as az

data = np.random.choice(a=[False, True], size=(4, 100))
az.plot_posterior(data)
```

utkarsh-maheshwari:
@OriolAbril Can the dtype of values be other than f, i or b?

sethaxen:
> @OriolAbril Can the dtype of values be other than f, i or b?

That's a good question. In principle a PPL could produces samples of strings or objects, though all I am aware of currently would require first assigning integers to the strings/objects and then produce integer samplers. Do we anywhere in arviz assert that the dtype is amount these types?

> We may also want to update a bit more the apperance of the plot? @sethaxen? i.e. showing the y axis in the boolean case, or hiding the mean and annotating both bars with `count (xx%)` on top of them, things like that

Yes, I think it makes sense to label the bars with count and percent instead mean, and if this is done, an axis is not necessary.
I like the gap between the bins.

ahartikainen:
True/False is an edge case from category dtype with 2 categories.

If we have a flag for category, then we could do this kind of (bar) plot for multiple categories too.

utkarsh-maheshwari:
> If we have a flag for category, then we could do this kind of (bar) plot for multiple categories too.

So, does that mean 
if dtype == " f " -> kde
elif dtype == " i " -> hist
else -> bar

OriolAbril:
> So, does that mean
if dtype == " f " -> kde
elif dtype == " i " -> hist
else -> bar

Sorry about the delay, I think this is a good starting point, we can then improve/refactor the else case to handle the category case too, but before that we have to decide on an api, have this info in inferencedata or as an argument or both which will probably take some time.

utkarsh-maheshwari:
`Added lines #L332 - L336 were not covered by tests ` 
How to resolve this error?

ahartikainen:
We need to add test that has boolean data. (check current tests and see if you can add boolean data as input for the test for this plot)

utkarsh-maheshwari:
I think I need to create a function `test_plot_posterior_boolean` below https://github.com/arviz-devs/arviz/blob/4d9caca88a9a3b4043062609cb9a401e3f08804d/arviz/tests/base_tests/test_plots_matplotlib.py#L946.
to test with a boolean data. Right?

ahartikainen:
yes

utkarsh-maheshwari:
I think now there are some errors not related to my code changes.

arviz/plots/backends/matplotlib/posteriorplot.py
```diff
@@ -319,18 +319,21 @@ def format_axes():
...
+            ax.set_xlim(xmin - 0.5, xmax + 0.5)
         kwargs.setdefault("align", "left")
         kwargs.setdefault("color", "C0")
         ax.hist(values, bins=bins, alpha=0.35, **kwargs)
+    elif values.dtype.kind == "b":
+        if bins is None:
+            bins = "auto"
+        kwargs.setdefault("color", "C0")
+        ax.bar(["False", "True"], [(~values).sum(), values.sum()], alpha=0.35, **kwargs)
+        hdi_prob = "hide"
```
OriolAbril, line 336: we should add an else and an error to make sure it fails informatively if anyone ever passed an array with object dtype for example

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -951,6 +951,12 @@ def test_plot_posterior(models, kwargs):
         assert axes.shape
 
 
+def test_plot_posterior_boolean():
+    data = np.random.choice(a=[False, True], size=(4, 100))
+    axes = plot_posterior(data)
+    assert axes
```
OriolAbril, line 957: can you also add an assert with `get_xticklabels` or something like that to make sure they are not 0 and 1? We don't need to check every aspect of the plot, I propose this one because I think it's the easiest one and will catch breaking changes if we update the plot and change the labels for example

arviz/plots/backends/matplotlib/posteriorplot.py
utkarsh-maheshwari: Adding `raise TypeError("Values Must be float, Integer or boolean")` to the else.

arviz/tests/base_tests/test_plots_matplotlib.py
utkarsh-maheshwari: `axes.get_xticklabels()` returns 
`[Text(0, 0, 'False'), Text(1, 0, 'True')]` in boolean case

Not very sure how to tackle this one. It needs some attention

arviz/plots/backends/matplotlib/posteriorplot.py
```diff
@@ -319,18 +319,23 @@ def format_axes():
             rug=False,
             show=False,
         )
-    else:
+    elif values.dtype.kind == "i":
```
OriolAbril, line 322: the initial condition is `if kind=kde and dtype=f` so this condition is too restrictive, I think it should be `if dtype=i or (dtype=f and kind=hist)`

utkarsh-maheshwari: Yea right. I should have done it that way. Thanks for pointing.

arviz/tests/base_tests/test_plots_matplotlib.py
utkarsh-maheshwari: Maybe 
`assert axes.get_xticklabels()[0].get_text()=='False'` this will do the job?

arviz/plots/backends/matplotlib/posteriorplot.py
OriolAbril: test exists because we all do this which is also why writing tests is important everywhere

utkarsh-maheshwari: Done.
One more observation. Now, that there is an added kind of plot `bar` for boolean values. So maybe we should set 
`kind = 'bar'` somewhere and should accept `kind = 'bar'`.

utkarsh-maheshwari:
I have no idea why tests are failing. It works fine when I write the same test in notebook.
Is there a better way to debug tests?

OriolAbril:
> Is there a better way to debug tests?

Were tests passing locally? Whenever I get stuck on one or a couple tests I run them locally, and only those so I can iterate fast. pytest has a lot of features to select which subset of tests to run: https://docs.pytest.org/en/6.2.x/reference.html#command-line-flags. To begin with you have to call `pytest dir_or_file` where `dir_or_file` can be a path to a folder, a path to a file or a glob pattern, you can also use special syntax after that with colons to select which class/test function to run but I find that very complicated and always use the `-k` flag to filter tests that fulfill a condition, in your case it could be something like `-k "posterior and boolean"` to run only the specific test you added

ahartikainen (approved):

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -951,6 +951,14 @@ def test_plot_posterior(models, kwargs):
         assert axes.shape
 
 
+def test_plot_posterior_boolean():
+    data = np.random.choice(a=[False, True], size=(4, 100))
+    axes = plot_posterior(data)
+    plt.draw()
```
ahartikainen, line 957:
```suggestion
```

ahartikainen (changes requested):

arviz/tests/base_tests/test_plots_matplotlib.py
ahartikainen: I think this is not needed for tests (or is it?)

utkarsh-maheshwari: I have an answer to this which may look weird at first as it was to me.
I was running an example on jupyter notebook. 
Calling `axes.get_xticklabels()` in the same cell `plot()` is called returns empty list.
However, I get the desired output when run `axes.get_xticklabels()` in next cell. 
I found a solution [here](https://github.com/matplotlib/matplotlib/issues/6103/) which suggested to add `plt.draw()`

utkarsh-maheshwari:
@OriolAbril This is much better approach. Thanks!

arviz/tests/base_tests/test_plots_matplotlib.py
utkarsh-maheshwari: Answer to this question may look weird at first but
On notebook, assert axes.get_xticklabels() runs fine when run in next cell but returns empty list when run on the same cell containing `plt.plot()`.
Found a solution [here](https://github.com/matplotlib/matplotlib/issues/6103/)


(My earlier comment got duplicated. When I deleted one, both got deleted 🤦🏼 )

ahartikainen: Ok, then this can be removed (we don't run tests in notebooks)

utkarsh-maheshwari: Tests were failing when I removed this because then get_xticklabels() returns empty list.

Interactive users when want to run `axes.get_xticklabels()` can do this by running it in different cell

But non interactive users have to call plt.draw() before using `axes.get_xticklabels()` 

More clear explanation : 
> in interactive mode in ipython, a draw() is automatically triggered after each function that has a plotting action. In non-interactive environments, the draw action is executed only at the end--this saves a lot of run time.


But maybe, we can call plt.draw() inside our posterior plot function and remove it from here.

ahartikainen: Yes, tests are really run from command line 

    python -m pytest path/to/tests

So keeping them clean is ok.

arviz/plots/backends/matplotlib/posteriorplot.py
```diff
@@ -319,18 +319,23 @@ def format_axes():
...
         kwargs.setdefault("color", "C0")
         ax.hist(values, bins=bins, alpha=0.35, **kwargs)
+    elif values.dtype.kind == "b":
+        if bins is None:
+            bins = "auto"
+        kwargs.setdefault("color", "C0")
+        ax.bar(["False", "True"], [(~values).sum(), values.sum()], alpha=0.35, **kwargs)
+        hdi_prob = "hide"
+    else:
+        raise TypeError("Values must be float, integer or boolean")
```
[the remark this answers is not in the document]
OriolAbril: This could be tested using an array of strings I think, but I'm not sure it's worth it. @canyon289?

utkarsh-maheshwari:
Unable to understand why a codecov shows a lot of code becomes uncovered through these changes. It shows whole matplotlib posterior plot function is becomes uncovered!

OriolAbril:
> Unable to understand why a codecov shows a lot of code becomes uncovered through these changes. It shows whole matplotlib posterior plot function is becomes uncovered!

If the azure job fails, nothing is uploaded to codecov, thus if all jobs fail, the coverage is 0

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -951,6 +951,13 @@ def test_plot_posterior(models, kwargs):
         assert axes.shape
 
 
+def test_plot_posterior_boolean():
+    data = np.random.choice(a=[False, True], size=(4, 100))
+    axes = plot_posterior(data)
+    assert axes
+    assert axes.get_xticklabels()[0].get_text() != 0
```
utkarsh-maheshwari, line 958: I know this is not the best way to see if xticklabels are correct. It should be like `assert axes.get_xticklabels()[0].get_text() == 'False'. 
When I tried it locally, tests work well locally only if I provide `show=True` to the plot function call and shows an assertion error when show=False (default).
`AssertionError: assert ' ' == 'False'`
Azure jobs fails with same error even if `show=True` is pushed.

arviz/plots/backends/matplotlib/posteriorplot.py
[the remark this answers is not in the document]
utkarsh-maheshwari: If it's not worth, can it be left uncovered?

arviz/tests/base_tests/test_plots_matplotlib.py
OriolAbril: maybe it could be solved using some non-interactive/directly writing to file backend? I don't reallly know what is happening, but it would not be surprising that Azure had some extra checks preventing guis, pop ups or other interactive objects

utkarsh-maheshwari: Already tried using `plt.draw()` in the matplotlib backend file, didn't work! I guess we need to call `plt.draw()` in the test function before assert or maybe leave it like that if its an "okayish" way to test if not good.

arviz/plots/backends/matplotlib/posteriorplot.py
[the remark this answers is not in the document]
canyon289: Hey sorry for the super long delay. Ideally itd be tested if its not too hard. In this case it dooesnt look like itll be too bad to write a test to check for the exception so Id do it

ahartikainen (approved):

ahartikainen:
Thanks for the fix

PR #1710, opened 2021-05-28
fixed plot_kde to take labels with kwargs.

### LINKED ISSUE #1703: plot_posterior doesn' accept label as kwarg
I think this is related to how we pass kwargs downstream to plot_dist or plot_kde... will look into it at some point unless someone beats me to it. My hunch is that plot_dist and plot_kde have a `label` argument and pass it explicitly to matplotlib instead of doing `setdefault` so if a label is passed via kwargs too it becomes a duplicated argument and errors out

Fixing this would allow using plot_posterior to overlay multiple distributions while having a legend.

utkarsh-maheshwari:
## Description
This pr solves the conflict that arises when `label` is passed as a kwarg. It conflicted with the argument `label`. 
Added `set_default` to the required args. 
Added relevant kwargs in the test module.
Also, now user can use plot_posterior to use plot multiple distributions and have a legend.

closes #1703

## Checklist

### DIFF
--- a/arviz/plots/backends/matplotlib/kdeplot.py
+++ b/arviz/plots/backends/matplotlib/kdeplot.py
@@ -144,11 +144,13 @@ def plot_kde(
         else:
             fill_kwargs.setdefault("alpha", 0)
             if fill_kwargs.get("alpha") == 0:
-                ax.plot(x, density, label=label, **plot_kwargs)
+                label = plot_kwargs.setdefault("label", label)
+                ax.plot(x, density, **plot_kwargs)
                 fill_func(fill_x, fill_y, **fill_kwargs)
             else:
+                label = fill_kwargs.setdefault("label", label)
                 ax.plot(x, density, **plot_kwargs)
-                fill_func(fill_x, fill_y, label=label, **fill_kwargs)
+                fill_func(fill_x, fill_y, **fill_kwargs)
         if legend and label:
             ax.legend()
     else:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@
 * Renamed `sample` dim to `__sample__` when stacking `chain` and `draw` to avoid dimension collision ([1647](https://github.com/arviz-devs/arviz/pull/1647))
 * Removed the `circular` argument in `plot_dist` in favor of `is_circular` ([1681](https://github.com/arviz-devs/arviz/pull/1681))
 * Fix `legend` argument in `plot_separation` ([1701](https://github.com/arviz-devs/arviz/pull/1701))
+* Removed testing dependency on http download for radon dataset ([1717](https://github.com/arviz-devs/arviz/pull/1717))
+* Fixed plot_kde to take labels with kwargs. ([1710](https://github.com/arviz-devs/arviz/pull/1710))
 
 ### Deprecation
 * Deprecated `index_origin` and `order` arguments in `az.summary` ([1201](https://github.com/arviz-devs/arviz/pull/1201))
--- a/arviz/plots/backends/bokeh/kdeplot.py
+++ b/arviz/plots/backends/bokeh/kdeplot.py
@@ -147,6 +147,8 @@ def plot_kde(
                     patch = ax.patch(patch_y, patch_x, **fill_kwargs)
                 glyphs.append(patch)
 
+            if label is not None:
+                plot_kwargs.setdefault("legend_label", label)
             if not rotated:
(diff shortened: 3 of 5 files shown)


arviz/plots/backends/matplotlib/kdeplot.py
```diff
@@ -144,11 +144,13 @@ def plot_kde(
         else:
             fill_kwargs.setdefault("alpha", 0)
             if fill_kwargs.get("alpha") == 0:
-                ax.plot(x, density, label=label, **plot_kwargs)
+                label = plot_kwargs.setdefault("label", label)
+                ax.plot(x, density, **plot_kwargs)
                 fill_func(fill_x, fill_y, **fill_kwargs)
             else:
+                label = fill_kwargs.setdefault("label", label)
```
OriolAbril, line 151: I think setdefault doesn't return anything, so it's output should not be stored anywhere.

OriolAbril (commented):
Two extra general comments. The bokeh version probably needs to be updated too and the issue being addressed should be only added to the PR description, not it's title. In the title we only see it's number which is generally no help at all because we don't remember issues by number, in the PR description however, github automatically generates a link to the issue and if you use [some specific keywords](https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords#about-issue-references) the issue will also be closed automatically when the PR is merged, making our work easier (we often don't remember to go and close the issue if we have to do so manually).

arviz/plots/backends/matplotlib/kdeplot.py
utkarsh-maheshwari: I referred [this](https://docs.python.org/3/library/stdtypes.html#dict.setdefault) link. I have used it stored its value because `label` is used in line 154.

utkarsh-maheshwari:
Sure, I'll keep this in mind. Thanks for pointing it out. 
I'll try to do it for bokeh as well.

utkarsh-maheshwari:
Not sure if it is to be done for contour_kwargs or fill_kwargs.
I guess since there is a loop to plot contour, `label` shouldn't be added there. Same applies to fill_kwargs.

arviz/plots/backends/matplotlib/kdeplot.py
OriolAbril: true, good catch

utkarsh-maheshwari:
Unlike matplotlib, passing `legend_label=None` generates an error in case of bokeh.

OriolAbril (commented):
changes so far are great and are actually fixing a bug we did not know we had, bokeh plot_kde currently ignores the label argument. 

It would be great to add `label` as kwarg in some of the cases parametrized at https://github.com/arviz-devs/arviz/blob/main/arviz/tests/base_tests/test_plots_matplotlib.py#L920 (adding an extra kwarg to one of the existing cases is enough it is independent of other arguments, no need to add a new case). Same for bokeh, I think the situation should be equivalent

I also noticed you removed the issue reference from the title but not added it to the description, so I have added it myself.

utkarsh-maheshwari:
Errors are because `label` is still unused in bokeh plot_kde. When passed `label` in bokeh plot_kde, it goes into `plot_kwargs`.

OriolAbril (commented):
> Errors are because label is still unused in bokeh plot_kde. When passed label in bokeh plot_kde, it goes into plot_kwargs.

`kwargs` are passed "as is" to the backend, so it's responsibility of the user to use the right ones. For bokeh it should not be label but `legend_label` I believe

utkarsh-maheshwari:
https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=4434&view=logs&j=5009c7ae-01dd-59da-6106-f3ac988cc395&t=0d1d07e1-2d84-5a43-33b6-ac11fbcc1860&l=94

Are these errors related to my code changes??

OriolAbril:
Do you want to add it to the changelog?

utkarsh-maheshwari:
Done!

OriolAbril:
xarray has modified how selection works internally so now what used to be a ValueError is a KeyError. Not completely sure what would be the best way forward but this is what's breaking the test

utkarsh-maheshwari:
Okayy. Wondering if changing the `ValueError` to https://github.com/arviz-devs/arviz/blob/33ec4e1582ab7b9c80841140fc3dcf004532db05/arviz/tests/base_tests/test_plot_utils.py#L145 to `KeyError` would work 🤔

PR #1713, opened 2021-06-01
Add init strategy to generate sample for ImproperUniform in NumPyro

### LINKED ISSUE #1704: NumPyroConverter makes too many assumptions about the model
The following code is not always runnable:

https://github.com/arviz-devs/arviz/blob/7ebedd2d1494f07513bc7b9ea21962bc11399a94/arviz/data/io_numpyro.py#L126-L133

This will fail for a model involving `ImproperUniform`, as sampling for it is not defined.

Would it be possible to add a keyword argument that specifies whether this automatic extraction of observations should be attempt or not, e.g. `extract_observations` or something? At the moment it's stopping me from being able to use arviz with my numpyro-model (unless I just comment out the lines above) :confused:

fehiepsi:
## Description
Resolves #1704. Currently, we run the model to inspect observed variables. However, numpyro does not allow to sample from `ImproperUniform` distribution. As a workaround, I add an init_strategy method to draw a sample from transform_to_support(Uniform(-2, 2)).

## Checklist

### DIFF
--- a/arviz/data/io_numpyro.py
+++ b/arviz/data/io_numpyro.py
@@ -124,7 +124,9 @@ class NumPyroConverter:
 
         observations = {}
         if self.model is not None:
-            seeded_model = numpyro.handlers.seed(self.model, jax.random.PRNGKey(0))
+            seeded_model = numpyro.handlers.substitute(
+                numpyro.handlers.seed(self.model, jax.random.PRNGKey(0)),
+                substitute_fn=numpyro.infer.init_to_sample)
             trace = numpyro.handlers.trace(seeded_model).get_trace(*self._args, **self._kwargs)
             observations = {
                 name: site["value"]


OriolAbril (approved):

arviz/data/io_numpyro.py
```diff
@@ -124,7 +124,10 @@ def arbitrary_element(dct):
 
         observations = {}
         if self.model is not None:
-            seeded_model = numpyro.handlers.seed(self.model, jax.random.PRNGKey(0))
+            seeded_model = numpyro.handlers.substitute(
+                numpyro.handlers.seed(self.model, jax.random.PRNGKey(0)),
```
canyon289, line 128: Can you add a comment explaining why this is here for whoever reads it next?

PR #1718, opened 2021-06-07
extend 'right' join bug fix

mjhajharia:
Bug: 

idata.extend(another_idata, join='right') returns two copies of every group  - i.e. posterior, prior 

image

Bug cause:
append should only happen if group not in self
image

Possible fix - (thanks @OriolAbril )
image

Note: extend(join='left') works just fine.

Feel free to suggest better fixes or tests or anything needed

### DIFF
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -1358,9 +1358,15 @@ class InferenceData(Mapping[str, xr.Dataset]):
             if dataset:
                 setattr(self, group, dataset)
                 if group.startswith(WARMUP_TAG):
-                    supported_order = [key for key in SUPPORTED_GROUPS_ALL if key in self._groups_warmup]
+                    supported_order = [
+                        key for key in SUPPORTED_GROUPS_ALL if key in self._groups_warmup
+                    ]
                     if (supported_order == self._groups_warmup) and (group in SUPPORTED_GROUPS_ALL):
-                        group_order = [key for key in SUPPORTED_GROUPS_ALL if key in self._groups_warmup + [group]]
+                        group_order = [
+                            key
+                            for key in SUPPORTED_GROUPS_ALL
+                            if key in self._groups_warmup + [group]
+                        ]
                         group_idx = group_order.index(group)
                         self._groups_warmup.insert(group_idx, group)
                     else:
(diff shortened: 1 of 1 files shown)


OriolAbril:
Regarding tests, I think this one https://github.com/arviz-devs/arviz/blob/main/arviz/tests/base_tests/test_data.py#L782 should be modified to check that not only the attributes are right but the list stored in `_groups` is too.

arviz/data/inference_data.py
```diff
@@ -1394,10 +1394,11 @@ def extend(self, other, join="left"):
                 )
             dataset = getattr(other, group)
             setattr(self, group, dataset)
-            if group.startswith(WARMUP_TAG):
-                self._groups_warmup.append(group)
-            else:
-                self._groups.append(group)
+            if not hasattr(self, group):
```
OriolAbril, line 1397: The current condition will always evaluate to false, the attribute is set in the previous line. We should check for it's presence in the `_groups` or `_groups_warmup` list with something like:

```
if group not in self._groups:
```
in which case it's probably easier to combine with the already present ifs

arviz/data/inference_data.py
```diff
@@ -1394,10 +1394,11 @@ def extend(self, other, join="left"):
                 )
             dataset = getattr(other, group)
             setattr(self, group, dataset)
-            if group.startswith(WARMUP_TAG):
-                self._groups_warmup.append(group)
-            else:
-                self._groups.append(group)
+            if group not in self._groups and self._groups_warmup:
```
ahartikainen, line 1397: This line tests if both of these are true

    bool(group not in self._groups)

    bool(self._group_warmup)

Is this what is wanted?

mjhajharia: i'm not very sure(looking at this PR after a while so lost some context, feel free to correct me) but i think we were trying to ensure no group is repeated twice when we do extend with 'right'. is the group_warmup list also used finally in the inferencedata object returned?

mjhajharia:
> Regarding tests, I think this one https://github.com/arviz-devs/arviz/blob/main/arviz/tests/base_tests/test_data.py#L782 should be modified to check that not only the attributes are right but the list stored in `_groups` is too.

i'm sorry, i'm not sure i understand

OriolAbril:
the code file probably changed and the line is no longer right. This test https://github.com/arviz-devs/arviz/blob/main/arviz/tests/base_tests/test_data.py#L791-L808 checks that the extended inferencedata has the right attributes with `hasattr` and that worked already. But having the right attributes is not enough for `extend` to have worked properly. We have to check that the `_groups` list is also correctly updated and matches the actual attributes in the inferencedata. The bug was due to `_groups` being wrong, so while the result inferencedata was "right" and had all the extended attributes, `_groups` (which is what is used for printing) was not, so it looked like there were wrong and/or repeated attributes.

ahartikainen:
I added some extra logic to keep the supported order for the groups (if groups are still in that order)

ahartikainen:
Hmm not sure.

I can add tests to my fix pr

ahartikainen:
If we do handle the order while printing, then I will remove that extra logic

OriolAbril:
> If we do handle the order while printing, then I will remove that extra logic

I remembered wrong, it's in __init__: https://github.com/arviz-devs/arviz/blob/main/arviz/data/inference_data.py#L142

ahartikainen:
Ok, I will leave in place. Thanks for checking it out

PR #1720, opened 2021-06-08
Fix smooth_data axis loo_pit

aloctavodia:
This PR computed loo-pit using a numpy vectorized version, instead of the wrapped _loo_pit function. Not necessary in favor of keeping this version, just pointing that the current version may be failing for some corner cases. Below, blue current version, black numpy version, gray `az.plot_bpv()`

For these two there is a large difference (both are logistic regressions)
logistic
test_loo_pit

For the rest the blue curve is hidden behind the black one (current and numpy versions agree)
miss_normal. 
radon
non_centered_eight

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -1569,6 +1569,8 @@ def loo_pit(idata=None, *, y=None, y_hat=None, log_weights=None):
                 "all 3 y, y_hat and log_weights must be array or DataArray when idata is None "
                 f"but they are of types {[type(arg) for arg in (y, y_hat, log_weights)]}"
             )
+        if isinstance(y, xr.DataArray):
+            y = y.values
 
     else:
         if y_hat is None and isinstance(y, str):
@@ -1578,6 +1580,8 @@ def loo_pit(idata=None, *, y=None, y_hat=None, log_weights=None):
         if isinstance(y, str):
             y_str = y
             y = idata.observed_data[y].values
+        elif isinstance(y, xr.DataArray):
+            y = y.values
         elif not isinstance(y, (np.ndarray, xr.DataArray)):
             raise ValueError(f"y must be of types array, DataArray or str, not {type(y)}")
         if isinstance(y_hat, str):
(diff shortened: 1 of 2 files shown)


arviz/stats/stats.py
```diff
@@ -1627,34 +1631,17 @@ def loo_pit(idata=None, *, y=None, y_hat=None, log_weights=None):
...
-        _loo_pit,
-        y,
-        y_hat,
-        log_weights,
-        ufunc_kwargs=ufunc_kwargs,
-        **kwargs,
-    )
-
+    mlw = np.max(log_weights)
+    pits = np.exp(mlw + np.log(np.sum(np.exp(log_weights.T - mlw) * (y_hat <= y), axis=0)))
```
junpenglao, line 1641: `np.log(np.sum(np.exp(...)))` isnt this the logsumexp?

junpenglao:
All these transpose is making me a bit nervous - seems it is possible to modify the `_loo_pit` to include `mlw`, why would that not fix the edge case?

arviz/stats/stats.py
aloctavodia: yes

aloctavodia:
Also nervous about the transposes :-)  `_logsumexp` used in `_loo_pit` is an stable version of logsumexp including the mlw term

junpenglao:
I think the root cause is in `smooth_data`:
https://github.com/arviz-devs/arviz/blob/57aeded2fff88a646939fc36e5b2a86a0cf718b0/arviz/stats/stats_utils.py#L570-L572

where smoothing should be done in the data dim, but the usage in `loo_pit` is incorrect.

In `bpvplot`, we have:

https://github.com/arviz-devs/arviz/blob/57aeded2fff88a646939fc36e5b2a86a0cf718b0/arviz/plots/backends/bokeh/bpvplot.py#L90-L94

Data dimension is at the end, so the current behavior of `smooth_data` makes sense (the smoothing is at the axis=1)

In `loo_pit` however the Data dimension is at axis=0, this is why your PR works by transposing. I think we should modify the `smooth_data` instead, and use the same reshaping as `bpvplot`

arviz/stats/stats_utils.py
```diff
@@ -561,14 +561,14 @@ def _circular_standard_deviation(samples, high=2 * np.pi, low=0, skipna=False, a
     return ((high - low) / 2.0 / np.pi) * np.sqrt(-2 * np.log(r_r))
 
 
-def smooth_data(obs_vals, pp_vals):
+def smooth_data(obs_vals, pp_vals, axis=1):
```
junpenglao, line 564:
```suggestion
def smooth_data(obs_vals, pp_vals, axis=-1):

```

arviz/stats/stats.py
```diff
@@ -1635,7 +1635,7 @@ def loo_pit(idata=None, *, y=None, y_hat=None, log_weights=None):
     ufunc_kwargs = {"n_dims": 1}
 
     if y.dtype.kind == "i" or y_hat.dtype.kind == "i":
-        y, y_hat = smooth_data(y, y_hat)
+        y, y_hat = smooth_data(y, y_hat, axis=0)
```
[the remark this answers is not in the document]
junpenglao: I think being a bit more explicit and always reshape so that the observation is only 1d might be good, like what we are doing in `bpvplot`:

https://github.com/arviz-devs/arviz/blob/57aeded2fff88a646939fc36e5b2a86a0cf718b0/arviz/plots/backends/bokeh/bpvplot.py#L90-L94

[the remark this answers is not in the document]
OriolAbril: The approach in bpvplot has two main drawbacks, which is what I was trying to point out above, and they are still relevant there, but quite negligible, imo here they are more important. 

First one is loosing all the shape and label info when doing that. It's not clear if and/or how pointwise or group loo pits can be useful but it seems like it could be useful so I would rather not loose it for good in all cases (float data already works with 4d observations and keeps shape and labels!), i.e. hierarchical model with 3 groups, each with multiple observations, 1st group has positive bias, 2nd group has good loo pit and 3rd group has negative bias so overall loo pit looks good but group level ones do not. 

Second one is that by doing the computation via xarray, it can easily be done under the hood by numpy or by dask, so everything will still work if we have many observations so that posterior predictive array doesn't fit in memory anymore.

agustinaarroyuelo:
@aloctavodia what would you like to do next with this PR? If there are any new developments to this discussion please share.

PR #1722, opened 2021-06-11
change error handling in get_coords

OriolAbril:
## Description
I think this will fix CI for xarray preview. Also sets the path to end up delegating errors to
xarray instead of catching them to improve the error message. From a couple cases, it looks
like errors in xarray have now improved and are much more informative.

## Checklist

- [ ] Does the PR follow [official](https://github.com/arviz-devs/arviz/blob/main/CONTRIBUTING.md#pull-request-checklist)
      PR format?
- [ ] Does the PR include new or updated tests to prevent issue recurrence (using [pytest fixture pattern](
- [ ] Is the code style correct (follows pylint and black guidelines)?
- [ ] Is the fix listed in the [Maintenance and fixes](https://github.com/arviz-devs/arviz/blob/main/CHANGELOG.md#maintenance-and-fixes)
      section of the changelog?


OriolAbril:
IIRC, we already merged an alternative PR to fix that a while ago so I'll close this.

Issue #1188, opened 2020-05-18
Progressively update docs references


OriolAbril:
Add links to related functions in docstrings (applies to both links to internal and external functions), also add See also sections where relevant. Follow up of #1184. Places where references are of particular importance are: `kwargs` description to ease consulting of the documentation relevant to the kwargs so users know what keys are supported; and functions that take as input the output of a previous one

For details on how to do so, read the [hyperlinks section](https://arviz-devs.github.io/arviz/contributing/developer_guide.html#hyperlinks) in the developer guide. Moreover, the docstring of `hdi` and `plot_hdi` are good examples.

## External references
An example of external reference can be found at `plot_hdi` page (see how [rendered page](https://arviz-devs.github.io/arviz/api/generated/arviz.plot_hdi.html) links to scipy, matplotlib and bokeh functions from [this source](https://github.com/arviz-devs/arviz/blob/main/arviz/plots/hdiplot.py#L35-L78)).

## Internal references
The same `plot_hdi` also has an internal reference to `hdi`. The links are the same as in the previous section. 

## See also section
An example of See also section can be found at [`compare`](https://arviz-devs.github.io/arviz/api/generated/arviz.compare.html) page

## Status
This should help tracking what has been done. We could open an issue for every single function but that would probably get out of hand fast and make keeping track harder. Functions where work is halfway will be listed below without checkmark, updated functions will be checked and all missing functions are free and unassigned. 

If you are interested in working on this, check the list to see what functions are available and comment which (1-4 functions per PR seem like a reasonable number, don't try to work on too many at the same time) would you like to tackle. 

* [ ] plot_hdi: nearly finished, only external references for `backend_kwargs` are missing.
* [x] hdi (maybe a reference would further improve the docstring)
* [x] InferenceData.sel/InferenceData.isel: should have all references to xr.Dataset.sel/isel with hyperlink and also a see also section linking them to each other and to the respective xarray versions
* [ ] Other InferenceData methods: intersphinx is present and linking to relevant places, a see also section would be appreciated. This one is **not** begginer friendly as it should be added programatically by `_extend_...` cc @percygautam 
* [ ] compare: extend see also section, maybe use references too? plot_separation is a good example on how to format references in docstrings
* all other functions...
   * [ ] plot_xyz
   * [ ] stats xyz
   * [ ] diagnostics xyz
   * [ ] ...

fairka:
Hey can I have this project?

OriolAbril:
[...]

PR #1727, opened 2021-06-15
[WIP] Initial prototype of plot_lm

utkarsh-maheshwari:
## Description
Here is my initial prototype for linear regression plot. I am checking if input data is correct in the the user API function. Backend function is quite small, only does the plotting part.
Example Notebook : https://gist.github.com/utkarsh-maheshwari/8d4cd2fd84c763bf85291c3f0881d588

High level design:
We accept 5 args

1. y_model : (Necessary) linear relationship in the form of `y = m1*x1 + m2*x2 + c`
2. idata : (Necessary if any of other args are string) should contain all the necessary variable names in any of the groups.
3. x_data : (Necessary because we cannot identify the x variable name from y_model) seq/str
4. y_data :  (Necessary if y_model is Sequence. We can identify the name from y_model) seq/str
5. y_ppc : (Necessary is y_model is sequence. Assumed to be same as y_data if absent) seq/str 

Function call:

`plot_lm(y_model = "kid_score ~ Intercept + slope * mom_iq",  idata = idata, x_data = "mom_iq")
`

Output : 
image

Shows uncertainty in y_data in orange, uncertainty in y_model in black, OLS regression line in yellow.
Should regression lines be plotted instead of filling the area?

-----------------------------------------------------------------------------------------------------

If y_model is sequence, we're good with plotting it. 
If y_model is string like `y ~ slope * x_name + Intercept`, it should be converted to sequence by 
- correctly identifying the variable names present in the string. 
- look for variable names in inferenceData groups.
- Compute y_model term by term to form desired sequence.

To begin with, we can do it like : 

```
rhs, lhs = y_model.split('~')
terms = rhs.split('+') : gives terms of the expression. 
var_name = terms.split('*') : gives variable names to be searched in data

```

Iterate over terms
calculate value of each term
add it to the y_model_values

It works well with parsing GLM because I am keeping track of functions in function stack.
Any better way to parse `y_model` to extract the variable names from the string?

Love to know suggestions/reccomendations.

## Checklist


arviz/plots/lmplot.py
```diff
@@ -0,0 +1,163 @@
...
+                var_values = getattr(data, var_group)[var_name]
+                while funcs:
+                    var_values = getattr(np, funcs.pop())(var_values)
+                temp = temp*var_values
+            y_model_values = y_model_values + temp
+
+    if y_data is None:
+        raise ValueError("y_data must be present")
+
+    x_values = x
```
canyon289, line 98: Renaming like this is odd. Is there a reason it needs to be done?

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,163 @@
...
+        if data is None:
+            raise ValueError("data must be present")
+
+        if x_group is None:
+            for group in groups:
+                item = getattr(data, group)
+                if x in item and x_group is None:
+                    x_group = group
+                elif x in item:
+                    print(
```
canyon289, line 109: Use warning module instead of print. Same for all print statements

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,163 @@
...
+            for group in groups:
+                item = getattr(data, group)
+                if y_data in item and y_data_group is None:
+                    y_data_group = group
+                elif y_data in item:
+                    print("Warning, duplicate variable names for y_data, "
+                    "using variable from group {}".format(y_data_group))
+        y_data_values = getattr(data, y_data_group)[y_data]
+
+    y_ppc_values = y_ppc
```
canyon289, line 131: Same comment here, why rename the variable?

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,163 @@
...
+            y_funcs = y_funcs + [lhs[:idx]]
+            y_var_name = y_var_name[idx+1:-1]
+
+        if y_data is None:
+            y_data = y_var_name
+
+        if y_ppc is None:
+            y_ppc = y_var_name
+
+        y_model_values=0
```
canyon289, line 71: I would consider splitting this into another function and testing it independently

canyon289 (dismissed):
At a high level this looks good, Please start adding tests for functionality so we can more fully shake out bugs

arviz/plots/backends/matplotlib/lmplot.py
```diff
@@ -0,0 +1,33 @@
+import matplotlib.pyplot as plt
+import numpy as np
+
+from ...hdiplot import plot_hdi
+
+def plot_lm(
+    y_model_values,
```
canyon289, line 7: Is there a way we can use inference data as an input and specify the x_values and such in reference to inf data?

canyon289:
> ## Description
> Here is my initial prototype for linear regression plot. I am checking if input data is correct in the the user API function. Backend function is quite small, only does the plotting part.

> 1. y_model : (Necessary) linear relationship in the form of `y = m1*x1 + m2*x2 + c`
> 2. idata : (Necessary if any of other args are string) should contain all the necessary variable names in any of the groups.
> 3. x_data : (Necessary because we cannot identify the x variable name from y_model) seq/str
> 4. y_data :  (Necessary if y_model is Sequence. We can identify the name from y_model) seq/str
> 5. y_ppc : (Necessary is y_model is sequence. Assumed to be same as y_data if absent) seq/str


The high level design api doesnt seem to match whats in the code. 

For the model specification instead of parsing it ourselves should we use patsy instead, and perhaps add it an optional dependency? @ahartikainen for a double check here

ahartikainen:
I'm not 100% we should parse formulas.

I think everything should be in the InferenceData and we call it.

For cases where we want to calculate the model mean, user given function is probably a safer option.

OriolAbril:
> I'm not 100% we should parse formulas.

I'm 90% sure we don't want to parse formulas at all, 100% sure we don't want to create any kind of parser (not only because it would be a lot of work and could even eat up the whole GSoC or even more time but also because we'd then have to maintain it!). 

AFAIK all formula parsing libraries we could depend on use dataframes as input which we don't have so we'd need to convert the data to a correctly formatted dataframe for things to work, and for formula parsing to be really useful we would not only need to parse the formula but also consider link functions for example. I think having a lower level thing that takes samples/functions is preferrable, and it should be easy for libraries like bambi to build on top of that and offer a formula based api to users.

> For cases where we want to calculate the model mean, user given function is probably a safer option.

You can take a look at the [refitting notebooks that use xarray](https://arviz-devs.github.io/arviz/user_guide/numpyro_refitting_xr_lik.html) for examples on a very similar computation.

arviz/plots/lmplot.py
utkarsh-maheshwari: Basically x could be a string or sequence. So, I was renaming it to conserve the string name before converting it to DataArray.

arviz/plots/lmplot.py
utkarsh-maheshwari: Sure!

arviz/plots/backends/matplotlib/lmplot.py
utkarsh-maheshwari: Then I guess y_model need to be stored somewhere in idata.

utkarsh-maheshwari:
> The high level design api doesnt seem to match whats in the code but i generally like the way its stated here. It would be nice if the idata was the only "data" argument, and the other arguments are specifiers for sets of data in the idata object

Do you mean that we should accept `x_data`, `y_data` and `y_ppc` only as string as the names of DataArray present in idata, and not as sequence? Yea, that would be better.

utkarsh-maheshwari:
> I think everything should be in the InferenceData and we call it.

Even the y_model? 
By this, what comes in my mind is: 

For example `y ~ mx + c` is the modeling equation. User will first find the posterior values of `m` and `c`. Then user will do something like:
`idata.posterior_predictive["y_model"] = idata.posterior["m"] * idata.constant_data["x"] + idata.posterior["c"]`
And pass it to the function.?

canyon289:
Just to add my 2 cents, if we dont need to parse formulas I dont think we should either. Itll add an extra dependency to ArviZ, and in the modeling workflow it will be very annoying to restate a complicated model in a formula when you've already done it in a PPL. Im not sure if we can get away without it but glad @OriolAbril and @ahartikainen  chimed in saying we probably dont need it!

utkarsh-maheshwari:
Sure then! Assuming the y_model is present in inferenceData is looking a good option.

utkarsh-maheshwari:
> I think having a lower level thing that takes samples/functions is preferrable, and it should be easy for libraries like bambi to build on top of that and offer a formula based api to users.

Not completely sure what this means. Are you talking about, for now, taking all of the necessary values as references to precomputed DataArrays, stored in idata. And later, extending it to take an equation string and let bambi handle the equation.

utkarsh-maheshwari:
As suggested by everyone, what I have come up with is very simple than what I expected.
HIgh level : 
```
plot_lm(
    idata : az.InferenceData object
    x : str
    y : str
    y_model : str
    y_ppc : str
    x_group : str
    y_data_group : str
    y_model_group : str
    y_ppc_group : str
    num_ppc_samples : int
)
```

Function call : 
```
plot_lm_userapi(
    idata = idata, 
    x = "mom_iq", 
    y = "kid_score", 
    y_model = "y_model", 
    y_ppc = "kid_score", 
    x_group = "constant_data", 
    y_group = "observed_data", 
    y_model_group = "posterior", 
    y_ppc_group = "posterior_predictive"
)
```

We can then move forward to reduce input arguments by making certain assumptions(specially for the `group` arguments) and accepting inputs as sequence too!

I agree on the comments made here and on slack regarding parsing the formula and totally convinced that making a parser is not a good idea.

canyon289:
Looking good! Youre gonna get so sick of me saying this. Cant wait to see the function implementation and some tests so we can try it out :D

utkarsh-maheshwari:
I have added a basic test. Not 100% sure if it's the right approach. Actually, there was no pre-built linear regression model so created one in the test function itself. Should it go in the helper?

canyon289:
> I have added a basic test. Not 100% sure if it's the right approach. Actually, there was no pre-built linear regression model so created one in the test function itself. Should it go in the helper?

Looks like a good start to the test. Your instinct is right I would move the "model" to helpers. I also would suggest instead of creating groups directly to use the inferencedata interface so if that changes itll be reflected in this test

https://github.com/arviz-devs/arviz/blob/main/arviz/tests/helpers.py#L22

OriolAbril:
I would instead propose:

```

**Comment by utkarsh-maheshwari** [comment]
> x : str or array_like, optional
         If str, it's a variable name from constant_data group so we get a dataarray. 
         Therefore it should not be much of a change to also take directly dataarrays or arrays. 
         It is optional because if the x values are the coordinate values from y I think the best is to 
         pass y only and no x to indicate that x should be the coordinates in y

I really like this approach. Thanks for putting this.


> y_model : str or array_like, optional


If optional, would we be plotting just x and y?


> Regarding the group arguments, I strongly feel we should _not_ have them

So, then we would assume variables to be present in a certain group. If any of the variable is not found in that particular group, we should not search in other groups and return immediately with an error.? or should we search in other groups too?
(for example should we search for x only in canstant_data? )

**Comment by OriolAbril** [comment]
> If optional, would we be plotting just x and y?

x, y and y_ppc (which we should probably call `y_hat` instead to keep the same naming convention from plot_ppc and plot_loo_pit

> So, then we would assume variables to be present in a certain group

Yes, I think it will make the code much simpler and easier to maintain with little to no drawback. You have the x in observed_data instead of constant_data? use `plot_lm(x=idata.observed_data["x_var_name"], ...)`

**Comment by utkarsh-maheshwari** [comment]
> What information is that part supposed to transmit?

According to me, It depicts the uncertainty of mean line just like spagetti plots do, but by plotting credible interval. We could do it like following as well.

About y_ppc we can plot posterior_predictive samples in different colour than observed values. 

image

**Comment by canyon289** [comment]
> > What information is that part supposed to transmit?
> 
> According to me, It depicts the uncertainty of mean line just like spagetti plots do, but instead by plotting credible interval. We could do it like following as well.
> 
> About y_ppc we can plot posterior_predictive samples in different colour than observed values.
> 
> image

It is possible to plot the uncertainty three ways using an arguments? Two of the ways you've shown, either an hdi band, or samples from the posterior predictive.  The third idea is a band but where the opacity changes as a function of the HDI. this last one dont implement in this PR just throwing the idea out there.


I can see different situations where a user would want the first two, so being able to switch, or plot both together, with an argument would be nice. The third Im less sure about but what to suggest it now while were in the early PR phase

File: arviz/plots/backends/matplotlib/lmplot.py [hunk 6]
```diff
@@ -0,0 +1,44 @@
+"""Matplotlib plot linear regression figure."""
+import matplotlib.pyplot as plt
+import numpy as np
+
+from ...hdiplot import plot_hdi
+
+
+def plot_lm(x, y, y_model, y_ppc, num_pp_samples, kind_pp, kind_model):
+    """Matplotlib Linear Regression."""
+    _, ax = plt.subplots(1, 1, figsize=(6, 4), dpi=100)
```
ahartikainen, line 10: This could call the function in plot_utils. I think other plots use that func too.

arviz/plots/backends/matplotlib/lmplot.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # plot data
+    ax.plot(x, y, marker=".", color="C3", lw=0, zorder=10)
```
ahartikainen, line 13: At some point these could be used with kwargs (with some setdefault magic)

arviz/plots/backends/matplotlib/lmplot.py
```diff
@@ -0,0 +1,44 @@
...
+
+    if y_ppc is not None:
+        arr = list(range(y_ppc.draw.size * y_ppc.chain.size))
```
ahartikainen, line 16:
```suggestion
        arr = np.arange(y_ppc.draw.size * y_ppc.chain.size)
```

arviz/plots/backends/matplotlib/lmplot.py
```diff
@@ -0,0 +1,44 @@
...
+        if kind_pp == "samples":
+            y_ppc_ = y_ppc.stack(sample=("chain", "draw"))[..., slicer]
+            for i in range(num_pp_samples):
+                ax.plot(x, y_ppc_[..., i], marker=".", lw=0, alpha=0.1, color="C1")
+        else:
+            plot_hdi(x, y_ppc, color="C1")
+
+    # plot uncertainty in mean line
+    if y_model is not None:
+        arr = list(range(y_model.draw.size * y_ppc.chain.size))
```
ahartikainen, line 28:
```suggestion
        arr = np.arange(y_model.draw.size * y_ppc.chain.size)
```

arviz/plots/backends/matplotlib/lmplot.py
```diff
@@ -0,0 +1,44 @@
...
+            for i in range(num_pp_samples):
+                ax.plot(x, y_model_[..., i], color="k", lw=0.5, alpha=0.5, zorder=9)
+        else:
+            plot_hdi(x, y_model, color="k", fill_kwargs={"alpha": 0.5, "lw": 0.5, "zorder": 9})
+
+        # plot mean line
+        y_model_mean_ = y_model.stack(sample=("chain", "draw"))[..., slicer].mean("sample")
+        ax.plot(x, y_model_mean_, color="y", lw=0.8, zorder=11)
+
+    for spine in ax.spines.values():
```
ahartikainen, line 41: I think this also needs some thinking when we want to do this and when not

arviz/plots/backends/matplotlib/lmplot.py
```diff
@@ -0,0 +1,44 @@
...
+        spine.set_visible(False)
+    ax.grid(True)
```
ahartikainen, line 43: This should also be some kwargs

ahartikainen (commented):
Added couple of comments

arviz/plots/backends/matplotlib/lmplot.py
utkarsh-maheshwari: I have done it in my local branch. I will soon push the changes

arviz/plots/backends/matplotlib/lmplot.py
utkarsh-maheshwari: Yes , I have just taken it from from your function. I need to think about it because I have less idea where is it useful.

arviz/plots/backends/matplotlib/lmplot.py
utkarsh-maheshwari: Sure, I have made the change in local branch. Actually, currently I am working to add argument `plot_dim`. I will push the changes as soon as I finish it

utkarsh-maheshwari:
> plot_dim/x_dim : str, optional
>         The dimension name that contains the x/y values. i.e. if y_ppc has dims (chain, draw, group, time) 
>         and we set plot_dim=time, then we'd generate n_group subplots, one per value of group coordinate. 
>         Note that we might have multiple covariates (x could be a list of str) 
>         maybe even multiple y instead of multidimensional y? so things could bet a bit tricky. 
>         This argument is required unless x/y are 1d and y_model/y_ppc are 3d 
>         (or 2d if dataarrays with `__sample__, plot_dim` dimensions? not sure if we should support this case, 
>         I think it could be useful for cases where generating pp samples is hard so users can use extract function and 
>         generate pp samples only for the small random subset).

@OriolAbril Can you share a model to test this argument?

arviz/plots/backends/matplotlib/lmplot.py
utkarsh-maheshwari: Different kwargs for observed data, PPC and model?

utkarsh-maheshwari:
I have added some tests. Function is now working fine for multidimensional y as well as x. Now some labelling and touch-ups are required so that graph looks nice. It gets complicated at some places in user API, which I think need some review because there may be better ways to process the inputs. Thank you for being so helpful so far. 😄

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,274 @@
...
+        Optional only if y is Sequence
+    x : str or tuple of strings or DataArray or array-like, Optional
+        If str or tuple, variable name from constant_data
+        If ndarray, could be 1D, or 2D for multiple plots
+        If none, coords name of y (y should be DataArray).
+    y_model : str or Sequence, Optional
+        If str, varibale name from posterior.
+        Its dimensions should be same as y plus added chains and draws.
+    y_ppc : str, Optional
+        If str, varibale name from posterior_predictive.
```
canyon289, line 54:
```suggestion
        If str, variable name from posterior_predictive.
```

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,274 @@
...
+        Its dimensions should be same as y plus added chains and draws.
+    num_pp_samples : int, Optional, Default 50
+    kind_pp : ["samples", "hdi"], Optional, Default "samples"
+    kind_model : ["lines", "hdi"], Optional, Default "lines"
+    plot_dim : str, Optional
+        Necessary if y is multidimentional.
```
canyon289, line 60:
```suggestion
        Necessary if y is multidimensional.
```

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,274 @@
...
+    backend : str, optional
```
canyon289, line 61:
```suggestion
    backend : str, Optional
```

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,274 @@
...
+        Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
+    y_kwargs : Passed to plot() in matplotlib and circle() in bokeh
+    y_ppc_plot_kwargs : Passed to plot() in matplotlib and circle() in bokeh
```
canyon289, line 64: Mark these as optional

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,274 @@
...
+    grid : bool, optional
+        Add grid to figure. By default True.
+
+
+    Returns
+    -------
+    axes
+    """
+    if kind_pp not in ("samples", "hdi"):
+        raise ValueError("kind_ppc could be either samples or hdi")
```
canyon289, line 90:
```suggestion
        raise ValueError("kind_ppc should be either samples or hdi")
```

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,274 @@
...
+
+    if kind_model not in ("lines", "hdi"):
+        raise ValueError("kind_model could be either lines or hdi")
```
canyon289, line 93:
```suggestion
        raise ValueError("kind_model should be either lines or hdi")
```

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,274 @@
...
+    if y_ppc is None and isinstance(y, str):
+        y_ppc = y
+
+    if isinstance(y, str):
+        y = idata.observed_data[y]
+    elif not isinstance(y, DataArray):
+        y = xr.DataArray(y)
+
+    if len(y.dims) > 1 and plot_dim is None:
+        raise ValueError("Plot_dim argument is needed in case of multidimensional data")
```
canyon289, line 104:
```suggestion
        raise ValueError("Argument plot_dim is needed in case of multidimensional data")
```

Make the capitalization more clear

canyon289:
Here's what I get for generated plots. They generally look nice thought I have some nitpicks. This list isnt exhaustive so as we progress I may have a couple more nitpicks

* The visual style doesnt seem to match the plots in the ArviZ gallery
* The "parallel plot style, uncertainty seems a bit hard to read. Im not totally opposed to it just curious in the motivation and thinking behind it
* The posterior predictive samples in the bottom plot are really hard to see by default


test_plot_lm kwargs2
test_plot_lm kwargs3
test_plot_lm kwargs4
test_plot_lm kwargs5
test_plot_lm_multidim
test_plot_lm kwargs0
test_plot_lm kwargs1

utkarsh-maheshwari:
> * The posterior predictive samples in the bottom plot are really hard to see by default

Adjusting zorder, increasing default value of num_pp_samples could help maybe?

canyon289:
Maybe increasing dot size and adding a little bit of x jitter. The grid lines obscure them a little bit too I think? Do what you think works and well provide some feedback!

Unfortunately with plots things can get a little subjective sometimes, its the nature of the beast

image

utkarsh-maheshwari:
image
Jitter effect is great. Also changed alpha and marker size little bit. Looks better now. Thanks!

> * The visual style doesnt seem to match the plots in the ArviZ gallery

I have less idea what that means. Is it about layout, labeling, placing of legend and variables? 
Maybe I missed visual style is defined somewhere.

canyon289:
> I have less idea what that means. Is it about layout, labeling, placing of legend and variables?
> Maybe I missed visual style is defined somewhere.

Admittedly I did a terrible job describe this, so before I get into I'll say this. Dont worry about it too much right we can focus on functionality. But to add more detail, all the plots in the gallery have gray backgrounds with white grid marks in each axes, and a white background in the plot itself. Perhaps this is just a style application, but again this is a minor thing, functionality other visual details are more important at the moment. Sorry to distract!

image

utkarsh-maheshwari:
Understood the visual style point.

> * The "parallel plot style, uncertainty seems a bit hard to read. Im not totally opposed to it just curious in the motivation and thinking behind it

It is difficult in this case, maybe because data is very random. If `y_model` is well defined according to actual model equation, it  looks much better and easy to read. I tried the function on some of the BMCP examples.

image

canyon289:
This new plot looks amazing!

utkarsh-maheshwari:
Definitely, legend would be useful. It's an old picture I sent when the legend was not added.

utkarsh-maheshwari:
> * The visual style doesnt seem to match the plots in the ArviZ gallery
> * The "parallel plot style, uncertainty seems a bit hard to read. Im not totally opposed to it just curious in the motivation and thinking behind it
> * The posterior predictive samples in the bottom plot are really hard to see by default

1. Really have very less idea how to achieve it. When I plot other plots' tests results, even they don't  follow the visual style it. But plots in examples gallery follows. grey bg and all. I guess calling `az.style.use("arviz-darkgrid")` before calling `plot_lm` will do the job? 

2. As I have mentioned earlier, it is because of random data. Should I add a well-defined model in helpers to tackle this?

3. Solved

utkarsh-maheshwari:
Added some lines of code in `examples/mpl_plot_lm` but I can't test it unless this pr is merged. Should I still push the changes or examples need to be added in a separate PR after this is merged?

examples/matplotlib/mpl_plot_lm.py
```diff
@@ -0,0 +1,24 @@
...
+import matplotlib.pyplot as plt
+import xarray as xr
+import numpy as np
+import arviz as az
+
+az.style.use("arviz-darkgrid")
+
+data = az.load_arviz_data("regression1d")
+x = xr.DataArray(np.linspace(0, 1, 100))
+data.add_groups({"constant_data": {"x1": x}})
```
canyon289, line 17: Not required just asking. Should we just moditfy regressionid to include these?

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,323 @@
...
+    y : str or DataArray or ndarray
+        If str, variable name from observed_data
+    idata : az.InferenceData object, Optional
+        Optional only if y is Sequence
+    x : str or tuple of strings or DataArray or array-like, Optional
+        If str or tuple, variable name from constant_data
+        If ndarray, could be 1D, or 2D for multiple plots
+        If none, coords name of y (y should be DataArray).
+    y_model : str or Sequence, Optional
+        If str, varibale name from posterior.
```
canyon289, line 52:
```suggestion
        If str, variable name from posterior.
```

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,323 @@
...
+    Plot regression data and uncertanity in hdi form
+
+    .. plot:
+        :context: close-figs
+
+        >>> az.plot_lm(
+        ...     idata=data, y="y", x="x", y_model="y_model", kind_pp="hdi", kind_model="hdi"
+        ... )
+
+    Plot regression data for multi-dimentional y using plot_dim
```
canyon289, line 124:
```suggestion
    Plot regression data for multi-dimensional y using plot_dim
```

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,323 @@
...
+    .. plot::
+        :context: close-figs
+
+        >>> import arviz as az
+        >>> idata = az.load_arviz_data('regrssion1d')
+        >>> x = xr.DataArray(np.linspace(0, 1, 100))
+        >>> data.posterior["y_model"] = data.posterior["intercept"] + data.posterior["slope"]*x
+        >>> az.plot_lm(idata=data, y="y", x="x")
+
+    Plot regression data and uncertanity in mean
```
canyon289, line 108:
```suggestion
    Plot regression data and uncertainty in mean
```

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,323 @@
...
+
+    if plot_dim is None:
+        skip_dims = list(y.dims)
+    elif isinstance(plot_dim, str):
+        skip_dims = [plot_dim]
+    elif isinstance(plot_dim, tuple):
+        skip_dims = list(plot_dim)
+
+    x = filter_plotters_list(
+        list(
```
canyon289, line 223: Nit, add keyword arguments here, itll make the code easier to read

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,323 @@
...
+                var_names=x_var_names,
+                skip_dims=set(x_skip_dims),
+                combined=True,
+            )
+        ),
+        "plot_lm",
+    )
+
+    y = filter_plotters_list(
+        list(
```
canyon289, line 235: Same comment here

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,323 @@
...
+            xarray_var_iter(
+                y,
+                skip_dims=set(skip_dims),
+                combined=True,
+            )
+        ),
+        "plot_lm",
+    )
+
+    # If there are multiple x and multidimensional y, we need total of len(x)*len(y) graphs
```
canyon289, line 245: Thanks for the comment!

arviz/tests/base_tests/test_plots_bokeh.py
```diff
@@ -1104,3 +1105,110 @@ def test_plot_bpv_discrete():
...
+
+def test_plot_lm_list():
+    y = [1, 2, 3, 4, 5]
+    assert plot_lm(y=y, x=np.arange(len(y)), show=False, backend="bokeh")
+
+
+def test_plot_lm_bad_kind(models, multidim_models):
+    idata1 = models.model_1
+    idata2 = multidim_models.model_1
+    with pytest.raises(ValueError):
```
canyon289, line 1161: Can this test be parameterized?

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -1441,3 +1442,88 @@ def test_plot_bpv_discrete():
...
+        y="y",
+        plot_dim="dim1",
+        show=True,
+        figsize=(4, 16),
+    )
+    assert np.all(axes)
+
+
+def test_plot_lm_bad_kind(models, multidim_models):
+    idata1 = models.model_1
```
canyon289, line 1496: This one isnt your fault, but because we split our tests by plotting backend we get tons of repeated code. No need to change here. @ahartikainen  curious what your thoughts are

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,323 @@
...
+        >>> az.plot_lm(idata=data, y="y", x="x")
+
+    Plot regression data and uncertanity in mean
+
+    .. plot:
+        :context: close-figs
+
+        >>> az.plot_lm(idata=data, y="y", x="x", y_model="y_model")
+
+    Plot regression data and uncertanity in hdi form
```
canyon289, line 115:
```suggestion
    Plot regression data and uncertainty in hdi form
```

canyon289:
> > * The visual style doesnt seem to match the plots in the ArviZ gallery
> > * The "parallel plot style, uncertainty seems a bit hard to read. Im not totally opposed to it just curious in the motivation and thinking behind it
> > * The posterior predictive samples in the bottom plot are really hard to see by default
> 
>     1. Really have very less idea how to achieve it. When I plot other plots' tests results, even they don't  follow the visual style it. But plots in examples gallery follows. grey bg and all. I guess calling `az.style.use("arviz-darkgrid")` before calling `plot_lm` will do the job?
> 
>     2. As I have mentioned earlier, it is because of random data. Should I add a well-defined model in helpers to tackle this?
> 
>     3. Solved

1. I think it looks fine now. Thanks for fixing it.
2. Sorry for missing that. I agree Lets defer this for now
3. Solved

Overall this code looks great and I like the plots. This is fantastic add to the ArviZ toolset. Youre nearly most of the way there just some comments below.

# Need to fix or address comments
## Returned object should always be a plot
This examples from the docs returns a numpy array. It should return a plot
```
data = az.from_dict(observed_data = { "y": np.random.normal(size=(5, 7)) }, 
                     posterior_predictive = {"y": np.random.randn(4, 1000, 5, 7) / 2},
                     dims={"y": ["dim1", "dim2"]}, coords={"dim1": range(5), "dim2": range(7)} )
```
image

## Extra code documentation
There are a couple code comments that are very helpful. Add a couple more like the one noted above. nothing fancy, just one liners on major sections of logic. If youd like i can note where to help out

# Nitpicky Plot comment
This doesnt need to be fixed in this pr but the legend will need a box around it. Its really hard to read otherwise. I know we dont need this across all plots but I think we really should as some of our other plots suffer as well.
https://arviz-devs.github.io/arviz/examples/plot_parallel_rank.html
image

# Future PRs: Lots of Documentation
First off thanks for adding both a complete api documentation, code examples, and tests. It is a fantastic start. Some of the arguments are not particularly intuitive, which is not you fault just the nature of the best, so documentation will help.

In a followup PR can you add docs that show the following? Perhaps start with a notebook for prototyping and we can see how complicated it is and decide the best final home for the code

Things that will help are
* Showing more complicated y_models that include multiple covariates
* Non linear models
* Are counterfactuals supported?

canyon289:
Also this branch might need a rebase on main to fix the failing tests. Once we have passing tests we can assess the test coverage as well

arviz/plots/lmplot.py
canyon289: Add more comments like this in various places. Its very helpful to understand whats going on in the code!

examples/matplotlib/mpl_plot_lm.py
utkarsh-maheshwari: regression1d needs to be modified a bit. `y` dimensions and `y_model` dimensions do not match. y is aginst [0,100] while latter one is [0,1]

arviz/tests/base_tests/test_plots_bokeh.py
utkarsh-maheshwari: How about adding 3 different methods for testing ValueError, TypeError and UserWarning?

arviz/tests/base_tests/test_plots_matplotlib.py
utkarsh-maheshwari: Didn't get your point here. Confused what's not my fault 😆

arviz/tests/base_tests/test_plots_bokeh.py
utkarsh-maheshwari: I have tried to make these tests patrameterized.

utkarsh-maheshwari:
> This examples from the docs returns a numpy array

This one is interesting. This example is plotting multiple plots. So, it's returning an array of plots. Checked it for other plots like plot_trace, it too returns NumPy array. 

But it is even returning an array of shape (1,1) in case there is only one plot. It should return a plot, not an array of 1 plot. This needs some thinking, why is it happening

canyon289:
> > This examples from the docs returns a numpy array
> 
> This one is interesting. This example is plotting multiple plots. So, it's returning an array of plots. Checked it for other plots like plot_trace, it too returns NumPy array.
> 
> But it is even returning an array of shape (1,1) in case there is only one plot. It should return a plot, not an array of 1 plot. This needs some thinking, why is it happening

Oh youre right ignore my comment, no need to return a 1d array

utkarsh-maheshwari:
[arviz/data/io_pyro.py:111:20: E1102: pyro.poutine.trace is not callable (not-callable)](https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=4568&view=logs&j=af9a1d3a-ff8f-5483-6823-86518c773a9b&t=91d5999d-c3e7-502d-15b0-1abfa51eab6f&l=15)

Is it an error from pyro's side?

canyon289:
> [arviz/data/io_pyro.py:111:20: E1102: pyro.poutine.trace is not callable (not-callable)](https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=4568&view=logs&j=af9a1d3a-ff8f-5483-6823-86518c773a9b&t=91d5999d-c3e7-502d-15b0-1abfa51eab6f&l=15)
> 
> Is it an error from pyro's side?

Yea this is a linting error from pyro. Did you rebase on master? I think all the issues are fixed on master? It was passing as of 13 days ago, not sure if this is a new issue
https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=4558&view=results

utkarsh-maheshwari:
> Did you rebase on master

Yes, I did. I think this is some new error. Other PR's are also failing.

arviz/tests/base_tests/test_plots_matplotlib.py
canyon289: Theres a lot of duplication of code between `test_plots_matplotlib`, and the bokeh equivalent. Ideally we would not want to repeat things like the parameterization etc etc, and instead make the backend itself a parameter

arviz/tests/base_tests/test_plots_bokeh.py
```diff
@@ -1104,3 +1105,88 @@ def test_plot_bpv_discrete():
...
+        {
+            "x": ("x1", "x2"),
+            "kind_pp": "hdi",
+            "kind_model": "hdi",
+            "y_model_fill_kwargs": {"color": "red"},
+            "y_hat_fill_kwargs": {"color": "cyan"},
+        },
+    ],
+)
+def test_plot_lm(models, kwargs):
```
canyon289, line 1133: Add docstrings to all these tests. Even if a basic one

arviz/tests/base_tests/test_plots_bokeh.py
canyon289: Theres still many things being tested in the same test. Can we split these tests apart so each `pytest.raises` is in its own test? Or is this not possible?

utkarsh-maheshwari: Okay! Got it! making the changes..

arviz/tests/base_tests/test_plots_matplotlib.py
utkarsh-maheshwari: So does that mean, testing for both backends in a single file?

canyon289:
Master is fixed. Please rebase or merge so we can see if CI passes

arviz/tests/base_tests/test_plots_matplotlib.py
utkarsh-maheshwari: Removed `test_plot_lm_bad_kwargs` from bokeh tests.

utkarsh-maheshwari:
All green fInally!! Good to see passed tests after a long time xD. Thanks to all who worked to fix CI!

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,331 @@
...
+    legend=True,
+    grid=True,
+):
+    """Plot linear regression figure to visualize uncertaininty in mean and ppc values.
+
+    Parameters
+    ----------
+    y : str or DataArray or ndarray
+        If str, variable name from observed_data
+    idata : az.InferenceData object, Optional
```
OriolAbril, line 45: I think it should be `InferenceData` alone, or alternatively `arviz.InferenceData`, object is not necessary, and I think `az` can be confusing and redundant (arviz.InferenceData is as redundant, but IIUC, as arviz is the valid package name it will be rendered as a link to inferencedata api docs)

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,331 @@
...
+        Optional only if y is Sequence
```
OriolAbril, line 46: sequence is not a valid type according to `y` docstring, so I'm not sure what this means.

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,331 @@
...
+    x : str or tuple of strings or DataArray or array-like, Optional
```
OriolAbril, line 47:
```suggestion
    x : str, tuple of strings, DataArray or array-like, optional
```

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,331 @@
...
+        If ndarray, could be 1D, or 2D for multiple plots
+        If none, coords name of y (y should be DataArray).
+    y_model : str or Sequence, Optional
+        If str, variable name from posterior.
+        Its dimensions should be same as y plus added chains and draws.
+    y_hat : str, Optional
+        If str, variable name from posterior_predictive.
+        Its dimensions should be same as y plus added chains and draws.
+    num_pp_samples : int, Optional, Default 50
+    kind_pp : ["samples", "hdi"], Optional, Default "samples"
```
OriolAbril, line 58: for arguments taking only a set of known values, the convention is to use `{` instead of `[`. Moreover, adding `default "value"` already indicated that this is an optional argument in numpydoc:

```suggestion
    kind_pp : {"samples", "hdi"}, default "samples"
```

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,331 @@
...
+    kind_model : ["lines", "hdi"], Optional, Default "lines"
+    plot_dim : str, Optional
+        Necessary if y is multidimensional.
+    backend : str, Optional
+        Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
+    y_kwargs : dict, optional
+        Passed to plot() in matplotlib and circle() in bokeh
```
OriolAbril, line 65: can you make plot and circle hyperlinks using intersphinx? same for all other kwargs.

see https://arviz-devs.github.io/arviz/contributing/developer_guide.html#hyperlinks and #1188 for guidance and examples.

OriolAbril (commented):
The comments don't necessarily need to be addressed in this PR, but they need to be addressed at some point.

> This doesnt need to be fixed in this pr but the legend will need a box around it. Its really hard to read otherwise. I know we dont need this across all plots but I think we really should as some of our other plots suffer as well.
https://arviz-devs.github.io/arviz/examples/plot_parallel_rank.html

This is actually an issue with the arviz-darkgrid theme afaik: https://github.com/arviz-devs/arviz/blob/main/arviz/plots/styles/arviz-darkgrid.mplstyle#L11

arviz/plots/lmplot.py
utkarsh-maheshwari: If any of y, x, y_hat, y_model is str, then idata is necessary. Yea, This line is unclear.

arviz/plots/lmplot.py
utkarsh-maheshwari: Done!

arviz/plots/lmplot.py
OriolAbril: This explanation should be the one on the docstring! It's different from the current one too which seeing this looks wrong

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,334 @@
...
+        If none, coords name of y (y should be DataArray).
+    y_model : str or Sequence, Optional
+        If str, variable name from posterior.
+        Its dimensions should be same as y plus added chains and draws.
+    y_hat : str, Optional
+        If str, variable name from posterior_predictive.
+        Its dimensions should be same as y plus added chains and draws.
+    num_pp_samples : int, Optional, Default 50
+    kind_pp : {"samples", "hdi"}, Default "samples"
+    kind_model : {"lines", "hdi"}, Default "lines"
```
OriolAbril, line 59: these need an explanation, even if a single sentence.

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,334 @@
...
+    plot_dim : str, Optional
+        Necessary if y is multidimensional.
+    backend : str, Optional
+        Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
+    y_kwargs : dict, optional
+        Passed to :meth:`mpl:matplotlib.axes.Axes.plot` in matplotlib
+        and :meth:`bokeh:bokeh.plotting.figure.Figure.circle` in bokeh
```
OriolAbril, line 66: according to sphinxinv, the objects in bokeh docs containing circle are these: 

```
:py:class:`bokeh.models.glyphs.Circle`
:py:attribute:`bokeh.models.glyphs.Circle.x`
:py:attribute:`bokeh.models.glyphs.Circle.y`
:py:function:`bokeh.models.markers.Circle`
:py:function:`bokeh.models.markers.CircleX`
:py:function:`bokeh.models.markers.CircleY`
:py:method:`bokeh.plotting.Figure.circle`
:py:method:`bokeh.plotting.Figure.circle_x`
:py:method:`bokeh.plotting.Figure.circle_y`
:py:method:`bokeh.plotting.GMap.circle`
:py:method:`bokeh.plotting.GMap.circle_x`
:std:label:`bokeh.models.glyphs.circle`
:std:label:`bokeh.models.markers.circle`
:std:label:`bokeh.models.markers.circledot`
:std:label:`bokeh.models.markers.circlex`
:std:label:`bokeh.models.markers.circley`
```

the one you used is not an option, maybe it was in an old version?

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,334 @@
...
+    y_hat_fill_kwargs : dict, optional
+        Passed to az.plot_hdi()
+    y_model_plot_kwargs : dict, optional
+        Passed to :meth:`mpl:matplotlib.axes.Axes.plot` in matplotlib
+        and :meth:`bokeh:bokeh.plotting.figure.Figure.line` in bokeh
+    y_model_fill_kwargs : dict, optional
+        Passed to az.plot_hdi()
+    backend_kwargs : dict, optional
```
OriolAbril, line 77: this also needs some explanation as well as the intersphinx references.

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,340 @@
...
+    backend : str, Optional
+        Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
+    y_kwargs : dict, optional
+        Passed to :meth:`mpl:matplotlib.axes.Axes.plot` in matplotlib
+        and :meth:`bokeh:bokeh.plotting.Figure.circle` in bokeh
+    y_hat_plot_kwargs : dict, optional
+        Passed to :meth:`mpl:matplotlib.axes.Axes.plot` in matplotlib
+        and :meth:`bokeh:bokeh.plotting.Figure.circle` in bokeh
+    y_hat_fill_kwargs : dict, optional
+        Passed to az.plot_hdi()
```
OriolAbril, line 75:
```suggestion
        Passed to {func}`~arviz.plot_hdi`
```

arviz/plots/lmplot.py
```diff
@@ -0,0 +1,340 @@
...
+    y_model_plot_kwargs : dict, optional
+        Passed to :meth:`mpl:matplotlib.axes.Axes.plot` in matplotlib
+        and :meth:`bokeh:bokeh.plotting.Figure.line` in bokeh
+    y_model_fill_kwargs : dict, optional
+        Significant if `kind_model` is "hdi". Passed to :func:`~arviz.plot_hdi`
+    backend_kwargs : dict, optional
+        These are kwargs specific to the backend being used. For additional documentation
+        check the plotting method of the backend.
```
OriolAbril, line 83: the links to the methods where these are passed are still missing, you can take them from #1728 to avoid having to track them down in the source code.

arviz/plots/lmplot.py
utkarsh-maheshwari: I don't know how I missed this one!

PR #1728, opened 2021-06-17
[WIP] Initial prototype for Dot Plots

Rishabh261998:
## Description

This is the initial prototype for `dot plots`. I have implemented Wilkinson's algorithm for the plot. 

The only argument which is necessary for this plot is a  `values` argument of type array. These are the values for which the dots will be plotted.

Function Call:

`az.plot_dots(values = values)`

Output:

Screenshot from 2021-06-17 15-26-41

`az.plot_dots(values = values, point_interval = True)`

Output:

Screenshot from 2021-06-17 15-26-49

`az.plot_dots(values = values , quantiles = 50, point_interval = True)`

Output:

Screenshot from 2021-06-17 15-27-28

## Checklist


arviz/plots/dotplot.py
```diff
@@ -0,0 +1,142 @@
...
+    plot_kwargs=None,
+    interval_kwargs=None,
+    backend=None,
+    backend_kwargs=None,
+    **kwargs
+):
+
+    """Plot distribution as dot plot.
+
+    If quantiles is specified then it will plot the quantile dot plot.
```
aloctavodia, line 32: and if not?

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,142 @@
...
+    quantiles=None,
+    point_interval=None,
+    plot_kwargs=None,
+    interval_kwargs=None,
+    backend=None,
+    backend_kwargs=None,
+    **kwargs
+):
+
+    """Plot distribution as dot plot.
```
aloctavodia, line 30: Plot distribution as quantile dot plot

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,142 @@
...
+        Line width throughout. If None it will be autoscaled based on figsize.
+    markersize : int
+        Markersize throughout. If None it will be autoscaled based on figsize.
+    hdi_prob : float
+        Valid only when point_interval is True. Plots HDI for chosen percentage of density. 
+        Defaults to 0.94.
+    rotated : bool
+        Whether to rotate the dot plot by 90 degrees.
+    quantiles : int
+        Number of quantiles to plot, used for quantile dot plots
```
aloctavodia, line 65: I think I am missing something. What is the alternative if quantiles is none?

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,142 @@
...
+    stackratio=1,
+    hdi_prob=None,
+    rotated=False,
+    dotcolor="grey",
+    intervalcolor="red",
+    markersize=None,
+    figsize=None,
+    linewidth=None,
+    point_estimate="auto",
+    quantiles=None,
```
aloctavodia, line 21: could the default value be a small integer like 20?

arviz/plots/dotplot.py
OriolAbril: "... as a dot plot or quantile dot plot" I think would be best, plus as you commented above, references to both dotplot and quantile dotplot explanations.

arviz/plots/dotplot.py
OriolAbril: I agree that it would probably be better to have a quantile dotplot as a default, especially if the dotplot with a lot of samples is slow.

arviz/plots/__init__.py
```diff
@@ -49,4 +50,5 @@
     "plot_trace",
     "plot_violin",
     "plot_separation",
+    "plot_dots",
```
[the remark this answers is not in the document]
Rishabh261998: I'll rename it to `plot_dot`

arviz/plots/backends/matplotlib/dotplot.py
```diff
@@ -0,0 +1,86 @@
...
+    if markersize is None:
+        markersize = auto_markersize
+
+    if interval_kwargs is None:
+        interval_kwargs = {}
+        interval_kwargs.setdefault("color", intervalcolor)
+
+    _, ax = create_axes_grid(1, backend_kwargs=backend_kwargs,)
+
+    ## Wilkinson's Algorithm
```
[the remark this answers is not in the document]
Rishabh261998: Sure

arviz/plots/backends/matplotlib/plotpointinterval.py
```diff
@@ -0,0 +1,75 @@
+import numpy as np
+
+from ...plot_utils import calculate_point_estimate
+from ....stats import hdi
+
+
+def plot_point_interval(
```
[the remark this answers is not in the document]
Rishabh261998: As Oriol had mentioned that we could use it for `plot_posterior`, not sure though. I think it should be better to add it to plot_utils

arviz/plots/dotplot.py
Rishabh261998: Then it will set quantiles as the total number of points in data and then plot.

arviz/plots/dotplot.py
Rishabh261998: I also think "Plot distribution as dot plot or quantile dot plot" sounds better

arviz/plots/dotplot.py
Rishabh261998: As I mentioned above it will set quantiles as the number of samples. Basically, it will use the actual data to plot. Otherwise, if quantiles are given and are less than the total number of samples then it would generate quantile number of points and use them to plot. This is done because if we randomly sample quantile number of points from our values and plot them then it may be that it won't represent the actual distribution. [This](https://github.com/mjskay/when-ish-is-my-bus/blob/master/quantile-dotplots.md) explains it much better.

arviz/plots/dotplot.py
Rishabh261998: Ok, I can make quantile dot plot as default. Should I keep the default value as 20? or a bit more

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,142 @@
...
+        point_estimate = rcParams["plot.point_estimate"]
+    elif point_estimate not in {"mean", "median", "mode", None}:
+        raise ValueError("The value of point_estimate must be either mean, median, mode or None.")
+
+    if quantiles:
+        if quantiles >= values.shape[0]:
+            quantiles = values.shape[0]
+        else:
+            qlist = np.arange(100 / (2 * quantiles), 100, 100 / (quantiles))
+            values = np.percentile(values, qlist)
```
OriolAbril, line 107: I think it would be more natural to use a linspace and np.quantiles here.

arviz/plots/backends/matplotlib/plotpointinterval.py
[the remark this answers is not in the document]
OriolAbril: I think we should already start to progressively move towards https://github.com/arviz-devs/arviz/wiki/Plot-hierarchy. This plot_point_interval could be the plot_atomic_box for example. Otherwise every plot needs to 

1) reimplement every single component in both backends (plot_posterior plots and labels the hdi interval, plot_violin has the exact same line/boxplot like formatting we have here yet we can't use that here, we need to reimplement it again for the dotplot; plot forest also does something very similar again completely reimplemented from scratch). 
2) when we make changes to facetting (with the skip dims PR for example or if we wanted to support what Tomas asked the other day on slack) we have to update every single plot, instead of updating the facet/grid level functions.

Rishabh261998:
So @OriolAbril what other do I need to make to the `plot_point_interval`? one thing which I need to add is that `plot_posterior` plots a text for the `point_estimate` whereas `plot_dot` plots a dot and also an argument which enables us to only plot the HDI or HDI + some interval(default is the 25-75 quartile). Apart from these do I need to make any more changes?

Rishabh261998:
Also the half-eyed plot which I would be implementing in the future would also need the `plot_point_interval`

OriolAbril:
Like plot_posterior does internally, I would separate point estimate from interval. I think for now you can focus on ehat you need, we can then suggest extensions and see if we can simplify code in other parts of the library by using that

Is the half eye plot the same as plot_violin but without duplicating the kde?

Rishabh261998:
Ok sure, yes the `half_eye` plot is same as violin without the duplication

arviz/plots/dotplot.py
sethaxen: I agree quantile dotplot should be the default. I would check what [ggdist](https://mjskay.github.io/ggdist/) uses for the default number of dots and emulate that, since the author Matthew Kay does research in uncertainty visualizations.

arviz/plots/dotplot.py
sethaxen: Should probably note that the minimum of this number and the number of data points is used for plotting.

arviz/plots/backends/matplotlib/dotplot.py
sethaxen, line 55: For modularity, could you split this out into its own function?

sethaxen (changes requested):
Nice work so far!

arviz/plots/dotplot.py
Rishabh261998: Yes, if quantile dot plot is set as default, so what should I set the default value of `quantiles` because 20 seems too little for me

arviz/plots/dotplot.py
Rishabh261998: Sure

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,142 @@
...
+    else:
+        if not 1 >= hdi_prob > 0:
+            raise ValueError("The value of hdi_prob should be in the interval (0, 1]")
+
+    if point_estimate == "auto":
+        point_estimate = rcParams["plot.point_estimate"]
+    elif point_estimate not in {"mean", "median", "mode", None}:
+        raise ValueError("The value of point_estimate must be either mean, median, mode or None.")
+
+    if quantiles:
```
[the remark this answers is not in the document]
Rishabh261998: Sure, Thanks

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def plot_dots(
+    values=None,
+    binwidth=None,
+    dotsize=1,
+    stackratio=1,
+    hdi_prob=None,
+    rotated=False,
+    dotcolor="grey",
```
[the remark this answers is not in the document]
Rishabh261998: `ggdist` uses gey as the default hence I went for that

arviz/plots/backends/matplotlib/dotplot.py
Rishabh261998: Sure

Rishabh261998:
I have done these changes, the minor changes which @aloctavodia and @sethaxen  suggested, moved `plot_point_interval` to `plot_utils`, set quantile dot plots as default with a default value of `quantile` as 50 for the time being.

arviz/plots/dotplot.py
aloctavodia: My understanding is that dot-plots are useful (or at leas more common) for small dataset, otherwise an histogram or KDE would be a better choice. And that quantile dot plots are generally used with a small number of quantiles because you expect people to actually be able to count the dots. So a small number like 20 seems like a reasonable default. But of course I may be missing something and there are other reasons to pick other default values.

arviz/plots/dotplot.py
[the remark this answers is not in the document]
aloctavodia: In ArviZ we use "C0" as the default color.

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,142 @@
...
+    -------
+    axes : matplotlib axes
+    
+    """
+
+    if values is None:
+        raise ValueError("Please provide the values array for plotting")
+
+    values = np.asarray(values)
+    values = np.sort(values)
```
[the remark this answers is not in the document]
aloctavodia: `np.sort` already call `np.asarray`

arviz/plots/dotplot.py
sethaxen: In http://users.eecs.northwestern.edu/~jhullman/busUncertaintyVis.pdf, which presented quantile dotplots, they found that users were more precise at estimating probabilities from quantile dotplots with 20 dots vs 100. On the other hand, in https://dl.acm.org/doi/10.1145/3173574.3173718 they found that when it came to decison making using quantile dotplots, users performed better with 50 dots vs 20 (they did not compare 100). Which would perhaps indicate that 50 is a reasonable default.

arviz/plots/dotplot.py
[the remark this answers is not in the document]
Rishabh261998: Yes I have removed the `np.asarray` line

arviz/plots/dotplot.py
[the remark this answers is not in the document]
Rishabh261998: Ok will change this to "C0", currently I have set it to "C7" which is grey itself

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,153 @@
...
+        raise ValueError(
+            "quantiles must be of integer type, refer docs for further details"
+        )
+
+    
+    if quantiles >= values.shape[0]:
+        quantiles = values.shape[0]
+    else:
+        qlist = np.linspace(1 / (2 * quantiles), 1 - 1 / (2*quantiles), quantiles)
+        values = np.quantile(values, qlist)
```
sethaxen, line 119: Overwriting `values` here means that the point interval will always show the HDI of the quantiles, but it should probably show the HDI of the raw data instead.

arviz/plots/backends/bokeh/dotplot.py
```diff
@@ -0,0 +1,117 @@
...
+        ax = plot_point_interval(
+            ax, values, point_estimate, hdi_prob, quartiles, linewidth, markersize, rotated, interval_kwargs, 'bokeh'
+        )
+        
+    show_layout(ax)
+    
+    return ax
+
+
+def wilkinson_algorithm(
```
sethaxen, line 85: Do you need two implementations of Wilkinson's algorithm, or can they be unified? Also, wouldn't it be cleaner to have the unified Wilkinson's algorithm to just return stack locations and counts per stack, and then other backend-specific functions could handle the construction and rotation of the stack?

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,153 @@
...
+    backend_kwargs=None,
+    **kwargs
+):
+
+    """Plot distribution as dot plot or quantile dot plot.
+
+    This function uses the Wilkinson's Algorithm
+    (https://www.cs.uic.edu/~wilkinson/Publications/dotplots.pdf)to allot dots to bins.
+    The quantile dot plots was inspired from the paper(When (ish) is My Bus?:
+    User-centered Visualizations of Uncertainty in Everyday).
```
sethaxen, line 36: Can we add these as references with DOI's?

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,153 @@
...
+    linewidth : int
+        Line width throughout. If None it will be autoscaled based on figsize.
+    markersize : int
+        Markersize throughout. If None it will be autoscaled based on figsize.
+    hdi_prob : float
+        Valid only when point_interval is True. Plots HDI for chosen percentage of density.
+        Defaults to 0.94.
+    rotated : bool
+        Whether to rotate the dot plot by 90 degrees.
+    quantiles : int
```
sethaxen, line 68: `plot_kde` takes `quantiles` as a list of quantiles to plot. I wonder if we should use a different kwarg name than `quantiles` here for consistency. Perhaps `nquantiles` or `numquantiles`, which would also be more descriptive?

arviz/plots/plot_utils.py
```diff
@@ -385,6 +386,108 @@ def calculate_point_estimate(point_estimate, values, bw="default", circular=Fals
...
+                    quantiles_interval[-(j + 1)],
+                    linewidth=width,
+                    **interval_kwargs
+                )
+
+        if point_estimate:
+            point_value = calculate_point_estimate(point_estimate, values)
+            if rotated:
+                ax.plot(
+                    0, point_value, "o", markersize=markersize, color="black",
```
sethaxen, line 454: Can the color be made a keyword argument, perhaps defaulting to one of the standard colors like `C0`? (for dark-background themes, defaulting to "black" is not ideal)

arviz/plots/plot_utils.py
```diff
@@ -385,6 +386,108 @@ def calculate_point_estimate(point_estimate, values, bw="default", circular=Fals
...
+    interval_kwargs : dict
+        Keyword passed to the point interval
+    """
+
+    endpoint = 100 * (1 - hdi_prob) / 2
+    if quartiles:
+        qlist_interval = [endpoint, 25, 75, 100 - endpoint]
+    else:
+        qlist_interval = [endpoint, 100 - endpoint]
+    quantiles_interval = np.percentile(values, qlist_interval)
```
sethaxen, line 424: Using `quantile` is a little easier to follow when the variables record quantiles:
```suggestion
    endpoint = (1 - hdi_prob) / 2
    if quartiles:
        qlist_interval = [endpoint, .25, .75, 1 - endpoint]
    else:
        qlist_interval = [endpoint, 1 - endpoint]
    quantiles_interval = np.quantile(values, qlist_interval)
```

arviz/plots/plot_utils.py
sethaxen: Likewise, can the marker be a keyword argument?

sethaxen (changes requested):
This is looking great! I tested both backends a bit and only had some minor comments on functionality; most of my comments are regarding the internals.

Rishabh261998:
@sethaxen I make the suggested changes. Apart from this I had a doubt, so when I am using black to format the files then some more files which I haven't touched are also getting reformatted. So what should I do?, should I commit without running black or should I commit with the extra reformatted files? black version 19.10b0

sethaxen:
> @sethaxen I make the suggested changes.
Great! I'll re-review.

> Apart from this, I had a doubt, so when I am using black to format the files then some more files which I haven't touched are also getting reformatted. So what should I do?, should I commit without running black or should I commit with the extra reformatted files? black version 19.10b0

You should only commit formatting changes for parts of files that you've touched.  I'm guessing this is happening because there's a mismatch between the version of black being used by the CI and the version of black you have installed, but I'm not certain. @OriolAbril why do you think this is happening?

ahartikainen:
CI has `black==21.6b0`

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,153 @@
...
+    Returns
+    -------
+    axes : matplotlib axes
+    
+    """
+
+    if values is None:
+        raise ValueError("Please provide the values array for plotting")
+
+    values = np.sort(values)
```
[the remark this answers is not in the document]
Rishabh261998: This isn't working as .sort() works for lists and after applying flatten() values is still a numpy array

arviz/plots/dotplot.py
Rishabh261998: Going with `nquantiles`

Rishabh261998:
@sethaxen made all the changes and also added tests, please go through them. Thanks

arviz/plots/dotplot.py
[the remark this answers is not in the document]
sethaxen: I would then try something like
```suggestion
    values = np.ravel(values)
    values.sort()
```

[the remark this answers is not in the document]
Rishabh261998: Thanks

arviz/plots/backends/matplotlib/dotplot.py
```diff
@@ -0,0 +1,139 @@
...
+        y.append(y_coord)
+        for _ in range(num_dots_stack):
+            x.append(x_coord)
+            y.append(y_coord)
+            if rotated:
+                x_coord += binwidth + (stackratio - 1) * (binwidth)
+            else:
+                y_coord += binwidth + (stackratio - 1) * (binwidth)
+
+    return x, y
```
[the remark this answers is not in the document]
Rishabh261998: Yeah, this looks good. Thanks

Rishabh261998:
@sethaxen also can you look at the tests and suggest if I need to add anything. Thanks

sethaxen:
> @sethaxen also can you look at the tests and suggest if I need to add anything. Thanks

Sure thing, I'll look at them tomorrow.

Rishabh261998:
Thanks also I have added some examples. @aloctavodia if you get time could you also go through it? Thanks

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,229 @@
...
+        >>> import arviz as az
+        >>> values = np.random.normal(0, 1, 500)
+        >>> az.plot_dist(values)
+
+    Manually adjust number of quantiles to plot
+
+    .. plot::
+        :context: close-figs
+
+        >>> az.plot_dist(values, nquantiles=100)
```
sethaxen, line 118: Here and below:
```suggestion
        >>> az.plot_dot(values, nquantiles=100)
```

sethaxen (commented):
The tests look good to me, but someone who contributes more to the Python package should also take a look at them. It would be good to add a test for `plot_point_interval`, and a unit test for the Wilkinson algorithm as well.

aloctavodia (approved):

arviz/plots/dotplot.py
Rishabh261998: yeah, my bad will change that

Rishabh261998: @sethaxen `wilkinson_algorithm` function is in the `dotplot.py` file in plots directory so in which test file should I include the test for it?

arviz/plots/backends/bokeh/dotplot.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+
+    if nquantiles > values.shape[0]:
+        warnings.warn(
+            "nquantiles must be less than or equal to the number of data points", UserWarning
+        )
+        nquantiles = values.shape[0]
+    else:
+        qlist = np.linspace(1 / (2 * nquantiles), 1 - 1 / (2 * nquantiles), nquantiles)
+        values = np.quantile(values, qlist)
```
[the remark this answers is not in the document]
sethaxen: I think this can wait for a future PR though.

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -1441,3 +1444,98 @@ def test_plot_bpv_discrete():
...
+    values = np.sort(data)
+    ax = plot_point_interval(ax, values, **kwargs)
+    assert ax
+
+
+def test_wilkinson_algorithm(continuous_model):
+    data = continuous_model["x"]
+    values = np.sort(data)
+    _, stack_counts = wilkinson_algorithm(values, 0.5)
+    assert np.sum(stack_counts) == len(values)
```
sethaxen, line 1541: Maybe a test designed after Figure 4 of Wilkinson's paper.
```suggestion
    assert np.sum(stack_counts) == len(values)
    stack_locs, stack_counts = wilkinson_algorithm(([0.0, 1.0, 1.8, 3.0, 5.0], 1.0)
    @assert stack_locs == np.array([0.0, 1.4, 3.0, 5.0])
    @assert stack_counts == np.array([1, 2, 1, 1])
```

sethaxen (approved):
With just one more test of Wilkinson's algorithm, looks good to me! Great work!

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,229 @@
...
+    backend=None,
+    backend_kwargs=None,
+    **kwargs
+):
+
+    """Plot distribution as dot plot or quantile dot plot.
+
+    This function uses the Wilkinson's Algorithm
+    (Leland Wilkinson (1999) Dot Plots, The American Statistician, 53:3, 276-281,
+    DOI: 10.1080/00031305.1999.10474474)to allot dots to bins.
```
OriolAbril, line 37: How about using citation like references? 

```suggestion
    This function uses the Wilkinson's Algorithm [1]_ to allot dots to bins.
```

plus the complete reference in the `References` section like in https://arviz-devs.github.io/arviz/_modules/arviz/plots/separationplot.html#plot_separation. Same for the paper below, maybe leaving the title and reference number in the text?

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,229 @@
+import numpy as np
+
+from ..rcparams import rcParams
+from .plot_utils import get_plotting_function
+
+
+def plot_dot(
+    values=None,
```
OriolAbril, line 8: shouldn't values be a required argument instead of having a default of none and raise an error if it's none?

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,229 @@
...
+    DOI: 10.1080/00031305.1999.10474474)to allot dots to bins.
+    The quantile dot plots was inspired from the paper(Matthew Kay, Tara Kola, Jessica R. Hullman,
+    and Sean A. Munson. 2016. When (ish) is My Bus? User-centered Visualizations of Uncertainty in
+    Everyday, Mobile Predictive Systems. DOI:https://doi.org/10.1145/2858036.2858558).
+
+    Parameters
+    ----------
+    values : array-like
+        Values to plot
+    binwidth : float
```
OriolAbril, line 46:
```suggestion
    binwidth : float, optional
```

and same with the rest of optional parameters

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,229 @@
...
+        Width of the bin for drawing the dot plot.
+    dotsize : float
+        The size of the dots relative to the bin width. The default, 1, makes dots be
+        just about as wide as the bin width.
+    stackratio : float
+        The distance between the center of the dots in the same stack relative to the bin height.
+        The default, 1, makes dots in the same stack just touch each other.
+    point_interval : bool
+        Plots the point interval. Uses hdi_prob to plot the HDI interval
+    point_estimate : Optional[str]
```
OriolAbril, line 56:
```suggestion
    point_estimate : str, optional
```

not all the codebase follows the [numpydoc convention](https://arviz-devs.github.io/arviz/contributing/contributing_prs.html#docstring-formatting-and-type-hints) (the standard to be followed wasn't clear originally), but new code should stick to it now that we have. The `Optional[str]` syntax is for type hints, that go in the function definition, not in the docstring.

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,229 @@
...
+        Plot point estimate per variable. Values should be ‘mean’, ‘median’, ‘mode’ or None.
+        Defaults to ‘auto’ i.e. it falls back to default set in rcParams.
+    dotcolor : string
+        The color of the dots
```
OriolAbril, line 60: This should indicate which values are valid (same applies to all kwargs below). I think the best way to do so is by indicating to which matplotlib and bokeh function this is eventually passed to and using intersphinx to link there. Alternatively, if using `to_hex` it can also be simply "must be a valid matplotlib color", because we use a matplotlib helper function to convert the color to hex so all matplotlib colors also work on bokeh. 

Example: https://arviz-devs.github.io/arviz/_modules/arviz/plots/hdiplot.html#plot_hdi (bokeh patch seems to have been deleted/moved and the link is no longer generated correctly)

Guidance on intersphinx: https://arviz-devs.github.io/arviz/contributing/developer_guide.html#hyperlinks

OriolAbril (commented):
As this has already been approved twice, I have only reviewed the docstring

Rishabh261998:
Done @OriolAbril

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,229 @@
...
+    show=None,
+    plot_kwargs=None,
+    backend=None,
+    backend_kwargs=None,
+    **kwargs
+):
+    """Plot distribution as dot plot or quantile dot plot.
+
+    This function uses the Wilkinson's Algorithm[1] to allot dots to bins.
+    The quantile dot plots was inspired from the paper[2].
```
OriolAbril, line 36:
```suggestion
    This function uses the Wilkinson's Algorithm [1]_ to allot dots to bins.
    The quantile dot plots was inspired from the paper [2]_.
```

The final underscore i necessary to make those clickable links.

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,229 @@
...
+    markersize : int, optional
+        Markersize throughout. If None it will be autoscaled based on figsize.
+    markercolor: string, optional
+        The color of the marker when plot_interval is True. Should be a valid matplotlib color.
+    marker: string, optional
+        The shape of the marker. Valid for matplotlib backend
+        Defaults to "o"
+    hdi_prob : float, optional
+        Valid only when point_interval is True. Plots HDI for chosen percentage of density.
+        Defaults to 0.94.
```
OriolAbril, line 70: defaults to rcParams value

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,229 @@
...
+    nquantiles : int, optional
+        Number of quantiles to plot, used for quantile dot plots
+        Defaults to 50.
+    quartiles : bool, optional
+        If True then the quartile interval will be plotted with the HDI.
+        Defaults to True.
+    figsize : tuple, optional
+        Figure size. If None it will be defined automatically.
+    plot_kwargs : dict, optional
+        Keywords passed for customizing the dots.
```
OriolAbril, line 82: Add to which functions these kwargs are passed. scatter for matplotlib? or plot?

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,229 @@
...
+    backend: str, optional
+        Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
+    ax : axes, optional
+        Matplotlib axes or bokeh figures.
+    show: bool, optional
+        Call backend show function.
+    backend_kwargs: bool, optional
+        These are kwargs specific to the backend being used. For additional documentation
+        check the plotting method of the backend.
```
OriolAbril, line 91: Same here, add where these kwargs are passed to. Ideally users won't need to read the source of this function to see where these are passed, then go to plot_utils to the right function and then go to the matplotlib/bokeh docs of the function/class used only to know what is valid here, the same docstring should have the links to matplotlib/bokeh docs directly to the right function. We can then comment on #1188 to have this docstring be an example of good practices

side note, it says bool, I think it should be a dictionary.

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,231 @@
...
+    backend: str, optional
+        Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
+    ax : axes, optional
+        Matplotlib axes or bokeh figures.
+    show: bool, optional
+        Call backend show function.
+    backend_kwargs: dict, optional
+        These are kwargs specific to the backend being used, passed to
+        :meth:`mpl:matplotlib.axes.Axes.plot` or
+        :meth:`bokeh:bokeh.plotting.figure.Figure.patch`.
```
OriolAbril, line 93: the `figure.Figure` structure is no longer used in bokeh so this isn't a valid reference, and I think backend kwargs are not passed to patch but to a figure generation function.

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,231 @@
...
+    show=None,
+    plot_kwargs=None,
+    backend=None,
+    backend_kwargs=None,
+    **kwargs
+):
+    """Plot distribution as dot plot or quantile dot plot.
+
+    This function uses the Wilkinson's Algorithm [1]_ to allot dots to bins.
+    The quantile dot plots was inspired from the paper [2]_.
```
OriolAbril, line 36:
```suggestion
    The quantile dot plots was inspired from the paper *When (ish) is My Bus?* [2]_.
```

arviz/plots/dotplot.py
```diff
@@ -0,0 +1,231 @@
...
+        in matplotlib and :meth:`bokeh:bokeh.plotting.Figure.circle` in bokeh
+    backend: str, optional
+        Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
+    ax : axes, optional
+        Matplotlib axes or bokeh figures.
+    show: bool, optional
+        Call backend show function.
+    backend_kwargs: dict, optional
+        These are kwargs specific to the backend being used, passed to
+        :meth:`mpl:matplotlib.axes.Axes.plot` or
```
utkarsh-maheshwari, line 92: In matplotlib case, backend kwargs are passed to `create_axes_grid`, which passes it to `pyplot.subplots`. So, shouldn't it be :meth: `mpl:matplotlib.pyplot.subplots`

utkarsh-maheshwari, line 92: In matplotlib case, backend kwargs are passed to create_axes_grid, which passes it to pyplot.subplots. So, shouldn't it be :meth: mpl:matplotlib.pyplot.subplots

Rishabh261998: Yes, you are right. Thanks for pointing it out

Rishabh261998:
@OriolAbril I am having a bit of difficulty in solving the merge conflicts when I am rebasing. So when I rebase an old commit of this dot plot PR is trying to merge with the `plot_lm` commit.

arviz/plots/dotplot.py
OriolAbril: Actually, I thinl subplots is a function not a method,  so it should be `:func:` instead of meth

sethaxen:
Since @aloctavodia and I have already approved this PR, I think it falls to @OriolAbril to give the final approval.

PR #1731, opened 2021-06-20
use relative imports

MarcoGorelli:
## Checklist

- [ ] Follows [official](https://github.com/arviz-devs/arviz/blob/main/CONTRIBUTING.md#pull-request-checklist) PR format
- [ ] Includes a sample plot to visually illustrate the changes (only for plot-related functions)
- [ ] New features are properly documented (with an example if appropriate)?
- [ ] Includes new or updated tests to cover the new feature
- [ ] Code style  correct (follows pylint and black guidelines)
- [ ] Changes are listed in [changelog](https://github.com/arviz-devs/arviz/blob/main/CHANGELOG.md#v0xx-unreleased)

From #258  (cc @canyon289 ) :

> Also name all arviz imports relative imports
>
> This one was only was mostly due to my OCD. I hope you guys don't judge me too much

If relative imports are preferred here, why not automate them?

---

disclaimer: I'm the author of absolufy-imports

PR #1738, opened 2021-07-05
Revert "use relative imports"

canyon289:
Reverts arviz-devs/arviz#1731

For some reason merging the above pr seems to be breaking master CI. Seeing if revert fixes


canyon289:
Nevermind things are just broken

Issue #1685, opened 2021-04-30
Get kde to fail gracefully when given constant data


OriolAbril:
In some cases kde gets arrays that represent a constant value, all samples have the exact same value. AFAIK, this currently works (it may print a warning but it definitely works) but as expected it generates very tall spikes _a la_ dirac delta. It would be great to have it return nans or something like that, mostly for `plot_trace` in compact mode. 

See discussion in https://github.com/arviz-devs/arviz/pull/1641

PR #1744, opened 2021-07-19
KDE friendly warnings for constant data

aloctavodia:
Provides a more user-friendly warning when the data used to compute a KDE is constant (or has non finite values). Fixes https://github.com/arviz-devs/arviz/issues/1685. For those cases the plot will look empty.

### DIFF
--- a/arviz/stats/density_utils.py
+++ b/arviz/stats/density_utils.py
@@ -36,9 +36,9 @@ def _bw_silverman(x, x_std=None, **kwargs):  # pylint: disable=unused-argument
 def _bw_isj(x, grid_counts=None, x_std=None, x_range=None):
     """Improved Sheather-Jones bandwidth estimation.
 
-    Improved Sheather and Jones method as explained in [1]_. This is an internal version to be used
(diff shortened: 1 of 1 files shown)


arviz/stats/density_utils.py
```diff
@@ -36,10 +36,9 @@ def _bw_silverman(x, x_std=None, **kwargs):  # pylint: disable=unused-argument
 def _bw_isj(x, grid_counts=None, x_std=None, x_range=None):
     """Improved Sheather-Jones bandwidth estimation.
 
-    Improved Sheather and Jones method as explained in [1]_.
-    This is an internal version pretended to be used by the KDE estimator.
-    When used internally computation time is saved because things like minimums,
-    maximums and the grid are pre-computed.
+    Improved Sheather and Jones method as explained in [1]_. This is an internal version to be used
```
agustinaarroyuelo, line 39: Improved Sheather and Jones method as explained in [1]_. This method is used internally by the KDE estimator, resulting in saved computation time as minimums, maximums and the grid are pre-computed.

Issue #1560, opened 2021-02-13
Debug print left in release version


rpgoldman:
**Describe the bug**

While invoking `from_pymc3` I got unexpected output:

```
0, dim: AND rows, 1376 =? 1376
0, dim: NAND rows, 1799 =? 1799
0, dim: NOR rows, 1699 =? 1699
0, dim: OR rows, 1370 =? 1370
0, dim: XNOR rows, 1510 =? 1510
0, dim: XOR rows, 1674 =? 1674
```

The strings after `dim:` are names of dimensions in the `InferenceData` object being created. 

I find this in `arviz/data/base.py`:

    print(f"{i}, dim: {dim}, {dim_size} =? {len(coords.get(dim, []))}") 

See https://github.com/arviz-devs/arviz/blob/718eb31ed844c3bb71d564736af54e203ca0dc3a/arviz/data/base.py#L116

This line should either be deleted or replaced by an equivalent logging function call.

This line in the code is gated by the parameter `skip_event_dims` which is not explained in the docstring.  See https://github.com/arviz-devs/arviz/blob/718eb31ed844c3bb71d564736af54e203ca0dc3a/arviz/data/base.py#L73

I'm actually not even sure what an event dimension is, as opposed to any other kind of dimension, so some more information about this parameter would be very helpful.

OriolAbril:
Extra info on `event_dim` in addition to the explanations in the PR (https://github.com/arviz-devs/arviz/pull/1429) that introduced the changes:

I took the naming from tensorflow probability, they define:

> * Event shape describes the shape of a single draw from the distribution; it may be dependent across dimensions. For scalar distributions, the event shape is []. For a 5-dimensional MultivariateNormal, the event shape is [5].
> * Batch shape describes independent, not identically distributed draws, aka a "batch" of distributions.
> * Sample shape describes independent, identically distributed draws of batches from the distribution family.

So for example, if we model our 50 observations as 3d multivariate normals, and took 4 chains, 200 draws from the posterior, then the shape of the posterior predictive would be (4, 200, 50, 3) with (4, 200) being the sample shape, 50 the batch shape and 3 the event shape. log likelihood however will not have the same shape, as the event shape is reduced. Before that PR, things worked as long as no dims were passed to the converter (the defaults are generated on a per group basis) but would fail if dims were passed given that the variables in posterior predictive and in log likelihood have the same name

OriolAbril:
[...]

PR #1746, opened 2021-07-25
Disallow print statements for debugging

MarcoGorelli:
## Checklist

Following up on @canyon289 's suggestion - this is to avoid problems like https://github.com/arviz-devs/arviz/issues/1560

### DIFF
--- a/.azure-pipelines/azure-pipelines-external.yml
+++ b/.azure-pipelines/azure-pipelines-external.yml
@@ -133,6 +133,10 @@ jobs:
     displayName: 'Use relative imports'
 
   - script: |
+      python -m madforhooks.no_print_statements $(find arviz -name '*.py')
+    displayName: 'Disallow debugging print statements (use `file=sys.stdout` if not debugging)'
+
+  - script: |
       python -m mypy .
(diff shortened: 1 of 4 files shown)


canyon289:
I think we should probably disable prints entirely and just use logging. A little outside of this PR but thoughts?

canyon289 (approved):

OriolAbril:
> One question related I think we should probably disable prints entirely and just use logging. A little outside of the scope of PR but thoughts?

I think we already use only logging and warnings for all our messages. IIUC, the two modified lines that add `file=sys.stdout` are the only print occurrences in the codebase, and they are two tests.

ahartikainen:
Is this PR adding a check that there are no print statements in the codebase?

Why do we need to edit tests? They are hidden by default and when test fails, the printed parts are shown (also -s should show the prints when running locally)

Can we skip checking test folder?

MarcoGorelli:
Sure, done (bit unwieldy without pre-commit, but `-not -path` works)

---

(did I really accidentally open a PR from my `main` branch? oops :facepalm: )

Issue #1132, opened 2020-03-31
Update ArviZ data examples


OriolAbril:
## Tell us about it

inference data objects loaded with `load_arviz_data` have to be updated to move log likelihood data to `log_likelihood` group, add `constant_data` where relevant (i.e. x in regression, sigma in eight schools), maybe also add a couple of examples having predictions and warmup data? And an example from PyStan with different names in observed_data and posterior predictive?

smit-s:
If no one is working on it I would like to work on it. 
Also, I would like to get a more elaborate explanation please.

OriolAbril:
ArviZ has several inference data objects that can be loaded with `load_arviz_data`. These are useful for examples documentation, testing, showcasing ArviZ capabilities and best practices. They can also be used by prospective new users to experiment with ArviZ without having to invest too much learning about converters nor running simulations to visualize. 

Some examples come from the Cookbook and Quickstart guide from ArviZ docs and are packaged in [ArviZ itself](https://github.com/arviz-devs/arviz/tree/master/arviz/data/_datasets), some others come from PyMC3 docs or other sources and are retrieved from figshare. cc @ColCarroll 

All datasets have to be updated to move `log_likelihood` data from `sample_stats` group to the new `log_likelihood` group, some of them should also add the `constant_data` group to inference data object (i.e. x values in regression, sigma in eight schools...). 

Moreover, the set of available datasets should be extended (or use the current examples to take all examples into account) to include examples of the following:
* `predictions` and `predictions_constant_data` group
* need to use `data_pairs` in [`plot_ppc`](https://arviz-devs.github.io/arviz/generated/arviz.plot_ppc.html#arviz.plot_ppc) (this is a must in PyStan)
* warmup data -> https://github.com/arviz-devs/arviz/pull/1126

This issue requires being comfortable with converter functions, inference data (and the [schema specification](https://arviz-devs.github.io/arviz/schema/schema.html)) and inference libraries (at least PyMC3 and PyStan), ArviZ alone is not enough.

smit-s:
Thanks for the explanation @OriolAbril . To be honest, I might take some time to get familiar to PyMC3 and PyStan so, if this issue is suppose to be solved urgently, maybe I would not consider myself best choice for solving it otherwise I would like to give it a try.

canyon289:
Doesnt need to be solved urgently! Take your time. I'll assign this issue to you to ensure folks know its being addressed.

Thanks in advance @smit-s

[...]

PR #1747, opened 2021-07-26
[WIP] Prototype for timeseries plots

utkarsh-maheshwari:
## Description

Timeseries and components plots prototype. 
High level design:

- idata : az.InferenceData object
- y : str, variable name from observed_data
- x : str, Optional. If none, coords name of y (y should be DataArray).
- components : list, variables from posterior, Optional. Shape of the components must be like (chain, draw, *)
- holdout : int, Optional. No of samples from end. (Could be modified to take particular coordinate val, then find the value in `x` )
- (Doesn't handle multidim data till now. Plot_dim should be added? Plotting components could be complicated. Because posterior dims may vary from observed_data dims. )

Function call:
`plot_ts(idata=idata, y="y", components=["trend", "seasonality"], holdout = 120)`

Plots:
image

Ps: default kwargs are yet to be added in matplotlib function.

## Checklist


utkarsh-maheshwari:
Also, I guess I am not authorized to add labels. Hence, this and other PR is label-less.

arviz/plots/tsplot.py
```diff
@@ -0,0 +1,142 @@
...
+            component_mean = component_xr.mean(("chain", "draw"))
+            components_plotters += list(
+                xarray_var_iter(
+                    component_mean,
+                    skip_dims=set(component_mean.dims),
+                    combined=True,
+                )
+            )
+
+            components_satcked = component_xr[..., -holdout:].stack(sample=("chain", "draw"))[
```
OriolAbril, line 77: this assumes that the dim on which holdout should be done is the last one, which may not be the case. It should be `.isel(dim_name=slice(None, holdout)`

Also, the name of the stacked dimension should be `__sample__` instead to avoid name clashes. see #1613 for context

arviz/plots/tsplot.py
```diff
@@ -0,0 +1,142 @@
...
+                ..., pp_sample_ix
```
OriolAbril, line 78: Stacking makes sure the newly generated dimension is the last one, but even so I think using `isel` would be better.

OriolAbril:
> holdout : int, Optional. No of samples from end. (Could be modified to take particular coordinate val, then find the value in x )

I would have the holdout be separate arrays, both x and y (and optionally observations which I hardly ever expect to be known in this setting). The reason for that is that while it is possible that both posterior predictive samples for observed data and out of sample posterior predictive samples (often called predicions or forecasts) are stored in the same array, I expect them being in different arrays to be more common. Moreover, going from single to multiple array is a "simple" slice operation whereas from multiple to single requires concatenation/stacking.

utkarsh-maheshwari:
@OriolAbril 
You are right. I gave it a thought but implement this one first as it was easier 😛. 
If we use holdout like it then it should be like : 

If the user split data into train and test,

y: y_train, from observed_data, 
x: x_train, from constant_data or observed_data dims.
y_ppc: y_train_pred (bayesian models exclusive), from posterior_predictive
x_holdout: x_test, from constant_data or posterior_predictive dims
y_holdout: y_test, from observed_data (or constant data)
y_ppc_holdout: y_test_pred, from posterior_predictive 

We'll plot whatever information user provides.

arviz/plots/tsplot.py
```diff
@@ -0,0 +1,142 @@
...
+    holdout=-1,
+    num_samples=100,
+    backend=None,
+    backend_kwargs=None,
+    textsize=None,
+    figsize=None,
+    axes=None,
+    show=None,
+):
+    """Plot timeseries and it's components.
```
canyon289, line 23:
```suggestion
    """Plot timeseries and its components.
```

utkarsh-maheshwari:
Some points to be noted:


> * y: y_train, from observed_data,
> * x: x_train, from constant_data or observed_data dims.
> * x_holdout: x_test, from constant_data or posterior_predictive dims
> * y_holdout: y_test, from observed_data (or constant data)
> * y_ppc: y_train_pred (bayesian models exclusive), from posterior_predictive, default name same as y
> * forecast: y_test_pred, from posterior_predictive, default name same as y_holdout


Started implementing it. One possible drawback of using this design is we would be generating so many of plotters arrays.
Estimated number is 
1+1+1+1+2+2 (+4 plotters for components) = 12. 

Is this a concern with respect to complexity and maintainability ?

utkarsh-maheshwari:
It is getting a bit complex. So, I am assuming data is 1D for now. Will extend it for multi dim data once it is approved.

utkarsh-maheshwari:
Also not focusing on components plots for now. It is getting complex because there are a lot of cases to consider in this design. For example, when y_hat and y_forecast both are provided, we'll be plotting only y_forecast uncertainty (that is uncertainty in the forecasted data) and not y_hat (Uncertainty in the observed data).
However, if y_forecast is not provided, we'll be plotting uncertainty in observed data.

So, making the plotters to pass them to the backend function is getting complex.

arviz/plots/tsplot.py
```diff
@@ -0,0 +1,317 @@
...
+    figsize : tuple, optional
+        Figure size. If None it will be defined automatically.
+    textsize : float, optional
+        Text size scaling factor for labels, titles and lines. If None it will be
+        autoscaled based on figsize.
+
+
+    Returns
+    -------
+    axes: matplotlib axes or bokeh figures.
```
canyon289, line 64: Quick initial comment, we should include examples somewhere, whether here or in the examples folder

arviz/plots/tsplot.py
```diff
@@ -0,0 +1,430 @@
...
+    backend=None,
+    backend_kwargs=None,
+    y_kwargs=None,
+    y_hat_plot_kwargs=None,
+    y_mean_plot_kwargs=None,
+    vline_kwargs=None,
+    textsize=None,
+    figsize=None,
+    legend=True,
+    grid=False,
```
OriolAbril, line 30: same comment I did for `plot_lm`. Is the grid *vital* to the plot? The default should be to use or not use a grid based on the matplotlib style users have chosen, not only it's easy to set, it will affect all plots at once without needing any extra argument nor call to `plt.` after calling ArviZ. Adding grid here means we want the grid to be/not be present and believe this to be important enough to overwrite the defaults set by the user.

arviz/plots/tsplot.py
```diff
@@ -0,0 +1,430 @@
...
+    axes=None,
+    show=None,
+):
+    """Plot timeseries and it's components.
+
+    Parameters
+    ----------
+    idata : InferenceData
+    y : str, variable name from observed_data
```
OriolAbril, line 39:
```suggestion
    y : str
        variable name from observed_data
```

the "first line" should have only the argument name, type info and optional/default info. Any extra description should go in the "second line/paragraph" that is indented.

arviz/plots/tsplot.py
```diff
@@ -0,0 +1,430 @@
...
+    y_forecasts : str, from posterior_predictive, optional
+        Assumed shape (chain, draw, *)
+    x_holdout : str, from constant_data or y_holdout coords, optional
+        If None, coords of y_holdout or coords of y_forecast is chosen.
+    plot_dim: str, Optional
+        Necessary to choose x if x is None & if y is multidimensional.
+    holdout_dim: str, Optional
+        Necessary to choose x_holdout if x is None
+        and if y_holdout or y_forecasts is multidimensional.
+    num_samples : int, Optional, Default 100
```
OriolAbril, line 55:
```suggestion
    num_samples : int, default 100
```

in numpydoc style, using the `default xxx` idiom already indicates it's an optional argument because if it's not passed it will fall back to the default. The convention is also to use lowercase

arviz/plots/tsplot.py
```diff
@@ -0,0 +1,430 @@
...
+    backend : str, Optional
+        Select plotting backend {"matplotlib","bokeh"}. Default "matplotlib".
```
OriolAbril, line 57:
```suggestion
    backend : {"matplotlib","bokeh"}, default "matplotlib"
        Select plotting backend
```

OriolAbril (commented):
added some nitpicky comments, mostly about numpydoc standard: https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard. They should not be too lengthy to take care of

arviz/plots/tsplot.py
OriolAbril: examples should go in both places ideally, not necessarily in this PR. Also, if you liked one model from ROS or another resource and used it as example for testing, it could be added as an example to ArviZ so it's available via `load_arviz_data`. Tests can't use examples that require internet connection but docs do. See #1132

canyon289: More general outside of this PR @utkarsh-maheshwari it would be good to have a medium length blog post explaining this and plot_lm using real world examples. Itll be a good exercise for you to practice technical communication in a more general audience voice, and be beneficial to the community. Importantly it will help us advertise your work so people are both well aware of it and how to use it!

arviz/plots/tsplot.py
```diff
@@ -0,0 +1,430 @@
...
+    idata : InferenceData
+    y : str, variable name from observed_data
+    x : str, Optional
+        If none, coords of y dims
+    y_hat : str, from posterior_predictive, optional
+        Assumed to be of shape (chain, draw, *y.dims)
+    y_holdout : str, from observed_data, optional
+        Observed data after holdout
+    y_forecasts : str, from posterior_predictive, optional
+        Assumed shape (chain, draw, *)
```
canyon289, line 47: This is an example for my comment above. I'm not entirely certain what a holdout vs forecast is just reading the docstring

arviz/plots/tsplot.py
```diff
@@ -0,0 +1,430 @@
...
+    textsize : float, optional
+        Text size scaling factor for labels, titles and lines. If None it will be
+        autoscaled based on figsize.
+
+
+    Returns
+    -------
+    axes: matplotlib axes or bokeh figures.
+
+    Examples
```
canyon289, line 81: Thanks for adding these!

arviz/plots/tsplot.py
```diff
@@ -0,0 +1,430 @@
...
+        #     x_holdout_var_names = x_holdout
+        #     x_holdout = idata.constant_data
+
+    # Assign values to y_holdout
+    if isinstance(y_holdout, str):
+        y_holdout = idata.observed_data[y_holdout]
+        if len(y_holdout.dims) > 1 and holdout_dim is None:
+            raise ValueError("Argument holdout_dim is needed in case of multidimentional data")
+
+        # Assign values to x_holdout
```
canyon289, line 227: Same question here

arviz/plots/tsplot.py
utkarsh-maheshwari: Great advice! We have 10 days left for GSoC, which I would like to dedicate to documentation, examples and blog posts. Also, I will be available even after GSoC to maintain the functions.

arviz/plots/tsplot.py
utkarsh-maheshwari: Oh.. sure!

arviz/plots/tsplot.py
utkarsh-maheshwari: Oh, I miss that! I was over-concerned about the functionality and forgot to double-check the docstring! Thanks for pointing it out!

arviz/plots/tsplot.py
utkarsh-maheshwari: Oh, that's an important observation! Thanks!

arviz/plots/tsplot.py
```diff
@@ -0,0 +1,430 @@
...
+        if "posterior_predictive" not in idata.groups():
+            warnings.warn("posterior_predictive not found in idata", UserWarning)
+            y_forecasts = None
+        elif hasattr(idata.posterior_predictive, y_forecasts):
+            y_forecasts = idata.posterior_predictive[y_forecasts]
+        else:
+            warnings.warn("y_hat not found in posterior_predictive", UserWarning)
+            y_forecasts = None
+
+        # Assign values to x_holdout
```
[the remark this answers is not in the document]
utkarsh-maheshwari: No, I forgot to remove it. Was more concerned about the functionality!

arviz/plots/tsplot.py
utkarsh-maheshwari: Same answer here

arviz/plots/tsplot.py
utkarsh-maheshwari: Hmm, Didn't think it in "matplotlib style" way! I do not feel the need of it now.

PR #1753, opened 2021-08-01
Initial prototype of ecdf plot

### LINKED ISSUE #1213: Add plot_ecdf
## Tell us about it

It would be nice to have an empirical CDF (ECDF) plot function that isn't tied to `plot_ppc`. As far as information a `plot_ecdf` might display, I'm thinking it would behave pretty similarly to `plot_density`, but with an ECDF. One extra thing you might want to do is specify some quantiles that would be drawn on the plot. The most common style for ECDF is the line visualization.

ecdf

Another style that we might want to support is one that shows the intervals (e.g. see below from https://statisticsglobe.com/empirical-cumulative-distribution-function-ecdf-r)
R-Basics-the-dim-Function-in-R
However, in my experience, this is only useful for small sample sizes, and in general for posterior samples, the line is cleaner.

## Thoughts on implementation

It's already possible to compute the ECDF (we already have `arviz.plots.backends.matplotlib.ppcplot._empirical_cdf` and
`arviz.plots.backends.bokeh.ppcplot._empirical_cdf`), so this should be pretty straightforward to implement.

### LINKED ISSUE #1663: Implement simultaneous confidence bands for ECDF of PIT
This is related to the paper https://arxiv.org/abs/2103.10522 and the R code is https://github.com/TeemuSailynoja/simultaneous-confidence-bands

Rishabh261998:
## Description

Fixes #1213, Fixes #1663

This is the initial prototype of ECDF and ECDF-Difference Plots with simultaneous confidence for samples as well as PIT of sample bands. So the `values` argument is necessary as it is the sample. If we need to compare this sample to a distribution or another sample then we provide the `distribution` argument or the `values2` argument. 

Function call:

```
import arviz as az
from scipy.stats import uniform, binom, norm

sample = norm(0,1).rvs(1000)
distribution = norm(0,1)

az.plot_ecdf(sample)
```
Screenshot from 2021-08-02 01-24-51

`az.plot_ecdf(sample, distribution = distribution.cdf, confidence_bands = True)`

Screenshot from 2021-08-02 01-26-07

`az.plot_ecdf(sample, distribution = distribution.cdf, confidence_bands = True, difference = True)`

Screenshot from 2021-08-02 01-26-35

`az.plot_ecdf(sample, distribution = distribution.cdf, confidence_bands = True, difference = False, pit = True)`

Screenshot from 2021-08-02 01-27-16

`az.plot_ecdf(sample, distribution = distribution.cdf, confidence_bands = True, difference = True, pit = True)`

Screenshot from 2021-08-02 01-26-54

Similarly, graphs when comparing to `values2` instead of a `distribution`. Also, we can provide `distribution` or `values2` from a different distribution then the plots depict that the sample does not belong to that distribution.

For eg:

```
sample = norm(0,1).rvs(1000)
distribution = norm(0,2)

az.plot_ecdf(sample, distribution = distribution.cdf, confidence_bands = True, difference = True)
```
Screenshot from 2021-08-02 01-33-46

`az.plot_ecdf(sample, distribution = distribution.cdf, confidence_bands = True, difference = True, pit = True)`

Screenshot from 2021-08-02 01-34-07

## Checklist


OriolAbril (commented):
Did a very quick look through the code. The plots in the description look amazing and the flexibility of ecdf, diff and pit is awesome. 

Here are the main high level comments. It looks like plotting can basically be reduced to `step` for either ecdf or ecdf diff, and `fill_between` or `plot` for confidence bands. There are some small differences like the `/ n` term but it looks like all ecdf computation which is common for all backends can be moved to the general plotting function so that we don't need to duplicate the code.

Likewise for the `compute_ecdf` function which should go either in the general function or even in a higher level file (in case there were imports when using the `compute_ecdf` from plot_loo_pit or plot_ppc). I also don''t completely understand why we distinguish between array and list given that we return an array either way, so the first step could probably be `np.asarray` to make sure we always work with arrays. 

Performance wise two points to take into account. Like we commented on slack, I think the default should be to evaluate the ecdf on the samples themselves instead of on a different set of linspaced values, in which case the ecdf computation doesn't require any while. The compute_ecdf could easily distinguish between the two cases by making `z` optional and having an `if z is None` use the algorithm currently implemented in plot_ppc and plot_loo_pit, otherwise use this version. 

It also looks like the computation can be accelerated using numba here, but I'm not sure this is within the scope of your project. Both doing and not doing this are perfectly fine and valid option, I only point this out so you open an issue about using numba in compute_ecdf in case you don't plan to do it in this PR so we don't forget about it.

Rishabh261998:
Thanks, @OriolAbril  for the suggestions. First I am thinking to add all the computation code in the general function and writing the broken variation. After this, we can look into the performance of the compute_ecdf and also how numba can be used to speed up if time permits. Also, there is a more optimized code for computing gamma as presented in that paper, could also implement that.

Rishabh261998:
@sethaxen  and @aloctavodia please go through it whenever you get time. Thanks

arviz/plots/ecdfplot.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Returns
+    -------
+    axes : matplotlib axes or bokeh figures
+    """
+    if values2 is None and distribution is None and confidence_bands is True:
+        raise ValueError("For confidence bands you need to specify values2 or the distribution")
+
+    if distribution is not None and values2 is not None:
+        raise ValueError("To compare sample we need either distribution or values2 and not both")
```
sethaxen, line 70: Can you rewrite in 2nd person?

arviz/plots/ecdfplot.py
```diff
@@ -0,0 +1,107 @@
...
+    distribution : function, optional
+        Cumulative distribution function of the distribution we will compare our sample to
+    pit : bool, optional
+        If True plots the ECDF or ECDF-diff of PIT of sample
+    confidence_bands : bool, optional
+        If True plots the simultaneous confidence bands with 1 - alpha confidence level
+    granularity : int, optional
+        This denotes the granularity size of out plot
+    num_trails : int, optional
+        The number of trials to carry while finding gamma
```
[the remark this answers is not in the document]
sethaxen: It seems like it would be worthwhile to have an argument like `simultaneous=True` or `pointwise=False` that computes the pointwise (ie. marginal) confidence bands.

arviz/plots/ecdfplot.py
```diff
@@ -0,0 +1,107 @@
...
+import numpy as np
+
+from arviz.rcparams import rcParams
+from arviz.plots.plot_utils import get_plotting_function
+
+
+def plot_ecdf(
+    values,
+    values2=None,
+    distribution=None,
```
sethaxen, line 11: Since this is expected to be a CDF, what about naming this something like `cdf` or `cdf_compare`?

arviz/plots/backends/matplotlib/ecdfplot.py
```diff
@@ -0,0 +1,189 @@
...
+            values = (
+                distribution(values)
+                if distribution
+                else compute_ecdf(values2, values) / len(values2)
+            )
+        else:
+            x = np.linspace(values[0], values[-1], granularity)
+            z = distribution(x) if distribution else compute_ecdf(values2, x) / len(values2)
+
+        gamma = get_gamma(n, z, granularity, num_trials, alpha)
```
sethaxen, line 57: `compute_gamma` is more descriptive, since we're not merely getting it.
```suggestion
        gamma = compute_gamma(n, z, granularity, num_trials, alpha)
```

arviz/plots/backends/matplotlib/ecdfplot.py
```diff
@@ -0,0 +1,189 @@
...
+        lower, higher = get_lims(gamma, n, z)
+
+        x_coord, y_coord = [], []
+        if not difference:
+            for _, x_i in enumerate(x):
```
sethaxen, line 62: When plotting a non-difference ECDF where the first point is the minimum of the sample points, you want to add an additional point to the beginning so that the ECDF plot hits the y-axis.

arviz/plots/backends/matplotlib/ecdfplot.py
```diff
@@ -0,0 +1,189 @@
...
+                f_x_i = compute_ecdf(values, x_i) / n
+                x_coord.append(x_i)
+                y_coord.append(f_x_i)
+
+            plt.step(x_coord, y_coord, where="post")
+
+            if ecdf_fill:
+                plt.fill_between(x, lower / n, higher / n, color="C0", alpha=0.2)
```
sethaxen, line 70: Can the color and alpha be set from the `plot_ecdf` function somehow?

arviz/plots/backends/matplotlib/ecdfplot.py
```diff
@@ -0,0 +1,189 @@
...
+            else:
+                plt.plot(x, lower / n, x, higher / n, color="C0", alpha=0.2)
+        else:
+            for _, x_i in enumerate(x):
```
sethaxen, line 74: It would be preferable if the logic of the loop that incrementally computes the ECDF could be its own function. Then this complicated expression and the code below could be much simpler.

arviz/plots/backends/matplotlib/ecdfplot.py
```diff
@@ -0,0 +1,189 @@
...
+
+            plt.step(x_coord, y_coord, where="post")
+
+    if backend_show(show):
+        plt.show()
+
+    return ax
+
+
+def compute_ecdf(sample, z):
```
sethaxen, line 137: Technically this is not computing the ECDF; it's computing the rank of `z` in `sample` (because it doesn't divide by the denominator). Should the name reflect that?

sethaxen (changes requested):
In my first pass, the plots look great, and this PR is coming along nicely.

I had a number of suggestions. I also agree with @OriolAbril that as much generic code as possible should be moved out of the backend code. I also agree that when the user doesn't specify evaluation points, the ECDF can be computed much more efficiently without the `while` loop, and it would be nice to take advantage of that.

arviz/plots/backends/matplotlib/ecdfplot.py
```diff
@@ -0,0 +1,189 @@
...
+        if pit:
+            x = np.linspace(1 / granularity, 1, granularity)
+            z = x
+            values = (
+                distribution(values)
+                if distribution
+                else compute_ecdf(values2, values) / len(values2)
+            )
+        else:
+            x = np.linspace(values[0], values[-1], granularity)
```
Rishabh261998, line 54: @sethaxen even if we change this line and keep x as `values` still our main concern is the `get_gamma` function due to which the plot is slow apart from that the plot does not take even 0.2s. But that function would not be affected even if we set x as `values` rather than equally spaced points between values[0] and values[-1]. In get_gamma we need to get the ecdf values for the uniform sample w.r.t `z`. Please take a look into this and tell me if I am correct or not. So I think what @OriolAbril was saying would speed up some parts of the code apart from get_gamma but they are already fast so we wont achieve anything.

Rishabh261998:
Thanks @sethaxen I have started making the suggested changes and will soon make a commit.

arviz/plots/backends/matplotlib/ecdfplot.py
Rishabh261998: I will add plot_kwargs, fill_kwargs so that the plots can be customized

arviz/plots/backends/matplotlib/ecdfplot.py
```diff
@@ -0,0 +1,189 @@
...
+
+    if confidence_bands:
+        if pit:
+            x = np.linspace(1 / granularity, 1, granularity)
+            z = x
+            values = (
+                distribution(values)
+                if distribution
+                else compute_ecdf(values2, values) / len(values2)
+            )
```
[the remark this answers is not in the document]
Rishabh261998: Ok will do

arviz/plots/backends/matplotlib/ecdfplot.py
```diff
@@ -0,0 +1,189 @@
...
+    backend_kwargs.setdefault("figsize", figsize)
+    backend_kwargs["squeeze"] = True
+
+    if ax is None:
+        _, ax = create_axes_grid(1, backend_kwargs=backend_kwargs)
+
+    n = len(values)
+
+    if confidence_bands:
+        if pit:
```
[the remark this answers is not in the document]
Rishabh261998: Do you mean the `if confidence_bands` block?

arviz/plots/backends/matplotlib/ecdfplot.py
Rishabh261998: Just to be sure, here you mean that I can add a point before the minimum so that the ecdf at that point is 0?

arviz/plots/backends/matplotlib/ecdfplot.py
Rishabh261998: Do you mean the for loop on 74th line? Here I am finding the difference in ecdf and ecdf of uniform distribution

arviz/plots/backends/matplotlib/ecdfplot.py
Rishabh261998: Changed, I have added the it in the denominator rather than explicitly doing it outside

arviz/plots/ecdfplot.py
Rishabh261998: I think plot_ecdf is good as we are using the ecdf values to compare, not sure though.

arviz/plots/ecdfplot.py
```diff
@@ -0,0 +1,107 @@
...
+    alpha=0.95,
+    figsize=None,
+    ecdf_fill=True,
+    ax=None,
+    show=None,
+    backend=None,
+    backend_kwargs=None,
+    **kwargs
+):
+    """Plot ECDF and ECDF-Difference Plot with Confidence bands.
```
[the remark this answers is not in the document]
Rishabh261998: Done

arviz/plots/ecdfplot.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Parameters
+    ----------
+    values : array-like
+        Values to plot from an unknown continuous distribution
```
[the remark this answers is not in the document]
Rishabh261998: Thanks will change

arviz/plots/ecdfplot.py
```diff
@@ -0,0 +1,107 @@
...
+    values2 : array-like, optional
+        Values to compare to the original sample
```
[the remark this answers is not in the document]
Rishabh261998: So by default the `confidence_bands` argument is false so that when you are doing just plot_ecdf(x,y) it plots the normal ecdf plot of x rather than comparing it with y and plotting the confidence intervals. So should I set the default value of `confidence_bands` to true?

arviz/plots/ecdfplot.py
```diff
@@ -0,0 +1,107 @@
...
+    distribution : function, optional
+        Cumulative distribution function of the distribution we will compare our sample to
+    pit : bool, optional
+        If True plots the ECDF or ECDF-diff of PIT of sample
+    confidence_bands : bool, optional
```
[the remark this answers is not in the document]
Rishabh261998: The paper writes it as `Simultaneous confidence bands` and also we are finding the band for each evaluation point so I think bands is better

arviz/plots/ecdfplot.py
```diff
@@ -0,0 +1,107 @@
...
+        If True plots the simultaneous confidence bands with 1 - alpha confidence level
+    granularity : int, optional
+        This denotes the granularity size of out plot
```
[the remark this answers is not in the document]
Rishabh261998: *our plot, granularity size is the number of evaluation points for our ecdf or ecdf difference plot

arviz/plots/ecdfplot.py
```diff
@@ -0,0 +1,107 @@
...
+    num_trails : int, optional
+        The number of trials to carry while finding gamma
+    alpha : float, optional
```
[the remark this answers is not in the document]
Rishabh261998: Will need to look into how to define `alpha` not sure till now. In the paper, it is defined as this 
Screenshot from 2021-08-13 01-22-31

arviz/plots/ecdfplot.py
Rishabh261998: Done

arviz/plots/backends/matplotlib/ecdfplot.py
OriolAbril: > the plot does not take even 0.2s

But this is only one plot, imagine a user wants to plot the ecdf of 20 variables, also not sure how many samples you used for benchmarking, but we could easily have >1e6 samples together with 20, 40 or more variables to plot. In those cases the waiting time will be more than noticeable and it probably doesn't need to be. We want ArviZ to provide tools for direct exploratory analysis but also to provide building blocks for custom plots and workflows, which could even be online diagnostics during mcmc sampling or interactive tools for model checking or prior elicitation. 

> But that function would not be affected even if we set x as values rather than equally spaced points between values[0] and values[-1]. In get_gamma we need to get the ecdf values for the uniform sample w.r.t z

I don't have the algorithm in my mind, so I don't really know if `get_gamma` needs to evaluate the uniform samples on the same `z` as the ecdf of the provided samples, in which case we'd have 1 ecdf computed with the fast sorting algorithm for num_trials computed with the while algorithm, and the difference in computation time would be completely negligible, but I think this misses the point and the goal I was commenting above. In many cases we'll want an ecdf without bands (we already do that in plot_ppc for example) and again we can have many ecdfs to plot with potentially many samples.

OriolAbril (commented):
I am again focusing more on performance and computation because I think that other than documentation and Seth's comments, the functionality is already great. I answered one comment which I hope helps provide some more context as to why I think performance matters even if the plot is already fast as is.

I also noticed a lot of empty lists that are then filled during for loops. I don't know what are the ranges of possible lengths of these lists, but this is an antipattern with very bad performance memory wise for large lists. If we know the length/shape of these list/arrays, which looks like the case, they should be initialized as `np.empty` and then filled during the loop. 

Otherwise, what python ends up doing is at every iteration, create a new longer list, copy all the contents and then write the new item on the last position of the new list. We had several cases of this a while back which was often problematic and progressively updates as many as possible. https://github.com/arviz-devs/arviz/pull/1082#issuecomment-590073748 for example is one of the PRs updating this pattern and has some memory profiling too.

arviz/plots/backends/matplotlib/ecdfplot.py
Rishabh261998: So @OriolAbril  I did an analysis of the plot_ecdf with and without computing the bands with a sample size of 1000, granularity(number of evaluation points): 100, num_trials:500. 
Now without confidence bands(no compute_gamma) the entire plot takes around `0.0117s` for plotting ecdf of the sample or the PIT. 
Now when `confidence_bands=True` then the entire plot takes `7.7878s` and the compute_gamma in itself takes `7.7758s`. Now we were talking about using evaluation points s.t we could find compute ecdfs as suggested in `plot_ppc`, but after analysing I found that the `compute_ecdf` which is used inside the `compute_gamma` takes just `0.0812s` which is negligible as compared to the time taken by entire compute_gamma, so rather than thinking to optimize the `compute_ecdf` we should think of optimizing the part of `compute_gamma` apart from `compute_ecdf`.

Rishabh261998: Also I will use `np.empty` rather than initializing the list as empty and then filling it during the loop. Thanks for this

Rishabh261998:
@sethaxen I have made the suggested changes, just the performance of the plot is the main issue now. Currently analysing each part and seeing if something can be optimized

arviz/plots/backends/matplotlib/ecdfplot.py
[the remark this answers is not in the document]
sethaxen: No, the `if pit` block.

arviz/plots/ecdfplot.py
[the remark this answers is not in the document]
sethaxen: Ah, so you're saying that when `confidence_bands` is `True`, the confidence band is determined by comparing `values` and `values2`? What does this correspond to in the paper?

arviz/plots/ecdfplot.py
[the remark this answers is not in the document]
sethaxen: Yeah, in general, I think terms like "type I error rate" are technically correct but not descriptive, and for docstrings we want to be descriptive (while perhaps also technical). So I would recommend `fpr`.

arviz/plots/ecdfplot.py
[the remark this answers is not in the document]
sethaxen: Ah, okay, why not something like `npoints` to be more descriptive?

arviz/plots/ecdfplot.py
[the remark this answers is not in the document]
sethaxen: Okay, makes sense.

arviz/plots/ecdfplot.py
sethaxen: I mean the highlighted line, not the function name. Should we rename `distribution` to `cdf`?

arviz/plots/backends/matplotlib/ecdfplot.py
sethaxen: That's right, yes. Usually this is done by either adding `x=-Inf`, `y=0` (some step function implementations will just draw a vertical line to the `y`-axis in that case), or just duplicating the lowest `x` value and prepending `y=0`.

arviz/plots/backends/matplotlib/ecdfplot.py
sethaxen: Yes, but you can do that by 1. constructing the ECDFs and then 2. computing their difference. And all of the blocks that construct an ECDF by looping could be replaced with a single function call that does that, which would simplify the code. Also, that function is then a target for future efficiency improvements.

sethaxen:
@Rishabh261998 can you resolve the comments above that have been addressed so we know what is still open for discussion when we review the code?

arviz/plots/ecdfplot.py
Rishabh261998: yeah that makes sense, thanks

arviz/plots/ecdfplot.py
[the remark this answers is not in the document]
Rishabh261998: So in the paper, they had only plotted ecdf for PIT of the sample, so for calculation of PIT they have used either the distribution or the comparison sample which I have also done. But in my case I am also plotting ecdf for the sample, so for finding the evaluation points for the uniform distribution while calculating gamma I have used the cdf(x) where x is equally spaced points between min and max. So basically if we are provided the sample and cdf or the sample and comparision sample then if `confidence_bands` is false then there is no use of cdf or comparison sample and a normal ecdf plot is plotted for the sample.

Rishabh261998:
@sethaxen have made all the changes. Please go through it. Also will add tests in some time.

arviz/plots/ecdfplot.py
```diff
@@ -0,0 +1,341 @@
...
+    gamma = []
+    for _ in range(num_trials):
+        unif_samples = uniform.rvs(0, 1, n)
+        unif_samples = np.sort(unif_samples)
+        gamma_m = 1000
+        ## Can compute ecdf for all the z together or one at a time.
+        f_z = compute_ecdf(unif_samples, z)
+        for i in range(npoints):
+            curr = min(binom.cdf(n * f_z[i], n, z[i]), 1 - binom.cdf(n * f_z[i] - 1, n, z[i]))
+            gamma_m = min(2 * curr, gamma_m)
```
sethaxen, line 332: This vectorized version should be much faster
```suggestion
        f_z = compute_ecdf(unif_samples, z)
        gamma_m = 2 * min(np.amin(binom.cdf(n * f_z, n, z)), np.amin(1 - binom.cdf(n * f_z - 1, n, z)))
```

arviz/plots/ecdfplot.py
```diff
@@ -0,0 +1,341 @@
...
+    else:
+        ## if z is a list then follow this approach
+        f_z = np.empty(len(z))
+        u_idx = 0
+        for i, z_i in enumerate(z):
+            while u_idx < len(sample) and sample[u_idx] < z_i:
+                u_idx += 1
+            f_z[i] = u_idx + 1
+
+        return f_z / len(sample)
```
sethaxen, line 294: I believe this is computing the same thing as `searchsorted`, so we can avoid the Python for loop:
```suggestion
    return np.searchsorted(sample, z, side='right') / len(sample)
```

arviz/plots/ecdfplot.py
```diff
@@ -0,0 +1,341 @@
...
+        ## pseudo point at the start of ecdf plot so that it touches x-axis
+        x_coord[0], y_coord[0] = x[0], 0
+
+    for i, x_i in enumerate(x):
+        f_x_i = compute_ecdf(probs, x_i)
+        if difference:
+            x_coord[i], y_coord[i] = x_i, f_x_i
+        else:
+            x_coord[i + 1], y_coord[i + 1] = x_i, f_x_i
+    return x_coord, y_coord
```
sethaxen, line 312: What about something like this for simplicity?

```suggestion
    y = compute_ecdf(probs, x)

    if not difference:
        x = np.insert(x, 0, x[0])
        y = np.insert(y, 0, 0)
    return x, y
```

arviz/plots/ecdfplot.py
```diff
@@ -0,0 +1,341 @@
...
+        if not difference:
+            ## We store the coordinates of our ecdf in x_coord, y_coord
+            x_coord, y_coord = get_ecdf_points(x, probs, difference)
+        else:
+            ## Here we subtract the ecdf value as here we are plotting the ECDF-difference
+            x_coord, y_coord = get_ecdf_points(x, probs, difference)
+            for i, x_i in enumerate(x):
+                y_coord[i] = y_coord[i] - (
+                    x_i if pit else cdf(x_i) if cdf else compute_ecdf(values2, x_i)
+                )
```
sethaxen, line 211: Loops like this will not be as fast as just computing the reference array and then subtracting it, because NumPy's vectorized functions will be much faster than Python for loops.

sethaxen (changes requested):
Sorry for the delay in reviewing! Besides tests and the suggestions, this PR could use some slimming down. There's a fair amount of redundancy in the control flow that can probably be simplified or abstracted out to small functions for clarity.

In it's current state, the methods are really slow, but the good news is that `np.searchsorted` seems to do everything we need to eliminate the Python for loops.

michaelosthege:
Any news on this PR? It would be great to see this in ArviZ!
A few months ago I tried to implement this method myself, but failed miserably.

sethaxen:
Good question. @Rishabh261998, I'm sure you're busy with other things, but any chance you have the time to push this PR over the finish line? Else one of us can.

Rishabh261998:
Hi @sethaxen, I am really sorry for not completing it.  As of now, I am a bit busy as I have joined a company and also in parallel, I am preparing for an exam. I think you guys can take it as I am not sure when I will get time for this. Thanks

sethaxen:
> Hi @sethaxen, I am really sorry for not completing it. As of now, I am a bit busy as I have joined a company and also in parallel, I am preparing for an exam. I think you guys can take it as I am not sure when I will get time for this. Thanks

No problem at all, @Rishabh261998. Thanks to your excellent work on this, I don't think it will take us much to finish it up. Best of luck on the exam!

arviz/plots/ecdfplot.py
ahartikainen: This is something we can make faster in future PR if needed

ahartikainen:
Thanks for the PR

Issue #1466, opened 2020-12-16
from_cmdstan should accept observed_data in json or as variable


jsevo:
The below fails when providing a path to the data in `.json` format used by `cmdstan` itself. Per docstring, only data in `Rdump` format can be used. `Rdump` is not an ideal default data format.

`from_cmdstan` is lacking documentation; I could not find a workaround.

```
idata = az.from_cmdstan(posterior="trace.csv",
                                           obeserved_data="data.json"
                             )
```

ahartikainen:
Hi, yes we should update our `from_cmdstan` functionality. Would you be interested in adding a json support?

For short term solution I recommend you get posterior with `from_cmdstan` and then add data manually.

    import json
    with open("data.json", "w") as f:
        data_dict = json.load(f)
    idata_data = az.from_dict(observed_data=data_dict)
    idata_cmdstan = az.from_cmdstan(posterior="trace.csv")
    idata = az.concat((idata_cmdstan, idata_data))

cc. @percygautam & @OriolAbril did we have option to use merge too?

OriolAbril:
We added an [`add_groups`](https://arviz-devs.github.io/arviz/api/generated/arviz.InferenceData.add_groups.html#arviz.InferenceData.add_groups) not long ago which I think will be convenient in this case. 

Something like the code below should work.

```
import json
with open("data.json", "w") as f:
    data_dict = json.load(f)
idata = az.from_cmdstan(posterior="trace.csv")
idata.add_groups(observed_data=data_dict)
```

OriolAbril:
closed by 1762

PR #1762, opened 2021-08-23
Add support for JSON format

erocarrera:
This small change adds support to read Stan's JSON data in addition to Rdump. This was requested in #1466 and I've needed it myself. It updates `_read_data()` to parse as JSON any input files with `.json` extension, defaulting to Rdump otherwise.
Hence `observed_data`, `constant_data`, and `predictions_constant_data` all accept JSON inputs now.
I've run `black` and `pylint` on the changes.

Also added a missing bracket in a docstring.

### DIFF
--- a/arviz/data/io_cmdstan.py
+++ b/arviz/data/io_cmdstan.py
@@ -1,5 +1,6 @@
 # pylint: disable=too-many-lines
 """CmdStan-specific conversion code."""
+import json
 import logging
 import os
 import re
@@ -838,7 +839,7 @@ def _process_data_var(string):
 
 
 def _read_data(path):
-    """Read Rdump output to dictionary.
+    """Read Rdump or JSON output to dictionary.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


arviz/data/io_cmdstan.py
```diff
@@ -1,5 +1,6 @@
 # pylint: disable=too-many-lines
 """CmdStan-specific conversion code."""
+import json
```
ahartikainen, line 3:
```suggestion
try:
    import ujson as json
except ImportError:
    import json
```

ahartikainen (commented):
This is a great addition.

We have ujson as an optional dependency so I think we could always try to import it first

arviz/data/io_cmdstan.py
erocarrera: Oh yes! that's great!

ahartikainen:
Great.

Should we add some test for this? something similar that we have for rdump?

erocarrera:
> Great.
> 
> Should we add some test for this? something similar that we have for rdump?

Most definitely. I've added a json version of one of the Rdump files in tests and updated the tests to make sure both pass meet them same assertions.

PR #1764, opened 2021-08-24
Remove constrained_layout from arviz rcparams

### LINKED ISSUE #1763: ArviZ uses `constrained_layout=True` even when disabling it in `az.rcParams`
I'm using the `"arviz-darkgrid"` style and I'm getting a constrained layout even when I attempt to disable it via 

```python
az.rcParams["plot.matplotlib.constrained_layout"] = False
```

**The code**

```python
import arviz as az

az.style.use("arviz-darkgrid")
az.rcParams["plot.matplotlib.constrained_layout"] = False

fig, axes = plt.subplots(1, 2)

axes[0].scatter([1, 2], [1, 2])
axes[1].scatter([1, 2], [1, 2])
fig.subplots_adjust(bottom=0.2)
```

**The result**

image

**The environment**
```python
%load_ext watermark
%watermark -n -u -v -iv -w
```

```
Last updated: Mon Aug 23 2021

Python implementation: CPython
Python version       : 3.8.5
IPython version      : 7.18.1

matplotlib: 3.4.3
arviz     : 0.11.2
json      : 2.0.9

Watermark: 2.1.0
```

aloctavodia:
Fixes #1763 and this is also related to #1247


arviz/plots/backends/matplotlib/elpdplot.py
```diff
@@ -36,7 +36,6 @@ def plot_elpd(
         **backend_kwarg_defaults(),
         **backend_kwargs,
     }
-    backend_kwargs.setdefault("constrained_layout", not xlabels)
```
OriolAbril, line 39: Not sure this has to be removed. IIRC, I added this because when `xlabels=True` `plot_elpd` uses the `fig.autofmt_time` matplotlib helper function which is not compatible with constrained layout, so that aimed to override the style defaults unless set explicitly.

Issue #1767, opened 2021-08-26
Matplotlib backend in JupyterLab Notebooks shows plots even when `show=False`


forsakendaemon:
**Describe the bug**
When generating plots in a Jupyter Notebook, because `create_axes_grid` calls `matplotlib.pyplot.subplots`, the resultant figure is automatically displayed even when `show=False`.

**To Reproduce**
```(python)
import arviz as az
idata = az.from_netcdf("path/to/netcdf.nc")
axarr = az.plot_ess(idata, show=False)
```

**Expected behavior**
Plots created in this way should not be displayed.

**Additional context**
arviz: 0.11.2
matplotlib: 3.4.3
JupyterLab: 3.0.16

Chirag3841:
@forsakendaemon I guess you have to add add plt.close(fig) when show=False.

OriolAbril:
We have removed the show argument in `arviz-plots` so I will close the issue as I also think this is out of scope for `arviz-plots`. We have a `.show` method for cases we need to call it explicitly so it has a similar goal to the show argument, but I think how jupyter interacts with each plotting backend regarding initialized figures where the backend `.show` method is not called should be controlled from the backend or jupyter settings.

PR #1768, opened 2021-08-26
Instantiate Figure and call `.subplots()` on that instead of using the `pyplots` API

forsakendaemon:
## Description
Instead of calling `matplotlib.pyplot.subplots`, this change instantiates a `matplotlib.figure.Figure` directly and then calls `.subplots()` on that Figure. This addresses Issue #1767, where Figures are displayed when `show=False` in JupyterLab notebooks, due to the `pyplot` API doing a bunch of magic in the background.

## Checklist
-->


ahartikainen:
Hmm

now our tests fail due to problem with the use of `.gca()` in plot_kde?

https://github.com/arviz-devs/arviz/blob/b54e0226003c5e4ea49095b29a2ceac6af57a9b5/arviz/plots/backends/matplotlib/kdeplot.py#L60
```
        fig_manager = _pylab_helpers.Gcf.get_active()
        if fig_manager is not None:
            ax = fig_manager.canvas.figure.gca()
```

OriolAbril:
using this `_pylab_helper` feels very sketchy. From reading

* https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.gcf.html
* https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.get_fignums.html#matplotlib.pyplot.get_fignums
* https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.get_current_fig_manager.html

I think we could (or should?) change that to first check with `get_fignums` and then either 1) retrieve the current figure (either with `gcf` or `get_current_fig_manager` which seems like it would need less changes to the code) or 2) call `create_axes_grid`.

arviz/plots/backends/matplotlib/__init__.py
```diff
@@ -52,7 +52,11 @@ def create_axes_grid(length_plotters, rows=1, cols=1, backend_kwargs=None):
 
     backend_kwargs = {**backend_kwarg_defaults(), **backend_kwargs}
 
-    fig, axes = subplots(rows, cols, **backend_kwargs)
+    subplots_kws = {"nrows", "ncols", "sharex", "sharey", "squeeze", "subplot_kw", "gridspec_kw"}
+    fig_kwargs = {key: backend_kwargs[key] for key in backend_kwargs.keys() - subplots_kws}
+    backend_kwargs = {key: backend_kwargs[key] for key in backend_kwargs.keys() & subplots_kws} 
+    fig = Figure(**fig_kwargs)
+    axes = fig.subplots(rows, cols, **backend_kwargs)
```
OriolAbril, line 59: Is this needed to solve the issue? I don't see how this changes the behaviour to users and makes the code more complex and prone to needing maintenance.

ahartikainen: I think only difference is that now jupyter does not automatically plot the figure.

Everything else should work as is

ahartikainen: Hmm, maybe my solution does not actually work, will need to check it

PR #1769, opened 2021-08-26
ppc_plot specify colors for each element

aloctavodia:
Now we can pass a list of 3 colors instead of a single string

### DIFF
--- a/arviz/plots/ppcplot.py
+++ b/arviz/plots/ppcplot.py
@@ -217,6 +217,12 @@ def plot_ppc(
     if colors is None:
         colors = ["C0", "k", "C1"]
 
+    if isinstance(colors, str):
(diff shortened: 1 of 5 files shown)


arviz/plots/ppcplot.py
```diff
@@ -19,7 +19,7 @@ def plot_ppc(
     alpha=None,
     mean=True,
     observed=True,
-    color="C0",
+    colors=None,
```
OriolAbril, line 22: can we keep `color` for a bit and deprecate it so users know they need to change from singular to plural? I expect this change to be quite confusing otherwise.

ahartikainen (commented):
Everything looks ok, but where is the test for making sure that colors list has atleast 3 items?

PR #1775, opened 2021-08-30
Updated Docker Development 

roshnaeem:
**Description:** 
Updated the [Development in Docker](https://arviz-devs.github.io/arviz/contributing/contributing_prs.html#building-documentation-with-docker) part. 

**Note:** I think we should move this part to a separate HTML page, and have the heading on the left sidebar. It will be easy to navigate for newcomers.


canyon289:
The note sounds good! I dont know what it will take, but perhaps something we can explore as part of the docs project?

canyon289 (approved):

doc/source/contributing/contributing_prs.md
```diff
@@ -235,31 +236,33 @@ They are executed on the following order: clear-cache, build, test, docs, shell,
 
 ## Testing in Docker
 Testing the code using docker consists of executing the same file 3 times (you may need root privileges to run it).
-First run `./scripts/container.sh --clear-cache`. Then run `./scripts/container.sh --build`. This starts a local docker image called `arviz`. Finally run the tests with `./scripts/container.sh --test`. This should be quite close to how the tests run on TravisCI.
+First run `./scripts/container.sh --clear-cache`. Then run `./scripts/container.sh --build`. This starts a local docker image called `arviz`. Finally run the tests with `./scripts/container.sh --test`. This should be quite close to how the tests run on [TravisCI](https://travis-ci.org/).
```
OriolAbril, line 239: Minor nit: tests no longer run on Travis. General comment, now that you are modifying this, I would be much more explicit about what each flag does. as an example `--build` only needs to be executed once (or once in a while to update the docker image), **not** every time we want to test or build the docs. I am not sure this is clear to newcomers by reading this alone.

OriolAbril (commented):
I completely agree on moving that to it's own file. 

And we could also do other restructuring to the contributing_prs. For example:

Start the contributing code via pull requests with the info that would be considered "tutorial" according to diataxis. IMO this means the "steps before starting to work" and the "steps" sections which could be merged together too to some extent. 

Then the "how-to" content, the "PR checklist" section. Users who are familiar with git+github, python and ArviZ probably won't need to read the steps section or will only need to do so once. However, the checklist should be read and checked every time a PR is opened, even by core devs, the same way professional airplane pilots go over the take off checklist.

The "development process" summary I think corresponds to "explanation" and could probably be it's own file too (not really sure about that though) that summarizes the (for now very incomplete) developer guide. Again, once contributors are familiar with the content there, they won't really need to read it every time they open a PR unlike the checklist.

Note: I am using diatixis here even though I said that homepage, community, about us and contributing are not within the diatixis scope, which is true but omits some information. diataxis organizes the content around the user/reader, the axis are about how people interact with the content and when the content is useful to them. Therefore, ortanizing the documentation with diatixis means organizing it around _ArviZ users_ needs, and these four pages fall outside the scope. 

At the same time however, the contributing section is also documenting a technical process or craft. It's a different one and has a different target audience, but the contributing section by itself can also follow diataxis. We don't need to obsess over diatiaxis and ensure we follow it to the letter, but I find it a useful tool to keep in mind and I think these diataaxis inspired subdivisions can make the content more clear.

doc/source/contributing/contributing_prs.md
roshnaeem: Yeah, for docs testing we need to use the `--docs` flag. It is stated in the [Building documentation with Docker](https://arviz-devs.github.io/arviz/contributing/contributing_prs.html#building-documentation-with-docker) section. 

Here, do I need to make any changes? Should I remove the TravisCI line?

roshnaeem:
> I completely agree on moving that to it's own file.
> 
> And we could also do other restructuring to the contributing_prs. For example:
> 
> Start the contributing code via pull requests with the info that would be considered "tutorial" according to diataxis. IMO this means the "steps before starting to work" and the "steps" sections which could be merged together too to some extent.
> 
> Then the "how-to" content, the "PR checklist" section. Users who are familiar with git+github, python and ArviZ probably won't need to read the steps section or will only need to do so once. However, the checklist should be read and checked every time a PR is opened, even by core devs, the same way professional airplane pilots go over the take off checklist.
> 
> The "development process" summary I think corresponds to "explanation" and could probably be it's own file too (not really sure about that though) that summarizes the (for now very incomplete) developer guide. Again, once contributors are familiar with the content there, they won't really need to read it every time they open a PR unlike the checklist.
> 
> Note: I am using diatixis here even though I said that homepage, community, about us and contributing are not within the diatixis scope, which is true but omits some information. diataxis organizes the content around the user/reader, the axis are about how people interact with the content and when the content is useful to them. Therefore, ortanizing the documentation with diatixis means organizing it around _ArviZ users_ needs, and these four pages fall outside the scope.
> 
> At the same time however, the contributing section is also documenting a technical process or craft. It's a different one and has a different target audience, but the contributing section by itself can also follow diataxis. We don't need to obsess over diatiaxis and ensure we follow it to the letter, but I find it a useful tool to keep in mind and I think these diataaxis inspired subdivisions can make the content more clear.

Sounds good. We can work on it step by step and open new PRs for every step.

doc/source/contributing/contributing_prs.md
OriolAbril: > Should I remove the TravisCI line?

Travis has to be deleted because it is very outdated at this point. 

> Here, do I need to make any changes?

In general when reading the file, is it clear to you that you have to first build the docker image, then run the tests, or build the docs, or open a shell... and that you don't need to run build every time you do one of these? Was it clear to you when you set up docker? Is it clear that you have to rebuild the docs every time you make a change? Things like that. I think some changes would improve the doc but I'm not sure how many changes nor which ones

roshnaeem: No, it was not clear, and yeah, I thought about changing the docs section, then the thought that we are going to add sections like "How to read/write ArviZ docs" changed my mind. That might be adding the details in that section and adding its link here would be fine. But yes we can add the details here as well. What do you say?

For testing in Docker, we should change this part. Currently, it is giving the feeling that we need to build every time we want to test. But as you are saying that we need to run the `build` once, and only the `test` command again and again for testing. Right?

OriolAbril: Build needs to be executed only "once". It creates the docker image that can then be used to run any of the other commands as many times as desired. It only needs to be reexecuted to update the docker image if for example there have been new releases that we want to incorporate into the image like xarray or sphinx. 

The only command that is a bit special is test. IIUC, you can test with docker as many times as you want but if you test with docker and also locally you should run the clear cache command before every docker test because the cache generated locally is not valid for docker (and maybe viceversa too?)

The need for rebuilding the docs after every change is not docker specific so it doesn't need to be here, especially if it's already in the general doc guide, but I think it won't hurt to repeat that. Everything else does need to be here though because it's docker specific.

PR #1776, opened 2021-08-30
Updated Installation Guide

roshnaeem:
**Description:** Updated the[ Installation guide](https://arviz-devs.github.io/arviz/getting_started/Installation.html) part. 

**Note**: In order to add `python>=3.6`, should I update the `requirements.txt` or add it separately here?

### DIFF
--- a/doc/source/getting_started/Installation.rst
+++ b/doc/source/getting_started/Installation.rst
@@ -60,6 +60,12 @@ The required dependencies for installing ArviZ are:
 
 .. literalinclude:: ../../../requirements.txt
 
+and
+
+.. code:: bash
+
+    python>=3.6
+
 .. _Optional-dependencies:
 
 Optional dependencies
--- a/doc/source/getting_started/Installation.rst
+++ b/doc/source/getting_started/Installation.rst
@@ -24,7 +24,7 @@ Use the below pip command to install ArviZ with all of its :ref:`Optional-depend
 
 .. code:: bash
 
-    pip install arviz all
+    pip install arviz[all]
 
 Using conda-forge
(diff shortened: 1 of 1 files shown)


aloctavodia:
adding it here is OK.

canyon289 (approved):

doc/source/getting_started/Installation.rst
```diff
@@ -20,11 +20,11 @@ Using pip
 
     pip install arviz
 
-Use the below pip command to install ArviZ with all of it's :ref:`Optional-dependencies`.
+Use the below pip command to install ArviZ with all of its :ref:`Optional-dependencies`.
 
 .. code:: bash
 
-    pip install arviz[all]
+    pip install arviz all
```
OriolAbril, line 27: This doesn't work for me at all. I think this change is motivated by zsh shells not working with `pip install arviz[all]` unlike bash shells that do work. I think the fix for that is `pip install "arviz[all]"` instead

OriolAbril (changes requested):

doc/source/getting_started/Installation.rst
roshnaeem: Done.

OriolAbril: just rechecked that and `pip install arviz[all]` doesn't work on a zsh shell. It has to be `pip install "arviz[all]"`

roshnaeem: Yes, I tried it too. Changed, it is working. Thank you.

OriolAbril (approved):

PR #1778, opened 2021-08-30
Miscellanea and doc fixes

OriolAbril:
## Checklist

- [ ] Is the fix listed in the [Documentation](https://github.com/arviz-devs/arviz/blob/main/CHANGELOG.md#documentation) section of the changelog?


OriolAbril:
This is ready to merge. This simplifies the documentation build process so that it depends only on the packages listed on requirements and requirements-docs (where I have added bokeh, the only optional dependency we need). This doesn't mean that we don't show code with external or optional dependencies, only that we do so as notebooks, so doc building doesn't need to execute the code. Therefore, the BuildDocs job now only depends on the BaseTests, which should speed up CI completion. 

I am also adding readthedocs! This is for now experimental but as you can see, at the very least it will generate the cool previews of the docs for all PRs. Going forward we can use it for versioned docs, but we should discuss how to handle that and what to do with the domain.

It also fixes several issues with docs, mostly related to the new `plot_dot, `plot_lm` and `plot_ts`.

aloctavodia (approved):

ahartikainen (approved):

PR #1779, opened 2021-08-31
Updated Quickstart

roshnaeem:
**Description:**
Updated the [Arviz Quickstart](https://arviz-devs.github.io/arviz/getting_started/Introduction.html) part.

### DIFF
--- a/doc/source/getting_started/Introduction.ipynb
+++ b/doc/source/getting_started/Introduction.ipynb
@@ -104,7 +104,7 @@
    "source": [
     "## ArviZ rcParams\n",
     "\n",
-    "You may have noticed that for both `plot_posterior` and `plot_forest`, the Highest Density Interval (HDI) is 94%, which you may find weird at first. This particular value is a friendly reminder of the arbitrary nature of choosing any single value without further justification, including common values like 95%, 50% and even our own default, 94%. ArviZ includes default values for a few parameters, you can access them with `az.rcParams`. To change the default HDI value to let's say 90% you can do:"
+    "You may have noticed that for both [plot_posterior](https://arviz-devs.github.io/arviz/api/generated/arviz.plot_posterior.html) and [plot_forest](https://arviz-devs.github.io/arviz/api/generated/arviz.plot_forest.html), the Highest Density Interval (HDI) is 94%, which you may find weird at first. This particular value is a friendly reminder of the arbitrary nature of choosing any single value without further justification, including common values like 95%, 50% and even our own default, 94%. ArviZ includes default values for a few parameters, you can access them with `az.rcParams`. To change the default HDI value to let's say 90% you can do:"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


doc/source/getting_started/Introduction.ipynb
```diff
@@ -104,7 +104,7 @@
    "source": [
```
OriolAbril: <p>It is great to make these references to arviz functions link to their respective api page, but you should not do that with manual markdown links but with sphinx references instead. We have some docs on that at https://arviz-devs.github.io/arviz/contributing/developer_guide.html#documentation. In this case, you are referencing functions, so you should change the</p><p> </p><pre class="ql-syntax" spellcheck="false">[plot_posterior](https://arviz-devs.github.io/arviz/api/generated/arviz.plot_posterior.html)
</pre><p> </p><p>for</p><p> </p><pre class="ql-syntax" spellcheck="false">{func}~arviz.plot_posterior
</pre><p> </p><p>The same applies to all other references you are adding in this and future files. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/arviz-devs/arviz/pull/1779/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

OriolAbril: <p>There are some extra <code>&lt;/span&gt;</code>  in the reference which I think will break them. Can you check locally that they work?</p><p> </p><p>Also note that after https://github.com/arviz-devs/arviz/pull/1778 you can now rebase on main to get previews of the documentation rendered for each PR. I think this will help us review future PRs</p>

doc/source/getting_started/Introduction.ipynb
```diff
@@ -24,6 +24,13 @@
     "import numpy as np"
```
OriolAbril, line 24: <p>Here I would use </p><p> </p><pre class="ql-syntax" spellcheck="false">{doc}<code>examples/styles</code>
</pre><p> </p><p>There is also the option to use </p><p> </p><pre class="ql-syntax" spellcheck="false">{ref}<code>mpl_styles</code>
</pre><p> </p><p>which would also work, but I think we should update the gallery generator to change that before using it. mpl and bokeh used to be on different files but are not tabs within the same file, so now we don't need a tag for each file, only a tag for the example. If you prefer to use ref instead of doc, you can change the line https://github.com/arviz-devs/arviz/blob/main/doc/sphinxext/gallery_generator.py#L283 for </p><p> </p><pre class="ql-syntax" spellcheck="false">return f"example_{self.basename}"
</pre><p> </p><p>so that you can then use</p><p> </p><pre class="ql-syntax" spellcheck="false">{ref}<code>example_styles</code>
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/arviz-devs/arviz/pull/1779/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

roshnaeem: In place of "examples of style sheets [here](https://arviz-devs.github.io/arviz/examples/styles.html)" ?

doc/source/getting_started/Introduction.ipynb
roshnaeem: I checked the number of `<span` and `</span>` is the same in this file.

OriolAbril: when I look at reviewnb, I see `` {func}`~arviz.</span>plot_posterior` `` which is not a valid sphinx id. this pr has no readthedocs preview so I'm not sure if it's an nbviewer error or a mistake copy pasting from my earlier comment (copy pasting from reviewnb comments can be a nightmare). But it's something that needs to be double checked

doc/source/getting_started/Introduction.ipynb
OriolAbril: yes, you can still use custom text for the links even if using sphinx cross references instead of html hyperlinks. i.e.

```
examples of style sheets {doc}`examples/styles`
```

Will use as text for the hyperlink the title of the linked page, in this case "Matplotlib styles" (https://arviz-devs.github.io/arviz/examples/styles.html). Whereas

```
examples of style sheets {doc}`here <examples/styles>`
```

will show "here" as text.

OriolAbril (approved):

PR #1790, opened 2021-09-08
Updated pymc3 example

roshnaeem:
**Description:**
Updated the[ pymc3 example](https://arviz-devs.github.io/arviz/schema/PyMC3_schema_example.html) part.

- Tried to use simple language to leave no room for confusion.
- Used bullet points to make it more readable.

### DIFF
--- a/doc/source/schema/PyMC3_schema_example.ipynb
+++ b/doc/source/schema/PyMC3_schema_example.ipynb
@@ -327,7 +327,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "In this example, each variable for its dimension has a combination of the following three: `chain`, `draw` and `developer`. Moreover, each dimension has specific coordinate values. For example:\n",
+    "In this example, each variable for its dimension has a combination of the following three: `chain`, `draw` and `developer`. Moreover, each dimension has specific coordinate values.\n",
     "\n",
     "* In the case of `chain` and `draw` it is an integer identifier starting at `0`.\n",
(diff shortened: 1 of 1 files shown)


OriolAbril:
can you rerun the notebook? You'll have seen in the other notebooks that inferencedata now has an html repr where you can expand groups clicking on them, so there is no need anymore for a cell per group so everything is printed. 

There are also only 3 dimensions involved, so the explanation of coordinate values is not really an example but a complete description.

doc/source/schema/PyMC3_schema_example.ipynb
```diff
@@ -327,7 +327,10 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "In this example, each variable has as dimension a combination of the following 3: `chain`, `draw` and `developer`. Moreover, each dimension has specific coordinate values. In the case of `chain` and `draw` it is an integer identifier starting at `0`; in the case of `developer` dimension, its coordinate values are the following strings: `[\"Alice\", \"Bob\", \"Cole\", \"Danielle\", \"Erika\"]`."
+    "In this example, each variable for its dimension has a combination of the following three: `chain`, `draw` and `developer`. Moreover, each dimension has specific coordinate values.\n",
```
OriolAbril, line 330: > In this example, each variable for its dimension has a combination of the following three: `chain`, `draw` and `developer`.

I am not sure this is an improvement over the old wording, I think this can now be more confusing now and doesn't really follow the terminology defined in the schema. "dimension" should probably be plural (either in this new wording or the original one). This alone might be enough, I am sometimes overwary of constructions that differ too much from catalan or spanish ones.

PR #1796, opened 2021-09-13
Updated Refitting PyMC3 models with ArviZ

roshnaeem:
**Description:**
Updated the [Refitting PyMC3 models with ArviZ](https://arviz-devs.github.io/arviz/user_guide/pymc3_refitting.html#
) part.

### DIFF
--- a/doc/source/user_guide/pymc3_refitting.ipynb
+++ b/doc/source/user_guide/pymc3_refitting.ipynb
@@ -34,7 +34,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "For the example we will use linear regression."
+    "For the example we will use a linear regression model."
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


doc/source/user_guide/pymc3_refitting.ipynb
```diff
@@ -76,9 +76,12 @@
...
    "metadata": {},
    "source": [
-    "Now we will write the PyMC3 model, keeping in mind that 1) data must be modifiable (both `x` and `y`) and 2) the model must be recompiled in order to be refitted with the modified data. We therefore have to create a function that recompiles the model when it's called.\n",
+    "Now we will write the PyMC3 model, keeping in mind the following two points:\n",
+    "1. Data must be modifiable (both `x` and `y`).\n",
+    "2. The model must be recompiled in order to be refitted with the modified data.\n",
+    "We, therefore, have to create a function that recompiles the model when it's called.\n",
     "\n",
-    "Note how we would generally use `pm.Data` to populate `xyz_data` groups in the InferenceData result, however, given that we need to recompile the model every time, we are consciously skipping this step to speed up the recompilation. "
+    "Note how we would generally use `pm.Data` to populate `xyz_data` groups in the {ref}`InferenceData <xarray_for_arviz>` result, however, given that we need to recompile the model every time, we are consciously skipping this step to speed up the recompilation. "
```
OriolAbril, line 84: we don't really have the guide available yet, but following up on what we were commenting the other day, given that here we are talking about populating groups of inferencedata using `pm.Data` which is pymc specific and this page is an advanced one, I would link from here to the pymc3 conversion page once it's created instead of linking to the intro.

OriolAbril (commented):
also same comments as the pystan2 one. the extra comment here is non-actionable general one about how and why to link pages between them

PR #1797, opened 2021-09-14
Updated Refitting NumPyro models with ArviZ

roshnaeem:
**Description:**

Updated the[ Refitting NumPyro models with ArviZ](https://arviz-devs.github.io/arviz/user_guide/numpyro_refitting.html) part.


OriolAbril (changes requested):
marking as request changes for now until stan references are sorted out

OriolAbril (approved):

PR #1806, opened 2021-09-20
Autoupdate f strings with flynt

canyon289:
Seeing if we can get this all fixed in one shot


.pylintrc
```diff
@@ -74,7 +74,8 @@ disable=missing-docstring,
         no-member,
         #TODO: Remove this once todos are done
         fixme,
-        consider-using-with
+        consider-using-with,
+        consider-using-f-string
```
canyon289, line 78: Pylint disable for now so we can unblock PRs without a failing master. We can add an issue to track if we merge

PR #1814, opened 2021-09-29
Fix CI PyMC3 github install to v3

canyon289:
Fixing PyMC3 github source test to the PyMC3 branch in pymc3

This means PyMC (not 3) is not being tested right now so that'll have to get added later


canyon289:
Merging this so we can keep a green CI

@roshnaeem rebase your PRs on top

PR #1818, opened 2021-09-29
Added kde page in stats

roshnaeem:
## Description
The [plot_kde](https://arviz-devs.github.io/arviz/api/generated/arviz.plot_kde.html) was connected to itself in "see also" section. It should be instead connected to `kde` page. 
Added the `kde` page and fixed its identation.

### DIFF
--- a/arviz/stats/density_utils.py
+++ b/arviz/stats/density_utils.py
@@ -401,80 +401,80 @@ def kde(x, circular=False, **kwargs):
     .. plot::
         :context: close-figs
 
-    >>> import numpy as np
-    >>> import matplotlib.pyplot as plt
-    >>> from arviz import kde
-    >>>
-    >>> rvs = np.random.gamma(shape=1.8, size=1000)
-    >>> grid, pdf = kde(rvs)
-    >>> plt.plot(grid, pdf)
-    >>> plt.show()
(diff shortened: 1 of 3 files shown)


arviz/stats/density_utils.py
```diff
@@ -401,80 +401,80 @@ def kde(x, circular=False, **kwargs):
     .. plot::
```
OriolAbril, line 401: There is no empty line between the text aand the `.. plot::` directive so it's not rendering properly: https://arviz-devs.github.io/arviz/api/generated/arviz.kde.html

doc/source/api/stats.rst
```diff
@@ -17,3 +17,4 @@ Stats
     r2_score
     summary
     waic
+    kde
```
OriolAbril, line 20: can you move it so everything keeps being alphabetically sorted

OriolAbril (commented):
there should be a follow-up PR to fix rendering

doc/source/api/stats.rst
roshnaeem: Sure

arviz/stats/density_utils.py
roshnaeem: Okay

Issue #1815, opened 2021-09-29
Image not found


roshnaeem:
**Describe the bug**
The image of the first blog in the [Blogs and example compilations](https://arviz-devs.github.io/arviz/community.html#blogs-and-example-compilations) on the community page is missing. 

<img width="625" alt="Screenshot 2021-09-29 at 11 25 30 AM" src="https://user-images.githubusercontent.com/47316899/135214368-88997a28-d621-44a9-a7bc-a070e7f341de.png">

Source file: https://github.com/arviz-devs/arviz/blob/main/doc/source/external_resources.md

PR #1819, opened 2021-09-29
Fixed pymc img path

roshnaeem:
Fixes: https://github.com/arviz-devs/arviz/issues/1815


OriolAbril:
@aloctavodia @twiecki there used to be a rocket only logo which is the one we used here. Now there seem to only be complete logo+text images.

Old image: https://github.com/pymc-devs/pymc/blob/v3/docs/logos/PyMC3.png
Current one: https://github.com/pymc-devs/pymc/blob/main/docs/logos/PyMC.png

PR #1837, opened 2021-10-04
Add docs references to plot_dist docstring

JulianPedro:
## Description
Add summary in plot_dist function docstring and fix some functions references. Issue [https://github.com/arviz-devs/arviz/issues/1824](https://github.com/arviz-devs/arviz/issues/1824)

## Checklist

### DIFF
--- a/arviz/plots/distplot.py
+++ b/arviz/plots/distplot.py
@@ -65,8 +65,8 @@ def plot_dist(
     bw: Optional[float or str]
         If numeric, indicates the bandwidth and must be positive.
         If str, indicates the method to estimate the bandwidth and must be
-        one of "scott", "silverman", "isj" or "experimental" when `is_circular` is False
-        and "taylor" (for now) when `is_circular` is True.
+        one of "scott", "silverman", "isj" or "experimental" when ``is_circular`` is False
+        and "taylor" (for now) when ``is_circular`` is True.
         Defaults to "default" which means "experimental" when variable is not circular
         and "taylor" when it is.
     quantiles : list
@@ -88,14 +88,14 @@ def plot_dist(
         Ignored for 2D KDE
     rug_kwargs : dict
         Keywords passed to the rug plot. Ignored if rug=False or for 2D KDE
-        Use `space` keyword (float) to control the position of the rugplot. The larger this number
+        Use ``space`` keyword (float) to control the position of the rugplot. The larger this number
         the lower the rugplot.
     contour_kwargs : dict
         Keywords passed to the contourplot. Ignored for 1D KDE.
     contourf_kwargs : dict
-        Keywords passed to ax.contourf. Ignored for 1D KDE.
+        Keywords passed to :func:`ax.contourf`. Ignored for 1D KDE.
     pcolormesh_kwargs : dict
-        Keywords passed to ax.pcolormesh. Ignored for 1D KDE.
+        Keywords passed to :func:`ax.pcolormesh`. Ignored for 1D KDE.
     hist_kwargs : dict
         Keywords passed to the histogram.
     is_circular : {False, True, "radians", "degrees"}. Default False.
@@ -117,6 +117,12 @@ def plot_dist(
     Returns
     -------
     axes : matplotlib axes or bokeh figures
+
+    See Also
+    --------
+    plot_posterior : Plot Posterior densities in the style of John K. Kruschke’s book.
+    plot_density : Generate KDE plots for continuous variables and histograms for discrete ones.
+    plot_kde : 1D or 2D KDE plot taking into account boundary conditions.
 
     Examples
     --------


arviz/plots/distplot.py
```diff
@@ -88,14 +88,14 @@ def plot_dist(
         Ignored for 2D KDE
     rug_kwargs : dict
         Keywords passed to the rug plot. Ignored if rug=False or for 2D KDE
-        Use `space` keyword (float) to control the position of the rugplot. The larger this number
+        Use ``space`` keyword (float) to control the position of the rugplot. The larger this number
         the lower the rugplot.
```
OriolAbril, line 92: All the kwarg type arguments should have references to the functions they are eventually passed to. I'll do a couple of examples with links here, but all need to be updated with the same pattern. 

`rug_kwargs` is passed to `plot_dist` (defined in https://github.com/arviz-devs/arviz/blob/main/arviz/plots/distplot.py#L28) and then on to the backend plotting function (see [this line](https://github.com/arviz-devs/arviz/blob/main/arviz/plots/distplot.py#L195) and [these lines](https://github.com/arviz-devs/arviz/blob/main/arviz/plots/distplot.py#L211-L212)). 

### matplotlib
The matplotlib version passes it to `plot_kde` (https://github.com/arviz-devs/arviz/blob/main/arviz/plots/distplot.py#L211-L212). You'll see that in some cases, the kwargs are modified with setdefaults or things like that. You can ignore all that except for **pop** clauses, the only part relevant for these PRs is where they are passed to, no need to understand the code either. It is enough to say there are pop clauses and ask for help if you see them, again, no need to understand exactly what the code is doing.

### bokeh
Bokeh version also passes it to `plot_kde`. 

---

Therefore, in this case, the only change needed is adding a sentence to the argument description along the lines of 

    Passed to :func:`arviz.plot_kde` as ``rug_kwargs``

and it would be `plot_kde` that would have more detail on that.

arviz/plots/distplot.py
```diff
@@ -88,14 +88,14 @@ def plot_dist(
...
     contour_kwargs : dict
         Keywords passed to the contourplot. Ignored for 1D KDE.
     contourf_kwargs : dict
-        Keywords passed to ax.contourf. Ignored for 1D KDE.
+        Keywords passed to :func:`ax.contourf`. Ignored for 1D KDE.
     pcolormesh_kwargs : dict
-        Keywords passed to ax.pcolormesh. Ignored for 1D KDE.
+        Keywords passed to :func:`ax.pcolormesh`. Ignored for 1D KDE.
     hist_kwargs : dict
         Keywords passed to the histogram.
```
OriolAbril, line 100: The 1st example passed the kwargs to another arviz function, and had the same behaviour for both bokeh and matplotlib, but this won't be the case often. This particular one however will be among the most complicated kwarg arguments to document. In plot_dist, most are passed to plot_kde as is so don't let it throw you back.

We start again like with rug_kwargs, taken and passed directly to the backend plotting function, see https://github.com/arviz-devs/arviz/blob/main/arviz/plots/distplot.py#L32, https://github.com/arviz-devs/arviz/blob/main/arviz/plots/distplot.py#L199 and https://github.com/arviz-devs/arviz/blob/main/arviz/plots/distplot.py#L211-L212).

### matplotlib
If you search for `hist_kwargs` in this file you'll see many ocurrences in https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/matplotlib/distplot.py#L66-L77. These are one example of what I was saying before. Ignore all that, you only have to follow the argument as it is passed downstream. 

Here is is passed to `_histplot_mpl_op` in https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/matplotlib/distplot.py#L84. This is an internal function that you'll see doesn't appear in the [docs](https://arviz-devs.github.io/arviz/api/plots.html), so the job is not finished yet.

[`_histplot_mpl_op` is defined below](https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/matplotlib/distplot.py#L125) and it passes the hist kwargs to the matplotlib hist function in https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/matplotlib/distplot.py#L155. 

### bokeh 
basically the same pattern as before but now passing them to the `_histplot_bokeh_op` at https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/bokeh/distplot.py#L79. Here this is eventually passed to the [`quad` function](https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/bokeh/distplot.py#L178). But there are some pop clauses: https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/bokeh/distplot.py#L127, https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/bokeh/distplot.py#L132, https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/bokeh/distplot.py#L136 and https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/bokeh/distplot.py#L138. Each of these pop things should have an explanatory sentence.

---
Proposal for the description of this argument:

```
Keyword arguments used to customize the histogram. Ignored when plotting a KDE. They are passed to
:meth:`matplotlib.axes.Axes.hist` if using matplotlib, or to :meth:`bokeh.plotting.Figure.quad` if
using bokeh. In bokeh case, the following extra keywords are also supported:

* ``color``: replaces the ``fill_color`` and ``line_color`` of the ``quad`` method
* ``bins``: taken from ``hist_kwargs`` and passed to :func:`numpy.histogram` instead
* ``density``: normalize histogram to represent a probability density function, Defaults to ``True``
* ``cumulative``: plot the cumulative counts. Defaults to ``False``
```

roshnaeem:
https://github.com/arviz-devs/arviz/pull/1894 will close this PR.

PR #1870, opened 2021-10-11
Added searchbar on homepage

roshnaeem:
Description:
The homepage does not have a search bar. Added search after the first short paragraph. We can move it above or below the text if a change of location will be required.


roshnaeem:
@aloctavodia, I have moved it above, can you check which location do you prefer? Let me know.

PR #1927, opened 2021-11-04
Added refernces in plot_kde

roshnaeem:
## Description
* Added references 
* Added backticks
* Added kwargs functions

Note: `fill_kwargs` are being passed to `fill_func` at a few [lines](https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/matplotlib/kdeplot.py#L153), but I could not find it in matplotlib docs. 

Fixes: https://github.com/arviz-devs/arviz/issues/1853


OriolAbril:
some arguments only have reference to matplotlib or to bokeh methods they are passed to, not to both or saying it is ignored in some cases, but this is already an amazing improvement over the current state and you should focus on the write arviz guide and bigger picture things.

Issue #1474, opened 2020-12-23
TFP is pinned to 0.11.0


canyon289:
A deprecation in Edward in TFP 0.12.0 is causing issues with our converter and CI. 

**Expected Resolution**
We figure out if were going to have support new and old TFP, just one and then fix ArviZ appropriately

_Originally posted by @canyon289 in https://github.com/arviz-devs/arviz/issues/1472#issuecomment-749960630_

canyon289:
See https://github.com/arviz-devs/arviz/issues/1471 for more details

Issue #1890, opened 2021-10-16
TFP converter and tests need urgent updating


OriolAbril:
The tfp converter is still tf1 based, using the compatibility helpers to be able to run. We have other issues open related to that, but doing so has already generated multiple issues and incompatibilities in CI. IIRC, it forced us to stop testing the pyjags converter and now it also forced to pin cloudpickle to an old version

I think that if the tfp converter can be fixed and updated in the near future we should remove it for the time being. I am not even sure it is needed after the latest changes in tfp. cc @ColCarroll @canyon289 @junpenglao 

related to #812 and #1474

OriolAbril:
Superseeded by #2216

PR #1928, opened 2021-11-04
Fix CI: remove tfp converter and fix mypy

roshnaeem:
## Description
Replaced the current `mypy.ini` file with `mypy.ini` downloaded from the artifacts.

EDIT (Oriol): Fixes CI that was broken due to outdated tfp code. Closes https://github.com/arviz-devs/arviz/issues/1474


OriolAbril:
Removed the tfp converter to fix CI (which also allowed reactivating tests for pyjags that had to be skipped when pinning down tfp). I don't think much of the tfp related code was useful now, so it's probably better to take converters in tfp docs or the one in #1890 to substitute the current one completely

canyon289:
Ill approve this given the TF version is old and its holding up the entire ArviZ project. If we need to add it back in we have this pr as reference and its easy enough to revert

canyon289 (approved):

arviz/plots/dotplot.py
```diff
@@ -222,9 +222,7 @@ def layout_stacks(stack_locs, stack_counts, binwidth, stackratio, rotated):
     binradius = binwidth / 2
 
     x = np.repeat(stack_locs, stack_counts)
-    y = np.hstack(
-        np.array([dotheight * np.arange(count) + binradius for count in stack_counts], dtype=object)
-    )
+    y = np.hstack([dotheight * np.arange(count) + binradius for count in stack_counts])
```
canyon289, line 225: What happened here? Just curious

canyon289:
This can be fixed on followup PRs but we should remove these other tensorflow references as well
https://github.com/arviz-devs/arviz/blob/main/.azure-pipelines/azure-pipelines-external.yml#L31

arviz/plots/dotplot.py
OriolAbril: numpy added type hints, and array was not on the accepted list of input types

PR #1903, opened 2021-10-21
Write ArviZ docs guide

roshnaeem:
Description:

* Restructuring developer guide to separate documentation guide from it. 
* Updating doc guide to add more sections.

PR #1936, opened 2021-11-09
Fixed heading structure to maintain symmetry

roshnaeem:
## Description
On the [Contributing page](https://arviz-devs.github.io/arviz/contributing/index.html), the links in the bullet points are not on the same level. Heading-1 should only be the title of the page, but `contributing_prs.md` consisted of a lot of first-level headings. Fixed the heading levels to maintain the symmetry.


OriolAbril:
I think it would be best to do it as part of #1903, because there the hierarchy is also modified. Here or there works fine by me as the end result will be the same, but doing two separate PRs will probably mean rebases and git conficts on your end

roshnaeem:
It is fine, you can merge it, I'll rebase the other PR with main.

PR #1939, opened 2021-11-16
remove manual setting of 2d kde lims

OriolAbril:
## Description
A few lines above we set `x_x` and `y_y` using the min and max along each dimension, so this has no visual effect on the plot, _as long as you are plotting models independently_. I think it is important to remove in order to allow using the kde mode of plot_pair to compare models. Otherwise only the last kde to be plotted is guaranteed to be seen complete, the rest are generally cut or lay completely outside the limits.

## Checklist

- [ ] Does the PR include new or updated tests to prevent issue recurrence (using [pytest fixture pattern](https://docs.pytest.org/en/latest/fixture.html#fixture))? No, could think of adding one, not sure if necessary
- [ ] Is the fix listed in the [Maintenance and fixes](https://github.com/arviz-devs/arviz/blob/main/CHANGELOG.md#maintenance-and-fixes) section of the changelog?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Maintenance and fixes
 * Drop Python 3.6 support ([1430](https://github.com/arviz-devs/arviz/pull/1430))
 * Bokeh 3 compatibility. ([1919](https://github.com/arviz-devs/arviz/pull/1919))
+* Remove manual setting of 2d KDE limits ([1939](https://github.com/arviz-devs/arviz/pull/1939))
 
 ### Deprecation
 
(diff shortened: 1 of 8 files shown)


ahartikainen:
is there a possibility to check data limits?

Or does mpl this for us?

OriolAbril:
both matplotlib and bokeh do that automatically, you can see in the example gallery preview for example how everything looks the same: https://arviz--1939.org.readthedocs.build/en/1939/examples/plot_pair_kde.html or https://arviz--1939.org.readthedocs.build/en/1939/examples/plot_pair_kde_hdi.html

PR #1948, opened 2021-12-08
Unpin typing_extensions

### LINKED ISSUE #1947: Remove typing_extensions "<4" version guard
Is there a reasoning for the `typing_extensions<4` version guard in the [requirements.txt file](https://github.com/arviz-devs/arviz/blob/c45b92478333c6d79c409e65d4f97b9ad5ca7d4d/requirements.txt#L9)? If not, would it be possible to remove this `<4` limitation?

I'm running into issues with some other packages (non-public) due to this version constraint.

Duplicate of https://github.com/arviz-devs/arviz/issues/1945 which was closed without considering other use cases.

ColCarroll:
Fixes #1947. I also had some trouble installing some `requirements-external.txt`, but I didn't include those in the change so far. 

I'll tag @obi1kenobi who originally added this pin, I think it was just defensive against the next major version, but if the mypy checks pass, I think we can merge.

### DIFF
--- a/arviz/tests/base_tests/test_data_zarr.py
+++ b/arviz/tests/base_tests/test_data_zarr.py
@@ -2,6 +2,7 @@
 import os
 import shutil
 from collections.abc import MutableMapping
+from typing import Mapping
 
 import numpy as np
 import pytest
@@ -26,7 +27,8 @@ class TestDataZarr:
         class Data:
             # fake 8-school output
             obj = {}
-            for key, shape in {"mu": [], "tau": [], "eta": [8], "theta": [8]}.items():
+            shapes: Mapping[str, list] = {"mu": [], "tau": [], "eta": [8], "theta": [8]}
+            for key, shape in shapes.items():
                 obj[key] = np.random.randn(chains, draws, *shape)
 
         return Data
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,4 +6,4 @@ packaging
 pandas>=0.23
 xarray>=0.16.1
 netcdf4
-typing_extensions>=3.7.4.3,<4
+typing_extensions>=3.7.4.3


requirements.txt
```diff
@@ -6,4 +6,4 @@ packaging
 pandas>=0.23
 xarray>=0.16.1
 netcdf4
-typing_extensions>=3.7.4.3,<4
+typing_extensions>=3.7.4.3
```
obi1kenobi, line 9: Consider this as an alternative?
```suggestion
typing_extensions>=3.7.4.3,<5
```

I have no strong feelings one way vs the other — broadly, it's a tradeoff between "prevent possibly-non-working combinations, have to bump version bounds manually" and "users run into non-working combinations and open issues that don't really point to the root cause, setting you up for some fun debugging time".

arviz/tests/base_tests/test_data_zarr.py
```diff
@@ -26,7 +27,8 @@ def data(self, draws, chains):
         class Data:
             # fake 8-school output
             obj = {}
-            for key, shape in {"mu": [], "tau": [], "eta": [8], "theta": [8]}.items():
+            shapes: Mapping[str, list] = {"mu": [], "tau": [], "eta": [8], "theta": [8]}
```
obi1kenobi, line 30: Consider something more specific than `list` so it captures the types within the list? I'm not sure what `np.random.randn` requires the type of `shape` to be, exactly.

obi1kenobi (commented):
Yup, purely defensive — agreed that it's probably safe to merge if tests pass.

Pointing out two minor things for you to consider. I don't have strong feelings one way or the other, just wanted to point them out to make sure it's an intentional decision whichever way it goes.

arviz/tests/base_tests/test_data_zarr.py
ColCarroll: i was actually surprised this popped up -- this is a literal dict used once, so `List[int]` (or something...) seems like it should have been deduced.

requirements.txt
ColCarroll: I'm going to leave this as is, since we don't adopt this pinning elsewhere (but helpful to think about the multiple failure modes)

gothicVI:
The fix is great but can we also get a new release?

PR #1953, opened 2021-12-15
adapt code to bokeh 3.0

OriolAbril:
## Description
`source` argument is not needed (can be retrieved from the parent source) and has in
fact been removed in 3.0 version. This should fix the preview tests.

## Checklist

- [x] Is the fix listed in the [Maintenance and fixes](https://github.com/arviz-devs/arviz/blob/main/CHANGELOG.md#maintenance-and-fixes)
      section of the changelog?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Maintenance and fixes
 * Drop Python 3.6 support ([1430](https://github.com/arviz-devs/arviz/pull/1430))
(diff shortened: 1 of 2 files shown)


ahartikainen:
Maybe we should update our Bokeh code to 3.0

OriolAbril:
> Maybe we should update our Bokeh code to 3.0

it is still pre-release only though, I think for now the best we can do is progressively update things as they break, we do need to keep 2.x compatibility. I didn't find any 3.0 changelog either :thinking:

ahartikainen:
Yeah, we probably want to pin to 2 and then hop to 3 when it is released

PR #1965, opened 2022-01-17
[WIP] add draw priors function for prior elicitation

aloctavodia:
This adds a simple tool to allow users draw 1D distributions using a grid as input. This follows the [Roulette mode](http://optics.eee.nottingham.ac.uk/match/uncertainty.php)

https://user-images.githubusercontent.com/1338958/149752176-74a96dba-216e-4ce2-adf3-a6a48cffd91e.mp4

Each rectangle in the grid is clickable, once some rectangles are selected the pdf is drawn in the bottom panel. This is triggered by leaving the figure (moving the pointer out of the figure). By default the best distribution fitting the users input is automatically selected, alternatively the user can pick a single distribution. The size of the grid (x and y bins) can be adjusted, as well as the x-scale. Two quantiles are shown by default 0.33 and 0.66.

Not sure where is a good name or a good place for this function. Also I am planing on adding more features for prior elicitation, also based on user interaction and visualization. 

Some things to be done (not necessary as part of this PR):

* As can be seen from the demo, there is considerable delay from the moment the user click on a rectangle and the moment the rectangle is highlighted  in orange. 
* Instead of radio buttons it will be nice to have check-buttons, and allow selecting such as  "all but beta distribution". In that case instead of an "automatic" option we should have an "all" option)
* Further checks to ensure the automatic method is good enough.   
* Add more distributions.
* The implemented distributions match the PyMC definitions, we could have an option for other ppls. 
* Free hand drawing?

### DIFF
--- a/arviz/plots/draw_prior.py
+++ b/arviz/plots/draw_prior.py
@@ -1,0 +1,251 @@
+"""Prior elicitation using roulette mode."""
+
+import matplotlib.pyplot as plt
+import numpy as np
+import scipy.stats as stats
+import matplotlib.patches as patches
+from matplotlib.widgets import RadioButtons, TextBox
+from math import ceil, floor
+
+from ..rcparams import rcParams
+
+
+def create_grid(x_min=0, x_max=1, nrows=10, ncols=10, ax=None):
+    xx = np.arange(0, ncols)
+    yy = np.arange(0, nrows)
+
+    if ncols < 10:
+        num = ncols
+    else:
+        num = 10
+
+    ax[0, 0].set(
+        xticks=np.linspace(0, ncols, num=num),
+        xticklabels=[f"{i:.1f}" for i in np.linspace(x_min, x_max, num=num)]
+        ,
+    )
+
+    coll = np.zeros((nrows, ncols), dtype=object)
+    for idx, xi in enumerate(xx):
+        for idy, yi in enumerate(yy):
+            sq = patches.Rectangle((xi, yi), 1, 1, fill=True, facecolor="0.8", edgecolor="w")
+            ax[0, 0].add_patch(sq)
+            coll[idy, idx] = sq
+    return coll
+
+
+class Rectangles:
+    def __init__(self, fig, coll, nrows, ncols, ax):
+        self.fig = fig
+        self.coll = coll
+        self.nrows = nrows
+        self.ncols = ncols
+        self.ax = ax
+        self.weights = {k: 0 for k in range(0, ncols)}
+        fig.canvas.mpl_connect("button_press_event", self)
+
+    def __call__(self, event):
+        if event.inaxes == self.ax[0, 0]:
+            x = event.xdata
+            y = event.ydata
+            idx = floor(x)
+            idy = ceil(y)
+            if self.weights[idx] >= idy:
+                idy -= 1
+                for row in range(self.nrows):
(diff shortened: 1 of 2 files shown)


arviz/plots/draw_prior.py
```diff
@@ -0,0 +1,251 @@
...
+            la = np.repeat((k / ncols * x_range) + x_min + ((x_range / ncols) / 2), v*100+1)
+            samples.extend(la)
+
+        x = np.linspace(x_min, x_max, 500)
+        dist_name = radio.value_selected
+        if dist_name == "automatic":
+            sum_pdf_old = -np.inf
+            for dist in [stats.beta, stats.norm, stats.gamma, stats.lognorm]:
+                if dist.name == "beta":
+                    a, b, _, _ = dist.fit(samples, floc=x_min, fscale=x_range)
```
ColCarroll, line 83: this should work with just `dist_args = dist.fit(samples, ...)` and `can_dist = dist(*args)`. I guess you're restricting the distribution to have support in `(x_min, x_max)`, but then the fit could be arbitrarily bad, I think!

arviz/plots/draw_prior.py
```diff
@@ -0,0 +1,251 @@
...
+
+    if backend is None:
+        backend = rcParams["plot.backend"]
+    backend = backend.lower()
+    if backend == "bokeh":
+        raise TypeError("Draw prior is only supported with matplotlib backend.")
+
+    try:
+        shell = get_ipython().__class__.__name__
+        if shell == "ZMQInteractiveShell" and get_backend() != "nbAgg":
```
ColCarroll, line 180: `plt.get_backend`, I think

ColCarroll (commented):
I spent some time trying to get this to run on colab and on jupyter with no success! Specifically, I copy/pasted all of draw_prior.py, and got rid of the 6 lines with rcParams in them.

- On colab, I get a non-responsive widget with `%matplotlib inline`, and nothing using `nbagg` or `notebook`.
image

- On jupyter, I get 
image

Let me know if there's a better way to give this a try!

arviz/plots/draw_prior.py
aloctavodia: This needs a little bit more care so we can match scipy parameterization with PyMC (and other ppls parameterization). The fit seems to work well in practice. I compared with results from the MATCH site and with scipy, but I admit that these tests were not exhaustive.

aloctavodia:
It works for me with ` %matplotlib notebook`. I tried with a jupyter notebook, and a jupyter notebook inside vscode. Are you using Jupyter lab? If that is the case you need to use `%matplolib widget`

aloctavodia:
It seems that backend agnostics Buttons and TextBox are very slow in matplotlib. One solution is to remove these elements (the range, bins and quantiles can be defined when calling the function). Another one is to use tkinter, that is available by default in Linux and Mac, but not on Windows. Or at least that used to be the case.

ahartikainen:
What if we just use bokeh for this?

aloctavodia:
If possible we should use both as with other plots.

aloctavodia:
I am going to update this to work with Tkinter and separate the plotting code from the fitting code so we can also have a bokeh version and avoid code duplication.

aloctavodia:
Closing in favor of https://github.com/arviz-devs/preliz

PR #1972, opened 2022-01-30
Enabled type checking as described by PEP 561

michaelosthege:
[PEP 561](https://www.python.org/dev/peps/pep-0561/) describes how packages can tell tools such as `mypy` that they support typechecking.
Essentially there are three options:
* Separately released "stub packages"
* `.pyi` files with stub definitions
* inline type hints and a `py.typed` file

This PR adds a `py.typed` file as shown in the https://github.com/ethanhs/pep-561 example project.

The project structure here is a little different. I hope the line in `MANIFEST.in` is correct?

This way type checking in PyMC etc. should become more exact, even if there are still some inconsistencies left.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,6 +2,7 @@ include *.md
 include *.txt
 include *.toml
 include LICENSE
+include arviz/py.typed
 
 recursive-include arviz *.mplstyle
 recursive-include arviz *.nc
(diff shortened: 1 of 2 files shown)


OriolAbril:
I have no idea what the manifest is, but everything else looks good, I think we do have some type hints that users could take advantage of. 

Can you try pinning black to `black<22.1.0`? lints are fail fast and black failing is preventing mypy and typing copilot to run, which are relevant to this PR.

michaelosthege:
Nice!
I'm looking forward to the mypy nitpicks. For McBackend I'm optimistic, but I bet this will uncover some fragility in PyMC.

PR #1977, opened 2022-02-05
Allow `log_weights` to specify the log_likelihood variable name in `loo_pit` (and `plot_loo_pit`)

erocarrera:
It may not always be the case that variables in the log_likelihood group match exactly the names in observed or posterior predictive groups. This change allows to pass a str as "log_weights" which points to the log_likelihood variable to pair with the observed and posterior predictive ones.
This is helpful with InferenceData coming from Stan which won't allow, as far as I know, to use the same name for observed and log-likelihood variables in the same model file.
I've added the corresponding test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Add new convenience function `arviz.extract_dataset` ([1725](https://github.com/arviz-devs/arviz/pull/1725))
 * [experimental] Enable dask chunking information to be passed to `InferenceData.from_netcdf` with regex support ([1749](https://github.com/arviz-devs/arviz/pull/1749))
 * Allow kwargs to customize appearance of the mean in `plot_lm`
+* Allow `log_weights` to specificy the log_likelihood variable name in `loo_pit` (and `plot_loo_pit`)
 
 ### Maintenance and fixes
 * Drop Python 3.6 support ([1430](https://github.com/arviz-devs/arviz/pull/1430))
--- a/arviz/plots/loopitplot.py
+++ b/arviz/plots/loopitplot.py
@@ -46,8 +46,9 @@ def plot_loo_pit(
         extra dimension at the end of size n_samples (chains and draws stacked). If str or
         None, ``idata`` must contain the posterior predictive group. If None, ``y_hat`` is taken
         equal to y, thus, y must be str too.
-    log_weights : array or DataArray
-        Smoothed log_weights. It must have the same shape as ``y_hat``
+    log_weights: array, DataArray, or str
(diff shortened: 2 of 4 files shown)


ahartikainen:
I wonder if we should use same logic as in plotting interface?

erocarrera:
> I wonder if we should use same logic as in plotting interface?

Could you point me to the code? I could take a look and update the PR if it's a better solution.

ahartikainen:
https://github.com/arviz-devs/arviz/blob/eda14cd758fefd46dcb14944ba6a7d77c0d7904d/arviz/plots/ppcplot.py#L79

erocarrera:
From what I see, `data_pairs` is used only in plotting functions but not in their stats counterparts. How would you prefer this to be addressed? An options might be:

Add `data_pairs` to `plot_loo_pit` and leave `log_weigths` as it is (not taking a str), but allow the str `log_weights` parameter in `loo_pit`. `plot_loo_pit` would then read `data_pairs` and pass the variable name in `log_weights` to `loo_pit`. This strikes me as the simpler and most compatible way but it introduces inconsistencies in the signatures of `plot_loo_pit` and `loo_pit`.
In this case `data_pairs` would be an alternative to using `y_hat` and `log_weights` in `plot_loo_pit`.

Let me know your thoughts.

OriolAbril:
Thanks, I thought that was possible already, 

> This is helpful with InferenceData coming from Stan which won't allow, as far as I know, to use the same name for observed and log-likelihood variables in the same model file.

This is already possible though. log_likelihood can be a dictionary instead of only a string/list to take care of the renaming. See https://arviz-devs.github.io/arviz/getting_started/CreatingInferenceData.html#from-pystan

How would the dictionary work here? IIRC we discussed that when adding loo_pit, but here we have 3 quantities with potentially differnt names: observed, posterior predictive and log likelihood. It can't be the same as posterior predictive (even if we name the argument data_pairs.

erocarrera:
Thanks Oriol. I had overlooked that option. It is indeed possible via the import `from_pystan` but not via the one for `from_cmdstan`.

I have quickly implemented the equivalent support in `from_cmdstan` to take a dictionary to describe the renaming and it works. I have now a log_likelihood group with variable names identical to those in the observed_data group. I believe that's the intended behavior, right?

If so I could drop this PR with the changes to `loo_pit`, `plot_loo_pit` and send another with the update to `from_cmdstan`.

erocarrera:
The change to `from_cmdstan` to accept dict has worked great for me. I am closing this PR and will send another with those changes. Thanks Oriol for pointing me in the right direction.

PR #1978, opened 2022-02-07
Add dict to `from_cmdstan`'s log_likelihood parameter accepted types 

erocarrera:
This works identical as in `from_pystan`, allowing to map observed variable names to their corresponding log-likelihoods.
* Added test.

## Checklist

### DIFF
--- a/arviz/data/io_cmdstan.py
+++ b/arviz/data/io_cmdstan.py
@@ -441,7 +441,7 @@ class CmdStanConverter:
                 columns = {
                     log_lik_to_obs_name[col](varidx): idx
                     for col, varidx, idx in (
-                        (*col.split("."), idx) for col, idx in self.posterior_columns.items()
+                        (*col.split(".", 1), idx) for col, idx in self.posterior_columns.items()
                     )
                     if any(item == col for item in log_likelihood.values())
                 }
--- a/arviz/data/io_cmdstan.py
+++ b/arviz/data/io_cmdstan.py
@@ -437,17 +437,13 @@ class CmdStanConverter:
             data_warmup = _unpack_ndarrays(chain_data_warmup, columns, self.dtypes)
         else:
             if isinstance(log_likelihood, dict):
-                log_lik_to_obs_name = {
-                    v: lambda idx: f"{k}.{{}}".format(idx) if idx else k
-                    for k, v in log_likelihood.items()
-                }
+                log_lik_to_obs_name = {v: k for k, v in log_likelihood.items()}
                 columns = {
-                    log_lik_to_obs_name[col](varidx): idx
-                    for col, varidx, idx in (
-                        (*(col.split(".", 1) if "." in col else (col, None)), idx)
-                        for col, idx in self.posterior_columns.items()
+                    col.replace(col_name, log_lik_to_obs_name[col_name]): idx
+                    for col, col_name, idx in (
+                        (col, col.split(".")[0], idx) for col, idx in self.posterior_columns.items()
                     )
-                    if any(item == col for item in log_likelihood.values())
+                    if any(item == col_name for item in log_likelihood.values())
                 }
             else:
                 if isinstance(log_likelihood, str):
--- a/arviz/data/io_cmdstan.py
+++ b/arviz/data/io_cmdstan.py
@@ -439,10 +439,11 @@ class CmdStanConverter:
             if isinstance(log_likelihood, dict):
                 log_lik_to_obs_name = {v: f"{k}.{{}}".format for k, v in log_likelihood.items()}
                 columns = {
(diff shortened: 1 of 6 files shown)


ahartikainen:
This looks great.

I wonder if we should implement these dict structures for posterior/prior predictive group too? If simple copy-paste does not work for that change, then maybe we can do another PR later for it.

arviz/data/io_cmdstan.py
```diff
@@ -430,13 +436,22 @@ def log_likelihood_to_xarray(self):
...
-                for col, idx in self.posterior_columns.items()
-                if any(item == col.split(".")[0] for item in log_likelihood)
-            }
+            if isinstance(log_likelihood, dict):
+                log_lik_to_obs_name = {v: f"{k}.{{}}".format for k, v in log_likelihood.items()}
+                columns = {
+                    log_lik_to_obs_name[col_varidx[0]](col_varidx[1]): idx
+                    for col, idx in self.posterior_columns.items()
+                    if any(item == col.split(".")[0] for item in log_likelihood.values())
+                    and (col_varidx := col.split("."))
```
ahartikainen, line 445:
```suggestion
                    and (col_varidx == col.split("."))
```

I guess here is a typo?

OriolAbril:
> I wonder if we should implement these dict structures for posterior/prior predictive group too?

I think so (and thought that was supported already :flushed: ). I was just about to comment on #1977 that it is fine to do both changes in my opinion, but if we have to choose only one then I'd prefer the change in the converter because it allows renaming once instead of needing to pass data_pairs or y_hat explicitly every time ppc/loo_pit/lm even? are called.

arviz/data/io_cmdstan.py
erocarrera: It's not a typo. It always evaluates to true, the assignment expression makes `col_varidx` the result of the split. I need both parts to compose the key in the dictionary, so this seemed neater than making two calls to `split`.

ahartikainen: Ok, that came into python in 3.8 so our 3.7 tests are failing

erocarrera:
> > I wonder if we should implement these dict structures for posterior/prior predictive group too?
> 
> I think so (and thought that was supported already 😳 ). I was just about to comment on #1977 that it is fine to do both changes in my opinion, but if we have to choose only one then I'd prefer the change in the converter because it allows renaming once instead of needing to pass data_pairs or y_hat explicitly every time ppc/loo_pit/lm even? are called.

That's why closed #1977, it made calling `loo_pit` rather cumbersome and it seemed to provide no additional value. I think these *dict*s are a cleaner interface to shape up Stan data into a clean InferenceData object.

I see that `from_pystan` only takes dicts for `coords`, `dims`, `log_likelihood`, and `dtypes`. After this change, `from_cmdstan` will also support *dict*s for those same parameters.
Neither Stan interface takes *dict*s for posterior/prior predictive groups, although that'd be neat to do away with the `*_hat` alternate names in those groups, it would likely also simplify calls to stats/plot functions that may need hints about which `*_hat` to pick. But extending those to use *dict*s sounds like a more involved PR.

ahartikainen:
Ok, then let's keep this clean

ahartikainen (approved):

erocarrera:
Ok, that last commit should fix the issue.

arviz/data/io_cmdstan.py
```diff
@@ -430,13 +436,23 @@ def log_likelihood_to_xarray(self):
...
-                col: idx
-                for col, idx in self.posterior_columns.items()
-                if any(item == col.split(".")[0] for item in log_likelihood)
-            }
+            if isinstance(log_likelihood, dict):
+                log_lik_to_obs_name = {v: f"{k}.{{}}".format for k, v in log_likelihood.items()}
+                columns = {
+                    log_lik_to_obs_name[col](varidx): idx
+                    for col, varidx, idx in (
+                        (*col.split("."), idx) for col, idx in self.posterior_columns.items()
```
ahartikainen, line 444: Is there a possibility for something like `log_lik.1.2.3`?

erocarrera: Thanks @ahartikainen , good catch! I guess that's a possibility. I've just committed a tweak to split only on one dot into two strings: column name and indices. I am not sure though how that is handled down the line, does it look correct to you?

ahartikainen: It looks good.

ahartikainen (approved):

erocarrera:
I did get carried away and I wrote overly complicated code, pylint caught an issue with the closure. I simplified it.

OriolAbril (approved):

Issue #1981, opened 2022-02-16
`plot_pair` marginals have inconsistent axes


wm1995:
**Describe the bug**
When plotting a `plot_pair` with marginal distributions, the tick marks do not fall in locations consistent with the joint distribution plots beneath, and there are no axis labels so it's impossible to tell what the marginal should look like. If the axes are generated using `plt.subplots` with `sharex='col'`, the axis tick marks are aligned correctly, but all of the labels are absent.

**To Reproduce**
```python
import arviz as az
import matplotlib.pyplot as plt
import numpy as np

# Generate some random data from a multivariate normal
rng = np.random.default_rng(42)
data = rng.multivariate_normal([0, 0, 0], cov=np.diag([0.1, 1, 1]), size=(1, 2000))

# Naive plot_pair
az.plot_pair(data, kind="kde", marginals=True)
plt.savefig("naive.png")

# plot_pair with shared x axis in each column
fig, ax = plt.subplots(3, 3, sharex="col")
az.plot_pair(data, kind="kde", marginals=True, ax=ax)
plt.savefig("sharex.png")
```

The first plot has inconsistent x axes between the joint distribution plots and the marginal distribution plots - note how the tick marks are displaced in the leftmost and central marginal distribution plots relative to the joint plots beneath them:
![naive.png](https://user-images.githubusercontent.com/30499074/154265594-089955cf-90e3-49ff-bc49-0984d109353d.png)

The second plot has consistent x axes but no x axis labels:
![sharex.png](https://user-images.githubusercontent.com/30499074/154265600-9f8a64bd-2c13-4528-a773-811b910db655.png)

**Expected behavior**
I'd expect the first plot to have consistent axes for all subplots including the marginals, or at least provide axis labels for the subplots that have different axes. I'd definitely expect the second result to not get rid of axis labels.

**Additional context**
```
# pip freeze
arviz @ git+https://github.com/arviz-devs/arviz@5af74712950c99d486cc0f62db66537670e3bfcc
black==22.1.0
cftime==1.5.2
click==8.0.3
cycler==0.11.0
fonttools==4.29.1
kiwisolver==1.3.2
matplotlib==3.5.1
mypy-extensions==0.4.3
netCDF4==1.5.8
numpy==1.22.2
packaging==21.3
pandas==1.4.1
pathspec==0.9.0
Pillow==9.0.1
platformdirs==2.5.0
pyparsing==3.0.7
python-dateutil==2.8.2
pytz==2021.3
[...]

PR #1985, opened 2022-02-21
Fix bug in pair plot axis handling

wm1995:
## Description

This is an attempt at fixing #1981, in which the subplot axes do not have the same range by default but also lack unique labels. In doing so, I hope to correct some of the counterintuitive behaviour around passing "sharex" or "sharey" as backend parameters. I've written a fix, ~and am in the process of adding unit tests for said fix~ and unit tests for the fix. 

**Update** __(23/02/22)__:
I generated an example plot of the marginals, based on the plotpair example gallery code, to illustrate that the marginal axes now fall in the correct place.
example

Cf. the old behaviour, where the marginal plot axis ticks are not aligned with the plots beneath them, but the labels are also missing.

broken

## Checklist

(I removed the "new features" documentation from the list because this isn't a new feature, just a bug fix to correct unexpected behaviour)

### DIFF
--- a/arviz/plots/backends/matplotlib/pairplot.py
+++ b/arviz/plots/backends/matplotlib/pairplot.py
@@ -4,7 +4,6 @@ from copy import deepcopy
 
 import matplotlib.pyplot as plt
 import numpy as np
-from matplotlib.ticker import NullFormatter
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 
 from ....rcparams import rcParams
@@ -324,18 +323,29 @@ def plot_pair(
                                 **reference_values_kwargs,
                             )
 
-                if j != vars_to_plot - 1:
-                    ax[j, i].axes.get_xaxis().set_major_formatter(NullFormatter())
-                else:
+                if j == vars_to_plot - 1:
                     ax[j, i].set_xlabel(f"{flat_var_names[i]}", fontsize=ax_labelsize, wrap=True)
-                if i != 0:
-                    ax[j, i].axes.get_yaxis().set_major_formatter(NullFormatter())
-                else:
+                elif backend_kwargs.get("sharex") is None:
+                    try:
+                        ax[j, i].axes.sharex(ax[-1, i])
+                        plt.setp(ax[j, i].get_xticklabels(), visible=False)
+                    except ValueError:
+                        # Raised if user has provided shared axes
+                        pass
+                if i == 0:
                     ax[j, i].set_ylabel(
                         f"{flat_var_names[j + not_marginals]}",
                         fontsize=ax_labelsize,
                         wrap=True,
                     )
+                elif backend_kwargs.get("sharey") is None:
+                    try:
+                        if i != j:
+                            ax[j, i].axes.sharey(ax[j, 0])
+                        plt.setp(ax[j, i].get_yticklabels(), visible=False)
+                    except ValueError:
+                        # Raised if user has provided shared axes
+                        pass
                 ax[j, i].tick_params(labelsize=xt_labelsize)
 
     if backend_show(show):
--- a/arviz/plots/backends/matplotlib/pairplot.py
+++ b/arviz/plots/backends/matplotlib/pairplot.py
@@ -245,11 +245,19 @@ def plot_pair(
         point_estimate_marker_kwargs.setdefault("s", markersize + 50)
 
         if ax is None:
+            backend_kwargs.setdefault("sharex", "col")
+            if not_marginals:
+                backend_kwargs.setdefault("sharey", "row")
             fig, ax = plt.subplots(
                 vars_to_plot,
                 vars_to_plot,
                 **backend_kwargs,
             )
+            if backend_kwargs.get("sharey") is None:
+                for j in range(0, vars_to_plot):
+                    for i in range(0, j):
+                        ax[j, i].axes.sharey(ax[j, 0])
+
         hexbin_values = []
         for i in range(0, vars_to_plot):
             var1 = plotters[i][-1].flatten()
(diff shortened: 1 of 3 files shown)


wm1995:
Right, I think this PR is ready for review now. The test I put together isn't strictly necessary, I think, but it is useful to make sure that all the axes you'd expect to be shared are indeed shared.

wm1995:
Another example of the fix introduced by this PR is in the `plot_pair_point_estimate` example - cf. [the original](https://arviz-devs.github.io/arviz/examples/plot_pair_point_estimate.html) and [the updated](https://arviz--1985.org.readthedocs.build/en/1985/examples/plot_pair_point_estimate.html) pages in the docs - the point estimates now actually line up in the marginal distribution and the joint distribution plots.

arviz/plots/backends/matplotlib/pairplot.py
```diff
@@ -246,11 +245,19 @@ def plot_pair(
         point_estimate_marker_kwargs.setdefault("s", markersize + 50)
 
         if ax is None:
+            backend_kwargs.setdefault("sharex", "col")
```
OriolAbril, line 248: should we allow not sharing axes? or sharing the axis along a whole row with marginals?

OriolAbril (commented):
Thanks for the improvement

arviz/plots/backends/matplotlib/pairplot.py
wm1995: Both of these should be allowed by specifying `backend_kwargs`. In the first instance, it would be 
```python
backend_kwargs = {"sharex": False, "sharey": False}
```
and in the second instance
```python
backend_kwargs = {"sharey":"row"}
```

OriolAbril: yes, sorry I wasn't very clear. I am not sure making this possible benefits anyone, if a user sets share# to False, they will still not one set of ticklablels per plot yet it is supported and it makes ArviZ codebase bigger and harder to test and maintain (my main concern here). I think one could even pass `sharex="row"` if they wanted to see everything burn.

I'd personally hardcode sharing on a row/column basis to keep things simple as I think we all agree that should be the default, but I'm open to be convinced that maybe sharing set to True could also be useful and we might want to allow some extra freedom.

Side note, given how the code works and is tested we might want to keep this as is and defer to https://github.com/arviz-devs/GSoD/blob/main/2021/viz_evaluation.md on the question of "to what extent should bad practices be possible when using ArviZ?"

wm1995: OK, so just to be sure I understand, you're proposing two different options:
1. To hardcode axis sharing and ignore any sharing provided by the user (perhaps issuing a warning if sharex/sharey are provided in backend_kwargs to say that they will be ignored?)
2. To allow the user to set sharex=sharey=True and only that, so that all joint and marginal plots share the same axes? Perhaps through a new keyword? (I suggest a new keyword because sharey=True would break the marginal plots)
3. To close the PR and accept this bad practice? (I wasn't sure how to interpret "as is" in what you said)

I'd argue strongly against the third option, as at the moment it's impossible to produce plots that could be wrangled into publishable form with the code as is. If this is the desired option, I'd argue that the changes to L328 and L332 should be retained, because at least then the user can provide axes that are shared properly and not have the axis labels clobbered by the NullFormatter.

wm1995: Also side note, if we go down the opinionated route, I'd get rid of the y axis tick marks and labels on the marginal plots, but I didn't put that in here because it felt too opinionated to do without discussion.

OriolAbril: By "keeping this as is" I meant the PR, merging it without extra changes, definitely not keeping this bug. 

Regarding option 2, I would use `True` to not diverge too much from matplotlib. It is directly used when no marginals are present, and should be popped and handled correctly when marginals are present. As you say, sharey=True breaks the plot and should not be an option in my opinion.

wm1995: Do you have a preferred way forward here? Here are the changes I would make for each option:

1. If either sharex or sharey are provided, pop them and warn the user they'll be ignored
2. Allow sharex = True; if sharey = True is provided, allow it if marginal plots are off, otherwise share the y axis on all non-marginal plots
3. Make no changes to the PR

My reservations about the second option are that I'm not sure we should be going against default matplotlib behaviour; it could be counterintuitive if the user passes something expecting a certain output and arviz does something else, even if the input from the user would generate nonsensical output (such as sharing the marginal y axes). I could add a `shareall` keyword to the method that would internally set sharex=True, share all non-diagonal y axes, and warn the user that if they've provided sharex/sharey, those options would be ignored when `shareall=True`... but this does add to the code somewhat, so I can see that not being a desirable option.

It should be noted that under the current PR behaviour, if the user wanted all the joint plots to share a single y axis, they would have to create axes and set up the sharing manually, then pass those axes to arviz. So it's possible but fiddly...

Let me know what your preferred option is and I'll have a go at coding it up.

OriolAbril: I prefer option 1 which technically would still allow sharex=True or even other stranger configurations provided the user generates the axis grid manually and uses the `ax` argument but makes our code and testing significantly less complex.

wm1995: Right, I've had a go at implementing option 1 - please let me know your thoughts/suggestions/corrections

arviz/plots/backends/matplotlib/pairplot.py
```diff
@@ -246,11 +245,29 @@ def plot_pair(
...
+                    UserWarning,
+                )
+            if backend_kwargs.pop("sharey", None) is not None:
+                warnings.warn(
+                    "'sharey' keyword is ignored. For non-standard sharing, provide 'ax'.",
+                    UserWarning,
+                )
+            backend_kwargs["sharex"] = "col"
+            if not_marginals:
+                backend_kwargs.setdefault("sharey", "row")
```
OriolAbril, line 260:
```suggestion
                backend_kwargs["sharey"] = "row"
```

It makes no difference because the key if present has been removed, but I'd use the same method to set both sharex and sharey.

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -594,6 +594,42 @@ def test_plot_pair_shapes(marginals, max_subplots):
     assert ax.shape == (side, side)
 
 
+@pytest.mark.parametrize("sharex", ["col", None])
```
OriolAbril, line 597: I'd prefer if testing was with backend_kwargs having a sharex key or not having the key. The fact that the key with a value of None is equivalent to no key is an implementation detail that users should not use. We could parametrize with True/False and then do something like

```
backend_kwargs = {}
if sharex:
    backend_kwargs["sharex"] = ...
...
```

same would happen also with sharey.

OriolAbril (commented):
I left two minor comments, thank you so much for your contribution and patience while we discussed the approach to implement.

arviz/plots/backends/matplotlib/pairplot.py
wm1995: Whoops, I actually meant to do that, just forgot to change that line!

arviz/tests/base_tests/test_plots_matplotlib.py
wm1995: That makes perfect sense (and is how I implemented it the first time, I can't remember why I changed it)... I'll revert it to mirror your suggestion now

wm1995:
I've made the requested changes, will just make sure that the CI runs alright and then if all is OK I'll leave it with you. Thanks for your help in putting this together and for being so responsive!

OriolAbril (approved):

PR #1986, opened 2022-02-22
[WIP] add find_prior function

aloctavodia:
This is a proof of concept of a new ArviZ function, tentatively named `find_prior`. This is inspired in a recent function added to PyMC, but here I am trying to extend the functionality and offer a solution that works outside of PyMC.

The function is called

```python
az.find_prior("normal", lower=-3, upper=3, mass=0.90)
```

In this case we are asking to return the parameters of a normal distribution with 90% mass between `lower` and `upper`. The function returns 3 objects:

* axes: matplotlib axes
* rv_frozen : scipy.stats.distributions.rv_frozen
* opt: scipy.optimize.OptimizeResult

Having access to the `rv_frozen`, allow the user to then ask for other quantities like the mean, std, quantiles. make other plots, etc. 

Optionally, the function also generates a plot.

for example 
```python
lower = 0
upper = 1.5
ax, st, opt = az.find_prior("lognormal", lower=lower, upper=upper, mass=0.95)
ax, st, opt = az.find_prior("lognormal", lower=lower+0.5, upper=upper, mass=0.95, ax=ax)
```
index

Another example. The `extra` argument is used to indicate the degrees of freedom of the Student-T distribution.

```python
ax, st, opt = az.find_prior("normal", lower=-3, upper=3, mass=0.90)
ax, st, opt = az.find_prior("student", lower=-3, upper=3, mass=0.90, extra=5, ax=ax)
```
index

The function has a `parametrization` argument, currently it only has two options "pymc" and "scipy". The reason for having this  arguments is that distributions are parametrized in different ways in different PPLs/packages. 

Probably still a lot of room for improvement but I would like to see if we can agree on a minimal set of features for a first merge. Also, I am suggesting to host at least part of the new "prior elicitation" functionality inside a new `priors` module/folder with a new `prior_utils.py`, that could be shared with other upcoming prior-related functions like those in #1965.

### DIFF
--- a/arviz/__init__.py
+++ b/arviz/__init__.py
@@ -32,6 +32,7 @@ _log = Logger("arviz")
 from .data import *
 from .plots import *
 from .plots.backends import *
+from .priors import *
 from .stats import *
 from .rcparams import rc_context, rcParams
 from .utils import Numba, Dask, interactive_backend
--- a/arviz/priors/__init__.py
+++ b/arviz/priors/__init__.py
@@ -1,0 +1,9 @@
+# pylint: disable=wildcard-import
+"""Functions to assist prior elicitation"""
+from .find_prior import find_prior
+from .prior_utils import *
+
+
+__all__ = [
+    "find_prior",
+]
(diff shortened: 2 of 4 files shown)


aloctavodia:
closing in favor of https://github.com/arviz-devs/preliz

PR #1988, opened 2022-02-26
Added filter_vars argument to plot_density

### LINKED ISSUE #1225: plot_density has no filter_vars argument
I think it should have the argument. It may be a little more complicated given the multiple idata input but is should still be feasible.

prahasR:
## Description

`filter_vars` argument added to `plot_density`.
Fixes #1225 

## Checklist

### DIFF
--- a/arviz/tests/base_tests/test_plots_matplotlib.py
+++ b/arviz/tests/base_tests/test_plots_matplotlib.py
@@ -108,25 +108,6 @@ def data_list():
 
 
 @pytest.mark.parametrize(
-    "var_args",
-    [
-        (["ta"], ["beta1", "beta2", "theta"], "like"),
-        (["~beta"], ["phi", "theta"], "like"),
-        (["beta[0-9]+"], ["beta1", "beta2"], "regex"),
-        (["^p"], ["phi"], "regex"),
-        (["~^t"], ["beta1", "beta2", "phi"], "regex"),
-    ],
-)
-def test_var_names_filter(var_args):
-    samples = np.random.randn(10)
-    data1 = dict_to_dataset({"beta1": samples, "beta2": samples, "phi": samples})
-    data2 = dict_to_dataset({"beta1": samples, "beta2": samples, "theta": samples})
-    data = [data1, data2]
-    var_names, expected, filter_vars = var_args
-    assert _var_names(var_names, data, filter_vars) == expected
-
-
-@pytest.mark.parametrize(
     "kwargs",
     [
         {"point_estimate": "mean"},
--- a/arviz/tests/base_tests/test_plots_matplotlib.py
+++ b/arviz/tests/base_tests/test_plots_matplotlib.py
@@ -10,7 +10,7 @@ from matplotlib import animation
 from pandas import DataFrame
 from scipy.stats import gaussian_kde
 
-from ...data import from_dict, load_arviz_data, dict_to_dataset
+from ...data import from_dict, load_arviz_data
 from ...plots import (
     plot_autocorr,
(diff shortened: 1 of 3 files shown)


prahasR:
@OriolAbril, I have added the code in `densityplot.py` and `test_plots_matplotlib.py`. Also I am not sure that how do I run tests to check the code. Can you please help me.

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -107,6 +107,25 @@ def data_list():
     return list(range(11, 31))
 
 
+@pytest.mark.parametrize(
```
OriolAbril, line 110: these kwargs don't affect the behaviour of `filter_vars` so they should be removed to keep tests as short and to the point as possible. See also comment below.

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -107,6 +107,25 @@ def data_list():
...
+        {"point_estimate": "mean"},
+        {"hdi_prob": 0.94},
+        {"outline": True},
+        {"colors": ["g", "b", "r", "y"]},
+        {"colors": "k"},
+        {"hdi_markers": ["v"]},
+        {"shade": 1},
+    ],
+)
+@pytest.mark.parametrize("filter_vars", [None, "like", "regex"])
```
OriolAbril, line 122: The test should check that filter vars is behaving as expected, especially when there are multiple inferencedata objects as inputs, we already have tests for the other cases. Now thinking about these tests I realized in order to keep tests short and to the point, it is best to test only the behaviour of `_var_names`, no need to run it _through_ plot_density. 

You can use the pattern in https://github.com/arviz-devs/arviz/blob/main/arviz/tests/base_tests/test_utils.py#L69 (with less cases, 2 per value of filter_var is probably enough) but using an iterable of 2 inferencedata as the 2nd (data) input. These two inferencedata inputs should have some variable names in common but also some that differ (there is also no need for as many variables in the first example, 2 in common and 1 different for each should be good)

I'd also recommend using `plot_density` after installing your local version on an interactive interpreter or notebook before writing the tests.

OriolAbril (commented):
You can run the tests using `pytest` as a command line executable and providing the path to the tests, something like `pytest arviz/tests/base_tests/test_utils.py` (after the updates)

prahasR:
@OriolAbril Thankyou for the suggestions. I have applied the changes pls review.
Also I wanted to ask that you suggested that for testing our `filter_vars` argument we should test the behaviour of  `_var_names` instead of  `plot_density` to keep the test short, but we already have a test for this in `test_utils.py` as you have mentioned. Or I misunderstood something here?

OriolAbril:
The test looks good like this, but it should be in `test_utils.py`. 

> but we already have a test for this in test_utils.py as you have mentioned.

we have similar tests but not the same tests, the other tests don't check the behaviour with multiple input datasets, they check the behaviour of filter_like on a single dataset.

OriolAbril:
you have to remove the (now unused) imports in test_matplotlilb for CI to pass: https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=5507&view=logs&j=7f538728-4589-5563-0e87-d0823eb40420&t=6753be35-af0d-5ca7-e801-8ec60efd9201&l=14

prahasR:
Yes I got that, Thanks. I think now there will be no issue.

OriolAbril (approved):

OriolAbril:
Thanks @prahasR !

Issue #1993, opened 2022-03-10
arviz.concat fails if a variable is named "shift"


jt-lab:
**Describe the bug**
If I concatenate two traces, arviz.concat fails if a variable is named "shift" with the error message:
"'function' object has no attribute 'dims'". 

I have traced the issue down to the following problem: concat uses getattr(group_data, var).dims to get the dimensions of variables. Apparently the group_data object contains a method called "shift" and instead of the variable this method is returned by getattr.

Once I rename my variable to something else, it works.

Since "shift" is not an unlikely name for a variable, it might be good to have something to safeguard against this. Perhaps data groups have further functions that could conflict with other variable names.

**To Reproduce**
To reproduce, concatenate two InferenceData objects that contain a variable named "shift".

**Expected behavior**
Successful concatenation of InferenceData regardless the name of the variables.

**Additional context**
Arviz 0.11.2

ahartikainen:
Hi,

yeah, sounds like a bug.

I think we should change our code a bit to fix this.

https://github.com/arviz-devs/arviz/blob/4544dd0bceb2d85496d3ff962db009f93abfa768/arviz/data/inference_data.py#L1995

jt-lab:
I have already thought about how to handle this but have no good idea so far.
Prepending any variable names internally with some string before concatenating could work. Or perhaps making sure the data group's functions have uncritical names (but perhaps they inherit some). Using another mechanism to access attributes that can ignore functions would be great.

OriolAbril:
does `group_data[var].dims` work? I think the issue is that getattr is not the way to subset variables in a dataset and fails if the variable has the same name as one of its methods (I'd expect the same to happen for `mean`, `quantile`, `dims`... which should all be valid variable names from the xarray side)

jt-lab:
Yes, `group_data[var].dims` works. Was there a reason to use `getattr` in the first place? Is it faster?

ahartikainen:
No real reason (just a bad habid), looking a git blame I see myself there so :D

OriolAbril:
I might have been a lapsus between dataset variable retrieval and inferencedata group retrieval, especially with the attr retrieval some lines below. Not sure speed is relevant on this operation, we have many other places where we should look at optimizations before considering any analysis or benchmark here

OriolAbril:
I see now @ahartikainen already answered, @jt-lab do you want to send a PR with the fix?

jt-lab:
OK!

PR #1995, opened 2022-03-10
Changed attribute access from getattr to []

jt-lab:
Changed attribute access from getattr to [] in concat to fix arviz-devs/arviz#1993.
I have not changed it in lines 2012/2089 and 2017/2094 where "attrs" are handled.
But I did test concatenating with variables named "attrs" which seems to work.

### DIFF
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -1918,14 +1918,14 @@ def concat(*args, dim=None, copy=True, inplace=False, reset_dim=True):
                         " combine InferenceData with overlapping groups"
                     )
                     raise TypeError(msg)
-                group_data = arg[group]
+                group_data = getattr(arg, group)
                 args_groups[group] = deepcopy(group_data) if copy else group_data
         # add arg0 to args_groups if inplace is False
         # otherwise it will merge args_groups to arg0
         # inference data object
         if not inplace:
             for group in arg0_groups:
-                group_data = arg0[group]
+                group_data = getattr(arg0, group)
                 args_groups[group] = deepcopy(group_data) if copy else group_data
 
         other_groups = [group for group in args_groups if group not in SUPPORTED_GROUPS_ALL]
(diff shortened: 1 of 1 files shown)


arviz/data/inference_data.py
```diff
@@ -2066,21 +2066,21 @@ def concat(*args, dim=None, copy=True, inplace=False, reset_dim=True):
...
                     if not inplace:
                         group0_data = deepcopy(group0_data)
                     group0_vars = group0_data.data_vars
 
                     for var in group_vars:
                         if var not in group0_vars:
-                            var_data = getattr(group_data, var)
-                            getattr(arg0, group)[var] = var_data
+                            var_data = group_data[var]
+                            arg0[group][var] = var_data
```
OriolAbril, line 2080: I think that setitem is not implemented for Inferencedata, so this probably won't work.

arviz/data/inference_data.py
```diff
@@ -1918,14 +1918,14 @@ def concat(*args, dim=None, copy=True, inplace=False, reset_dim=True):
                         " combine InferenceData with overlapping groups"
                     )
                     raise TypeError(msg)
-                group_data = getattr(arg, group)
+                group_data = arg[group]
```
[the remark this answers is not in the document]
jt-lab: Makes sense!

arviz/data/inference_data.py
jt-lab: Oh, I see.

jt-lab: How do I update the PR?

OriolAbril: this specific line should not be updated, it is already using the right `[var]` and should keep the `getattr` for the `group`. As a general rule, I think that if what is being retrieved is a group it should be getattr, [] otherwise. The use of [] and getattr is not consistent and a bit confusing, but it looks like the "index" use is consistent: `getattr(group_data, var)` is using var as "index" and should use [] instead, if using `getattr(..., group)` then getattr should be kept

OriolAbril: ooh, sorry. You should add a new commit to the same branch and then push. The PR will be updated with the new commit automatically

arviz/data/inference_data.py
```diff
@@ -1980,7 +1980,7 @@ def concat(*args, dim=None, copy=True, inplace=False, reset_dim=True):
                     if not inplace and group in inference_data_dict:
                         group0_data = inference_data_dict[group]
                     else:
-                        group0_data = getattr(arg0, group)
+                        group0_data = getattr(arg, group)
```
OriolAbril, line 1983:
```suggestion
                        group0_data = getattr(arg0, group)
```

OriolAbril:
Fixed a typo, now it should be ready to merge I think if tests pass. Thanks for raising the issue and sending the PR!

jt-lab:
Thanks for your help @OriolAbril!

PR #1997, opened 2022-03-17
Allow varnames that aren't strings to be used in InferenceData

zaxtax:
This updates the util functions so that variables don't need to be represented as strings. As far as I know
this is the only place in the codebase that assumes it.

### DIFF
--- a/arviz/utils.py
+++ b/arviz/utils.py
@@ -50,7 +50,7 @@ def _var_names(var_names, data, filter_vars=None):
         else:
             all_vars = list(data.data_vars)
 
-        all_vars_tilde = [var for var in all_vars if var.startswith("~")]
+        all_vars_tilde = [var for var in all_vars if repr(var).startswith("~")]
         if all_vars_tilde:
             warnings.warn(
                 """ArviZ treats '~' as a negation character for variable selection.
@@ -94,7 +94,7 @@ def _subset_list(subset, whole_list, filter_items=None, warn=True):
         if isinstance(subset, str):
             subset = [subset]
 
-        whole_list_tilde = [item for item in whole_list if item.startswith("~")]
+        whole_list_tilde = [item for item in whole_list if repr(item).startswith("~")]
         if whole_list_tilde and warn:
             warnings.warn(
                 "ArviZ treats '~' as a negation character for selection. There are "
@@ -105,7 +105,7 @@ def _subset_list(subset, whole_list, filter_items=None, warn=True):
             )
 
         excluded_items = [
-            item[1:] for item in subset if item.startswith("~") and item not in whole_list
+            item[1:] for item in subset if repr(item).startswith("~") and item not in whole_list
         ]
         filter_items = str(filter_items).lower()
         not_found = []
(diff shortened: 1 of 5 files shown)


OriolAbril:
Thanks! I have a couple requests:

1. can you add some tests too to make sure we don't go back to string only behaviour? I think ideally a direct test using _var_names, then one using plot_posterior and one with plot_forest or plot_density with multiple models (that also use var names to align the models)
2. can you update https://arviz-devs.github.io/arviz/user_guide/plots_arguments_guide.html#var-names to show it works and how it works with non string variables?

zaxtax:
@OriolAbril I think that about does it.

arviz/utils.py
```diff
@@ -105,7 +112,9 @@ def _subset_list(subset, whole_list, filter_items=None, warn=True):
             )
 
         excluded_items = [
-            item[1:] for item in subset if item.startswith("~") and item not in whole_list
+            item[1:]
+            for item in subset
+            if _repr_str(item).startswith("~") and item not in whole_list
```
OriolAbril, line 117: Do we want the `~` effect to work on non-string variables? Can it really be used in those cases?

doc/source/user_guide/plots_arguments_guide.md
```diff
@@ -83,6 +83,21 @@ Use `var_names` to indicate which variables to exclude with the `~` prefix:
 az.plot_posterior(centered_eight, var_names=['~mu', '~theta']);
 ```
 
+Variables do not need to be strings to be used
```
OriolAbril, line 86: What are the conditions needed for that to work? Am I right to assume they are only two:
1. Having `__repr__` defined
2. Being a hashable (aka [valid xarray variable name](https://xarray.pydata.org/en/stable/user-guide/terminology.html#term-name))

which brings to the 2nd question. Could this example use int or tuple variable names instead of defining a custom class?

OriolAbril: and maybe 3rd question, if the `~` were to only work on strings, could the condition be reduced to being a valid xarray name only? it looks like the _repr helper is only used right before using `startswith`, maybe is could be updated to be `check_tilde_start` or something like this to always return False if not str (I think, it is late and my logic and double negations already went to sleep, I only know it would need to be constant), check `.startswith` if str.

arviz/utils.py
zaxtax: I think the subset feature probably shouldn't work on non-string variables but I think the idea of extending similar functionality to them.

arviz/utils.py
```diff
@@ -16,6 +16,13 @@
 STATIC_FILES = ("static/html/icons-svg-inline.html", "static/css/style.css")
 
 
+def check_tilde_start(x):
+    if isinstance(x, str) and x.startswith("~"):
+        return True
+    else:
+        return False
```
OriolAbril, line 23:
```suggestion
    return False
```

OriolAbril (approved):

aloctavodia (approved):

zaxtax:
Yep sounds good

Issue #1298, opened 2020-07-16
Add helper methods for azsummary


canyon289:
## Tell us about it
Adding median and median absolute deviation would be to ArviZ summary would be nice, particular as in Regression and Other Stories those are the two summary statistics used most often.

See this ticket from Bambi
https://github.com/bambinos/bambi/issues/223

https://github.com/bambinos/bambi/issues/223
![image](https://user-images.githubusercontent.com/7213793/87624032-8e8d8c00-c6db-11ea-80cd-d184645c3928.png)

## Thoughts on implementation
Maybe we add a module thats like convenient functions or something? Whatever code we use would either go into the summary method below, or be located somewhere else. Open to ideas

https://github.com/arviz-devs/arviz/blob/master/arviz/stats/stats.py#L1153

ahartikainen:
We could add some predefined but optional stats.

Have you tried `stat_funcs` kw?

But for example mad is already defined for ess / mcse then we could have option to spit out mad stats only

We could have options for the following

    mean stats --> mean, sd, ess_mean, ess_sd, mcse...
    quantile stats --> quantiles (list input), ess_quantiles, mcse_q...
    mad stats --> median, mad, (iqr?)...

AlexAndorra:
Adding the median seems nice, especially to easily spot potentially skewed distribution. 
I'm not sure about adding MAD though, as the HDI is already in `summary` IIRC, and the latter is more robust to distributional assumptions and interpretable even when we're not working with Normal distributions -- unlike MAD estimates, IIUC

canyon289:
Only proposing MAD because its the main metric used in ROS to characterize variation in parameters, but am not strongly tied to it myself

aloctavodia:
I like @ahartikainen proposal.

virajone:
Hey, I'd like to know if I can pickup this issue ?

OriolAbril:
I think the goal here would be to add a new argument to `az.summary` to set which statistics and diagnostics to use. 

Users can already use the `stats_funcs` like in the code from the first comment to do this, but some of the combinations are common enough that having a more convenient way to get a custom summary would be a valuable feature.

I think that there is still some loose ends to tie before being able to implement it such as the argument name and the combinations to add, @ahartikainen already started with a proposal above which should be a great start, from here on we should define all the columns of each of the alternatives

virajone:
[...]

PR #1998, opened 2022-03-18
[WIP] Add new argument to az.summary

prahasR:
## Description
To solve #1298 
A new argument:`stat_focus` is added in `summary` of `stats.py`. Currently the  argument is implemented such that it will allow user to change the "focus" of `az.summary` from `mean` to `median`.  

String should be passed to `stat_focus`. If nothing is passed then `az.summary` will give default output.
Eg:
```
>>>data = az.dict_to_dataset({
                            "theta": [np.random.normal(0,1,100), np.random.normal(0,2,100)], 
                            "alpha": [np.random.normal(7,1.5,100), np.random.normal(7,3,100)]
                           })
>>>az.summary(data,  stat_focus = "median")
       median    mad
theta   0.079  0.888
alpha   7.196  1.391
``` 
 

## Checklist

- [x] ~Includes a sample plot to visually illustrate the changes (only for plot-related functions)~


OriolAbril:
I like the idea of a stats_focus argument. Can you update the description or issue with how it should work and give some examples/screenshots of its use and generated outputs?

prahasR:
Thankyou for the review!
> Can you update the description or issue with how it should work and give some examples/screenshots of its use and generated outputs?

prahasR:
Sorry for closing this, It accidently got closed due to some issues, I will try to resolve this.

arviz/plots/densityplot.py
```diff
@@ -19,6 +19,7 @@ def plot_density(
     var_names=None,
     filter_vars=None,
     combine_dims=None,
+
```
OriolAbril, line 22: This change should be undone, and it looks like it's part of why black is failing CI for this PR. See https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=5559&view=logs&j=7f538728-4589-5563-0e87-d0823eb40420&t=af0f9f62-4f87-5492-4555-4ab2d5002559&l=20

arviz/stats/stats.py
```diff
@@ -1269,6 +1276,11 @@ def summary(
     extra_metrics = []
     extra_metric_names = []
 
+    if stat_focus is not None:
+        if stat_focus == "median":
+            stat_funcs = {"median": np.median, "mad": st.median_abs_deviation}
+            extend = False
```
OriolAbril, line 1282: Does summary only return the median and the mad when `stat_focus="median"`?

arviz/stats/stats.py
```diff
@@ -1259,6 +1262,10 @@ def summary(
     if not isinstance(kind, str) or kind not in kind_group:
         raise TypeError(f"Invalid kind: '{kind}'. Kind options are: {kind_group}")
 
+    focus_group = ("mean", "median")
+    if not isinstance(stat_focus, str) or (stat_focus not in focus_group):
+        raise TypeError(f"Invalid format: '{stat_focus}'. Focus options are: {focus_group}")
```
OriolAbril, line 1267: We could make stat_focus default to `None` and when this happens take the default from rcParams, not necessarly in this PR but trying to discuss and agree on an api. @canyon289?

arviz/stats/stats.py
prahasR: Yes currently it gives median and mad only, I have added the example in the Description.

arviz/stats/stats.py
prahasR: Current implementation of `stat_focus` would only allow `"mean"` and `"median"` to be passed to `stat_focus`. If nothing is passed then the summary returns the default output that is it returns mean focused summary ( therefore I set the default to `"mean"`). And if any other input is passed to `stat_focus` then `az.summary` will produce this Type Error.
Should I change it to `None`? Also should I modify the error handling part here? 

Also what should be the desired output if we provide a dictionary in `stat_funcs` in addition to providing a string to `stat_focus`?

arviz/plots/densityplot.py
prahasR: I have fixed this.

OriolAbril (commented):
I took a long look at this and have several ideas. They mean some changes, sorry about that. Adding new functionality is always tricky, especially choosing the API part.

I think it would be best to keep the stat_focus and the stat_kwargs separate. I am adding links to the code I refer too in the text so you know what I mean by each section. Right now,
summary starts [computing quantities](https://github.com/arviz-devs/arviz/blob/main/arviz/stats/stats.py#L1290-L1339), trying to compute only those necessary. Then it [initializes
two lists that track the names and the values](https://github.com/arviz-devs/arviz/blob/main/arviz/stats/stats.py#L1341-L1376) _from scratch_ and by using the named quantities that have been computed. After that are [some substitutions for circular variables](https://github.com/arviz-devs/arviz/blob/main/arviz/stats/stats.py#L1378-L1386) (back on that later) and the [extension with stat_func metrics](https://github.com/arviz-devs/arviz/blob/main/arviz/stats/stats.py#L1388-L1389) as last step. After that it is just dataframe/dataset wrangling to return the results well rounded and in a pretty format.

I think the best way to go about this would be to restructure a bit these steps and instead do:

1. define the lists containing metric names and values as empty lists
2. add the metric names and values to these lists as soon as they are calculated, they are already calculated in the right order. Here is where the stat_focus is taken care of
3. make the circular variable corrections (with a bit more care)
4. the stat_funcs extensions are already the same.

Implementation details. That would mean for example that the [block that computes
the "stats"](https://github.com/arviz-devs/arviz/blob/main/arviz/stats/stats.py#L1290-L1297) would change to:

```python
# note that in this scenario, the variables metrics and metrics_names have already been defined before this block of code
if extend and kind in ["all", "stats"]:
    if stat_focus == "mean":
        mean = dataset.mean(dim=("chain", "draw"), skipna=skipna)

        sd = dataset.std(dim=("chain", "draw"), ddof=1, skipna=skipna)

        hdi_post = hdi(dataset, hdi_prob=hdi_prob, multimodal=False, skipna=skipna)
        hdi_lower = hdi_post.sel(hdi="lower", drop=True)
        hdi_higher = hdi_post.sel(hdi="higher", drop=True)
        metrics.extend((mean, sd, hdi_lower, hdi_higher))
        metric_names.extend(("mean", "sd", f"hdi_{100 * alpha / 2:g}%", f"hdi_{100 * (1 - alpha / 2):g}%"))
    elif stat_focus == "median":
        median = dataset.median(dim=("chain", "draw"), skipna=skipna)

        mad = dataset.mad(dim=("chain", "draw"), ddof=1, skipna=skipna)  # not sure if this exists

        eti_post = dataset.quantile((alpha/2, 1-alpha/2), dim=("chain", "draw"), skipna=skipna)
        eti_lower = eti_post.isel(quantile=0, drop=True)
        eti_higher = eti_post.isel(quantile=1, drop=True)
        metrics.extend((median, mad, eti_lower, eti_higher))
        metric_names.extend(("median", "mad", f"eti_{100 * alpha / 2:g}%", f"eti_{100 * (1 - alpha / 2):g}%"))
```

We could then get rid of all the initializing metric and metric_names code and the dependency on positional things like `[4:]` but we'd need to be careful to replace the right things in the circular variable block.

prahasR:
I have applied the changes please review. In the circular variable block I found that the stats are related to `mean` only. So in case when `stat_focus == median` then circular variables should not be calculated. Right?

OriolAbril:
> In the circular variable block I found that the stats are related to mean only. So in case when stat_focus == "median" then circular variables should not be calculated. Right?

It is fine to ignore `stat_focus` for circular variables yes.

> I have applied the changes please review.

Have you tried any of the changes? I did not provide working code in my comment. I tried to explain the restructuring and all the pieces involved in it. Then to ease the understanding of the changes outlined I provided code that _should_ work (but did not test) _for one of the blocks only_, there are still many changes left to do. Moreover, within that block there is the `.mad` line.

prahasR:
I understand! I am still working over my recent commits , I will let you know my queries. Thankyou

arviz/stats/stats.py
```diff
@@ -1353,42 +1363,26 @@ def summary(
...
-                ess_bulk,
-                ess_tail,
-                r_hat,
-            )
-        elif kind == "stats":
-            metrics_ = (mean, sd, hdi_lower, hdi_higher)
-            metrics_names_ = metrics_names_[:4]
-        elif kind == "diagnostics":
-            metrics_ = (mcse_mean, mcse_sd, ess_bulk, ess_tail, r_hat)
-            metrics_names_ = metrics_names_[4:]
```
prahasR, line 1374: As we are taking care of the situation when `kind=="stats"` already and when kind equals `"diagnostics"` or `"all"` then we will have to add these quantities in both the cases, therefore I removed the if/elif conditions here. Can you please review this part.

arviz/stats/stats.py
```diff
@@ -1299,14 +1294,29 @@ def summary(
...
+            hdi_higher = hdi_post.sel(hdi="higher", drop=True)
+            metrics.extend((mean, sd, hdi_lower, hdi_higher))
+            metric_names.extend(("mean", "sd", f"hdi_{100 * alpha / 2:g}%", f"hdi_{100 * (1 - alpha / 2):g}%"))
+        elif stat_focus == "median":
+            median = dataset.median(dim=("chain", "draw"), skipna=skipna)
 
-        hdi_post = hdi(dataset, hdi_prob=hdi_prob, multimodal=False, skipna=skipna)
-        hdi_lower = hdi_post.sel(hdi="lower", drop=True)
-        hdi_higher = hdi_post.sel(hdi="higher", drop=True)
+            mad = dataset.mad(dim=("chain", "draw"), ddof=1, skipna=skipna)  # not sure if this exists
```
prahasR, line 1313: I checked the functions in this block and found that all of them are working except the `.mad`. I am not sure about where should this function be added, can you please help.

arviz/stats/stats.py
```diff
@@ -1341,42 +1363,26 @@ def summary(
     # Combine metrics
     metrics: List[xr.Dataset] = []
     metric_names: List[str] = []
```
OriolAbril, line 1365: these two lines are resetting the metrics and metric_names lists, so all the stats that were added to them above are lost.

arviz/stats/stats.py
```diff
@@ -1341,42 +1363,26 @@ def summary(
...
-    if extend:
-        metrics_: Tuple[xr.Dataset, ...]
+    if extend and kind in ["all", "diagnostics"]:
```
OriolAbril, line 1366: This block (and the block above where these are computed is hardcoded to the mean stat_focus. The diagnostics shown should also depend on the stat_focus as shown in https://github.com/arviz-devs/arviz/pull/1998/files#r832644368.

The idea I tried to explain in the super long comment should be applied to both stats (more or less done in the provided code example) and diagnostics (no sample code provided)

arviz/stats/stats.py
OriolAbril: this should be removed yes, but there are still some changes needed to recover the "original" output

arviz/stats/stats.py
OriolAbril: Let's use the one in xarray-einstats that will work directly on xarray datasets like `hdi`.

prahasR:
Can you please help me in getting the tests right. It is showing module not found error in `xarray_einstats`.

arviz/stats/stats.py
```diff
@@ -1331,52 +1353,57 @@ def summary(
...
+                output_core_dims=tuple([] for _ in range(5)),
+            )
+        elif stat_focus == "median":
+            mcse_median, ess_median, ess_tail, r_hat = xr.apply_ufunc(
+                _make_ufunc(_multichain_statistics, n_output=4, ravel=False),
+                dataset,
+                kwargs=dict(focus="median"),
+                input_core_dims=(("chain", "draw"),),
+                output_core_dims=tuple([] for _ in range(4)),
+            )
```
OriolAbril, line 1370:
```suggestion
            )
        metrics.extend(metrics_)
        metric_names.extend(metrics_names_)
```

arviz/stats/stats.py
```diff
@@ -1331,52 +1353,57 @@ def summary(
...
+                "ess_median",
+                "ess_tail",
+                "r_hat",
+            )
         metrics.extend(metrics_)
         metric_names.extend(metrics_names_)
 
     if circ_var_names:
-        if kind != "diagnostics":
+        if kind != "diagnostics" and stat_focus == "mean":
```
OriolAbril, line 1406: This should work in both cases. We only need to be careful to replace the right positions in `metrics`. I don't think there is a clear definition of a circular median (nor implementation in scipy for example) so it is fine for circular variables to report the mean instead. We only need to document that properly.

arviz/stats/stats.py
```diff
@@ -1287,14 +1295,28 @@ def summary(
...
+            median = dataset.median(dim=("chain", "draw"), skipna=skipna)
+
+            mad = stats.median_abs_deviation(dataset, dims=("chain", "draw"))
+            eti_post = dataset.quantile((alpha/2, 1-alpha/2), dim=("chain", "draw"), skipna=skipna)
+            eti_lower = eti_post.isel(quantile=0, drop=True)
+            eti_higher = eti_post.isel(quantile=1, drop=True)
+            metrics.extend((median, mad, eti_lower, eti_higher))
+            metric_names.extend(("median", "mad", f"eti_{100 * alpha / 2:g}%", f"eti_{100 * (1 - alpha / 2):g}%"))
 
     if circ_var_names:
```
OriolAbril, line 1321: Now circmean and circstd can use the ones from xarray_einstats directly instead of coding the wrappers here.

OriolAbril (commented):
> Can you please help me in getting the tests right. It is showing module not found error in xarray_einstats.

We are introducing xarray-einstats as a new dependency of ArviZ, so it needs to be added to the requirements, otherwise it is not installed and it can't be found when trying to run the tests.

You should add `xarray-einstats>=0.2` to https://github.com/arviz-devs/arviz/blob/main/requirements.txt

arviz/stats/stats.py
```diff
@@ -1330,60 +1358,67 @@ def summary(
...
-            ):
+            if stat_focus == "mean":
+                for metric, circ_stat in zip(
+                    # Replace only the first 5 statistics for their circular equivalent
+                    metrics[:5],
+                    (circ_mean, circ_sd, circ_hdi_lower, circ_hdi_higher, circ_mcse),
+                ):
+                    for circ_var in circ_var_names:
+                        metric[circ_var] = circ_stat[circ_var]
+            else:
```
OriolAbril, line 1404: I'll try to come back to interaction between stat_focus and circular variables at some point next week.

OriolAbril: After some deliberation it is best if we (at least for now) raise an error if `stat_focus!="mean"` and `circ_var_names in not None` (which should happen at the beginning, like the check to ensure stat_focus is mean or median.

arviz/stats/diagnostics.py
```diff
@@ -923,24 +923,31 @@ def _mc_error(ary, batches=5, circular=False):
...
     ary = np.atleast_2d(ary)
     if _not_valid(ary, shape_kwargs=dict(min_draws=4, min_chains=1)):
-        return np.nan, np.nan, np.nan, np.nan, np.nan
-    # ess mean
+        if focus == "mean":
+            return np.nan, np.nan, np.nan, np.nan, np.nan
+        return np.nan, np.nan, np.nan, np.nan
+    # ess mean/ ess median
     ess_mean_value = _ess_mean(ary)
+    ess_median_value = _ess_median(ary)
```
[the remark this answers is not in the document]
OriolAbril: still pending

OriolAbril (commented):
I think we can already start looking into tests, also make sure to run `pip install -r requirements-dev.txt` to make sure you have the right version of black, then check `lint` script runs (can be run as shown in https://python.arviz.org/en/latest/contributing/pr_checklist.html)

prahasR:
I have applied changes.  

> also make sure to run `pip install -r requirements-dev.txt` to make sure you have the right version of black

Also I have run this command and the black tests are passing in my local machine, I checked pytest-black version and it is showing the latest version.

arviz/stats/stats.py
```diff
@@ -1287,38 +1298,53 @@ def summary(
...
-            _make_ufunc(st.circmean),
-            dataset,
-            kwargs=dict(high=np.pi, low=-np.pi, nan_policy=nan_policy),
-            input_core_dims=(("chain", "draw"),),
+        circ_mean = stats.circmean(
+            dataset, dims=["chain", "draw"], high=np.pi, low=-np.pi, nan_policy=nan_policy
         )
         _numba_flag = Numba.numba_flag
-        func = None
+        circ_sd = None
```
OriolAbril, line 1336: no need to define `circ_sd` here, it is defined from scratch in both branches of the if

OriolAbril (commented):
Looks great, have only left some nits to review. Also, can you add an example of using `stat_focus` on the docstring?

> Also I have run this command and the black tests are passing in my local machine, I checked pytest-black version and it is showing the latest version.

Sorry to continue on this, but we need CI to pass in order to be able to merge and [the black check](https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=5691&view=logs&j=7f538728-4589-5563-0e87-d0823eb40420&t=af0f9f62-4f87-5492-4555-4ab2d5002559&l=19) is still failing on Azure even if it passes locally. 

Which black version do you have? You can use `black --version` to check. Azure has [22.3.0](https://dev.azure.com/ArviZ/ArviZ/_build/results?buildId=5691&view=logs&j=7f538728-4589-5563-0e87-d0823eb40420&t=a4781dd5-117c-57bc-6ede-0df74fc9da75&l=26). If you also have the same, let me know once you have done these last changes and I'll take a look and make some commits to your PR to see if I can fix it.

arviz/stats/stats.py
```diff
@@ -1209,6 +1213,14 @@ def summary(
            ...:     extend=False
            ...: )
 
+    Use ``stat_focus`` to change the focus of summary statistics obatined to median:
+
+    .. ipython::
+
+        In [1]: import arviz as az
+           ...: data = az.load_arviz_data("centered_eight")
+           ...: az.summary(data, stat_focus="median")
```
OriolAbril, line 1222:
```suggestion
        In [1]: az.summary(data, stat_focus="median")
```

the lines of code executed previously _in the same docstring_ are remembered, it is not necessary to import arviz again or to create `data`.

OriolAbril (commented):
CI is passing :tada: 

I did a 2nd pass at the docstring. I left one comment and noticed that the return section of the docstring also has to be updated. You can replace 

```
        Return value will contain summary statistics for each variable. Default statistics are:
        `mean`, `sd`, `hdi_3%`, `hdi_97%`, `mcse_mean`, `mcse_sd`, `ess_bulk`, `ess_tail`, and
        `r_hat`.
        `r_hat` is only computed for traces with 2 or more chains.
```

with 

```
        Return value will contain summary statistics for each variable. Default statistics depend on
        the value of ``stat_focus``:

        - ``stat_focus="mean"``: `mean`, `sd`, `hdi_3%`, `hdi_97%`, `mcse_mean`, `mcse_sd`, `ess_bulk`, `ess_tail`, and `r_hat`
        - ``stat_focus="median"``: `median`, `mad`, `eti_3%`, `eti_97%`, `mcse_median`, `ess_median`, `ess_tail`, and `r_hat`

        `r_hat` is only computed for traces with 2 or more chains.
```

Do check that lines are under 100 characters, but it is important to keep an empty line both before and after the bullet point list for rst to render correctly and without warnings.

Finally, I also updated the PR description checklist. The comments above will take care of the documentation point and only adding to the changelog (which you can already do) will be missing. This is an important change so it needs to be on the changelog :smile:

prahasR:
> The comments above will take care of the documentation point and only adding to the changelog (which you can already do) will be missing. This is an important change so it needs to be on the changelog 😄

So I have to list this change in **v0.x.x Unreleased**, under **New Features** section? Right?
And do I have to provide description also ?

OriolAbril:
> So I have to list this change in v0.x.x Unreleased, under New Features section? Right?
And do I have to provide description also ?

Yes, it should have a one sentence description like the others and the link to this PR. It doesn't need to be super clear or contain all the changes/context either, something like "add `stat_focus` argument to `az.summary`" would already be fine.

OriolAbril:
Thanks @prahasR!

PR #2007, opened 2022-03-31
Remove logos

aloctavodia:
Remove logos in favour of hosting them in governance repository arviz-devs/arviz_governance#6


sethaxen:
Since the ReadMe links to https://arviz-devs.github.io/arviz/_static/logo.png, how will this appear when GitHub is in dark mode?

https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#specifying-the-theme-an-image-is-shown-to seems to be the documented solution to have theme-specific images, but for me it only worked when the image itself was hosted on GitHub.

OriolAbril:
We should update the docs and readmes so nothing points to github pages anymore. We now have to host on readthedocs (to have versioned docs) and serve in arviz.org subdomains

aloctavodia:
The *_fav.* versions, i.e. those without the name and with the "A" in the middle work well for both dark and light modes

OriolAbril:
Sorry for realizing it this late, shouldn't all the logos be stored in arviz-governance repo (that could be renamed to arviz-project or something) and then only copy the used files (or use urls to files in the repo with the logos I am not sure what differenced the two options have) to the specific repositories?

canyon289:
Instead of removing id vote putting them in an archive directory of sorts just in case they need to be referenced for some reason

canyon289:
Nevermind I see we're removing. from. this repo. That makes sense, in that case we should add to other repo

README.md
```diff
@@ -1,4 +1,4 @@
-<img src="https://arviz-devs.github.io/arviz/_static/logo.png" height=100></img>
+<img src="https://github.com/arviz-devs/arviz_governance/blob/main/arviz_logos/ArviZ_fav.png" height=100></img>
```
sethaxen, line 1: I'm not certain whether it's preferably to link to the `raw` subdomain directly or not. Either way, because they're on GitHub, we could link to the full logo and activate the right one depending on the theme setting.
```suggestion
<img src="https://raw.githubusercontent.com/arviz-devs/arviz_governance/main/arviz_logos/ArviZ.png#gh-light-mode-only" width=200></img>
<img src="https://raw.githubusercontent.com/arviz-devs/arviz_governance/main/arviz_logos/ArviZ_white.png#gh-dark-mode-only" width=200></img>
```

If we use the full logo with the name, then maybe we should not have the ArviZ heading below.

OriolAbril (approved):

Issue #1968, opened 2022-01-28
plot_hdi fails when x data is time series


drbenvincent:
When calling `az.plot_hdi` with x being a datetime type, an error is given

This (from the docs) does work...
```python
import numpy as np
import arviz as az
import pandas as pd

x_data = np.random.normal(0, 1, 100)
y_data = np.random.normal(2 + x_data * 0.5, 0.5, size=(2, 50, 100))
az.plot_hdi(x_data, y_data)
```

This does not work...
```python
x_data = np.random.normal(0, 1, 100)
y_data = np.random.normal(2 + x_data * 0.5, 0.5, size=(2, 50, 100))
my_dates = pd.date_range(start='1/1/2018', periods=100)
az.plot_hdi(my_dates, y_data)
```

It errors with 
```
UFuncTypeError                            Traceback (most recent call last)
/var/folders/pp/w61sshp15wx7s7k0ghpjv4700000gn/T/ipykernel_17597/4186978889.py in <module>
      6 y_data = np.random.normal(2 + x_data * 0.5, 0.5, size=(2, 50, 100))
      7 my_dates = pd.date_range(start='1/1/2018', periods=100)
----> 8 az.plot_hdi(my_dates, y_data)

~/opt/anaconda3/envs/pymc-dev-py39/lib/python3.9/site-packages/arviz/plots/hdiplot.py in plot_hdi(x, y, hdi_prob, hdi_data, color, circular, smooth, smooth_kwargs, figsize, fill_kwargs, plot_kwargs, hdi_kwargs, ax, backend, backend_kwargs, show)
    170         smooth_kwargs.setdefault("window_length", 55)
    171         smooth_kwargs.setdefault("polyorder", 2)
--> 172         x_data = np.linspace(x.min(), x.max(), 200)
    173         x_data[0] = (x_data[0] + x_data[1]) / 2
    174         hdi_interp = griddata(x, hdi_data, x_data)

<__array_function__ internals> in linspace(*args, **kwargs)

~/opt/anaconda3/envs/pymc-dev-py39/lib/python3.9/site-packages/numpy/core/function_base.py in linspace(start, stop, num, endpoint, retstep, dtype, axis)
    125     # Convert float/complex array scalars to float, gh-3504
    126     # and make sure one can use variables that have an __array_interface__, gh-6634
--> 127     start = asanyarray(start) * 1.0
    128     stop  = asanyarray(stop)  * 1.0
    129 

[...]

PR #2016, opened 2022-04-17
plot_hdi: add exception if `x` is type `np.datetime64` and `smooth=True`

drbenvincent:
In #1968 I pointed out that `az.plot_hdi` failed if `x` was of type `datetime` when `smooth=True`. This PR adds an exception to hopefully make this easier for a user to work out what is going wrong. 

FYI this is my first PR to `arviz`, so apologies if I've missed some steps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New features
 
 ### Maintenance and fixes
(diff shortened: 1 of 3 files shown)


OriolAbril:
Looks great, thanks! Can you add a test similar to https://github.com/arviz-devs/arviz/blob/main/arviz/tests/base_tests/test_plots_matplotlib.py#L1188 to check this error is thrown when datetime data is passed (and we therefore don't remove it unkowingly in the future)

PR #2019, opened 2022-04-23
Fix BASE_FMT text to clearly specify sample size

elusenji:
## Description
Based on issue #2019 
This PR fixes the BASE_FMT text so that it explicitly states the number of samples and number of points. For instance, instead of `Computed from 4000 by 262 log-likelihood matrix` the message should read `Computed from 4000 samples and 262 points log-likelihood matrix`

## Checklist

### DIFF
--- a/arviz/stats/stats_utils.py
+++ b/arviz/stats/stats_utils.py
@@ -439,7 +439,8 @@ def get_log_likelihood(idata, var_name=None):
         return log_likelihood
 
 
-BASE_FMT = """Computed from {{n_samples}} samples and {{n_points}} points log-likelihood matrix.
+BASE_FMT = """Computed from {{n_samples}} posterior samples and \
+{{n_points}} observations log-likelihood matrix.
 
 {{0:{0}}} Estimate       SE
 {{scale}}_{{kind}} {{1:8.2f}}  {{2:7.2f}}
--- a/arviz/stats/stats_utils.py
+++ b/arviz/stats/stats_utils.py
@@ -439,7 +439,7 @@ def get_log_likelihood(idata, var_name=None):
         return log_likelihood
 
 
-BASE_FMT = """Computed from {{n_samples}} by {{n_points}} log-likelihood matrix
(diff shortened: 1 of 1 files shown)


arviz/stats/stats_utils.py
```diff
@@ -439,7 +439,7 @@ def get_log_likelihood(idata, var_name=None):
         return log_likelihood
 
 
-BASE_FMT = """Computed from {{n_samples}} by {{n_points}} log-likelihood matrix
+BASE_FMT = """Computed from {{n_samples}} samples and {{n_points}} points log-likelihood matrix.
```
OriolAbril, line 442:
```suggestion
BASE_FMT = """Computed from {{n_samples}} samples and {{n_points}} observations log-likelihood matrix.
```

OriolAbril (approved):
Thanks! 

Do you want to add it to the changelog?

elusenji:
I'm not sure how to go about that. Also, there seems to be some tests that are failing- not sure why.

ahartikainen:
Pylint says

```
************* Module arviz.stats.stats_utils
arviz/stats/stats_utils.py:442:0: C0301: Line too long (102/100) (line-too-long)

```

arviz/stats/stats_utils.py
OriolAbril, line 442:
```suggestion
BASE_FMT = """Computed from {{n_samples}} posterior samples and \
{{n_points}} observations log-likelihood matrix.
```

I find points ambiguous and not much of an improvement over saying nothing. The `\` will allow breaking the line in the source code (therefore keeping lines below 100 characters) while keeping the printed message all in the same line.

OriolAbril (commented):
Adding to the changelog means adding a bullet point in https://github.com/arviz-devs/arviz/blob/main/CHANGELOG.md#maintenance-and-fixes following the pattern of the other changes

arviz/stats/stats_utils.py
ahartikainen: One other thing that also works is using `(`


    string1 = "abcd"
    string2 = (
        "a"
        "b"
        "c"
        "d"
    )
    string1 == string2

PR #2020, opened 2022-04-24
Update BASE_FMT message

elusenji:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 ### Documentation
 
+## v0.12.1 (2022 Apr 24)
+
+### Maintenance and fixes
+* Update `BASE_FMT` message to clearly specify sample size. ([2019](https://github.com/arviz-devs/arviz/pull/2019))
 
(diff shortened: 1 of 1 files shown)


OriolAbril:
Hi @elusenji, we add the changes to the changelog in their same PR. Only if for some reason (even though we have a check point in the PR template) we forget to add some big changes we make a follow up PR to update the changelog afterwards. Moreover, having follow-up PRs to update the changelog for all changes would nearly duplicate CI time and also increase the time we need to dedicate to reviewing.

I think it would be best to close this PR and, either leave this change out of the changelog as it is stylistic and there have already been many similar changes we did not list, or if you work on another issue, add the two changes to the changelog in the other PR. As a note, if you decide for the 2nd option, items should be added to the changelog as bullet points only on the unreleased section, the headers are only edited when making a release.

Also, as always, if you have trouble finding an issue to work on don't hesitate to ask here or on gitter, if you tell us your interests we'll do our best to find issues that match them.

elusenji:
This is noted. I will close the PR and add those changes if I get to work on another issue.

Issue #1646, opened 2021-03-30
add test for from_cmdstanpy to distinguish between vectors of length 1 and scalars


jburos:
**EDIT (@OriolAbril): See https://github.com/arviz-devs/arviz/issues/1646#issuecomment-1125464400 for an up to date description of the issue.**

---

**Describe the bug**
When fitting a model using cmdstanpy and reading it into an arviz.InferenceData object, I sometimes end up with a vector parameter with length 1. 

When this happens, the io_cmdstanpy code incorrectly ignores the last dimension, causing an error when I provide coord/dims for that parameter.

**To Reproduce**
I have created a reproducible example [here]():

```python
# To add a new cell, type '# %%'
# To add a new markdown cell, type '# %% [markdown]'
# %%

import cmdstanpy
import arviz as az

# %% [markdown]
# # Review Stan model code

# %%
model_code = '''
// from https://arviz-devs.github.io/arviz/notebooks/InferenceDataCookbook.html#From-CmdStanPy
data {
    int<lower=0> J;
    real y[J];
    real<lower=0> sigma[J];
    int<lower = 0, upper=1> prior_only;
}
parameters {
    real mu;
    real<lower=0> tau;
    real theta_tilde[J];
}
transformed parameters {
    real theta[J];
    for (j in 1:J)
        theta[j] = mu + tau * theta_tilde[j];
}
model {
    mu ~ normal(0, 5);
    tau ~ cauchy(0, 5);
    theta_tilde ~ normal(0, 1);
    if (prior_only == 0) {
        y ~ normal(theta, sigma);
    }
}
generated quantities {
    vector[J] log_lik;
    vector[J] y_hat;
    for (j in 1:J) {
        log_lik[j] = normal_lpdf(y[j] | theta[j], sigma[j]);
        y_hat[j] = normal_rng(theta[j], sigma[j]);
    }
}
'''

# %% [markdown]
# # Fitting the model using cmdstanpy

# %%
stan_file = "eight_schools.stan"
with open(stan_file, 'x') as f:
    f.write(model_code)
model = cmdstanpy.CmdStanModel(stan_file = stan_file)


# %%
stan_data8 = dict(
            J=8,
            y=[28, 8, -3, 7, -1, 1, 18, 12],
            sigma=[15, 10, 16, 11, 9, 11, 10, 18],
            prior_only=0,
        )
stan_data1 = dict(
            J=1,
            y=[28],
            sigma=[15],
            prior_only=0,
        )
[...]

PR #2023, opened 2022-04-30
Fix io_cmdstanpy error when importing a fit with vector parameter of length 1

jburos:
## Description

This addresses #1646, which causes an error when importing from cmdstanpy, particularly regarding the distinction between a scalar and vector of length 1. My guess is that this is legacy code from previous versions where the dimension or type had to be inferred rather than known. 

## Checklist


ahartikainen:
What cmdstanpy version causes the error?

ahartikainen:
@OriolAbril why did we have this special case originally?

OriolAbril:
Not sure, looking at the issue discussion I previously guessed it was to properly distinguish between scalar and 1d vector, but I am not sure those two cases can be distinguished in old cmdstanpy cases. I think we should merge this and add a new test on cmdstanpy data that checks scalar and 1d vector data (here or in a follow-up PR)

OriolAbril:
Thanks @jburos! 

As a PSA and for future me, I think new cmdstanpy should be prioritized over old ones (that we have tested work correctly with old ArviZ versions also). Now going to the issue which I won't close yet to signal a test is needed for this case

PR #2025, opened 2022-05-01
Adding plotting guides

symeneses:
## Description

Adds plotting guide per backend. It solves #1488 

## Checklist

- [ x] Follows [official](https://github.com/arviz-devs/arviz/blob/main/CONTRIBUTING.md#pull-request-checklist) PR format

### DIFF
--- a/doc/source/user_guide/plotting_with_bokeh.md
+++ b/doc/source/user_guide/plotting_with_bokeh.md
@@ -34,8 +34,6 @@ az.output_notebook()
 To display the plots as output in a Jupyter notebook, it's required to call the function  {func}`arviz.output_notebook` beforehand.
 :::
 
-:::{warning} Page in construction
-:::
 
 ## Customizing plots
 
--- a/doc/source/user_guide/plotting_with_bokeh.md
+++ b/doc/source/user_guide/plotting_with_bokeh.md
@@ -37,7 +37,7 @@ To display the plots as output in a Jupyter notebook, it's required to call the 
 
 ## Customizing plots
 
-
+(bokeh_backend_kwargs)=
 ### Using `backend_kwargs`
 
 The `backend_kwargs` argument can be very useful for some specific configuration. That is parameters available in {class}`bokeh:bokeh.plotting.figure.Figure`. As the options available depends on the backend, this parameter is not as flexible as creating custom axes.
@@ -57,11 +57,13 @@ az.plot_posterior(
                     "background_fill_color": "#d3d0e3"});
 ```
 
+(bokeh_show)=
 ### The parameter `show`
 
 The parameter `show` is used to control whether the plot is displayed or not. This behavior can be useful when a plot will be displayed in a grid as the example in the following section.
 
 
+(bokeh_ax)=
 ### Defining custom axes
 
 The `ax` argument of any `plot` function allows to use created axes manually. In the example, this parameter allows to arrange 2 different plots in a grid, set limit to the x axis and share the axes' ranges between the two plots.
--- a/doc/source/user_guide/plotting_with_bokeh.md
+++ b/doc/source/user_guide/plotting_with_bokeh.md
@@ -95,10 +95,10 @@ Arviz plot returns a {class}`~bokeh.plotting.Figure` object, therefore different
 data = az.load_arviz_data('regression1d')
 X = data.observed_data.y_dim_0.to_numpy()
 Y = data.observed_data.y.to_numpy()
-predictions = data.posterior_predictive.y.to_numpy()
+y_pp = data.posterior_predictive.y.values
 # plot
 f1 = figure(plot_width=600, plot_height=600, toolbar_location="below")
-az.plot_hdi(X, predictions, color="#b5a7b6", show=False, ax=f1)
+az.plot_hdi(X, y_pp, color="#b5a7b6", show=False, ax=f1)
 f1.scatter(X, Y, marker="circle", fill_color="#0d7591")
 
 show(f1)
@@ -113,8 +113,8 @@ observed_data = data.observed_data.y.to_numpy()
 f1 = figure(plot_width=400, plot_height=400, toolbar_location="below")
 f2 = figure(plot_width=400, plot_height=400, toolbar_location="below")
 # plot
-az.plot_hdi(X, predictions, color="#b5a7b6", show=False, ax=f1)
-f1.line(X, predictions.mean(axis=(0, 1)), color="black")
+az.plot_hdi(X, y_pp, color="#b5a7b6", show=False, ax=f1)
+f1.line(X, y_pp.mean(axis=(0, 1)), color="black")
 f2.scatter(X, Y, marker="circle", fill_color="#0d7591")
 
 show(row(f1, f2))
--- a/doc/source/user_guide/plotting_with_bokeh.md
+++ b/doc/source/user_guide/plotting_with_bokeh.md
@@ -87,7 +87,7 @@ az.show_layout([[f1], [f2]])
 ```
 
 :::{note}
-{func}`arviz.show_layout` creates a bokeh layout and calls shows if `show=True`. The default value of `show` is set by the rcParam `plot.bokeh.show` (which defaults to `True`).
+{func}`arviz.show_layout` creates a bokeh layout and calls shows if `show=True`, which is the default value.
 :::
 
 ### Extending ArviZ-Bokeh plots
--- a/doc/source/user_guide/plotting_with_bokeh.md
+++ b/doc/source/user_guide/plotting_with_bokeh.md
@@ -84,7 +84,7 @@ az.plot_ppc(data, group="posterior", num_pp_samples=100, show=False, ax=f2)
 show(gridplot([[f1], [f2]]))
 ```
 
-## Combining ArviZ and Bokeh plots
+### Extending ArviZ-Bokeh plots
 
 Arviz plot returns a {class}`~bokeh.plotting.Figure` object, therefore different Bokeh plots can be combined into a plot created using Arviz:
 
--- a/doc/source/user_guide/plotting_with_bokeh.md
+++ b/doc/source/user_guide/plotting_with_bokeh.md
@@ -127,7 +127,7 @@ show(row(f1, f2))
 :::{note}
 In Juptyer notebooks, Arviz offers {func}`arviz.output_notebook` which is a wrapper over {func}`bokeh.io.output.output_notebook`.
 
-To use Bokeh with JupyterLab, JupyterLab and widgets, please refer to this {doc}`user guide <bokeh:docs/user_guide/jupyter>`.
+To use Bokeh with JupyterLab, JupyterLab and widgets, please refer to this {ref}`user guide <bokeh:userguide_jupyter>`.
 :::
 
 
@@ -135,5 +135,5 @@ To use Bokeh with JupyterLab, JupyterLab and widgets, please refer to this {doc}
 
 :::{note}
 In the examples above, bokeh is being used in `Standalone` mode. 
-Bokeh can be also used to create interactive web applications. For more details see {doc}`Running a Bokeh server <bokeh:docs/user_guide/server>`.
+Bokeh can be also used to create interactive web applications. For more details see {ref}`Running a Bokeh server <bokeh:userguide_server>`.
 :::
--- a/doc/source/user_guide/plotting_with_bokeh.md
+++ b/doc/source/user_guide/plotting_with_bokeh.md
@@ -1,0 +1,135 @@
+---
+jupytext:
+  text_representation:
+    extension: .md
+    format_name: myst
+kernelspec:
+  display_name: Python 3 (ipykernel)
+  language: python
+  name: python3
+---
+
+(plotting_with_bokeh)=
+
+# Plotting with Bokeh
+
+Arviz has the option to use {doc}`Bokeh <bokeh:index>` as backend which offers interactive plots. Although most of the functions in the {ref}`plot <plot_api>` module work seamlessly with any backend, some advanced plots may require the use of backend specific features. In this guide, advanced plotting with Bokeh will be covered.
+
+This page can be downloaded as a {jupyter-download:script}`Python script <plotting_with_bokeh>`
+or as a {jupyter-download:nb}`Jupyter notebook <plotting_with_bokeh>`.
+
+```{code-cell} ipython3
+import arviz as az
+import matplotlib.pyplot as plt
+import numpy as np
+
+az.style.use("arviz-darkgrid")
+
+# Confgure Bokeh as backend
+az.rcParams["plot.backend"] = "bokeh"
+az.output_notebook()
+```
+
+:::{note}
+To display the plots as output in a Jupyter notebook, it's required to call the function  {func}`arviz.output_notebook` beforehand.
+:::
+
+:::{warning} Page in construction
+:::
+
+## Customizing plots
+
+
+### Using `backend_kwargs`
+
+The `backend_kwargs` argument can be very useful for some specific configuration. That is parameters available in {class}`bokeh:bokeh.plotting.figure.Figure`.
+
+As an example, the following code changes the color background and the plot width.
+
+```{code-cell} ipython3
+# load data
+data = az.load_arviz_data('radon')
+```
+
+```{code-cell} ipython3
+az.plot_posterior(
+    data,
+    var_names=["g"],
+    backend_kwargs={"width": 350, 
+                    "background_fill_color": "#d3d0e3"});
+```
+
+### The parameter `show`
+
+The parameter `show` is used to control whether the plot is displayed or not. This behavior can be useful when a plot will be displayed in a grid as the example in the following section.
+
+
+### Defining custom axes
+
+The `ax` argument of any `plot` function allows to use created axes manually. In the example, this parameter allows to arrange 2 different plots in a grid, set limit to the x axis and share the axes' ranges between the two plots.
+
+
+```{code-cell} ipython3
+from bokeh.io import show
+from bokeh.layouts import gridplot, row
+from bokeh.plotting import figure
+
+# load data
+observed_data = data.observed_data.y.to_numpy()
+# create axes
+f1 = figure(x_range=(observed_data.min() - 1, observed_data.max() + 1))
+f2 = figure(x_range=f1.x_range, y_range=f1.y_range)
+# plot
+az.plot_ppc(data, group="prior", num_pp_samples=100, show=False, ax=f1)
+az.plot_ppc(data, group="posterior", num_pp_samples=100, show=False, ax=f2)
+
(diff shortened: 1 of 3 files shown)


doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+:::{note}
+To display the plots as output in a Jupyter notebook, it's required to call the function  {func}`arviz.output_notebook` beforehand.
+:::
+
+:::{warning} Page in construction
+:::
+
+## Customizing plots
```
OriolAbril, line 40: I think the level 2 headers here and in "combining ArviZ and bokeh plots" might be a bit confusing. Even if generated through arviz, all plots generated here are also bokeh plots, and customizing could also include adding further elements to the plot, not only using the arguments available in arviz functions.

Seeing the differences in the separation I think the titles could be: "Bokeh specific arguments" and "Extending ArviZ-Bokeh plots" or something of the sort. Then we should also go to the plots arguments guide and replace the `ax, backend, show` sections with two links to the bokeh and matplotlib specific sections.

That being said, the level 3 headings look fine, I would not change them to be the argument name only like in the plot argument guide. They should however have their own anchors though, like in the plot argument guide. The end goal is to then update the docstrings to have links to the right sections in the corresponding parameter description and keep the examples in the docstring to plot specific arguments and examples.

doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,135 @@
...
+The `ax` argument of any `plot` function allows to use created axes manually. In the example, this parameter allows to arrange 2 different plots in a grid, set limit to the x axis and share the axes' ranges between the two plots.
+
+
+```{code-cell} ipython3
+from bokeh.io import show
+from bokeh.layouts import gridplot, row
+from bokeh.plotting import figure
+
+# load data
+observed_data = data.observed_data.y.to_numpy()
```
OriolAbril, line 78:
```suggestion
observed_data = data.observed_data.y.values
```

not completely sure what `to_numpy()` method does, but afaik, the recommended ways to access the underlying arrays in xarray.DataArray is `.values` if we want the data as a numpy array or `.data` if we want the actual array structure used by xarray to store the data in that case (defaults to numpy but can already be dask or pydata/sparse and they are working to extend it to arbitrary array libraries)

doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,135 @@
...
+## Combining ArviZ and Bokeh plots
+
+Arviz plot returns a {class}`~bokeh.plotting.Figure` object, therefore different Bokeh plots can be combined into a plot created using Arviz:
+
+```{code-cell} ipython3
+# load data
+data = az.load_arviz_data('regression1d')
+X = data.observed_data.y_dim_0.to_numpy()
+Y = data.observed_data.y.to_numpy()
+predictions = data.posterior_predictive.y.to_numpy()
```
OriolAbril, line 98:
```suggestion
y_pp = data.posterior_predictive.y.to_numpy()
```

or `post_pred`, `pps`, `y_post_pred`, inverted order... The comment is basically about avoiding the use of the word "predictions" in that context as it can be confusing. In ArviZ we have a group called `posterior_predictive` and a group called `predictions`. Both are used to store posterior predictive samples that can also/always be called predictions if you want, but we make this distinction to reduce the complexity of arviz code. 

The posterior_predictive group should store pp samples aiming to predict the same data that was already observed (kind of -but not really- as if it hadn't been observed) which might seem a bit pointless but is very useful for model criticism. These samples are therefore used to perform posterior predictive checks: comparing the actual observed data to these samples because if the model is correct, it should not be possible to distinguish these synthetic/simulated observations from the real ones.

The predictions group stored pp samples that aim to predict on new/unobserved data

doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,135 @@
...
+observed_data = data.observed_data.y.to_numpy()
+# create axes
+f1 = figure(plot_width=400, plot_height=400, toolbar_location="below")
+f2 = figure(plot_width=400, plot_height=400, toolbar_location="below")
+# plot
+az.plot_hdi(X, predictions, color="#b5a7b6", show=False, ax=f1)
+f1.line(X, predictions.mean(axis=(0, 1)), color="black")
+f2.scatter(X, Y, marker="circle", fill_color="#0d7591")
+
+show(row(f1, f2))
```
OriolAbril, line 120: Would `show_layout` help here? https://python.arviz.org/en/latest/api/utils.html#bokeh-utils. Keep in mind I am not an avid bokeh user. I don't know if you didn't know about it, if it is not documented enough (which in this case seems like a safe bet to at least be a factor) and/or if it relies too much on rcparams and lacks arguments to override them once when calling it.

Side note: all these should be documented under plot utils unlike the labels that should actually be in utils. I am also quite sure I was the one to add them all in the wrong place, no need to fix it in this PR, I can open an issue and come back to this or hope someone fixes it before.

doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+## Bokeh with Jupyter
```
OriolAbril, line 123: probably here instead of on the note at the top of the document we might want to keep shorter and to the point. We should probably mention the `az.output_notebook` is nothing more than a wrapper over the bokeh function with the same (or similar?) name.

doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,135 @@
...
+
+:::{note}
+To use Bokeh with JupyterLab, JupyterLab and widgets, please refer to this {doc}`user guide <bokeh:docs/user_guide/jupyter>`.
```
OriolAbril, line 126: note to self, check if this could be a `ref` type cross-reference.

OriolAbril (commented):
Looks great so far, thanks! I'll continue reviewing as soon as possible

doc/source/user_guide/plotting_with_bokeh.md
symeneses: I tried `ref` something like ``{ref}`user guide <bokeh:jupyter>`` but it didn't work

doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,135 @@
...
+az.rcParams["plot.backend"] = "bokeh"
+az.output_notebook()
+```
+
+:::{note}
+To display the plots as output in a Jupyter notebook, it's required to call the function  {func}`arviz.output_notebook` beforehand.
+:::
+
+:::{warning} Page in construction
+:::
```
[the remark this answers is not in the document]
symeneses: Deleted

doc/source/user_guide/plotting_with_bokeh.md
symeneses: I changed the name and level of the header. Once the docs are ready, we can add the links in `Plots’ arguments guide`.

doc/source/user_guide/plotting_with_bokeh.md
symeneses: Variable renamed

symeneses: And now I got the names...a bit confusing 😅

doc/source/user_guide/plotting_with_bokeh.md
symeneses: `to_numpy()` was replaced by `values`.

doc/source/user_guide/plotting_with_bokeh.md
symeneses: Added `show_layout` instead of `show(gridplot([[f1], [f2]]))` in the other example. If this one is modified, the toolbar is visible in only one plot.

doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,135 @@
...
+  name: python3
+---
+
+(plotting_with_bokeh)=
+
+# Plotting with Bokeh
+
+Arviz has the option to use {doc}`Bokeh <bokeh:index>` as backend which offers interactive plots. Although most of the functions in the {ref}`plot <plot_api>` module work seamlessly with any backend, some advanced plots may require the use of backend specific features. In this guide, advanced plotting with Bokeh will be covered.
+
+This page can be downloaded as a {jupyter-download:script}`Python script <plotting_with_bokeh>`
```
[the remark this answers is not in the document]
OriolAbril: didn't find any relevant difference, I'll assume it's an issue with the preview and if it breaks we'll fix it

doc/source/user_guide/plotting_with_bokeh.md
OriolAbril: sounds great, you have now used `show_layout` more than I have :see_no_evil:

doc/source/user_guide/plotting_with_bokeh.md
OriolAbril: the `ref` type target for this page should be `userguide_jupyter`. It is a bit inconvenient, you have to look at the source to see which anchors are defined: https://raw.githubusercontent.com/bokeh/bokeh/branch-3.0/sphinx/source/docs/user_guide/jupyter.rst

doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,139 @@
...
+f2 = figure(x_range=f1.x_range, y_range=f1.y_range)
+# plot
+az.plot_ppc(data, group="prior", num_pp_samples=100, show=False, ax=f1)
+az.plot_ppc(data, group="posterior", num_pp_samples=100, show=False, ax=f2)
+
+az.show_layout([[f1], [f2]])
+```
+
+:::{note}
+{func}`az.show_layout` creates a bokeh layout and calls shows if `shows` is `True` which is the default behavior.
```
OriolAbril, line 88:
```suggestion
{func}`arviz.show_layout` creates a bokeh layout and calls shows if `show`. The default value of `show` is set by the rcParam `plot.bokeh.show` (which defaults to `True`)
```

doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,139 @@
...
+:::
+
+### Extending ArviZ-Bokeh plots
+
+Arviz plot returns a {class}`~bokeh.plotting.Figure` object, therefore different Bokeh plots can be combined into a plot created using Arviz:
```
OriolAbril, line 93:
```suggestion
Arviz plot returns a {class}`~bokeh.plotting.Figure` object, therefore different Bokeh plots can be added to a plot created using Arviz:
```

doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,139 @@
...
+y_pp = data.posterior_predictive.y.values
+# plot
+f1 = figure(plot_width=600, plot_height=600, toolbar_location="below")
+az.plot_hdi(X, y_pp, color="#b5a7b6", show=False, ax=f1)
+f1.scatter(X, Y, marker="circle", fill_color="#0d7591")
+
+show(f1)
+```
+
+Similarly, custom axes allow to display Arviz and Matplotlib plots in the same grid. In this example, the plot in `ax1` has an Arviz plot modified with a Matplotlib plot and the `ax2` has a `scatter` created using Matplotlib.
```
OriolAbril, line 109:
```suggestion
Similarly, custom axes allow to display Arviz and Bokeh plots in the same grid. In this example, the plot in `f1` has an Arviz plot extended with a Bokeh plot and `f2` has a `scatter` created using Bokeh directly.
```

doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,139 @@
...
+
+To use Bokeh with JupyterLab, JupyterLab and widgets, please refer to this {doc}`user guide <bokeh:docs/user_guide/jupyter>`.
+:::
+
+
+## Server mode
+
+:::{note}
+In the examples above, bokeh is being used in `Standalone` mode. 
+Bokeh can be also used to create interactive web applications. For more details see {doc}`Running a Bokeh server <bokeh:docs/user_guide/server>`.
```
OriolAbril, line 138:
```suggestion
Bokeh can be also used to create interactive web applications. For more details see {ref}`Running a Bokeh server <bokeh:userguide_server>`.
```

ref: https://raw.githubusercontent.com/bokeh/bokeh/branch-3.0/sphinx/source/docs/user_guide/server.rst

doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,139 @@
...
+
+:::{note}
+To display the plots as output in a Jupyter notebook, it's required to call the function  {func}`arviz.output_notebook` beforehand.
+:::
+
+
+## Customizing plots
+
+
+### Using `backend_kwargs`
```
OriolAbril, line 41:
```suggestion
(bokeh_backend_kwargs)=
### Using `backend_kwargs`
```

In the plot arguments guide we use `common_arg_name`, so following the same idea I think we should add these anchors here.

doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,139 @@
...
+
+```{code-cell} ipython3
+az.plot_posterior(
+    data,
+    var_names=["g"],
+    backend_kwargs={"width": 350, 
+                    "background_fill_color": "#d3d0e3"});
+```
+
+### The parameter `show`
```
OriolAbril, line 60:
```suggestion
(bokeh_show)=
### The parameter `show`
```

doc/source/user_guide/plotting_with_bokeh.md
```diff
@@ -0,0 +1,139 @@
...
+
+The parameter `show` is used to control whether the plot is displayed or not. This behavior can be useful when a plot will be displayed in a grid as the example in the following section.
+
+
+### Defining custom axes
```
OriolAbril, line 65:
```suggestion
(bokeh_ax)=
### Defining custom axes
```

doc/source/user_guide/plotting_with_matplotlib.md
```diff
@@ -0,0 +1,103 @@
...
+        "facecolor": "#d3d0e3",
+        "gridspec_kw": {
+            "width_ratios": [6,4]}});
+```
+:::{note}
+This parameter is however not as flexible as creating custom axes.
+:::
+:::{caution}
+The parameters `ncol` and `nrows` from {func}`matplotlib.pyplot.subplots` should not be set using the `backend_kwargs` parameter.
+:::
```
OriolAbril, line 56: I find the two consecutive boxes a bit overloaded. Can you try moving the caution box above the plot, between the two text paragraphs and making the note box regular text?

Also, a note on ncol and nrow caution box. I would add a 2nd sentence saying something like ``Use {ref}`common_grid` argument instead.``

doc/source/user_guide/plotting_with_matplotlib.md
```diff
@@ -0,0 +1,103 @@
...
+
+### The parameter `show`
+
+:::{note} The parameter `show` calls the show function of the backend. Matplotlib shows the plot in a Jupyter notebook even if this is set to `False`. For more details see {func}`matplotlib.pyplot.show`.
```
OriolAbril, line 60:
```suggestion
The parameter `show` calls {func}`matplotlib.pyplot.show` after all ArviZ plotting finishes. defaults rcparam... (like with bokeh but defaults to false)

:::{note} There are some Matplotlib backends like the inline (default) Jupyter backend that show the plot even if `plt.show` is not called. In such cases, plots will be shown automatically even if `show=False` as this only skips calling `plt.show` it doesn't do anything to prevent the plots from showing.
```

OriolAbril (commented):
Everything looks great, reviewed already stylistic choices and typos

doc/source/user_guide/plotting_with_bokeh.md
symeneses: Updated

doc/source/user_guide/plotting_with_bokeh.md
symeneses: I changed this part. However, seeing the [code](https://python.arviz.org/en/latest/_modules/arviz/plots/backends.html#show_layout), the default is `True`, only if `show` is `None` takes the value from the rcParam.

```
def show_layout(show=True, force_layout=False):
    """Create a layout and call bokeh show."""
    if show is None:
        print("Default value")
    if show:
        print("It shows")
        
show_layout(show=None)
show_layout()
```

doc/source/user_guide/plotting_with_bokeh.md
symeneses: Changed

doc/source/user_guide/plotting_with_bokeh.md
symeneses: That happens when you copy paste 🙈 
Changed

doc/source/user_guide/plotting_with_bokeh.md
symeneses: I added the 6 anchors. Let me know how we should refer to them in the arguments guide. Something like see `ref_matplotlib` or `ref_bokeh`?

doc/source/user_guide/plotting_with_matplotlib.md
symeneses: Done!

doc/source/user_guide/plotting_with_matplotlib.md
symeneses: I did some changes here. I am not sure if I understood the part about the `rcparam`, I couldn't find any similar  rcparam in matplotlib.

doc/source/user_guide/plotting_with_bokeh.md
OriolAbril: In the plot arguments guide I'd probably leave two links only, one per backend referencing the top of the file anchor. The argument specific ones should be referenced from the respective descriptions in the docstring (no need to do this here). Once we merge this I think the best would be to update one docstring as an example and then open an issue to signal other people they can also work on this.

doc/source/user_guide/plotting_with_bokeh.md
symeneses: All good! I just removed that

doc/source/user_guide/plotting_with_matplotlib.md
symeneses: Added

symeneses:
> Everything looks great, reviewed already stylistic choices and typos

Looks like all the conversations are solved now.

OriolAbril:
Thanks!

PR #2029, opened 2022-05-13
Make netCDF4 optional.

### LINKED ISSUE #2028: Make netcdf4 dependency optional
netcdf is used in only two methods (loading and saving to netcdf) and is a heavier dependency, could we make that optional?

background: I'm playing around with pyscript and the netcdf4 dependency makes arviz not work with micropip.

twiecki:
Closes #2028.

### DIFF
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -24,7 +24,6 @@ from typing import (
     overload,
 )
 
-import netCDF4 as nc
 import numpy as np
 import xarray as xr
 from packaging import version
@@ -356,6 +355,8 @@ class InferenceData(Mapping[str, xr.Dataset]):
         -------
         InferenceData object
         """
+        import netCDF4 as nc
+
         groups = {}
 
         try:
@@ -412,6 +413,8 @@ class InferenceData(Mapping[str, xr.Dataset]):
         str
             Location of netcdf file
         """
+        import netCDF4 as nc
+
         mode = "w"  # overwrite first, then append
         if self._groups_all:  # check's whether a group is present or not.
             if groups is None:
--- a/requirements-optional.txt
+++ b/requirements-optional.txt
@@ -1,5 +1,6 @@
 numba
 bokeh>=1.4.0,<3.0
+netcdf4
 ujson
 dask[distributed]
 zarr>=2.5.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,5 @@ scipy>=0.19
 packaging
 pandas>=0.23
 xarray>=0.16.1
-netcdf4
 typing_extensions>=3.7.4.3
 xarray-einstats>=0.2


OriolAbril:
This will break a lot of code and environments, we need to be careful about how to do this, I don't think removing the dependency straight away is a good idea.

The whole of ArviZ assumes netcdf files can be stored and read. For example, the doc failures are due to that, using `arviz.load_arviz_data` reads netcdf files (either packaged with the library or after downloading from figshare) and most documentation examples are based on this. Changing that will also mean that after doing `pip install arviz` people won't be able to save inferencedata as netcdf as they are now

twiecki:
OK, maybe it should stay a dependency then?

OriolAbril:
I think it should be possible to install ArviZ without installing libraries that are not required throughout the library (netcdf, but also matplotlib for example like I said before), but I am also worried about making `pip install arviz` default to an installation that can't save to file nor plot (in the extreme) unless some of the possible optional dependencies are installed manually as well as backward compatibility issues. 

Let's see what the rest of the team thinks. One thing we discussed was splitting the library in data/stats/plots but releasing in sync for simplicity and keeping `arviz` as the bundle of the three (plus maybe netcdf and/or matplotlib). But it might be better to rip the band aid and have `arviz` make a minimal install and recommend installing `arviz[things]`

ColCarroll:
I don't recall if it is possible to do like, `pip install arviz[minimal]` -- would that help with this? I'm pretty happy to allow bespoke `[special_install_requires]` if it allows usage.

canyon289:
Catching up. Agree with Oriol, We definitely can't do it this way this "hardline" way its going to break a lot of things. I think we have do it the way Colin recommends, which is netcdf by default, but it can be turned off with a flag.

Also note for the pyscript usecase we'll also need to compile both wheels everytime we release as well so its fetchable by the pyscript users. Outside of the scope of this PR but something to note

OriolAbril:
> I don't recall if it is possible to do like, pip install arviz[minimal] -- would that help with this?

That would be amazing but I don't think it is possible. afaik the [name] can only be used to _add_ dependencies to the set of requirements. 

> I'm pretty happy to allow bespoke [special_install_requires] if it allows usage.

Just to make sure we are all on the same page, we already provide `arviz[all]` to install all requirements and optional dependencies at once. See https://python.arviz.org/en/latest/getting_started/Installation.html

> Also note for the pyscript usecase we'll also need to compile both wheels everytime we release as well so its fetchable by the pyscript users. Outside of the scope of this PR but something to note

Whenever we release, the Azure job already builds a wheel and uploads it to pypi, see https://pypi.org/project/arviz/#files. Only one wheel is generated though and you mentioned "both wheels" @canyon289 can you create an issue with what is missing so we remember to update the release Azure job?

canyon289:
> Whenever we release, the Azure job already builds a wheel and uploads it to pypi, see https://pypi.org/project/arviz/#files. Only one wheel is generated though and you mentioned "both wheels" @canyon289 can you create an issue with what is missing so we remember to update the release Azure job?

Sure but we only need to do this if we go through with the proposal above which its uncertain if we will

varchasgopalaswamy:
What about using [h5netcdf](https://github.com/h5netcdf/h5netcdf)? That might make it a lot less heavy since the netCDF4 C binaries won't be necessary.

twiecki:
Closing in favor of https://github.com/arviz-devs/arviz/pull/2122.

Issue #689, opened 2019-06-02
request for Bayes Factor plot


drbenvincent:
I wonder whether it would be useful to have a Bayes Factor type plot, like this?
![IMG_4444](https://user-images.githubusercontent.com/6765047/58761225-c23e3e80-8539-11e9-8d9d-176db13e09b5.JPG)

ColCarroll:
It sounds like there is at least one person who would find it useful, and my impression is that they are standard (if somewhat controversial?)

We can leave this issue open for discussion, or for potential contributors to mention that they are working on it?

As an aside, this is the first time I've seen a sketch as a PR, and I love it.

drbenvincent:
It's not impossible that I might attempt it. But only after I've actually submitted a bunch of papers, late in the summer.

There's a bit of debate between estimation and CI's vs Bayesian hypothesis testing. I've not really dug into it, but good examples of the pro-estimation approach are:
- Kruschke, J. K. (2011). Bayesian Assessment of Null Values Via Parameter Estimation and Model Comparison. Perspectives on Psychological Science, 6(3), 299–312. http://doi.org/10.1177/1745691611406925
- Kruschke, J. K., & Liddell, T. M. (2015). The Bayesian New Statistics: Two historical trends converge, 1–21.

And some examples on the pro hypothesis testing side...

- Morey, R. D., Rouder, J. N., Verhagen, J., & Wagenmakers, E.-J. (2014). Why hypothesis tests are essential for psychological science: a comment on Cumming (2014). Psychological Science, 25(6), 1289–1290. http://doi.org/10.1177/0956797614525969
- Wagenmakers, E.-J., Verhagen, J., Ly, A., Matzke, D., Steringoever, H., Rouder, J. N., & Morey, R. D. (2015). The Need for Bayesian Hypothesis Testing in Psychological Science, 1–29.

I'm not sure, but I think the Bayes Factor camp might have the slight edge, certainly in my field because of the legacy of frequentist hypothesis testing and because of it's use in tools like [JASP](https://jasp-stats.org) which makes great plots like this...
![image4](https://user-images.githubusercontent.com/6765047/58763636-39350080-8555-11e9-91f1-c006a2bade82.png)
Personally I always report CI's or HDI's and optionally BF's when practical for the modelling context.

treszkai:
[...]

PR #2037, opened 2022-05-30
Add Bayes Factor plot via Savage-Dickey density ratio

orduek:
## Description

As requested in issue #689  - this is an attempt to add plot_bf function, that will take prior, idata (trace) and reference value, and output a Bayes Factor + plot (please see issue #689 for details and examples)


OriolAbril:
Hi, sorry for the very slow response. It looks like it is something that is probably useful to ArviZ users, but I had never heard about this plot before the issues was opened and haven't found the time to review the references (I am also quite slow to review PRs about things I do know about). I can leave some general comments for now and questions. Thanks for your patience.

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,48 @@
...
+    import matplotlib.pyplot as plt
+    import numpy as np
+    # test some elemtns
+    # varName should be string
+    if not isinstance(var_name, str):
+        print('varName is not a string')
+    # BFs based on density estimation (using kernel smoothing instead of spline)
+    # stack trace posterior
+    tr = trace.posterior.stack(draws=("chain", "draw"))
+    post = tr[var_name]
```
OriolAbril, line 15: This could be 

```suggestion
    post = extract(idata, var_names=var_name)
    prior = extract(idata, var_names=var_name, group="prior")
```

Ref: https://python.arviz.org/en/latest/api/generated/arviz.extract.html.

We also generally use `data` or `idata` instead of `trace` and get the samples from the multiple groups from there.

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,48 @@
...
+        prior_pdf = stats.gaussian_kde(prior)
+    elif family!='normal':
+        # for now and error of notImplemented
+        print("The method for {family} distribution is not implemented yet")
+    if ax is None:
+        fig, ax = plt.subplots()
+    ax.plot(
+        x, my_pdf(x), "--", lw=2.5, alpha=0.6, label="Posterior"
+    )  # distribution function
+    ax.plot(x, prior_pdf(x), "r-", lw=2.5, alpha=0.6, label="Prior")
```
OriolAbril, line 35: Is it necessary to use the same `x` when plotting the prior and the posterior? Most of the plotted line for the posterior will generally be the tail of the kde approximation. Does this method/figure (sorry for the lack of background again) work for bounded variables?

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,48 @@
...
+    posterior = my_pdf(ref_val) # this gives the pdf at ref_val
```
OriolAbril, line 36: What if `ref_val` is outside the `[min(post), max(post)]` range?

Side note, it looks like https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.gaussian_kde.html has a logpdf method. Having no idea about the expected ranges of values for this variable, could it make sense to compute them in the log space, substract them and exponentiate later for plotting the dot and reporting the values?

OriolAbril (commented):
I tried to focus on conceptual questions, but did want to advertize the `extract` function. It isn't released yet, but can be safely used here and should make things like this easier.

arviz/plots/bf_plot.py
orduek: Hi, Sorry for this very late response (I just started a position, so things are a bit hectic).
I think its better to use the same x, as we basically want to test the density under the same point of both prior and posterior.

arviz/plots/bf_plot.py
orduek: I've added these lines in the new version. But, as sometimes we would like to play with the prior, I've left the option to input a prior other than the one stated in the model

arviz/plots/bf_plot.py
orduek: An interesting question that doesn't seem to be unlikely usually, but I've added an error message if this case happens.

I'm not sure about using logpdf; what is the advantage?

orduek (commented):
I've tried to correct your comments. I hope this one is better.
Thank you and sorry for the late response.

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,55 @@
...
+    ax.plot(x, prior_pdf(x), "r-", lw=2.5, alpha=0.6, label="Prior")
+    if ref_val>np.max(post) | ref_val<np.min(post):
+        print('Reference value is out of bounds of posterior')
+    else:
+        posterior = my_pdf(ref_val) # this gives the pdf at ref_val
+    prior = prior_pdf(ref_val)
+    BF10 = posterior / prior
+    BF01 = prior / posterior
+    print("the Bayes Factor 10 is %.3f" % (BF10))
+    print("the Bayes Factor 01 is %.3f" % (BF01))
```
aloctavodia, line 47: this should use the logging system, or maybe add it to the title

aloctavodia (changes requested):

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,55 @@
+# Plotting and reporting Bayes Factor given idata, var name, prior distribution and reference value
+def plot_bf(trace, var_name, prior, family = 'normal',  ref_val=0, xlim=None, ax=None):
```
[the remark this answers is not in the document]
aloctavodia: We can add a description to the docstring like 

Bayes Factor approximated as the Savage-Dickey density ratio.

The Bayes factor is estimated by comparing a model against a model in which the parameter of interest has been restricted to a point-null.

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,55 @@
...
+    # BFs based on density estimation (using kernel smoothing instead of spline)
+    # stack trace posterior
+    post = extract(idata, var_names=var_name)
+    if prior is None:
+        # grab prior from the data in case it wasn't defined by the user
+        prior = extract(idata, var_names=var_name, group="prior")
+    post = tr[var_name]
+    if post.ndim > 1:
+        print("Posterior distribution has {post.ndim} dimensions")
+    if family=='normal':
```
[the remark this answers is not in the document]
orduek: Theoretically, one might use it for other all sorts of priors, although the normal distribution is the most common case (maybe a student's-t etc.)

arviz/plots/bf_plot.py
orduek: How can I use the logging system (sorry, a bit of a newbie here).

orduek:
OK, I've fixed the issues you've raised. 
Two things I'm not sure about:
1. The logging system 
2. Should we add more distributions (I would probably consult other about it before doing so)

arviz/plots/bf_plot.py
aloctavodia: See for example


You have to import logging

And then do
 https://github.com/arviz-devs/arviz/blob/9f4d7ac657e4170b7da6724041df726eb89e918d/arviz/plots/backends/matplotlib/ppcplot.py#L14

And then something like 

https://github.com/arviz-devs/arviz/blob/9f4d7ac657e4170b7da6724041df726eb89e918d/arviz/plots/backends/matplotlib/ppcplot.py#L88

But use .info instead of .warn

arviz/plots/bf_plot.py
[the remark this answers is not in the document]
aloctavodia: What I don't understand is was is particular about the normal in this implementation and how an implementationof a different family would look like. 

If I am understanding correctly we should only make a distinction between continuous and discrete distributions. For the continuous we use a kde to approximate the density of both prior and posterior. And for discrete we instead count how many samples we have for the ref_val normalized by the total number of samples.

orduek:
ok, I added logging. Let me know if you think it's ok now.
Thank you for all the guidance!

arviz/plots/bf_plot.py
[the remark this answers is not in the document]
orduek: @drbenvincent - what do you think?

aloctavodia:
I don't think we should have a family argument. Instead, we should be able to handle all kinds of distributions without user intervention. 

I see two simple solutions 

1) Ignore boundaries and use the KDE from scipy as you are currently doing
2) Use our KDE, which takes into account the boundaries. Our KDE implementation does not currently allow for evaluation of the pdf. But you get the densities evaluated at discrete points on a grid, so you can use those values if the grid is small enough.

aloctavodia:
yes we can leave it like this for this PR, and in the future think of a better method, not sure we will need to support more than one method.

orduek:
ok, great.
Let me know what else is needed to merge it.

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,67 @@
+# Plotting and reporting Bayes Factor given idata, var name, prior distribution and reference value
+from scipy import stats
+import matplotlib.pyplot as plt
+import numpy as np
+import logging
+# 
+_log = logging.getLogger(__name__)
+
+def plot_bf(idata, var_name, prior, family = 'normal',  ref_val=0, xlim=None, ax=None):
+    """
```
aloctavodia, line 10: check the doctring style from other plots

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,67 @@
...
+    :ref_val: reference value for BF testing
+    :xlim: limit the x axis (might be used for visualization porpuses sometimes)
+
+    # grab trace, a variable name to compute difference and prior.
+    # ref_val is the parameter we want to compare
+    # test some elemtns
+    # varName should be string
+    if not isinstance(var_name, str):
+        print('varName is not a string')
+    # BFs based on density estimation (using kernel smoothing instead of spline)
```
aloctavodia, line 28:
```suggestion
```

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,67 @@
...
+        # grab prior from the data in case it wasn't defined by the user
+        prior = extract(idata, var_names=var_name, group="prior")
+    if post.ndim > 1:
+        print("Posterior distribution has {post.ndim} dimensions")
+    # generate vector
+    if xlim is None:
+        x = np.linspace(np.min(prior), np.max(prior),prior.shape[0])
+    else:
+        x = np.linspace(xlim[0], xlim[1],prior.shape[0])
+    #x = np.linspace(np.min(post), np.max(post),prior.shape[0])
```
aloctavodia, line 41:
```suggestion
```

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    post = extract(idata, var_names=var_name)
+    if prior is None:
+        # grab prior from the data in case it wasn't defined by the user
+        prior = extract(idata, var_names=var_name, group="prior")
+    if post.ndim > 1:
+        print("Posterior distribution has {post.ndim} dimensions")
+    # generate vector
+    if xlim is None:
+        x = np.linspace(np.min(prior), np.max(prior),prior.shape[0])
```
aloctavodia, line 38: probably better to just have a "high" number of points, like 5000, 10000?

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,67 @@
+# Plotting and reporting Bayes Factor given idata, var name, prior distribution and reference value
+from scipy import stats
+import matplotlib.pyplot as plt
+import numpy as np
+import logging
+# 
+_log = logging.getLogger(__name__)
+
+def plot_bf(idata, var_name, prior, family = 'normal',  ref_val=0, xlim=None, ax=None):
```
aloctavodia, line 9:
```suggestion
def plot_bf(idata, var_name, prior,  ref_val=0, xlim=None, ax=None):
```

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,67 @@
...
+    else:
+        posterior = my_pdf(ref_val) # this gives the pdf at ref_val
+    prior = prior_pdf(ref_val)
+    BF10 = posterior / prior
+    BF01 = prior / posterior
+    _log.warning("the Bayes Factor 10 is %.3f" % (BF10)"
+                ("the Bayes Factor 01 is %.3f" % (BF01)
+    )
+    ax.plot(ref_val, posterior, "ko", lw=1.5, alpha=1)
+    ax.plot(ref_val, prior, "ko", lw=1.5, alpha=1)
```
aloctavodia, line 61:
```suggestion
    ax.plot(ref_val, posterior, "ko", lw=1.5)
    ax.plot(ref_val, prior, "ko", lw=1.5)
```

arviz/plots/bf_plot.py
aloctavodia: I think what Oriol is saying here is that we can build the kde using one x for the prior and one for the posterior to ensure we are placing points where there is some mass. Then we can evaluate those pdf for the same `ref_val`

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,67 @@
+# Plotting and reporting Bayes Factor given idata, var name, prior distribution and reference value
+from scipy import stats
+import matplotlib.pyplot as plt
+import numpy as np
+import logging
```
aloctavodia, line 5:
```suggestion
import logging

import matplotlib.pyplot as plt
import numpy as np
from scipy import stats
```

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,67 @@
...
+    my_pdf = stats.gaussian_kde(post)
+    prior_pdf = stats.gaussian_kde(prior)
+    if ax is None:
+        fig, ax = plt.subplots()
+    ax.plot(
+        x, my_pdf(x), "--", lw=2.5, alpha=0.6, label="Posterior"
+    )  # distribution function
+    ax.plot(x, prior_pdf(x), "r-", lw=2.5, alpha=0.6, label="Prior")
+    if ref_val>np.max(post) | ref_val<np.min(post):
+        print('Reference value is out of bounds of posterior')
```
aloctavodia, line 51: use logging

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,67 @@
...
+    else:
+        posterior = my_pdf(ref_val) # this gives the pdf at ref_val
+    prior = prior_pdf(ref_val)
+    BF10 = posterior / prior
+    BF01 = prior / posterior
+    _log.warning("the Bayes Factor 10 is %.3f" % (BF10)"
```
aloctavodia, line 57:
```suggestion
    _log.info("the Bayes Factor 10 is %.3f" % (BF10)"
```

arviz/plots/bf_plot.py
orduek: It will probably be very similar, but why not use the prior dimensions? to save resources?

aloctavodia:
We already have some logic to handle variables names see for example https://github.com/arviz-devs/arviz/blob/main/arviz/plots/autocorrplot.py#L120.

Something like 5000 points for the kde should be enough.

orduek:
I've changed the size to 5000 and tried to incorporate the _var_names_ method. 
Let me know if it looks ok now

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,83 @@
+# Plotting and reporting Bayes Factor given idata, var name, prior distribution and reference value
+import logging
+
+
+import matplotlib.pyplot as plt
+import numpy as np
+from scipy import stats
+#
```
aloctavodia, line 8:
```suggestion

```

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,83 @@
...
+        a :class:`arviz.InferenceData` object
+    :var_name: str 
+        Name of variable we want to test.
+    :prior: numpy.array, optional
+        In case we want to use diffent prior (for sensitivity analysis of BF), 
+        we can define one and sent it to the function.
+    :ref_val: int, default is 0
+        Reference value for BF testing
+    :xlim: numpy.array, optional
+        limit the x axis (might be used for visualization porpuses sometimes)
```
aloctavodia, line 35:
```suggestion
    idata : obj
        a :class:`arviz.InferenceData` object
    var_name : str 
        Name of variable we want to test.
    prior : numpy.array, optional
        In case we want to use different prior, for example for sensitivity analysis.
    ref_val : int
        Point-null for Bayes factor estimation. Defaults to 0.
    xlim:  numpy.array, optional
        Set the x limits, which might be used for visualization purposes.
```

arviz/plots/bf_plot.py
```diff
@@ -0,0 +1,83 @@
...
+
+    Returns
+    -------
+    Bayes Factor 10, BF01 and ax
```
aloctavodia, line 39: maybe be more explicit

aloctavodia:
After those changes I think this is ready to merge.

orduek:
Ok, I've added some more words (what the function returns). 
I also committed both of your suggestion (I'm still not sure how to work on the web with that).
Thank you!

aloctavodia (approved):

orduek:
OK, I've tried to fix most of pylint issues.
I have two questions:
(1) How can I load the extract function?
(2) How can I use pylint and run it (is it only via Azure? or locally?)

orduek:
OK, So I've solved how to run pylint :)
My only issue is with loading the extract function. In my code (my local computer), I don't have it at all. In the github, I can find it, under /arviz/data/utils.py. How can I load utils from /data folder?

aloctavodia:
yes, try with a relative import.

aloctavodia:
thanks @orduek!

Issue #230, opened 2018-09-12
Add unconstrained/transformed variables to InferenceData


ColCarroll:
@ahartikainen suggested adding unconstrained variables to the schema somewhere. It seems like they could go 

1. Make new `unconstrained_posterior` group
  Good:
    - Explicit is better than implicit: it is clear what this group is for
  Bad:
    - Should there also be `unconstrained_prior`? 
2. Add them to the existing `posterior` group
  Good:
    - Can reuse existing coords/dims
    - Can add to prior in the same way
  Bad:
    - Would need a naming convention, which we've managed to avoid so far
    - PyMC3 currently does this, and it is kind of a pain to have a flag everywhere to include or exclude these variables.
3. Don't include them at all
  Good:
    - No work :)
  Bad:
    - Bounds and domain information are not captured anywhere currently

aloctavodia:
I like option 3! :-) 

If the main purpose of including unconstrained variables is to explore things like divergences, then option 1 becomes the best option as there is no motivation (other than consistency) to add `unconstrained_prior`. 

BTW, shouldn't the bounds and domain information be part of a variable type (or something similar), currently we only have this for discrete and continuous variables (because we have integers and floats types). This will be super useful for circular variables as we can automatically compute proper statistics and plots for such variables.

ahartikainen:
I think option 2 could be the easiest. If we add an option to transform everything to unconstrained space if a user wants it. That way we do not really need to add anything to the schema. Unless there is a notion on metadata somewhere.

I don't think users normally would need these values.

junpenglao:
Vote for option 2 for the convenience of backward compatibility with pymc3 :grimacing:

dfm:
I just came across this because I often find myself wanting access to the transformed variables in a PyMC3 trace. I often need this in order to evaluate model components after sampling (for example, I often want to plot the model prediction on a fine grid that is too large/memory intensive to save as a deterministic, but is fine to compute at a few points in a thinned chain). Right now, it looks like it would be easy enough to allow opt-in support for PyMC3 here:

https://github.com/arviz-devs/arviz/blob/59575274a5921acae24930733b7125c96ff8e97b/arviz/data/io_pymc3.py#L257-L259

since `include_transformed` is hard-coded to `False`. I'd be keen to make this a parameter of the `PyMC3Converter` to at least have it as an option (this would be option 2 from above). Any feelings about this?

ColCarroll:
[...]

PR #2040, opened 2022-06-02
[WIP] Add marginal likelihood estimation via bridge sampling

karink520:
Co-authored-by: junpenglao <junpenglao@gmail.com>

## Description

Provides an estimate of the (log) marginal likelihood, estimated using bridge sampling (as described in Gronau, Quentin F., et al., 2017), building on an implementation from @junpenglao. This could be expanded to add Bayes factor functionality if desired.

The bridge sampler uses samples from the posterior, so the `log_marginal_likelihood_bridgesampling` function takes as a parameter and InferenceData object that has a posterior group, as well as an unnormalized log probability function (e.g. from a pymc model `model.logp_array`).

Because we fit a multivariate normal proposal distribution to the posterior samples, it is helpful to have samples that are transformed e.g. to have support on the real line instead of on a bounded interval. Although these transformed samples are created as part of the NUTS sampling, I believe they're not currently included in InferenceData (see issue #230 ). So, `log_marginal_likelihood_bridgesampling` currently takes a dict whose keys are variable names and whose values are the associated transformation (or the identity). You could get this from a pymc model with something like the following, although maybe there's a better way:
```
def get_transformation_dict_from_model(model):
    """
    Returns a dict giving the transformations for each variable

    Parameters:
    -----------
    model: a PyMC model

    Returns:
    --------
    transformation_dict: dict 
      Keys are (str) names of model variables (their pre-transformation names),
      Values are their associated transformation as a function that 
      (elementwise) transforms an array. If the variable has no transformation
      associated, we use the identity function.
    """

    transformation_dict = {}
    for var_name in model.named_vars:
        if not var_name.endswith('__'): 
            var = getattr(model, var_name)
            transformation = getattr(var, 'transformation', None)
            if transformation is not None:
              transformation_dict[var_name] = transformation.forward_val
            else: # if no transformation, use identity
              transformation_dict[var_name] = lambda x: x
    return transformation_dict
```

Curious to hear any thoughts or feedback!

## Checklist

- [x] New features are properly documented (with an example if appropriate)

### DIFF
--- a/arviz/stats/bridge_sampling.py
+++ b/arviz/stats/bridge_sampling.py
@@ -1,0 +1,337 @@
+"""Marginal likelihood estimation via bridge sampling."""
+import warnings
+import dataclasses
+
+import scipy.stats as st
+from numpy import dot
+from scipy.linalg import cholesky
+import numpy as np
+from scipy.linalg import lstsq
+from statsmodels.tsa.ar_model import AR
+
+from ..data import InferenceData
+from ..stats.diagnostics import ess
+from ..data.utils import extract_dataset
+
+__all__ = ["log_marginal_likelihood_bridgesampling", "bridgesampling_rel_mse_est"]
+
+
+def log_marginal_likelihood_bridgesampling(idata, logp, transformation_dict=None, maxiter=1000):
+    """
+    Log marginal likelihood estimated via bridgesampling.
+
+    Parameters
+    ----------
+    idata: InferenceData
+        :class:`arviz.InferenceData` object. Must have a posterior group
+    logp: callable
+        Unnormalized posterior log probability function.
+        (E.g. model.logp_array for model, a pymc model)
+    transformation_dict: dict
+      Keys are (str) names of model variables in idata , values are each one's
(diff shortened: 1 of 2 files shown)


ahartikainen:
Should we use a class approach like we do with reloo? This way different backends only need to create specific methods with uniform parameters and output.

cc @OriolAbril

OriolAbril:
I think we should define how to include unconstrained variables in InferenceData and solve #230 between options 1 or 2. It is turning out to be much more work and inconvenient to go with option 3.

I have skimmed the code and have many ideas, mostly related to using xarray more. But I am not sure it is worth it to start changing things yet until we have decided on the issue.

karink520:
Just checking in to see if there's anything I can be helpful with here!

OriolAbril:
Very sorry about the other PR taking so long, but it has finally been merged. I can take care of rebasing if it helps, then I'll try and add some high level comments

OriolAbril:
closing to continue work on arviz-stats, for now https://github.com/arviz-devs/arviz-stats/issues/225

PR #2043, opened 2022-06-08
Make plot_compare easy to understand

### LINKED ISSUE #1989: plot_compare needs a legend
There is a lot going on in the `plot_compare` output. There should be a legend displayed by default, as most users will not be intimately familiar with Statistical Rethinking. There should also be a better description in the function docstring describing what gets plotted.

aloctavodia:
This fixes #1989 

* Add a legend
* Make insample_dev False as default
* Add a label to the y-axis
* Make x-axis label clear, indicates what we are computing and the scale
* Add a title with a short description on how to interpret the plot. This is something we have discussed in the past to incorporate to ArviZ inspired by https://easystats.github.io/performance/index.html), in the future we could even have a global rcparams that turn this short guide on and off globally (or even have levels, beginner, intermedia, advanced)
* This also introduce changes in the names of the computed quantities, for example "elpd_loo" instead of just "loo". This is mainly to increase consistency with the loo R package. 
* Update the docstring a little. In the near future ArviZ is going to have an "educational" site with examples and recommendations on how to use and interpret arviz plots and diagnostics. Once that is ready the docstring will point to that.

~bokeh is not updated because at the moment I have problems displaying bokeh plots~

Old default 
old_loo

New default
new_loo

Suggestions for the text in the plot and/or docstrings are more than welcome, @fonnesbeck, @yannmclatchie, @PabloGGaray. Maybe we should reduce the times ELPD is mentioned :-) 

update after discussion with @yannmclatchie :
What about this as the default, no legend and two axis one for the "raw ELPDs", the other for the differences?

new_loo

### DIFF
--- a/arviz/plots/backends/matplotlib/compareplot.py
+++ b/arviz/plots/backends/matplotlib/compareplot.py
@@ -84,7 +84,7 @@ def plot_compare(
     else:
         ax.set_yticks(yticks_pos[::2])
 
-    scale = comp_df[f"scale"][0]
+    scale = comp_df["scale"][0]
 
     if insample_dev:
         p_ic = comp_df[f"p_{information_criterion.split('_')[1]}"]
--- a/arviz/plots/backends/matplotlib/compareplot.py
+++ b/arviz/plots/backends/matplotlib/compareplot.py
@@ -8,6 +8,7 @@ from . import backend_kwarg_defaults, backend_show, create_axes_grid
 def plot_compare(
     ax,
     comp_df,
+    legend,
     figsize,
     plot_ic_diff,
     plot_standard_error,
@@ -41,12 +42,38 @@ def plot_compare(
     if ax is None:
         _, ax = create_axes_grid(1, backend_kwargs=backend_kwargs)
 
+    if plot_standard_error:
+        ax.errorbar(
+            x=comp_df[information_criterion],
+            y=yticks_pos[::2],
+            xerr=comp_df.se,
+            label="ELPD",
+            color=plot_kwargs.get("color_ic", "k"),
+            fmt=plot_kwargs.get("marker_ic", "o"),
+            mfc=plot_kwargs.get("marker_fc", "white"),
+            mew=linewidth,
+            lw=linewidth,
+        )
+    else:
+        ax.plot(
+            comp_df[information_criterion],
+            yticks_pos[::2],
+            label="ELPD",
+            color=plot_kwargs.get("color_ic", "k"),
+            marker=plot_kwargs.get("marker_ic", "o"),
+            mfc=plot_kwargs.get("marker_fc", "white"),
+            mew=linewidth,
+            lw=0,
+            zorder=3,
+        )
+
     if plot_ic_diff:
         ax.set_yticks(yticks_pos)
         ax.errorbar(
             x=comp_df[information_criterion].iloc[1:],
             y=yticks_pos[1::2],
             xerr=comp_df.dse[1:],
+            label="ELPD difference",
             color=plot_kwargs.get("color_dse", "grey"),
             fmt=plot_kwargs.get("marker_dse", "^"),
             mew=linewidth,
@@ -56,30 +83,9 @@ def plot_compare(
     else:
         ax.set_yticks(yticks_pos[::2])
 
-    if plot_standard_error:
-        ax.errorbar(
-            x=comp_df[information_criterion],
-            y=yticks_pos[::2],
-            xerr=comp_df.se,
-            color=plot_kwargs.get("color_ic", "k"),
-            fmt=plot_kwargs.get("marker_ic", "o"),
-            mfc="None",
-            mew=linewidth,
-            lw=linewidth,
-        )
-    else:
-        ax.plot(
-            comp_df[information_criterion],
-            yticks_pos[::2],
-            color=plot_kwargs.get("color_ic", "k"),
-            marker=plot_kwargs.get("marker_ic", "o"),
-            mfc="None",
-            mew=linewidth,
-            lw=0,
-        )
+    scale = comp_df[f"scale"][0]
 
     if insample_dev:
-        scale = comp_df[f"{information_criterion}_scale"][0]
         p_ic = comp_df[f"p_{information_criterion}"]
         if scale == "log":
             correction = p_ic
@@ -90,6 +96,7 @@ def plot_compare(
         ax.plot(
             comp_df[information_criterion] + correction,
             yticks_pos[::2],
+            label="In-sample ELPD",
             color=plot_kwargs.get("color_insample_dev", "k"),
             marker=plot_kwargs.get("marker_insample_dev", "o"),
             mew=linewidth,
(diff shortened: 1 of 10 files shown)


arviz/plots/compareplot.py
```diff
@@ -22,33 +25,37 @@ def plot_compare(
     show=None,
 ):
     """
-    Summary plot for model comparison.
+    Summary plot for model comparison
 
-    This plot is in the style of the one used in the book Statistical Rethinking (Chapter 6)
-    by Richard McElreath.
+    Models are compared based on their expected log pointwise predictive density (ELPD),
```
yannmclatchie, line 30: Models are compared based on their expected log ~pointwise~ predictive density (ELPD),

arviz/plots/compareplot.py
```diff
@@ -22,33 +25,37 @@ def plot_compare(
...
     """
-    Summary plot for model comparison.
+    Summary plot for model comparison
 
-    This plot is in the style of the one used in the book Statistical Rethinking (Chapter 6)
-    by Richard McElreath.
+    Models are compared based on their expected log pointwise predictive density (ELPD),
+    the ELPD is estimated either by Pareto smoothed importance sampling leave-one-out
+    cross-validation (loo) or using the widely applicable information criterion (WAIC).
+    We recommend LOO.
```
yannmclatchie, line 33: We recommend LOO in line with the theory presented by Vehtari et al. (2016) available here: https://arxiv.org/abs/1507.04544.

arviz/plots/compareplot.py
```diff
@@ -22,33 +25,37 @@ def plot_compare(
...
 ):
     """
-    Summary plot for model comparison.
+    Summary plot for model comparison
 
-    This plot is in the style of the one used in the book Statistical Rethinking (Chapter 6)
-    by Richard McElreath.
+    Models are compared based on their expected log pointwise predictive density (ELPD),
+    the ELPD is estimated either by Pareto smoothed importance sampling leave-one-out
+    cross-validation (loo) or using the widely applicable information criterion (WAIC).
```
yannmclatchie, line 32: cross-validation (~loo~ LOO) or using the widely applicable information criterion (WAIC).

arviz/stats/stats.py
```diff
@@ -61,19 +61,20 @@ def compare(
...
-    r"""Compare models based on PSIS-LOO `loo` or WAIC `waic` cross-validation.
+    r"""Compare models based on  their expected log pointwise predictive density (ELPD).
 
-    LOO is leave-one-out (PSIS-LOO `loo`) cross-validation and
-    WAIC is the widely applicable information criterion.
-    Read more theory here - in a paper by some of the leading authorities
-    on model selection dx.doi.org/10.1111/1467-9868.00353
+
+    The ELPD is estimated either by Pareto smoothed importance sampling leave-one-out
+    cross-validation (loo) or using the widely applicable information criterion (waic).
```
yannmclatchie, line 68: cross-validation (~loo~LOO) or using the widely applicable information criterion (~waic~WAIC).

arviz/stats/stats.py
```diff
@@ -112,29 +113,29 @@ def compare(
...
-        If `scale` is `deviance` or `negative_log` smaller IC indicates
+    elpd: ELPD estimated either using (PSIS-LOO-CV `elpd_loo` or WAIC `elpd_waic`).
+        Higher ELPD indicates higher out-of-sample predictive fit ("better" model).
+        If `scale` is `deviance` or `negative_log` smaller values indicates
         higher out-of-sample predictive fit ("better" model).
     pIC: Estimated effective number of parameters.
-    dIC: Relative difference between each IC (PSIS-LOO `loo` or WAIC `waic`)
-          and the lowest IC (PSIS-LOO `loo` or WAIC `waic`).
-          The top-ranked model is always 0.
+    elpd_diff: the difference in elpd for two models.
```
yannmclatchie, line 124:
```suggestion
    elpd_diff: The difference in ELPD between two models.
```

arviz/stats/stats.py
```diff
@@ -343,7 +342,7 @@ def _calculate_ics(
     ic: Optional[ICKeyword] = None,
     var_name: Optional[str] = None,
 ):
-    """Calculate loo and waic information criteria only if necessary.
+    """Calculate loo or waic only if necessary.
```
yannmclatchie, line 345:
```suggestion
    """Calculate LOO or WAIC only if necessary.
```

yannmclatchie:
This looks like a good start, I particularly like the interpretable title and the future idea of a level-based guide through the workflow.

I wonder if both the x-axis label and legend are necessary if they repeat the fact that they show ELPD? The ELPD difference is still somewhat confusing to me, since it is shown on the same scale and axis as the raw ELPDs. If a user is looking at the credible interval of the ELPD difference, it is likely that they are looking to see if zero is included in that interval, which is not immediately obvious from this plot. 

What may be clearer is if we produce multiple separate plots showing:
1. The ELPD distributions of the ranked models with the same dashed line for the "best" model (the current default plot but without the differences) with no legend and the currently suggested x-axis label, and;
2. A similar plot of just the difference in ELPD distributions with another dashed line at zero (something of an equivalent to [`plot.vsel` in `projpred`](https://mc-stan.org/projpred/reference/plot.vsel.html) with `deltas=TRUE`) which is only shown alongside the raw ELPD plot if the user passes `plot_deltas = True` for instance, again without legend and with an appropriate x-axis label.

Let me know what you think :pray:

aloctavodia:
@yannmclatchie thanks for the suggestions, I incorporated those related to the docstring.
Regarding the plot I would like to keep this plot as close as possible to the one used in Statistical Rethinking. But maybe we can add a second plot like the one you are suggesting. Even when the scale is on the "raw ELPD" values, the ELPD difference are computed with respect to the top-ranked model and thus the vertical dashed line can be used as the reference point.

cfonnesbeck:
The "new default" in the first pane is a big improvement, so thanks! I'm not as enthusiastic about the second one--dual y-axes don't work all that well IMO. 

An even simpler version would be to remove the difference from the plot altogether and add it as a text annotation to the relevant point. 

... OR allow users to plot the differences *instead* of the raw ELPD scores, if that is what they are interested in.

arviz/plots/backends/matplotlib/compareplot.py
```diff
@@ -41,45 +43,50 @@ def plot_compare(
...
+        )
+
+    else:
+        ax.set_yticks(yticks_pos[::2])
+
+    scale = comp_df[f"scale"][0]
+
     if insample_dev:
-        scale = comp_df[f"{information_criterion}_scale"][0]
         p_ic = comp_df[f"p_{information_criterion}"]
```
OriolAbril, line 90: This continues to be `p_loo` but now the fstring generates `p_elpd_loo` which is an invalid key. Not sure why tests are all passing, but it is the reason the doc related builds are failing.

OriolAbril (commented):
I like the new default. I'd vote against double axis as it is quite confusing.

I am not very opinionated on with/without difference on the plot as default. I do think it is good to add/remove it with a boolean kwarg. Maybe it could be added as an rcParam so users can change the library default on their personal rctemplate and forget about using that argument every time the plot is called.

aloctavodia:
Thanks for the feedback. I will work on refining the "new default". We already have a keyword for removing the differences. I agree we can add a rcParam.

aloctavodia:
@OriolAbril can you provide help about the errors building the docs?

OriolAbril:
Should be fixed now. The "See Also" section has a special syntax. elements must be valid sphinx references and one can optionally add a description to the link after a ` : `. If the description is too long to fit in one line it needs to be on the line below _and indented_. Ref: https://numpydoc.readthedocs.io/en/latest/format.html#see-also

OriolAbril (approved):

PR #2044, opened 2022-06-11
Pin pylint version

aloctavodia:


requirements-dev.txt
```diff
@@ -3,7 +3,7 @@ astroid
 madforhooks
 numpydoc
 pydocstyle
-pylint
+pylint=2.10.2
```
charlesdong1991, line 6:
```suggestion
pylint==2.10.2
```

probably need double `=` to hard-pin?

btw, a small side note: i was trying to fix the CI with this PR: https://github.com/arviz-devs/arviz/pull/2046 🤗 

There seem to be several pylint and mypy errors leading to CI failures. It fails now but i _think_ the error is not related? 🤔

PR #2045, opened 2022-06-12
Correct a typo in allowed method in `compare`

charlesdong1991:
## Description

It seems arviz allows `pseudo-BMA` instead of `pseudo-MA` as options, this PR corrects this typo.

## Checklist

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -54,7 +54,7 @@ __all__ = [
 def compare(
     compare_dict: Mapping[str, InferenceData],
     ic: Optional[ICKeyword] = None,
-    method: Literal["stacking", "BB-pseudo-BMA", "pseudo-MA"] = "stacking",
+    method: Literal["stacking", "BB-pseudo-BMA", "pseudo-BMA"] = "stacking",
     b_samples: int = 1000,
     alpha: float = 1,
     seed=None,


OriolAbril:
@charlesdong1991 now that CI is fixed (again, thanks for the help), can you rebase the PR? I don't see any reason why CI would not pass with those changes but always best to check, specially for pylint/black/mypy, we might be able to improve the mypy config and ignore less errors for example

charlesdong1991:
@OriolAbril i rebased, and CI is still good.

Small question though: any idea why mypy wasn't giving any errors before on this typo? I would assume mypy will complain given how `Literal` is being designed. 🤔 I switched import of `Literal` from `type_extension` to `typing`, but seems the same (no error).

OriolAbril:
I have no idea.

We might be ignoring the error in https://github.com/arviz-devs/arviz/blob/main/mypy.ini or maybe the pseudo-BMA value is not used anywhere in the codebase? I'd think it has a test given its importance for the compare method, but we don't test all values of all arguments so I might be wrong here. 

Or, can mypy understand we have an `if not in` then raise error and compares the tuple there to the literal defined? That sounds a bit too good to be true to me

charlesdong1991:
kk, there is a test indeed for checking the value. mypy only analyzes code but doesn't execute, so i think this doesn't affect code execution at all.  Probably some mypy configuration issue indeed, quite weird 🤔

PR #2061, opened 2022-06-28
update versions

aloctavodia:
Not crucial, but maybe we want to list newer versions, for example SciPy 0.19 can be run on Python 2.

This update the version to packages releases around December 2021 (or  a little later), except for xarray-einstats that is updated to the latest version.


OriolAbril:
Sounds good, I did something similar on xarray-einstats in https://github.com/arviz-devs/xarray-einstats/pull/19 trying to follow [nep 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) and [spec 0](https://scientific-python.org/specs/spec-0000/) (draft from https://scientific-python.org/). 

We should also consequently update Azure to run on 3.8-3.10 python versions

Issue #2056, opened 2022-06-21
Council vote: unconstrained variables


OriolAbril:
#230 has been open for too long without any clear consensus. Following https://www.arviz.org/en/latest/governance/processes.html#council-decision-making-process I am opening this issue with a proposal on how to solve this and calling a vote of the RV council. cc @AlexAndorra @sethaxen @ColCarroll @ahartikainen @canyon289 @aloctavodia. 

### Proposal
Add a new `unconstrained_posterior` group (and other `unconstrained_<group>` as needed). I propose that the unconstrained representation of variables with a transformation (no need to duplicate those that are the same) are added to this group with the exact same name as they have in the posterior. The transform can be added as metadata either at variable level or as metadata of the unconstrained_... group and is left to the PPLs to add.

The whole unconstrained posterior can therefore be obtained by taking the variables in `unconstrained_posterior` and adding those missing using their values form the `posterior` group.

OriolAbril:
Note: We now enter the process of asking for clarifications on the _proposal's wording_, I am locking to prevent people not familiar with the governance process to join the discussion. Ref: https://www.arviz.org/en/latest/governance/processes.html#call-for-a-vote

sethaxen:
The transformation will be formulated as a transformation from the unconstrained to constrained variable, correct?

OriolAbril:
The proposal is to store transformed and untransformed samples and make this part of the inferencedata schema.

The formulation of the transformation and how to do it is left completely on the hands of the PPLs and the people writing the converters. It can be the actual function in either direction stored as metadata (probably won't work saving to netcdf or zarr then but metadata can be excluded from the writing process, and would work while using the results after sampling); it can be a keyword that identifies the transform in that PPL, it might also be possible to combine both and use a function like `turing_transforms_to/from_metadata`...

ahartikainen:
What groups might have this property?

- [ ] `posterior`
- [ ] `posterior_predictive` (??)
- [ ] `predictions`
- [ ] `prior`
- [ ] `prior_predictive` (??)

ColCarroll:
I think the idea is that this would be one or two new groups: 
- `posterior_transformed`
- `prior_transformed` (_maybe_ -- I don't know why I would want that, or to what end I would use it)

I agree that we should leave this mostly up to the PPLs to implement, but as a _suggestion_, we might have a posterior:

```
sd ~ HalfNormal(...)
mu ~ Normal(...)
[...]

PR #2086, opened 2022-08-07
start updating schema doc

### LINKED ISSUE #230: Add unconstrained/transformed variables to InferenceData

### LINKED ISSUE #2048: Supporting multiple posterior predictive draws for each posterior draw
For each posterior draw $n \in \{1, \ldots, N\}$, a user may want to generate and store $M$ posterior predictive draws in `InferenceData`. To support this, we should:
- decide how these should be stored, e.g. probably the $M$ draws should be stored as an additional dimension, but should we specify the name of that dimension?
- make sure that all functions that use posterior predictive draws do the right thing. In some cases, this might mean using one of the draws for each posterior draw and warning the user. In other cases, this might mean plotting all draws by default and allowing the user to select a subset of the draws to plot. In other cases, it might mean averaging over the draws.

### LINKED ISSUE #2064: Some changes to the InferenceData schema
## Tell us about it

With https://github.com/arviz-devs/ArviZ.jl/pull/191, ArviZ.jl has an independent implementation of the [`InferenceData` schema](https://python.arviz.org/en/latest/schema/schema.html), which has previously only been implemented using xarray in Python for the storage. This motivates a few minor additions and changes to the schema:

## Additions

- New default attribute: `arviz_language`: currently either "python" or "julia". This is necessary for interpreting the "arviz_version", since the Python and Julia versions do not correspond.
- Require that the `arviz_version` attribute is always included. Unless we intend for packages besides ArviZ to define their own implementations of the schema. In that case, there are other attributes we should add (e.g. `created_by`).

## Changes

- Remove the constraint that `inference_library` and `inference_library_version` are included. These maybe don't always make sense to include, e.g. when samples were generated exactly with scipy or maybe when multiple inference libraries were used together to generate samples. In Julia, the object storing MCMC chains is sometimes implemented in a different package from the PPL, and the actual inference algorithm might live in its own package (e.g. MCMCChains implements `chains`, which is used in both Turing and CmdStan/Stan), so the storage object itself is often not sufficient to infer the inference library, and it's also not clear which package the inference library should even be. 
- Remove the constraint that `[chain, dim]` be the _first_ dimensions. In general, I think we've agreed this should not be assumed by ArviZ's functions, and plots/diagnostics should consult the axes.

It might make sense to define the schema separately from the implementation. e.g. ArviZ.jl and Python ArviZ might both implement the same schema but with different conventions. e.g. ArviZ.jl and Python ArviZ could adopt different conventions RE where `[chain, dim]` are placed in the dimensions list. Or maybe ArviZ.jl would use an API where the `attrs` keyword is replaced by `metadata`, though they are both serialized to NetCDF attributes. So the schema could be defined at [arviz.org](https://www.arviz.org/en/latest/), and then each implementation could define any additional conventions their implementation uses.

OriolAbril:
## Description
The goal is to close #230, close #2048 and close #2064. So even though the fix for #230 has already been voted on #2056, I expect some discussion to happen, especially around the rest of the changes.

I have pushed the branch to arviz repo to make it easy for other people to make changes directly. cc @sethaxen 

I am also making the changes here for now, which should get the most visibility, we can move the page to the arviz-project repo after merging.

## Checklist

- ✅  Follows [official](https://github.com/arviz-devs/arviz/blob/main/CONTRIBUTING.md#pull-request-checklist) PR format
- ❌  New features are properly documented (with an example if appropriate)? It is probably time already to rerun the examples, but I don't plan to do it on this PR.
- 🔲  Changes are listed in [changelog](https://github.com/arviz-devs/arviz/blob/main/CHANGELOG.md#v0xx-unreleased)

----
:books: Documentation preview :books:: https://arviz--2086.org.readthedocs.build/en/2086/

### DIFF
--- a/doc/source/schema/schema.md
+++ b/doc/source/schema/schema.md
@@ -8,16 +8,19 @@ The `InferenceData` schema approach defines a data structure compatible with [Ne
 Currently there are **two beta implementations** of this design:
 * {ref}`ArviZ <homepage>` in **Python** which integrates with:
   - [emcee](https://emcee.readthedocs.io/en/stable/)
-  - [PyMC3](https://docs.pymc.io)
+  - [PyMC](https://www.pymc.io)
   - [Pyro](https://pyro.ai/) and [NumPyro](https://pyro.ai/numpyro/)
-  - [PyStan](https://pystan.readthedocs.io/en/latest/index.html), [CmdStan](https://mc-stan.org/users/interfaces/cmdstan) and [CmdStanPy](https://cmdstanpy.readthedocs.io/en/latest/index.html)
+  - [PyStan](https://pystan.readthedocs.io/en/latest/index.html),
+    [CmdStan](https://mc-stan.org/docs/cmdstan-guide/index.html) and
+    [CmdStanPy](https://mc-stan.org/cmdstanpy/)
   - [TensorFlow Probability](https://www.tensorflow.org/probability)
 * [ArviZ.jl](https://github.com/arviz-devs/ArviZ.jl) in **Julia** which integrates with:
   - [CmdStan.jl](https://github.com/StanJulia/CmdStan.jl), [StanSample.jl](https://github.com/StanJulia/StanSample.jl) and [Stan.jl](https://github.com/StanJulia/Stan.jl)
   - [Turing.jl](https://turing.ml/dev/) and indirectly any package using [MCMCChains.jl](https://github.com/TuringLang/MCMCChains.jl) to store results
 
 ## Terminology
-The terminology used in this specification is based on [xarray's terminology](http://xarray.pydata.org/en/stable/terminology.html), however, no xarray knowledge is assumed in this description. There are also some extensions particular to  the {ref}`InferenceData <xarray_for_arviz>` case.
+The terminology used in this specification is based on {ref}`xarray's terminology <xarray:terminology>`, however, no xarray knowledge is assumed in this description, nor xarray is needed to use or interact with the schema.
+There are also some extensions particular to  the {ref}`InferenceData <xarray_for_arviz>` case.
 
 * **Variable**: NetCDF-like variables are multidimensional labeled arrays representing a single quantity. Variables and their dimensions must be named. They can also have attributes describing it. Relevant terms related to `InferenceData` variables are following:
   - *variable_name*
@@ -36,24 +39,36 @@ The terminology used in this specification is based on [xarray's terminology](ht
 `InferenceData` stores all quantities that are relevant to fulfilling its goals in different groups. Different groups generally distinguish conceptually different quantities in Bayesian inference, however, convenience in {ref}`creation <creating_InferenceData>` and {ref}`usage <working_with_InferenceData>` of `InferenceData` objects also plays a role. In general, each quantity (such as posterior distribution or observed data) will be represented by several multidimensional labeled variables.
 
 ### Rules
-Following are a few rules which should be followed:
+Below are a few rules which should be followed:
 * Each group should have one entry per variable and each variable should be named.
-* When relevant, the first two dimensions of each variable should be the sample identifier (`chain`, `draw`).
-* For groups like `observed_data` or `constant_data`, the two initial dimensions(`chain`, `draw`) are omitted.
+* Dimension names `chain`, `draw`, `sample` and `post_pred_id` are reserved for
+  InferenceData use to indicate sample dimensions.
+* Dimensions in InferenceData (including sample dimensions) should be identified by name only.
+* For groups like `observed_data` or `constant_data`, all sample dimensions can be
+  omitted. For groups like `prior`, `posterior` or `posterior_predictive` either `sample` has to be
+  present or both `chain` and `draw` dimensions need to be present. Any combinations that follow
+  this are valid.
 * Dimensions must be named and share name with a coordinate specifying the index values, called coordinate values.
 * Coordinate values can be repeated and should not necessarily be numerical values.
 * Variables must not share names with dimensions.
-* Moreover, each group contains the following attributes:
-  - `created_at`: the date of creation of the group.
-  - `inference_library`: the library used to run the inference.
-  - `inference_library_version`: version of the inference library used.
 
 ### Relations
-`InferenceData` data objects contain any combination of the groups described below. There are also some relations (detailed below) between the variables and dimensions of different groups. Hence, whenever related groups are present they should comply with these relations.
+`InferenceData` data objects contain any combination of the groups described below. There are also some relations (detailed below) between the variables and dimensions of different groups. Hence, whenever related groups are present they should comply with these relations. Neither the presence of groups nor described below or the lack of some of the groups described below go against the schema.
 
 #### `posterior`
-Samples from the posterior distribution p(theta|y).
+Samples from the posterior distribution p(theta|y) in the parameter (also called constrained) space.
 
+#### `unconstrained_posterior`
+Samples from the posterior distribution p(theta_transformed|y) in the unconstrained (also called transformed) space.
+
+Only variables that undergo a transformation for sampling should be present here.
+Therefore, to get the samples for _all_ the variables in the unconstrained space,
+variables should be taken from the `unconstrained_posterior` group if present,
+and if not, then the values from the variable in the `posterior` group should be used.
+
+Both samples and variables (for those present only) should match between the `posterior` and the `unconstrained_posterior` groups. Note that as defined above, matching samples and variables impose constraints on dimensions and coordinates, not on the values which will be different.
+
+(schema/sample_stats)=
 #### `sample_stats`
 Information and diagnostics for each `posterior` sample, provided by the inference
 backend. It may vary depending on the algorithm used by the backend (i.e. an affine
@@ -61,8 +76,12 @@ invariant sampler has no energy associated). Therefore none of these parameters
 should be assumed to be present in the `sample_stats` group. The convention
 below serves to ensure that if a variable is present with one of these names
 it will correspond to the definition given in front of it.
+Moreover, some `sample_stats` may be constant throughout the sampling
+process; these variables don't need to have any sampling dimensions.
 
-The name convention used for `sample_stats` variables is the following:
+
+:::{dropdown} Naming convention used for `sample_stats` variables
+:icon: list-unordered
 
 * `lp`: The joint log posterior density for the model (up to an additive constant).
 * `acceptance_rate`: The average acceptance probabilities of all possible samples in the proposed tree.
@@ -79,6 +98,11 @@ additive constant).
 the accepted proposal.
 * `max_energy_error`: The maximum absolute difference in Hamiltonian energy between the initial point and all possible samples in the proposed tree.
 * `int_time`: The total integration time (static HMC sampler)
+* `mass_matrix`: Mass matrix (also known as _metric_) used in HMC samplers for the computation of the Hamiltonian.
+  When it is constant, the resulting implementation is known as Euclidean HMC;
+  in that case, the variable wouldn't need to have any sampling dimensions
+  even if part of the `sample_stats` group.
+:::
 
 
 #### `log_likelihood`
@@ -99,9 +123,6 @@ Model constants, data included in the model which is not modeled as a random var
 #### `prior`
 Samples from the prior distribution p(theta). Samples do not need to match `posterior` samples. However, this group will still follow the convention on `chain` and `draw` as first dimensions. It should have matching variables with the `posterior` group.
 
-#### `sample_stats_prior`
-Information and diagnostics for the samples in the `prior` group, provided by the inference backend. It may vary depending on the algorithm used by the backend. Variable names follow the same convention defined in `sample_stats`.
-
 #### `prior_predictive`
 Samples from the prior predictive distribution. Samples should match `prior` samples and each variable should have a counterpart in `posterior_predictive`/`observed_data`.
 
@@ -110,6 +131,37 @@ Out of sample posterior predictive samples p(y'|y). Samples should match `poster
 
 #### `predictions_constant_data`
 Model constants used to get the `predictions` samples. Its variables should have a counterpart in `constant_data`. However, variables in `predictions_constant_data` and their counterpart in `constant_data` can have different coordinate values.
+
+:::{admonition} Note on sample stats, warmup and unconstrained groups
+:class: note, dropdown
+
+The schema does not define which warmup or unconstrained groups exist or can exist
+by default. We recognize both the samplers and the models are continuously evolving.
+Some models already require the use of sampling algorithms to get prior samples,
+in which case we basically need to treat the prior and posterior groups in the same way.
+
+We define the prefixes and to allow libraries that use InferenceData to be aware
+of the potential relations and hopefully support as many cases as possible.
+Back to the case above, it might be necessary to generate a pair plot
+for prior samples generated with NUTS _and_ its associated divergences,
+which would then come from `sample_stats_prior`.
+:::
+
+#### Sample stats groups
+Information and diagnostics for the samples in any InferenceData group
+other than the posterior should be stored in a separate group with the
+`sample_stats_` prefix. For example `sample_stats_prior`.
+
+The same rules and conventions defined in {ref}`schema/sample_stats` apply to
+any sample stats group.
+
+#### Warmup groups
+Samples generated during the adaptation/warmup phases of algorithms like HMC
+can also be stored in InferenData. In such cases, the data/samples
+generated during the adaptation process should be stored in groups with
+the same name with the `warmup_` prefix, e.g. `warmup_posterior`, `warmup_sample_stats_prior`.
+
+#### Unconstrained groups
 
 ## Planned features
 The `InferenceData` structure is still evolving, with some feature being currently developed. This section aims to describe the roadmap of the specification.
--- a/doc/source/schema/schema.md
+++ b/doc/source/schema/schema.md
@@ -15,7 +15,7 @@ Currently there are **two beta implementations** of this design:
     [CmdStanPy](https://mc-stan.org/cmdstanpy/)
   - [TensorFlow Probability](https://www.tensorflow.org/probability)
 * [ArviZ.jl](https://github.com/arviz-devs/ArviZ.jl) in **Julia** which integrates with:
-  - [CmdStan.jl](https://github.com/StanJulia/CmdStan.jl), [StanSample.jl](https://github.com/StanJulia/StanSample.jl) and [Stan.jl](https://github.com/StanJulia/Stan.jl)
+  - [CmdStan.jl](https://github.com/StanJulia/CmdStan.jl), [Soss.jl](https://cscherrer.github.io/Soss.jl/stable/), [StanSample.jl](https://github.com/StanJulia/StanSample.jl) and [Stan.jl](https://github.com/StanJulia/Stan.jl)
   - [Turing.jl](https://turing.ml/dev/) and indirectly any package using [MCMCChains.jl](https://github.com/TuringLang/MCMCChains.jl) to store results
 
 ## Terminology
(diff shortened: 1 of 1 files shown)


doc/source/schema/schema.md
```diff
@@ -36,33 +39,73 @@ The terminology used in this specification is based on [xarray's terminology](ht
 `InferenceData` stores all quantities that are relevant to fulfilling its goals in different groups. Different groups generally distinguish conceptually different quantities in Bayesian inference, however, convenience in {ref}`creation <creating_InferenceData>` and {ref}`usage <working_with_InferenceData>` of `InferenceData` objects also plays a role. In general, each quantity (such as posterior distribution or observed data) will be represented by several multidimensional labeled variables.
 
 ### Rules
-Following are a few rules which should be followed:
+Below are a few rules which should be followed:
 * Each group should have one entry per variable and each variable should be named.
-* When relevant, the first two dimensions of each variable should be the sample identifier (`chain`, `draw`).
-* For groups like `observed_data` or `constant_data`, the two initial dimensions(`chain`, `draw`) are omitted.
+* Dimension names `chain`, `draw`, `sample` and `post_pred_id` are reserved for
```
ahartikainen, line 44:
```suggestion
* Dimension names `chain`, `draw`, `sample`, `posterior_pred_id` and `prior_pred_id` are reserved for
```

doc/source/schema/schema.md
```diff
@@ -36,33 +39,73 @@ The terminology used in this specification is based on [xarray's terminology](ht
...
-`InferenceData` data objects contain any combination of the groups described below. There are also some relations (detailed below) between the variables and dimensions of different groups. Hence, whenever related groups are present they should comply with these relations.
+### Metadata
+No metadata is _required_ to be present in order to be compliant with the InferenceData schema.
+However, it is recommended to store the following fields when relevant:
+* `name`: InferenceData objects represent multiple quantities related to Bayesian modelling,
+  but they are all tied to a single model. The model identifier can be added as metadata
+  to simplify the calls to model comparison functions.
+* `created_at`: the date of creation of the group.
+* `arviz_version`: the version of the ArviZ library that generated the InferenceData
+* `arviz_language`: the programming language from which ArviZ was used to create the InferenceData
```
ahartikainen, line 64: Could these be a bit more general?

If another program creates the InferenceData objects, it would be a good idea to save its version, language. So also the library name for it might be a good idea.

`creation_library`

doc/source/schema/schema.md
```diff
@@ -36,33 +39,73 @@ The terminology used in this specification is based on [xarray's terminology](ht
...
+(schema/unconstrained_posterior)=
+#### `unconstrained_posterior`
+Samples from the posterior distribution p(theta_transformed|y) in the unconstrained (also called transformed) space.
+
+Only variables that undergo a transformation for sampling should be present here.
+Therefore, to get the samples for _all_ the variables in the unconstrained space,
+variables should be taken from the `unconstrained_posterior` group if present,
+and if not, then the values from the variable in the `posterior` group should be used.
+
+Both samples and variables (for those present only) should match between the `posterior` and the `unconstrained_posterior` groups. Note that as defined above, matching samples and variables impose constraints on dimensions and coordinates, not on the values which will be different.
```
ahartikainen, line 84: What are these constraints?

E.g. constrained and unconstrained spaces might have different number of dimensions

doc/source/schema/schema.md
```diff
@@ -36,33 +39,73 @@ The terminology used in this specification is based on [xarray's terminology](ht
...
 
 #### `posterior`
-Samples from the posterior distribution p(theta|y).
+Samples from the posterior distribution p(theta|y) in the parameter (also called constrained) space.
+
+(schema/unconstrained_posterior)=
+#### `unconstrained_posterior`
+Samples from the posterior distribution p(theta_transformed|y) in the unconstrained (also called transformed) space.
+
+Only variables that undergo a transformation for sampling should be present here.
```
ahartikainen, line 79: This might be a problematic for Stan where it is easy to calculate the transformation but hard to make sure that transformation is not unity (or is comparing values against each other valid way to do this?)

doc/source/schema/schema.md
```diff
@@ -36,33 +39,73 @@ The terminology used in this specification is based on [xarray's terminology](ht
...
+
+Both samples and variables (for those present only) should match between the `posterior` and the `unconstrained_posterior` groups. Note that as defined above, matching samples and variables impose constraints on dimensions and coordinates, not on the values which will be different.
 
+:::{note}
+:class: dropdown
+
+Both InferenceData groups and variables can have metadata, which in the `unconstrained_posterior`
+case could be used to store the transformations each variable goes through to map between the
+constrained and unconstrained spaces. The schema leaves this completely up to the user
+and imposes on conventions or restrictions on such metadata.
```
ahartikainen, line 92:
```suggestion
and imposes no conventions or restrictions on such metadata.
```

?

doc/source/schema/schema.md
```diff
@@ -111,6 +156,44 @@ Out of sample posterior predictive samples p(y'|y). Samples should match `poster
...
+Information and diagnostics for the samples in any InferenceData group
+other than the posterior should be stored in a separate group with the
+`sample_stats_` prefix. For example `sample_stats_prior`.
+
+The same rules and conventions defined in {ref}`schema/sample_stats` apply to
+any sample stats group.
+
+#### Warmup groups
+Samples generated during the adaptation/warmup phases of algorithms like HMC
+can also be stored in InferenData. In such cases, the data/samples
```
ahartikainen, line 184:
```suggestion
can also be stored in InferenceData. In such cases, the data/samples
```

doc/source/schema/schema.md
```diff
@@ -111,6 +156,44 @@ Out of sample posterior predictive samples p(y'|y). Samples should match `poster
...
+generated during the adaptation process should be stored in groups with
+the same name with the `warmup_` prefix, e.g. `warmup_posterior`, `warmup_sample_stats_prior`.
```
ahartikainen, line 186: How would the naming go for unconstained warmup groups

    warmup_unconstrained_posterior

?

ahartikainen (commented):
Added couple of comments, questions and fixes

doc/source/schema/schema.md
sethaxen: One might also want a `pred_id` for `predictions` group. What if `pred_id` was the only reserved name and was used in `predictions`, `posterior_predictions`, and `prior_predictions`?

doc/source/schema/schema.md
sethaxen: I 2nd this, along with a `creation_library_version`. We might also want `creation_library_language`. On the Julia side we could append a `".jl"` to the name if a user provides a package object itself instead of just the name to differentiate, but `".py"` for Python makes less sense.

doc/source/schema/schema.md
```diff
@@ -8,16 +8,19 @@ The `InferenceData` schema approach defines a data structure compatible with [Ne
...
+  - [PyMC](https://www.pymc.io)
   - [Pyro](https://pyro.ai/) and [NumPyro](https://pyro.ai/numpyro/)
-  - [PyStan](https://pystan.readthedocs.io/en/latest/index.html), [CmdStan](https://mc-stan.org/users/interfaces/cmdstan) and [CmdStanPy](https://cmdstanpy.readthedocs.io/en/latest/index.html)
+  - [PyStan](https://pystan.readthedocs.io/en/latest/index.html),
+    [CmdStan](https://mc-stan.org/docs/cmdstan-guide/index.html) and
+    [CmdStanPy](https://mc-stan.org/cmdstanpy/)
   - [TensorFlow Probability](https://www.tensorflow.org/probability)
 * [ArviZ.jl](https://github.com/arviz-devs/ArviZ.jl) in **Julia** which integrates with:
   - [CmdStan.jl](https://github.com/StanJulia/CmdStan.jl), [StanSample.jl](https://github.com/StanJulia/StanSample.jl) and [Stan.jl](https://github.com/StanJulia/Stan.jl)
   - [Turing.jl](https://turing.ml/dev/) and indirectly any package using [MCMCChains.jl](https://github.com/TuringLang/MCMCChains.jl) to store results
```
sethaxen, line 19: [Soss.jl](https://github.com/cscherrer/Soss.jl) can be added here also.

doc/source/schema/schema.md
```diff
@@ -36,33 +39,73 @@ The terminology used in this specification is based on [xarray's terminology](ht
 `InferenceData` stores all quantities that are relevant to fulfilling its goals in different groups. Different groups generally distinguish conceptually different quantities in Bayesian inference, however, convenience in {ref}`creation <creating_InferenceData>` and {ref}`usage <working_with_InferenceData>` of `InferenceData` objects also plays a role. In general, each quantity (such as posterior distribution or observed data) will be represented by several multidimensional labeled variables.
 
 ### Rules
-Following are a few rules which should be followed:
+Below are a few rules which should be followed:
```
sethaxen, line 42:
```suggestion
Below are a few rules that should be followed:
```

doc/source/schema/schema.md
```diff
@@ -36,33 +39,73 @@ The terminology used in this specification is based on [xarray's terminology](ht
...
 * Each group should have one entry per variable and each variable should be named.
-* When relevant, the first two dimensions of each variable should be the sample identifier (`chain`, `draw`).
-* For groups like `observed_data` or `constant_data`, the two initial dimensions(`chain`, `draw`) are omitted.
+* Dimension names `chain`, `draw`, `sample` and `post_pred_id` are reserved for
+  InferenceData use to indicate sample dimensions.
+* Dimensions in InferenceData (including sample dimensions) should be identified by name only.
```
sethaxen, line 46: What does this mean?

doc/source/schema/schema.md
```diff
@@ -36,33 +39,73 @@ The terminology used in this specification is based on [xarray's terminology](ht
...
-  - `inference_library`: the library used to run the inference.
-  - `inference_library_version`: version of the inference library used.
+* Groups, variables or the InferenceData itself can have arbitrary metadata stored.
 
-### Relations
-`InferenceData` data objects contain any combination of the groups described below. There are also some relations (detailed below) between the variables and dimensions of different groups. Hence, whenever related groups are present they should comply with these relations.
+### Metadata
+No metadata is _required_ to be present in order to be compliant with the InferenceData schema.
+However, it is recommended to store the following fields when relevant:
+* `name`: InferenceData objects represent multiple quantities related to Bayesian modelling,
```
sethaxen, line 59: Would `model_name` be more explicit that this is meant to be the name of the model, not the name of the group or dataset?

sethaxen: Or is this metadata at the InferenceData level instead of at the group level?

doc/source/schema/schema.md
```diff
@@ -36,33 +39,73 @@ The terminology used in this specification is based on [xarray's terminology](ht
...
+  to simplify the calls to model comparison functions.
+* `created_at`: the date of creation of the group.
+* `arviz_version`: the version of the ArviZ library that generated the InferenceData
+* `arviz_language`: the programming language from which ArviZ was used to create the InferenceData
+* `inference_library`: the library used to run the inference.
+* `inference_library_version`: version of the inference library used.
+
+
+### Relations between groups
+`InferenceData` data objects contain any combination of the groups described below. There are also some relations (detailed below) between the variables and dimensions of different groups. Hence, whenever related groups are present they should comply with these relations. Neither the presence of groups nor described below or the lack of some of the groups described below go against the schema.
```
sethaxen, line 70:
```suggestion
`InferenceData` data objects contain any combination of the groups described below. There are also some relations (detailed below) between the variables and dimensions of different groups. Hence, whenever related groups are present they should comply with these relations. Neither the presence of groups not described below or the lack of some of the groups described below go against the schema.
```

doc/source/schema/schema.md
```diff
@@ -36,33 +39,73 @@ The terminology used in this specification is based on [xarray's terminology](ht
...
 
 #### `posterior`
-Samples from the posterior distribution p(theta|y).
+Samples from the posterior distribution p(theta|y) in the parameter (also called constrained) space.
```
sethaxen, line 73: Should this use mathjax?

doc/source/schema/schema.md
```diff
@@ -79,6 +122,11 @@ additive constant).
 the accepted proposal.
 * `max_energy_error`: The maximum absolute difference in Hamiltonian energy between the initial point and all possible samples in the proposed tree.
 * `int_time`: The total integration time (static HMC sampler)
+* `mass_matrix`: Mass matrix (also known as _metric_) used in HMC samplers for the computation of the Hamiltonian.
```
sethaxen, line 125: Isn't it more common to represent the inverse of the metric instead of the metric itself? I believe this is what Stan does, and this is also what Julia's HMC samplers do. In many/most cases the inverse metric is diagonal, so perhaps we should explicitly say that diagonal metrics can be represented as just the diagonal?

Personally I prefer the term metric over mass matrix.

ahartikainen: Stan also prefers metric (mass_matrix comes from physics)

doc/source/schema/schema.md
OriolAbril: I like `pred_id` much more! Thanks!

doc/source/schema/schema.md
OriolAbril: Yes, that is indeed much better, I'll update it

doc/source/schema/schema.md
OriolAbril: I was completely unaware of that 😓, so they would have the same sampling dimensions and variable name but not necessarly the same coords and dims, is that right? Would it be ok to allow for variables in unconstrained_posterior to have different shape than that same variable in posterior group, but that if they do have the same shape, then dims and coords should be shared?

doc/source/schema/schema.md
OriolAbril: This is only relative to the samples, so comparing values against each other is perfeclty fine. Even in edgecases like a transformation that is the unity in most of the space but not everywhere, if all the samples are the same then the transformation is not doing anything, we don't need to save the samples in both places; and if we generate new posterior samples, we also need to generate the associated unconstrained samples both of which are generated by the PPLs what will know if and which transformations to do.

Maybe we can write it in a bit more clear phrasing. Not the good style for the doc, but in a conversational style it could be something like: I have the posterior samples and I want to have the unconstrained samples. What are the variables whose
samples I need to replace with different ones? -> those variables and samples are the ones that go in this `unconstrained_sample` group.

doc/source/schema/schema.md
OriolAbril: yes

doc/source/schema/schema.md
OriolAbril: yes, I would make the `warmup_` prefix prepend the other two potential prefixes. I'll add that.

doc/source/schema/schema.md
OriolAbril: That the dimension called `chain` is identified by this and only this. It's position relative to other dimensions is not relevant

doc/source/schema/schema.md
OriolAbril: I was thinking about group level attributes here. I'll update the title. Maybe also add a note that the creation_library attrs might be stored in groups (they could be different between groups even though I don't expect this to be common)

doc/source/schema/schema.md
OriolAbril: yep, the docs already has mathjax loaded and supports the $ syntax

doc/source/schema/schema.md
OriolAbril: I'll update this then, it looks like using `inv_metric` would be best right? also feel free to suggest better definitions or editing it directly

doc/source/schema/schema.md
ahartikainen: That sounds like it should work

e.g https://mc-stan.org/docs/reference-manual/simplex-transform.html

cc @avehtari 

Aki do you remember if any constrained vs unconstrained transform had more complicated structure?

avehtari: > Aki do you remember if any constrained vs unconstrained transform had more complicated structure?

Correlation matrices, Covariance matrices, Cholesky factors of covariance matrices, and Cholesky factors of correlation matrices https://mc-stan.org/docs/reference-manual/variable-transforms.html

OriolAbril:
I would like to merge this before the holidays if someone can review 🙏🏿

ahartikainen (approved):
lgtm

doc/source/schema/schema.md
```diff
@@ -8,16 +8,19 @@ The `InferenceData` schema approach defines a data structure compatible with [Ne
...
   - [emcee](https://emcee.readthedocs.io/en/stable/)
-  - [PyMC3](https://docs.pymc.io)
+  - [PyMC](https://www.pymc.io)
   - [Pyro](https://pyro.ai/) and [NumPyro](https://pyro.ai/numpyro/)
-  - [PyStan](https://pystan.readthedocs.io/en/latest/index.html), [CmdStan](https://mc-stan.org/users/interfaces/cmdstan) and [CmdStanPy](https://cmdstanpy.readthedocs.io/en/latest/index.html)
+  - [PyStan](https://pystan.readthedocs.io/en/latest/index.html),
+    [CmdStan](https://mc-stan.org/docs/cmdstan-guide/index.html) and
+    [CmdStanPy](https://mc-stan.org/cmdstanpy/)
   - [TensorFlow Probability](https://www.tensorflow.org/probability)
 * [ArviZ.jl](https://github.com/arviz-devs/ArviZ.jl) in **Julia** which integrates with:
```
sethaxen, line 17:
```suggestion
* [InferenceObjects.jl](https://github.com/arviz-devs/InferenceObjects.jl) in **Julia** used in [ArviZ.jl](https://github.com/arviz-devs/ArviZ.jl), which integrates with:
```

doc/source/schema/schema.md
```diff
@@ -36,33 +39,79 @@ The terminology used in this specification is based on [xarray's terminology](ht
...
+  but they are all tied to a single model. The model identifier can be added as metadata
+  to simplify the calls to model comparison functions.
+* `created_at`: the date of creation of the group.
+* `creation_library`: the library used to create the InferenceData (might not necessarly be ArviZ)
+* `creation_library_version`: the version of `creation_library` that generated the InferenceData
+* `creation_library_language`: the programming language from which `creation_library` was used to create the InferenceData
+* `inference_library`: the library used to run the inference.
+* `inference_library_version`: version of the inference library used.
+
+Metadata can be stored at the whole `InferenceData` level but also at group level when needed.
```
sethaxen, line 70: Do we have any recommendations yet as to which metadata fields belong in groups vs `InferenceData`?

doc/source/schema/schema.md
```diff
@@ -111,6 +162,45 @@ Out of sample posterior predictive samples p(y'|y). Samples should match `poster
...
 
+:::{admonition} Note on sample stats, warmup and unconstrained groups
+:class: note, dropdown
+
+The schema does not define which warmup or unconstrained groups exist or can exist
+by default. We recognize both the samplers and the models are continuously evolving.
+Some models already require the use of sampling algorithms to get prior samples,
+in which case we basically need to treat the prior and posterior groups in the same way.
+
+We define the prefixes and to allow libraries that use InferenceData to be aware
```
sethaxen, line 173:
```suggestion
We define the prefixes to allow libraries that use InferenceData to be aware
```

sethaxen (approved):
LGTM, just a few minor (and optional) suggestions. Merge when you feel it's ready!

doc/source/schema/schema.md
```diff
@@ -36,33 +39,79 @@ The terminology used in this specification is based on [xarray's terminology](ht
 `InferenceData` stores all quantities that are relevant to fulfilling its goals in different groups. Different groups generally distinguish conceptually different quantities in Bayesian inference, however, convenience in {ref}`creation <creating_InferenceData>` and {ref}`usage <working_with_InferenceData>` of `InferenceData` objects also plays a role. In general, each quantity (such as posterior distribution or observed data) will be represented by several multidimensional labeled variables.
 
 ### Rules
-Following are a few rules which should be followed:
+Below are a few rules that should be followed:
 * Each group should have one entry per variable and each variable should be named.
-* When relevant, the first two dimensions of each variable should be the sample identifier (`chain`, `draw`).
-* For groups like `observed_data` or `constant_data`, the two initial dimensions(`chain`, `draw`) are omitted.
+* Dimension names `chain`, `draw`, `sample` and `pred_id` are reserved for
```
sethaxen, line 44: Oh, you know, `pred_id` is not defined anywhere. IIRC, we discussed using this when one needs multiple IID predictions per sample. Perhaps add a note about this to `posterior_predictive`?

doc/source/schema/schema.md
OriolAbril: not yet, I'll open an issue for this

doc/source/schema/schema.md
OriolAbril: will do, thanks for pointing it out

Issue #2076, opened 2022-07-25
Organize example gallery into sections and re-scale images


sarinac:
## Tell us about it

**Current State:**
The [ArviZ example gallery](https://python.arviz.org/en/latest/examples/index.html) currently lists plots in alphabetical order (based on API name), but there are a few critiques:
1. It gets difficult to browse or search by visual where there are ~50 graphs to browse through. This makes discovery pretty difficult (users might think a very specific chart doesn't exist in ArviZ because they can't find it easily). 
  a. Also, some charts (`az.plot_dist_comparison`) from the API Reference might be missing from the example gallery?
2. Some plots are hard to recognize by picture (e.g. "Compare Plot", "ELPD Plot", and "Quantile MCSE Error Plot" are too zoomed in or not cropped optimally).
3. The hover event is not helpful. For example, if I hover over the "Autocorrelation Plot" tile, I see the text  overlay "example_plot_autocorr". 

**Recommendation:**
1. Group all the current charts by use case. 
    *  Some great examples of galleries that do this are [matplotlib](https://matplotlib.org/stable/gallery/index#) and [python graph gallery](https://www.python-graph-gallery.com/). Both of these examples have a section header followed by relevant charts, so that it helps break up the information and makes it easier for the user to digest. For Matplotlib, you can browse sections using the right panel, which would also be great for ArviZ.
    * See bottom **Thoughts on implementation** section for a starting point on how I would section these, but I think it would great to get multiple opinions on the organization. 
2. Resize the images of example charts to fit in the thumbnail entirely (probably edit `doc/sphinxext/gallery_generator.py` file). Most charts in the example gallery should default to `figsize=[8, 6]` wherever makes sense (e.g. not "Separation Plot" because it makes more sense as a thin bar).
3. Change hover to show text from documentation or example from gallery. For example, if I hover over the "Autocorrelation Plot" tile, seeing the text "Bar plot of the autocorrelation function for a sequence of data" is more helpful. FYI Matplotlib does this with a tooltip instead of overlay. See screenshot below:
<img width="422" alt="image" src="https://user-images.githubusercontent.com/20257416/180714077-5da60f40-bc04-4da5-ba9d-94701364467b.png">

## Thoughts on implementation

1. Grouping charts by use case- are some sections I can think of but definitely open to other ideas. With my current approach, the "Distributions" section is pretty overloaded, and I'm not sure which section each graph belongs to.
    * `Distributions` - density, dist, dot plot, trace plots
[...]

PR #2087, opened 2022-08-07
Update layout for example gallery

sarinac:
## Description

Regarding issue #2076 : 
* Move CSS out of `gallery_generator.py` (all custom CSS should go in `custom.css`)
* Convert template `CONTENTS_ENTRY_TEMPLATE` from raw HTML to RST, so it follows sphinx template for creating grid-item-cards ([sphinx documentation](https://sphinx-design.readthedocs.io/en/latest/cards.html))
    * Reformatted toctree and contents a bit by adding `TOCTREE_START` and `CONTENTS_START`, and keeping those separate (instead of joining them together when inserting into `INDEX_TEMPLATE`).
    * **FYI:** Formatting was handled in the CSS. I inserted the `example-gallery` class in each card, so that the CSS only targets objects within that class (i.e. will only affect the Example Gallery page and not others if they are using cards).
* Resized images in example gallery. Instead of using thumbnails, it uses the charts generated by matplotlib.
    * **FYI:** Keeping the ratio intact while making sure images take up the same amount of real estate per card was handled in the CSS (`.example-gallery img.sd-card-img-top` and `.example-gallery .sd-card-body`). This isn't really the normal way to resize images but it works. The more normal would be to create a `<div>` container with fixed height and have the img fill up the container, but I don't think Sphinx has this flexibility.
    * Removed code for generating thumbnail and wherever else thumbnail code was without making the example gallery + subpages worse 🤞 

image

# Questions / Issues:
1. What is the purpose for `sphinxtag` ? I used to be the "example_chart_name" shown in the hover event, which isn't very useful as display text. We could remove it if it's not used.
2. Can I also remove all the `_thumb: .5, .5` related settings from the .py scripts (docstring) in `examples/matplotlib/*` and `examples/bokeh/*`? The gallery generator still has to remove them or else they will appear in the example pages.
3. 2 `_example_title`s in the example gallery are too long (see below). They are `plot_loo_pit_ecdf` and `plot_loo_pit_overlay`. I think there are 2 options:
    * **Option A:** Remove `_example_title` from everything and use the default title in each script. For example, "Plot LOO-PIT KDE overlaid on KDEs of uniform samples to assess predictive calibration." turns into "LOO-PIT Overlay Plot".
    * **Option B:** Change these 2 to a more concise name that can fit up to 3 lines. The card **must be set at a fixed height** (via CSS, which I set at 250px) or else the gallery will look uneven.
    * image

----
:books: Documentation preview :books:: https://arviz--2087.org.readthedocs.build/en/2087/

### DIFF
--- a/doc/source/_static/custom.css
+++ b/doc/source/_static/custom.css
@@ -29,7 +29,7 @@ html[data-theme="dark"] {
 
 .bd-content {
   flex-grow: 1;
-  max-width: 60em;
+  max-width: 70em; /* Override 60em default, can use 100% */
 }
 
 /* -------------------- HOMEPAGE -------------------- */
@@ -42,19 +42,10 @@ html[data-theme="dark"] {
   padding: 20px 0px 40px;
 }
 
-.home-img-plot {
-  width: 200px;
-  height: 120px;
-  display: flex;
-  justify-content: center;
-  align-items: center;
-  overflow: hidden
-}
-
 .home-img-plot,
 .home-img-plot-overlay {
-  width: 200px;
-  height: 120px;
+  width: 235px;
+  height: 130px;
   display: flex;
   justify-content: center;
   align-items: center;
@@ -83,11 +74,6 @@ html[data-theme="dark"] {
 }
 
 /* -------------------- EXAMPLE GALLERY OVERRIDES -------------------- */
-
-.bd-content .sd-card .sd-card-body,
-.bd-content .sd-card .sd-card-footer {
-  background-color: var(--pst-color-background);
-}
 
 /* Each card has same height (across all rows in the grid) */
 .sd-card.example-gallery {
@@ -129,11 +115,9 @@ html[data-theme="dark"] {
   font-size: 0.8rem;
 }
 
-/* -------------------- EXAMPLE PLOTS -------------------- */
+/* -------------------- EXAMPLE PLOTS - DOWNLOAD PYTHON SOURCE CODE -------------------- */
 
-/* Download Python Source Code */
 .example-plot-download {
-  background-color: #0fbcbc12;
   border: 0.05rem solid rgb(222, 222, 222);
   border-radius: 0.2rem;
   /* Center button */
@@ -142,14 +126,17 @@ html[data-theme="dark"] {
   box-shadow: 0 0.2rem 0.5rem rgb(0 0 0 / 5%), 0 0 0.0625rem rgb(0 0 0 / 10%);
 }
 .example-plot-download p a {
-  font-weight: 700;
-  font-size: 0.8rem;
-  text-align: center;
   padding: 0.4rem 2rem;
-  margin: 0;
 }
 .example-plot-download p { /* Override margin defaults for p */
   margin: 0;
+}
+.example-plot-download code.literal { /* Override defaults for code.literal */
+  border: 0px;
+  background-color: transparent;
+}
+.example-plot-download a.download:hover  { /* Override defaults for hover */
+  color: var(--pst-color-primary);
 }
 
 /* -------------------- TOCTREE OVERRIDES -------------------- */
--- a/doc/source/_static/custom.css
+++ b/doc/source/_static/custom.css
@@ -1,16 +1,87 @@
 .sphinx-codeautolink-a{
-    border-bottom-color: rgb(0, 0, 0);
-    border-bottom-style: dotted;
-    border-bottom-width: 1px;
+  border-bottom-color: rgb(0, 0, 0);
+  border-bottom-style: dotted;
+  border-bottom-width: 1px;
 }
 .sphinx-codeautolink-a:hover{
-    color: rgb(255, 139, 139);
+  color: rgb(255, 139, 139);
 }
 
 .supportbutton a {
- color: var(--pst-color-sidebar-link-active);
+color: var(--pst-color-sidebar-link-active);
 }
 
 .navbar-brand {
-  display: inline-block !important;
+display: inline-block !important;
+}
+
+
+/* -------------------- EXAMPLE GALLERY -------------------- */
+
+.example-gallery-figure {
+  position: relative;
+  float: left;
+  margin: 10px;
+  width: 180px;
+  height: 200px;
+}
+
+.example-gallery-figure img {
+  position: absolute;
+  display: inline;
+  left: 0;
+  width: 170px;
+  height: 170px;
+  opacity:1.0;
+  filter:alpha(opacity=100); /* For IE8 and earlier */
+}
+
+.example-gallery-figure:hover img {
+  -webkit-filter: blur(3px);
+  -moz-filter: blur(3px);
+  -o-filter: blur(3px);
+  -ms-filter: blur(3px);
+  filter: blur(3px);
+  opacity:1.0;
+  filter:alpha(opacity=100); /* For IE8 and earlier */
+}
+
+span.example-gallery-figure-label {
+  position: absolute;
+  display: inline;
+  left: 0;
+  width: 170px;
+  height: 170px;
+  background: #000;
+  color: #fff;
+  visibility: hidden;
+  opacity: 0;
+  z-index: 100;
+}
+
+.example-gallery-figure p {
+  position: absolute;
+  top: 45%;
+  width: 170px;
+  font-size: 110%;
+}
+
+.example-gallery-figure:hover span {
+  visibility: visible;
+  opacity: .4;
+}
+
+.example-gallery-figure .example-gallery-figure-title p {
+  position: relative;
+  top: 170px;
+  color: black;
+  visibility: visible;
+  text-align: center !important;
+  line-height: normal;
+}
+
+.example-gallery-figure .example-gallery-figure-title span {
+  top: 170px;
+  position: relative;
+  visibility: visible;
 }
--- a/doc/source/_static/custom.css
+++ b/doc/source/_static/custom.css
@@ -18,7 +18,7 @@ display: inline-block !important;
 /* -------------------- THEME OVERRIDES -------------------- */
 
 html[data-theme="light"] {
-  --pst-color-primary: rgb(0 192 191);
+  --pst-color-primary: rgb(11 117 145);
   --pst-color-secondary: rgb(238 144 64);
 }
 
@@ -44,6 +44,15 @@ html[data-theme="dark"] {
   padding: 20px 0px 40px;
 }
 
+.home-img-plot {
+  width: 200px;
+  height: 120px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  overflow: hidden
+}
+
 .home-img-plot,
 .home-img-plot-overlay {
   width: 200px;
@@ -65,7 +74,7 @@ html[data-theme="dark"] {
   position: absolute;
   border: 1px solid #dee2e6;  /* Same dimensions as img-thumbnail from pydata css */
   border-radius: 0.25rem;  /* Same dimensions as img-thumbnail from pydata css */
-  color: #fff;
+  color: var(--pst-color-background);
   opacity: 0;
   transition: .2s ease;
   text-align: center;
--- a/doc/source/_static/custom.css
+++ b/doc/source/_static/custom.css
@@ -15,18 +15,21 @@ color: var(--pst-color-sidebar-link-active);
 display: inline-block !important;
 }
 
-
 /* -------------------- EXAMPLE GALLERY OVERRIDES -------------------- */
 
-.example-gallery {
+.bd-content .sd-card .sd-card-body,
+.bd-content .sd-card .sd-card-footer {
+  background-color: transparent;
+}
+
+/* Each card has same height (across all rows in the grid) */
+.sd-card.example-gallery {
   height: 250px;
-  display: flex;
-  justify-content: space-between;
-  flex-direction: column;
   border-radius: 0.2rem;
   border: 0.6px solid rgba(255,255,255,0) !important;  /* Override Sphinx class */
 }
 
+/* Hover effects for each card */
 .example-gallery:hover {
   box-shadow: 0 0 0.5rem 0.2rem rgba(15, 187, 188, 0.3) !important;  /* Override Sphinx class */
   border: 0.6px solid #0fbbbc;
@@ -35,27 +38,22 @@ display: inline-block !important;
   color: #0fbbbc;
 }
 
-.example-gallery img.sd-card-img-top {
-  padding: 10px;
-  object-fit: contain;
-  max-height: 60%;
+/* Images in each card are scaled to fit 60% of the card */
+.example-gallery .sd-card-body {
+  height: 60%;
+  display: flex;  /* This vertically centers the img */
+}
+.example-gallery .sd-card-body img {
+  max-height: 100%;
   max-width: 100%;
-  flex-grow: 6;
+  margin: auto;  /* This keeps img ratio and horizontally centers the img */
 }
 
+/* Plot titles go in the footer and are aligned to fit 40% of the card */
+.example-gallery .sd-card-footer {
+  height: 40%;
+  width: 100%;
+}
 .example-gallery .sd-card-text {
   color: #086c86;
-}
-
-.bd-content .sd-card .sd-card-body,
-.bd-content .sd-card .sd-card-footer {
-  background-color: transparent;
-}
-
-.example-gallery .sd-card-body {
-  height: 40%;
-  flex-grow: 4;
-  position: absolute;
-  bottom: 0;
-  width: 100%;
 }
--- a/doc/source/_static/custom.css
+++ b/doc/source/_static/custom.css
@@ -15,33 +15,94 @@ color: var(--pst-color-sidebar-link-active);
 display: inline-block !important;
 }
 
+/* -------------------- THEME OVERRIDES -------------------- */
+
+html[data-theme="light"] {
+  --pst-color-primary: rgb(0 192 191);
+  --pst-color-secondary: rgb(238 144 64);
+}
+
+html[data-theme="dark"] {
+  --pst-color-primary: rgb(0 192 191);
+  --pst-color-secondary: rgb(238 144 64);
+}
+
+.bd-container .bd-container__inner {
+  justify-content: center;
+}
+.bd-content {
+  max-width: 60em;
+}
+
+/* -------------------- HOMEPAGE -------------------- */
+
+.home-flex-grid {
+  display: flex;
+  flex-flow: row wrap;
+  justify-content: center;
+  gap: 10px;
+  padding: 20px 0px 40px;
+}
+
+.home-img-plot,
+.home-img-plot-overlay {
+  width: 200px;
+  height: 120px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  overflow: hidden
+}
+
+.home-img-plot img {
+  flex-shrink: 0;
+  min-width: 100%;
+  min-height: 100%;
+}
+
+.home-img-plot-overlay {
+  background: var(--pst-color-primary);
+  position: absolute;
+  border: 1px solid #dee2e6;  /* Same dimensions as img-thumbnail from pydata css */
+  border-radius: 0.25rem;  /* Same dimensions as img-thumbnail from pydata css */
+  color: #fff;
+  opacity: 0;
+  transition: .2s ease;
+  text-align: center;
+}
+
+.home-img-plot-overlay:hover {
+  opacity: 90%;
+}
+
 /* -------------------- EXAMPLE GALLERY OVERRIDES -------------------- */
 
 .bd-content .sd-card .sd-card-body,
 .bd-content .sd-card .sd-card-footer {
-  background-color: transparent;
+  background-color: var(--pst-color-background);
 }
 
 /* Each card has same height (across all rows in the grid) */
 .sd-card.example-gallery {
-  height: 250px;
+  height: 200px;
+  padding: 0.2rem;
   border-radius: 0.2rem;
-  border: 0.6px solid rgba(255,255,255,0) !important;  /* Override Sphinx class */
+  border: 0.6px solid var(--pst-color-background); /* Same width as hover effect */
 }
 
 /* Hover effects for each card */
 .example-gallery:hover {
-  box-shadow: 0 0 0.5rem 0.2rem rgba(15, 187, 188, 0.3) !important;  /* Override Sphinx class */
-  border: 0.6px solid #0fbbbc;
+  border: 0.6px solid var(--pst-color-primary);
 }
 .example-gallery:hover p {
-  color: #0fbbbc;
+  color: var(--pst-color-primary);
 }
 
 /* Images in each card are scaled to fit 60% of the card */
 .example-gallery .sd-card-body {
-  height: 60%;
+  height: 70%;
   display: flex;  /* This vertically centers the img */
+  padding: 10px 10px 5px;
 }
 .example-gallery .sd-card-body img {
   max-height: 100%;
@@ -51,36 +112,16 @@ display: inline-block !important;
 
 /* Plot titles go in the footer and are aligned to fit 40% of the card */
 .example-gallery .sd-card-footer {
-  height: 40%;
+  height: 30%;
   width: 100%;
+  padding: 0px 10px 10px;
 }
 .example-gallery .sd-card-text {
-  color: #086c86;
+  color: var(--pst-color-text-muted);
+  font-size: 0.8rem;
 }
 
 /* -------------------- EXAMPLE PLOTS -------------------- */
-
-/* API Reference */
-.example-plot-api-container {
-  margin: 1.5em auto;
-  padding: 0 0.6rem 0.4rem;
-  border: 0.05rem solid rgb(222, 222, 222);
-  border-left: 0.2rem solid;
-  border-left-color: #0fbbbc;
-  border-radius: 0.2rem;
-  box-shadow: 0 0.2rem 0.5rem rgb(0 0 0 / 5%), 0 0 0.0625rem rgb(0 0 0 / 10%);
-}
-.example-plot-api-title {
-  background-color:#0fbcbc12;
-  padding: 0.4rem 0.6rem 0.4rem 1rem;
-  margin: 0 -0.6rem;  /* Same as padding from container */
-  font-weight: 700;
-  font-size: 0.8rem;
-}
-.example-plot-api-name p {
-  padding: 0.4rem 0.6rem 0.4rem 1.2rem;
-  margin: 0rem;
-}
 
 /* Download Python Source Code */
 .example-plot-download {
@@ -101,4 +142,15 @@ display: inline-block !important;
 }
 .example-plot-download p { /* Override margin defaults for p */
   margin: 0;
+}
+
+/* -------------------- TOCTREE OVERRIDES -------------------- */
+
+.toctree-wrapper a {
+  color: var(--pst-color-text-muted);
+}
+
+.toctree-wrapper a:hover {
+  text-decoration: none;  /* No underline */
+  color: var(--pst-color-primary);
 }
--- a/doc/source/_static/custom.css
+++ b/doc/source/_static/custom.css
@@ -29,7 +29,7 @@ html[data-theme="dark"] {
 
 .bd-content {
   flex-grow: 1;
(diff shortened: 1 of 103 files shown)


doc/sphinxext/gallery_generator.py
```diff
@@ -59,136 +59,33 @@
...
-    fig.savefig(thumbfile, dpi=dpi)
-    plt.close(fig)
+CONTENTS_START = """
+.. grid:: 3
+   :gutter: 2 2 3 3
+"""
+
+CONTENTS_ENTRY_TEMPLATE = """
+   .. grid-item-card:: 
+      :link: ./{htmlfilename}
```
OriolAbril, line 81: Is the htmlfilename used anywhere else? If we used the `sphinxtag` here together with `link-type: ref` I think we could use [sphinx-hoverxref](https://sphinx-hoverxref.readthedocs.io/en/latest/index.html) together with the [use of tooltips on all `ref` roles](https://sphinx-hoverxref.readthedocs.io/en/latest/usage.html#tooltip-on-all-ref-roles) might be a nice fix for the hover content

doc/sphinxext/gallery_generator.py
```diff
@@ -59,136 +59,33 @@
...
+      :img-top: ./matplotlib/_images/{imgfilename}
+      :text-align: center
+      :shadow: none
+      :class-card: example-gallery
+
+      {title}
+"""
```
OriolAbril, line 88: > FYI: Keeping the ratio intact while making sure images take up the same amount of real estate per card was handled in the CSS (.example-gallery img.sd-card-img-top and .example-gallery .sd-card-body). This isn't really the normal way to resize images but it works. The more normal would be to create a <div> container with fixed height and have the img fill up the container, but I don't think Sphinx has this flexibility.

I think this is possible. Cards in sphinx-design can add content in up to 3 separated divs (header, body, footer, the default being to use a single body div). I think these could be used for that. It is also possible to define divs (and their classes) as an rst directive, but that would be a bit more cumbersome. It looks like the issue here is that using img-top does not place the image inside a specific div. I think something like this would work:

```suggestion
CONTENTS_ENTRY_TEMPLATE = """
   .. grid-item-card:: 
      :link: ./{htmlfilename}
      :text-align: center
      :shadow: none
      :class-card: example-gallery
      :class-header: specific div size fixing class here
      
      .. image:: ./matplotlib/_images/{imgfilename}
         :alt: alt text would go here (optional for now, could be the description we discussed)
      
      ^^^

      {title}
"""
```

classes passed in `class-card` affect the outer card div, `class-header` classes affect only the content above the `^^^` and it is also possible to use `class-body` with classes that affect only the content below `^^^`.

OriolAbril (commented):
> Can I also remove all the _thumb: .5, .5 related settings from the .py scripts (docstring) in examples/matplotlib/* and examples/bokeh/*? The gallery generator still has to remove them or else they will appear in the example pages.

Yes, we can now remove all the thumb things. I might wait until we decide how to implement the subsections though. One option would be adding more folder hierarchy, but another could be using a `_category` element similar to thumb, in which case we could reuse the thumb reading code.

> 2 _example_titles in the example gallery are too long (see below). They are plot_loo_pit_ecdf and plot_loo_pit_overlay. I think there are 2 options

I think getting rid of the extended title and adding a description to all plots instead would be better. Moreover, if combined with the hoverxref thing, it won't really matter if the title in the example gallery is not clear enough, visitors will only need to hover to get some more details.

doc/source/_static/custom.css
```diff
@@ -1,16 +1,61 @@
...
 
 .navbar-brand {
-  display: inline-block !important;
+display: inline-block !important;
+}
+
+
+/* -------------------- EXAMPLE GALLERY OVERRIDES -------------------- */
+
+.example-gallery {
```
canyon289, line 21: Add comments as for what each blog it does, even if it seems obvious. The overlap betweeen bayesian expert and css experts is quite small

canyon289:
New gallery is looking nicer! For a first PR you picked a challenging one :)

image

doc/sphinxext/gallery_generator.py
```diff
@@ -59,136 +59,33 @@
...
-    plt.close(fig)
+CONTENTS_START = """
+.. grid:: 3
+   :gutter: 2 2 3 3
+"""
+
+CONTENTS_ENTRY_TEMPLATE = """
+   .. grid-item-card:: 
+      :link: ./{htmlfilename}
+      :img-top: ./matplotlib/_images/{imgfilename}
```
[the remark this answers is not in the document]
sarinac: This confused me a lot actually... I sort of guessed until it magically worked 😂 

If you open up the individual RST file, the image path is relative to that file. For example, `source/examples/index.rst` would need the image source as `./matplotlib/_images/*.png`. It will not match with the rendered page, which dumps all the images in `html/_images`.

doc/sphinxext/gallery_generator.py
sarinac: okay THANK YOU! This works better and now I can center align things easier with more legible CSS :)

sarinac:
Thanks for reviewing Oriol and Ravin! 

I think the things left to do for example gallery (listed in priority high to low) are:
* Remove extended title descriptions from .py files and use normal title for example gallery
* Remove `_thumb` from .py files, but comment out the thumb reading code in case we want to reuse for categories
* Add plot description (maybe using hoverxref)
    I haven't address this [comment](https://github.com/arviz-devs/arviz/pull/2087#discussion_r939648482) yet: 
    > Is the htmlfilename used anywhere else? If we used the sphinxtag here together with link-type: ref I think we could use [sphinx-hoverxref](https://sphinx-hoverxref.readthedocs.io/en/latest/index.html) together with the [use of tooltips on all ref roles](https://sphinx-hoverxref.readthedocs.io/en/latest/usage.html#tooltip-on-all-ref-roles) might be a nice fix for the hover content
* Section into categories (allowing overlapping)

The first 2 can be done tomorrow, but should the last 2 be in a separate PR? We currently don't have plot descriptions (nor a standard on how to write it) and not sure yet what is the best way to implement categories. With categories, if they exist as tags, I think keeping it in the individual example pages might be nice (screenshot below). 

image

doc/sphinxext/gallery_generator.py
OriolAbril: Glad I could help. As I said, I don't know much about css and no js. But I do know a lot of sphinx (using either rst or markdown) 😊

doc/sphinxext/gallery_generator.py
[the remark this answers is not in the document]
OriolAbril: that's true, sphinx moves all the images to `_images/` folder during the build process, independently of their original relative path inisde the source files, only the paths of the source files are preserved into the urls. Thanks for the reminder

OriolAbril:
> The first 2 can be done tomorrow, but should the last 2 be in a separate PR? We currently don't have plot descriptions (nor a standard on how to write it) and not sure yet what is the best way to implement categories. With categories, if they exist as tags, I think keeping it in the individual example pages might be nice (screenshot below).

I think either option is fine, whichever works best for you. But I also would not necessarly aim to update all the descriptions in a single PR. Once the structure of how this goes is defined we could merge the PR with 2-3 examples of how to write/add descriptions and progressively work on the others. The tags/subsections part should probably be done in one go instead, but I don't think it is so time intensive.

Re: tags, I would use level two headings for each section (to create separate grids for each category) and separate toctrees (all hidden but with different captions so they appear on the sidebar like the ones in https://python.arviz.org/en/latest/contributing/index.html

sarinac:
I made some minor CSS changes but I might change it later depending on what other info gets added (e.g. tags)... This screenshot below (plot_loo_pit_ecdf) is the only one with a description. The descriptions only need to be in the matplotlib file.
image
...
image

doc/source/_static/custom.css
```diff
@@ -1,16 +1,104 @@
...
+}
+
+/* Hover effects for each card */
+.example-gallery:hover {
+  box-shadow: 0 0 0.5rem 0.2rem rgba(15, 187, 188, 0.3) !important;  /* Override Sphinx class */
+  border: 0.6px solid #0fbbbc;
+}
+.example-gallery:hover p {
+  color: #0fbbbc;
+}
```
OriolAbril, line 39: Not sure what are the colors being used and where, I think it's the "culprit" is the `p` one, but commenting over all the style just in case. 

The text in the grid cards is rendered in blue and then switches to a lighter shade of blue when hovering, which I find a bit annoying for the following two reasons:

* neither the base blue nor the lighter "on hover" blue match the link color of the theme, the one in the `edit this page` text for example. I think at least one of the two should be using the theme defined colors: https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/customizing.html#color-variables.
* That behaviour doesn't match the theme behaviour. I have noticed 3 possible cases relative to the theme:
  - Text is black/text base color and becomes blue on hover (navbar and left sidebar behaviour)
  - Text is blue and keeps the same color but becomes underlined on hover (right sidebar behaviour, the donate is custom added and doesn't match the theme either and we should probably fix that too, it should not be considered an example here though)
  - Text is blue and becomes pink and underlined on hover (behaviour on body and toctrees, see https://arviz--2087.org.readthedocs.build/en/2087/user_guide/index.html)

The second point is not that important, the similar yet slightly differnet shades of blue I think should be changed.

OriolAbril (commented):
left two extra styling comments. after that we can merge and work on description and tags in a follow-up PR

OriolAbril:
Just realized that before merging we need to update the paths/styles of the showcased images in the homepage: [current](https://python.arviz.org/en/latest/) vs [preview from this PR](https://arviz--2087.org.readthedocs.build/en/2087/) and in the [readme](https://github.com/arviz-devs/arviz#gallery) that would otherwise break too

doc/source/_static/custom.css
sarinac: Thanks for pointing this out. I wasn't exactly sure which color should be used but it's sounds like we should ultimately keep it consistent. 
* [Color] The blue I put in the custom CSS was based on the teal in the ArviZ logo, but since the theme is pydata I could use those predefined colors. Any reason to keep the original theme colors instead of recoloring the blues to fit the ArviZ logo? 
* [Hover] I can take a look at these events and default the example gallery tile hover back to orange underline to keep consistency. LMK if you have a difference preference. I am personally not a huge fan of blue -> orange because it's a bit generic with data visualizations 😂 
* [General] How custom do you want the example gallery and overall website? Can you point me to any inspirations (what aspect you like from other docs) that you would like the docs to look like (or not look like) in a future state?

sarinac: Probably getting distracted, but on the topic of theme/homepage, are you open to different layouts? This mockup below (from pasting screenshots together) uses up real estate more efficiently. And some things I prefer are:
* Title header should go before search bar (So I think ideally it should be H1 > brief description > search bar)
* Example charts can go below. Currently the example charts are shown in the right, but there is nothing else on the right panel. This point is super minor.
* Having more headers (installation, contribution) helps split up the bodies of text! 
* Looking at other data viz documentations, they do not have "Installation/Contribution" details in their homepage even though it seems important here. Instead they have links on where to go ([see bokeh example](https://docs.bokeh.org/en/latest/index.html)). I'm not sure which direction is better... maybe keep what is currently on this page and add a section (before "Installation") on where to start (e.g. example gallery for ideas, QuickStart page, etc.)? 
image

OriolAbril: > [Color] The blue I put in the custom CSS was based on the teal in the ArviZ logo, but since the theme is pydata I could use those predefined colors. Any reason to keep the original theme colors instead of recoloring the blues to fit the ArviZ logo?

We can also update the predefined colors! I didn't mean that the color from the examples needs to be the one currenly used in the navbar hover, only that they should be the same. 

We can use a custom color scheme defined in our own `custom.css` (or `colors.css`, we can load multiple css files during the doc build). I think we'd only need to follow this: https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/customizing.html#base-variables, then the arviz teal color would be the highlight color for navbar links on hover and so on, but there would not be multiple shades of blue in similar roles.

From what I understand the theme uses two colors mainly, blue and pink in addition to the black/white text (depending on the mode being light or dark) and uses those two for hover effects and highlights and so on. The black to blue is black to primary color I think, the blue to pink is primary to secondary color, I don't really understand the always blue that only becomes underlined, it might not be configurable 🤔 

> [Hover] I can take a look at these events and default the example gallery tile hover back to orange underline to keep consistency. LMK if you have a difference preference. I am personally not a huge fan of blue -> orange because it's a bit generic with data visualizations

We already have at least 3 different hover effects _for text_ all of which indicate clickable links, I am not tied to any of the existing effects, but if possible I'd prefer to not add another hover text effect that also indicates link. 

> [General] How custom do you want the example gallery and overall website? Can you point me to any inspirations (what aspect you like from other docs) that you would like the docs to look like (or not look like) in a future state?

Not really sure how to answer to that, especially the first part. I'd say probably as customizable as possible without taking over a large maintenance burden. That is, we should take advantage of theme customizations and use custom css, we could even create a "theme extension" (like https://github.com/matplotlib/mpl-sphinx-theme) so we can use it in ArviZ, [preliz](https://preliz.readthedocs.io/en/latest/), we could also change the [xarray-einstats](https://einstats.python.arviz.org/en/latest/) one. However, creating our theme from scratch or forking it and renouncing to the improvements that would come with future releases to create a custom "forzen" theme would be a no-go; similarly, I could probably be convinced to add some js but I'd need a lot of convincing.

As for inspirations, here are some gallery pages we could get inspiration from:

* https://matplotlib.org/stable/gallery/index.html
  - Very similar to the current grid style, the images are a bit too small in my opinion, the 3 columns is better, I'll also try to get rid of the right sidebar probably. I'm still on the fence about having text under the image. I do not like the tooltip it has neither in style nor in content. IMO the tooltip should be short and direct, it should not have "this example shows" (the same way alt text should not have "an image of"). I also think it should not be cut half way and use `...`, for the style, on the fence about the tooltip appearing to the right, but I don't like the combination of still needing to read the example title (which is shown in orage over a white background when hovering) together with the tooltip description in white text over a black background.
* https://se.mathworks.com/products/matlab/plot-gallery.html
  - The images are also too little, not sure if still considered a tooltip, but I like the click on the image that generates a pop-up with the image in a larger size.
* http://seaborn.pydata.org/examples/index.html
  - (I think our example gallery was adapted originally from this one, or both were adapted from the same source) I like the no text at all option and perfectly square images, but the cropped square images I think are a case of aesthetics overriding functionality so I would not do/go back to this. But we might want to consider no-text options like here (or more like the example below) and have the text both title and description (as I said I'm partial to bigger images) appear only on hover.
* https://r-graph-gallery.com/density-plot.html
  - I like the style of this gallery, big images and little text, I do not like the hover content nor the hover style though. Given how all examples use viridis colormap, I find the orange to clash too much and also to not have enough contrast with the white text (which becomes even worse because it is also a bit transparent). There is no way this passes the accessibility color contrast checks. You proposes something similar on the issue too, if is is possible to get ti working with colors that allow reading the text easily then we could look into something like this. 
  - Not the thing of the example, but using the "modal style" that https://sphinx-hoverxref.readthedocs.io/en/latest/ supports will probably be simpler to implement (and not depend on us) and also look good, as the "opened" preview would show both description and larger plot, even have working links from the hover window/tooltip.
* https://altair-viz.github.io/gallery/index.html
  - too small images and too small example grid area. This is close to what I don't want the example gallery to look like.

OriolAbril: > are you open to different layouts?

yes!


>  Title header should go before search bar (So I think ideally it should be H1 > brief description > search bar)

The search bar should be gone. It is currently there because we believe all the pages should have the searchbar accessible somewhere, and it didn't fit in the navbar. The next version of the theme solves that, the search bar will be an icon on the navbar that expands/pops up when clicked on: https://pydata-sphinx-theme.readthedocs.io/en/latest/ (towards the right of the navbar, left of the links)

> Example charts can go below. Currently the example charts are shown in the right, but there is nothing else on the right panel. This point is super minor.

Sounds good, if we put them on a single line they could even be a carrousel: https://sphinx-design.readthedocs.io/en/sbt-theme/cards.html#card-carousels

> Looking at other data viz documentations, they do not have "Installation/Contribution" details in their homepage even though it seems important here. Instead they have links on where to go ([see bokeh example](https://docs.bokeh.org/en/latest/index.html)). I'm not sure which direction is better... maybe keep what is currently on this page and add a section (before "Installation") on where to start (e.g. example gallery for ideas, QuickStart page, etc.)?

We could look into restructuring this. I am also no convinced by the layout nor content but lack ideas on what to change too. 

A general note though. I don't think any text section should be as wide as the navbar. There is a reaso why the theme doesn't take all the available screen and why even if we remove the sidebar it doesn't expand to take all the width by default (see the theme page: https://pydata-sphinx-theme.readthedocs.io/en/latest/, no sidebars yet same width). It makes the text harder to read, bokeh does it and the elevator pitch right above the search bar is a bit hard to follow for me, but we can definitely change the layout. the matplotlib homepage for example: https://matplotlib.org/ is wider than the navbar yet the text sentences are all restricted to at most half the width.

doc/sphinxext/gallery_generator.py
```diff
@@ -30,17 +29,29 @@
...
+
+    <div class='example-plot-api-container'>
+        <p class='example-plot-api-title'>API Reference</p>
+        <div class='example-plot-api-name'>
+        
+{api_name}
+
+.. raw:: html
+
+    </div></div>
```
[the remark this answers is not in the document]
sarinac: Thanks for showing me this! I'm so used to editing with CSS that I forget/don't realize (truthfully, don't know 😅) I could do this with Sphinx. 

I'm going to propose a 3rd option which is to use [seealso] (https://www.sphinx-doc.org/en/master/extdev/nodes.html#sphinx.addnodes.seealso) which is already used in the [API Reference pages](file:///Users/sarinachen/Desktop/arviz/doc/build/html/api/generated/arviz.plot_dot.html). Seems like the "See Also" sections are meant to link to other sections of ArviZ documentation, so I think this is fitting. 
```
.. seealso::

    API Documentation: {api_name}
```

sarinac:
Here's a summary of what changes based on follow-ups:
* **[Colors]** Updated CSS primary color to use Arviz blue. I think the website has the same consistent blue (primary) and orange (secondary, didn't change). I think the pink you were mentioning may have only been applicable for syntax highlighting? 
* **[Hover]** I cleaned up some of the hover effects, but there are still multiple types. I have an explanation for this design:
    * Option 1: Black text default -> Blue text when hovering: visible in TOC (didn't change), index.html files (used to be option 2 below), and example gallery tiles (updated in this commit). This is used for links that are obvious to users (e.g. TOC, See next example, AKA things that belong to some structure); most links default to Option 2 but too many links would look a little annoying.
    * Option 2: Blue text default -> underlined orange text when hovering: visible for inline links and a few other places. This is used for links that are less obvious to users.
    * Option 3: image -> blue overlay with white text when hovering: only in homepage. 
        * I know you mentioned not really liking this example in the R-Graph-Gallery due to colors/lack of contrast, but would that also apply here? If we regenerate the example plots with a different style (e.g. remove gray background, matching blue), then would that change your mind, or do you also not like this style of hover due to the text blocking the image?
        * Here's an example of some implementation where hover will not block the image, in case what I have currently doesn't work out. 
          image
        * The alternative example I have would be to create a tooltip that appears outside of the image tile (maybe on top, white background, gray border, black text), and with only title as the text, it'll be less distracting than the one matplotlib has.
    * Other thoughts: I am not a fan of the sphinx hover extension for images in the example gallery because it feels bulky/excessive. However, I'm open to it for 
* **[Homepage]** I redid the layout a bit by adding `h2` headers to some things, which matches the style for the rest of the website. I personally think this is clean enough (has good spacing out for content). I also changed the images to use `flex-grid` which is responsive. I don't think a carousel is necessary because all these graphs can fit comfortably on the page, and they're pretty uniquely different. The carousel will hide some of those images (even though it saves on real estate), but I don't think that's really a net positive.
    * Just a note, the homepage images don't have text because I think each graph here is uniquely different. However, I don't think the example gallery is ready yet for the no-text approach. Some graphs like versions of the parallel plot are too similar that it'll be frustrating for the user to click on the right example on their first try - we should tag/categorize the example gallery before hiding text if we go with that approach. Will keep in mind that larger images/little text is sort of end goal though.
* **[Admonition Box]** I changed this to `seealso` since it already exists + in somewhat consistent with how API reference uses this. Plus, we are not 100% sure what else will go here so I think this is the lowest maintenance approach.

OriolAbril:
> I know you mentioned not really liking this example in the R-Graph-Gallery due to colors/lack of contrast, but would that also apply here? If we regenerate the example plots with a different style (e.g. remove gray background, matching blue), then would that change your mind, or do you also not like this style of hover due to the text blocking the image?

I like this hover effect, and prefer it to the other alternatives. I don't care that the text blocks the image because the whole image was visible and quite big already before the hover.

 However I do have a hard time reading the text shown in the hover here too. I also would have sweared the link overlays were pink 🤷🏿 , so I am not one to be trusted a lot when it comes to colors, but in this case we should probably follow the recommendations like https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=00C0BF when it comes to color contrast in elements we expect visitors to read. Maybe we could make the primary color a bit darker or the darkest in the logo to get better contrast. I find it easier to read the titles than the hovers, but the contrast checks seem to be symmetric and I find it even easier to read the regular text (that was already an issue with the theme preset colors, but while we are at it it would be nice to fix) 

We can use the current color for the dark theme where it does pass the contrast checks: https://webaim.org/resources/contrastchecker/?fcolor=00C0BF&bcolor=000000.

> Just a note, the homepage images don't have text because I think each graph here is uniquely different. However, I don't think the example gallery is ready yet for the no-text approach. Some graphs like versions of the parallel plot are too similar that it'll be frustrating for the user to click on the right example on their first try - we should tag/categorize the example gallery before hiding text if we go with that approach. Will keep in mind that larger images/little text is sort of end goal though.

I am not sure everyone agrees with me though! And they already look very good, having the title in small text is good too, especially for now I think it is great. I would stop here and only revisit/continue once we have the sections and descriptions that could make potential hover text...

---

Can you use `pydata-sphinx-theme @ git+https://github.com/pydata/pydata-sphinx-theme` in https://github.com/arviz-devs/arviz/blob/main/requirements-docs.txt#L7? Before merging we should update and change to `>=0.10.0` or `>=0.10.0.dev0` (if we merge before the stable release is published which seems likely) but this will already allow checking how things look with the "incoming" version of the theme (already pre-released).

OriolAbril:
From the error message, the CI failures look unrelated, something to do with httpstan probably, @ahartikainen do you recognize a quick fix?

sarinac:
Thanks @OriolAbril! 
I updated the colors. White + dark blue and Teal + Black does look better like you said, although I think the inline links are a little less subtle now (tradeoffs, but I think it's fine 😅). 
For updating the package version for pydata-sphinx-theme, can you show me how to do it? I am not good with anything setup related, so I don't trust myself with making changes correctly... I do think there are going to be a few CSS things that'll change. The necessary change will probably only be fixing the download button (which will look fine if I remove all the [custom CSS related to it](https://github.com/arviz-devs/arviz/blob/ec5b867c8394d46ddb6c8f22fd666523959c1536/doc/source/_static/custom.css#L133-L155)).

doc/source/_static/custom.css
```diff
@@ -1,16 +1,164 @@
...
+
+html[data-theme="dark"] {
+  --pst-color-primary: rgb(0 192 191);
+  --pst-color-secondary: rgb(238 144 64);
+}
+
+.bd-content {
+  flex-grow: 1;
+  max-width: 60em;
```
OriolAbril, line 32: maybe we can try 70em? it feels a bit too narrow still, especially in the example gallery with the sidebar taking extra space too

Using 100% as recommended in https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/layout.html#horizontal-spacing would also be great, but then we'd need to restructure the page to avoid having text that is too wide.

doc/source/_static/custom.css
```diff
@@ -1,16 +1,164 @@
...
+.home-img-plot-overlay:hover {
+  opacity: 90%;
+}
+
+/* -------------------- EXAMPLE GALLERY OVERRIDES -------------------- */
+
+.bd-content .sd-card .sd-card-body,
+.bd-content .sd-card .sd-card-footer {
+  background-color: var(--pst-color-background);
+}
```
OriolAbril, line 90: what is this part doing?

doc/source/_static/custom.css
```diff
@@ -1,16 +1,164 @@
...
+  max-width: 100%;
+  margin: auto;  /* This keeps img ratio and horizontally centers the img */
+}
+
+/* Plot titles go in the footer and are aligned to fit 40% of the card */
+.example-gallery .sd-card-footer {
+  height: 30%;
+  width: 100%;
+  padding: 0px 10px 10px;
+  border-top: none !important;
```
OriolAbril, line 125: I added this because with the new version of pydata-sphinx-theme, the footer and body of the cards are separated by a thin grey line which I didn't like. There are probably better ways to go about this though.

Ref: 
* footer has separation line: https://arviz--2087.org.readthedocs.build/en/2087/community.html#blogs-and-example-compilations
* no separation line: https://python.arviz.org/en/latest/community.html#blogs-and-example-compilations

doc/source/index.rst
```diff
@@ -1,13 +1,106 @@
-.. raw:: html
-
-    <form class="bd-search align-items-center" action="search.html" method="get">
-      <input type="search" class="form-control search-front-page" name="q" id="search-input" placeholder="&#128269; Search the docs ..." aria-label="Search the docs ..." autocomplete="off">
-    </form>
+:html_theme.sidebar_secondary.remove:
```
OriolAbril, line 1: this is what removes the right sidebar from the page, it needs to be set on a page by page basis.

doc/source/index.rst
```diff
@@ -1,13 +1,106 @@
...
+
+        <a href="examples/plot_autocorr.html">
+            <div class="home-img-plot img-thumbnail">
+                <img src="./_images/mpl_plot_autocorr.png">
+                <span class="home-img-plot-overlay">Autocorrelation Plot</span>
+            </div>
+        </a>
+    </div>
+
 |pypi|
```
OriolAbril, line 104: I would keep the badges right under the title or get rid of them. Their main (only?) goal is that you can get an idea about the state of the library at first glance (CI passing, good coverage, published on pypi, has a doi...), if it is not the first/2nd thing you see I think it looses most of the use.

It is also something very dev/contributor based, so it might be good to keep them on the readme but not on the homepage.

doc/source/index.rst
```diff
@@ -1,13 +1,106 @@
...
+
+ArviZ is a Python package for exploratory analysis of Bayesian models. Includes functions for posterior analysis, data storage, sample diagnostics, model checking, and comparison.
+
+The goal is to provide backend-agnostic tools for diagnostics and visualizations of Bayesian inference in Python,
+by first converting inference data into `xarray <https://xarray.pydata.org/en/stable/>`_ objects.
+See :ref:`here <xarray_for_arviz>` for more on xarray and ArviZ usage
+and :ref:`here <schema>` for more on ``InferenceData`` structure
+and specification.
+
+.. raw:: html
```
OriolAbril, line 16: note: I like the wider layout with the images on the right over this one. But I think it is probably best to merge this for now, and have a follow up pr for the homepage and another for the examples/tags/descriptions. There are many things we should take into account to update the homepage (i.e. sponsors should be there! like we do on our [project page](https://www.arviz.org/en/latest/#sponsors-and-institutional-partners))

OriolAbril (commented):
This PR is making amazing changes, thanks! The main question right now I think is fixing a couple nits like the 60/70em and merging already so visitors see the improved example gallery or continuing to work here. I think it is probably best to merge.

doc/source/_static/custom.css
sarinac: Nice! This works. I noticed the line too but didn't have much of an opinion on whether to include it. Now I'm glad it's not included.

doc/source/_static/custom.css
sarinac: I could have sworn this had a purpose at some point... going to remove 🤷‍♀️

OriolAbril (approved):
🚀

OriolAbril:
I am merging because I have no clue about how to fix CI but it should not block @sarinac's work.

PR #2095, opened 2022-08-15
Labellers testing module

### LINKED ISSUE #1574: Add tests to arviz.labels module
#1201 has done a significant refactor of plotting functions to rewrite the labeling process. Part of the code added is tested in the sense that plots use it and tests for plots have passed, but there should be a `test_labels.py` in the base tests folders to test all bundled classes and test specifically the labellers.

agustinaarroyuelo:
Co-authored-by: Oriol Abril-Pla <oriol.abril.pla@gmail.com>

## Description
Here we will add a new testing module specific for the labeller classes. It will fix #1574. 

## Checklist

----
:books: Documentation preview :books:: https://arviz--2095.org.readthedocs.build/en/2095/


arviz/tests/base_tests/test_labels.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self):
+            self.sel= {
+                "instrument": "a",
+                "experiment": 3,
+            }
+            self.isel= {
+                "instrument": 0,
+                "experiment": 4,
+            }
+
```
OriolAbril, line 22:
```suggestion
class Data():
    def __init__(self):
            self.sel= {
                "instrument": "a",
                "experiment": 3,
            }
            self.isel= {
                "instrument": 0,
                "experiment": 4,
            }

@pytest.fixture(scope="module")
def multidim_sels():
    return Data()
```

I think this will be the valid pytest pattern to get the fixtures to work and to do what we want them to do. Ref: [@pytest.fixture(scope="module")](https://docs.pytest.org/en/6.2.x/fixture.html#what-fixtures-are). Depending on how we organize things we can also set autouse on this one (maybe the labellers one below too): https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request.

arviz/tests/base_tests/test_labels.py
```diff
@@ -0,0 +1,70 @@
...
+        # for now simply skip it
+    
+        self.labellers = {"BaseLabeller": BaseLabeller(), 
+                          "DimCoordLabeller": DimCoordLabeller(),
+                          "IdxLabeller": IdxLabeller(),
+                          "DimIdxLabeller": DimIdxLabeller(),
+                          "MapLabeller": MapLabeller(),
+                          "NoVarLabeller": NoVarLabeller(),
+                          "NoModelLabeller": NoModelLabeller(),
+                          "mix_labellers": mix_labellers()}
```
OriolAbril, line 38: similar comment, in pytest, classes whose methods are tests we want to be executed need to start with `Test` (like test functions/methods need to start with `test_`) and are not initialized nor use `self`, they are a way to organize tests in order to limit the scope of a fixture (see https://docs.pytest.org/en/6.2.x/fixture.html#fixture-availability) or to group multiple tests that all have the same parametrize toghether (see https://stackoverflow.com/questions/50016862/grouping-tests-in-pytest-classes-vs-plain-functions or https://github.com/arviz-devs/xarray-einstats/blob/main/tests/test_stats.py#L126). 

I think I copied the idea in the issue from a test class in arviz so it came with the pylint disable and no self in the call signature instead of having self on the call signature but not using it in the test/function. From what I understand, the way to make some pre-prepared data available to tests is adding a method to the class decorated with the fixture thing, not initializing the test class with attributes.

agustinaarroyuelo:
Thanks for your comments @OriolAbril. I cleaned up the script trying to get a minimum working unit. If we agree on it, we can start adding more tests.

arviz/tests/base_tests/test_labels.py
```diff
@@ -0,0 +1,60 @@
...
+    "args",
+    [
+        ("BaseLabeller", "theta\na, 3"),
+    ],
+)
+class TestLabellers:
+    def test_make_label_vert(self, args, multidim_sels, labellers):
+        labeller_arg = labellers.labellers[args[0]]
+        label = labeller_arg.make_label_vert("theta", multidim_sels.sel, multidim_sels.isel)
+        assert label == args[1]
```
OriolAbril, line 60: very minor nit, for readability I (aka present me who seems a bit wiser than old me) would probably unpack the `args` into something like `name, expected_label = args` and then use those inside the test instead of `args[0]`.

ahartikainen (approved):

PR #2096, opened 2022-08-15
Use arviz_example_data submodule

sethaxen:
## Description

Replaces stored example data with arviz_example_data based on https://github.com/arviz-devs/arviz_example_data/pull/6. Note: this should not be merged until that PR is merged!

## Checklist

- [x] Follows [official](https://github.com/arviz-devs/arviz/blob/main/CONTRIBUTING.md#pull-request-checklist) PR format
- [ ] New features are properly documented (with an example if appropriate)?
- [x] Includes new or updated tests to cover the new feature
- [x] Code style  correct (follows pylint and black guidelines)
- [x] Changes are listed in [changelog](https://github.com/arviz-devs/arviz/blob/main/CHANGELOG.md#v0xx-unreleased)

----
:books: Documentation preview :books:: https://arviz--2096.org.readthedocs.build/en/2096/

PR #2105, opened 2022-08-27
Replace example data submodule with subtree

sethaxen:
## Description

As discussed on Slack, this replaces the git submodule approach with including example data in this repo (#2096) with an approach using git subtree. This makes management slightly more complicated on the developer side but makes it easier whenever one needs to download the repo, as special commands are no longer necessary to clone the repo. See https://stackoverflow.com/questions/31769820/differences-between-git-submodule-and-subtree for some comparisons between the two approaches.

## Checklist


ahartikainen (approved):

sethaxen:
We should document somewhere how developers should update the subtree of arviz_example_data is updated. Where's a good place for that?

sethaxen:
@ahartikainen thoughts on this question?

OriolAbril:
I would say some of the pages in https://python.arviz.org/en/latest/contributing/index.html. Maybe a new one in how-to guides?

sethaxen:
> I would say some of the pages in https://python.arviz.org/en/latest/contributing/index.html. Maybe a new one in how-to guides?

Done! https://arviz--2105.org.readthedocs.build/en/2105/contributing/updating_example_data.html

Unless anything on that page needs to change, this should be ready to merge.

Issue #2100, opened 2022-08-17
navbar now puts the two last elements inside a "more" field


OriolAbril:
The navbar used to have: Getting Started, Example Gallery, User Guide, API Reference, Community, Contributing, About US

now it has Getting Started, Example Gallery, User Guide, API Reference, Community, More (which unfolds and contains Contributing, About US) (see https://python.arviz.org/en/latest/).

There is enough space on the navbar, so I think we should show all elements or alternatively make use of the "more"+dropdown feature. I think the first option is better, it can be done following https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/header-links.html#navigation-bar-dropdown-links

sarinac:
Totally agree with option #1. I noticed this recently but wasn't sure how to change it so thanks for linking the how-to.

In general I lean on fewer tabs, so I would not consider option #2. Other thought is that is the "About" (Arviz.org redirect) necessary on the nav bar? I don't know if users click on that often compared to the other options, so removing it could declutter. Is it possible to place a) within the home/community page b) left panel below the table of contents c) footer?
<img width="82" alt="image" src="https://user-images.githubusercontent.com/20257416/186949529-e589f9fc-eadd-41ac-a070-4e2949d83aab.png">

OriolAbril:
If we find a good place for it on the homepage and we add it to the footer we could skip that part. But even if very few people click on it I think it is important to have and would personally prefer to keep it on the navbar. It may not be many people but we especially want the people interested in learning more about the project and its members to be able to find more info, find our personal websites/blogs/courses...

Issue #2039, opened 2022-06-02
R hat not showing in plot_forest?


cluhmann:
**Describe the bug**
When `plot_forest` is called with pymc-created InferenceData and `r_hat=True`, a subplot is created but the R hat values are not visible.

**To Reproduce**
```python
import pymc as pm
import arviz as az
import matplotlib.pyplot as plt
with pm.Model() as model:
    a = pm.Normal("a")
    obs = pm.Normal("obs", mu=a, observed=[2])
    idata = pm.sample()
print(az.summary(idata))
az.plot_forest(idata, r_hat=True)
plt.show()
```

The summary looks ok:

```python
    mean     sd  hdi_3%  hdi_97%  mcse_mean  mcse_sd  ess_bulk  ess_tail  r_hat
a  0.983  0.678  -0.296    2.241       0.02    0.014    1139.0    1436.0    1.0
```
But no R-hat is visible:
![Figure_1](https://user-images.githubusercontent.com/39170995/171563065-01cfb2d7-02fc-4ec7-aeea-83fb4745ef92.png)

Linux
arviz     : 0.12.1
pymc      : 4.0.0b6

OriolAbril:
I think it has to do with https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/matplotlib/forestplot.py#L646 (and same/similar repeated line in bokeh which is broken too) and its interaction with the `skip_dim`/`combine_dim` changes. Will have to track it down better and see how to fix it

Alexander-Serov:
I am having the same issue with pymc 3.11.4, unexpected indeed because I had the rhat values plotted before for more or less the same code.

I was wondering whether someone has been able to find a hotfix? Do you think downgrading to a previous version could work for the moment?

Also, just curious, @cluhmann what did you use to print the summary?

cluhmann:
> what did you use to print the summary?

It's in the example: `print(az.summary(idata))`

aloctavodia:
@ndmlny-qs This is the issue I mentioned today. Essentially we should have as many "point-intervals" as chains, but only one ESS and R-hat per parameter.  The issue is that the `value` variable that is passed to `_ess` and `_rhat` is a 1d vector https://github.com/arviz-devs/arviz/blob/main/arviz/plots/backends/matplotlib/forestplot.py#L635-L648 instead i.e. the chains have been split. 

[...]

PR #2108, opened 2022-09-02
Update gallery style

sarinac:
## Description

This PR contains some items from the conversation in #2076 : 
1. Add `arviz-doc` style with the goal of applying it to the example gallery, other parts of documentation, and having it generally available. First 2 colors are ArviZ blues. [All colors are here and are distinguishable under all colorblind types](https://coolors.co/107591-00c0bf-f69a48-fdcd49-8da798-a19368-525252-a6761d-7035b7-cf166e). I also copied some settings from the existing arviz styles, like fontsize. 
2. Update CSS in example gallery and refactoring so that homepage/gallery share same attributes. As a result, the example gallery CSS will have more in common with the homepage.
3. Split up charts in example gallery into categories.

### Some questions:
1. `arviz-doc` style: can we proceed with applying this for the example gallery? Any style-related concerns (e.g. color, grid, figsize)?
    a. Is there even a point in adding a default `figsize`? It seems to get overwritten a lot, and I'm not sure if it's worth setting the figsize in all example charts to make them relatively symmetric / same size in the example gallery card.
2. CSS: 
    a. Keep the hover for example gallery (Plot Title + function)? If not, then what text should it have instead?
    b. Any concerns with the border around chart for example gallery and homepage (they have consistent styles now)? FYI, the homepage in this PR does not stretch the chart image unlike in the current website (subtle difference). 
    c. Concerns around font size in example gallery or overall documentation? 
3. Keep or change this categorization approach? Each MPL chart will have 1 category as `_gallery_category` (default to "Miscellaneous" if none or not in a predefined list of categories). I created separate `toctree` directives for each category, but I don't know how to link the toctree title to the Category header if that is possible 🙈 If **keep**, then I'll continue categorizing/styling other charts and think of what else to add. If **change**, then I'll need some suggestions on categorizing or making the example gallery easier to navigate. 

@OriolAbril @canyon289


canyon289:
I have never been more excited for a scrollover transition

image

>  a. Is there even a point in adding a default figsize? It seems to get overwritten a lot, and I'm not sure if it's worth setting the figsize in all example charts to make them relatively symmetric / same size in the example gallery card.

Yes set a reasonable default, if people override it its fine. On your end dont stress out too hard about what it should be. Just pick something

> FYI, the homepage in this PR does not stretch the chart image unlike in the current website (subtle difference).

Thanks for not this change

> but I don't know how to link the toctree title to the Category header if that is possible

I don't fully understand what the practical difference would be here.


## Questions

image

Can the text such as trace_plot_rank bars be bigger? Is there a way for us to specify a custom text field here to make these nicer? for instance I think that would be better as Rank Bars Diagnostic with KDE

doc/sphinxext/gallery_generator.py
```diff
@@ -92,12 +93,27 @@
...
+         {overlay}
+
       .. image:: ./matplotlib/{pngfilename}
 
       +++
       {title}
 """
-#:img-top: ./matplotlib/{pngfilename}
+
+CATEGORIES = [
```
canyon289, line 106: These categories are fine. We can change them ourselves now that you've created the framework

doc/sphinxext/gallery_generator.py
```diff
@@ -216,25 +232,27 @@ def extract_docstring(self):
...
             # if m:
-            #     thumbloc = float(m.group(1)), float(m.group(2))
-            #     continue
-            # m = re.match(r"^_example_title: (.*)$", line)
-            # if m:
-            #     title = m.group(1)
-            #     continue
+                # thumbloc = float(m.group(1)), float(m.group(2))
+                # continue
+            m = re.match(r"^_gallery_category: (.*)$", line)
```
canyon289, line 244: Regex! Nice

doc/sphinxext/gallery_generator.py
```diff
@@ -216,25 +232,27 @@ def extract_docstring(self):
         # thumbloc = None
         # title: Optional[str] = None
         ex_title: str = ""
+        gallery_category = None
         for line in docstring.split("\n"):
-            # # we've found everything we need...
+            # we've found everything we need...
             # if thumbloc and title and ex_title != "":
```
canyon289, line 238: I wonder if we can delete all this old stuff now

doc/sphinxext/gallery_generator.py
```diff
@@ -216,25 +232,27 @@ def extract_docstring(self):
...
             if ex_title == "":
                 ex_title = line
         assert ex_title != ""
-        # if thumbloc is not None:
-        #     self.thumbloc = thumbloc
-        #     docstring = "\n".join([l for l in docstring.split("\n") if not l.startswith("_thumb")])
+        self.gallery_category = "Miscellaneous"  # Default to category-less
+        if gallery_category is not None:
+            self.gallery_category = gallery_category
+            docstring = "\n".join([l for l in docstring.split("\n") if not l.startswith("_gallery_category")])
```
canyon289, line 255: Add docstring. While a clever one liner, id make this its own method that says "get docstring" or something similar so its more clear what's going on

doc/sphinxext/gallery_generator.py
```diff
@@ -269,6 +287,7 @@ def contents_entry(self) -> str:
             pngfilename=self.pngfilename,
             sphinx_tag=self.sphinxtag,
             title=self.title,
+            overlay="{0}{1}".format(self.title, (" using " + self.apiname) if self.apiname != "No API Documentation available" else ""),
```
canyon289, line 290: Same thing with overlay. Create a dedicated method in the class.

You are a one liner pro though, I am so impressed by your ability to cram so much logic in 80 characters

doc/sphinxext/gallery_generator.py
```diff
@@ -342,12 +357,33 @@ def main(app):
...
+    contents = ""
+
+    # Sort and write toctree + contents
+    for category, entries in categorized_contents.get("toctree").items():
+        if len(entries) > 0:
+            toctree += TOCTREE_START.format(category=category)
+            entries.sort()
+            for entry in entries:
+                toctree += entry
+    for category, entries in categorized_contents.get("contents").items():
```
canyon289, line 377: Add docstring

examples/matplotlib/mpl_plot_density.py
```diff
@@ -1,12 +1,13 @@
 """
 Density Plot
 ============
+_gallery_category: Distributions
```
canyon289, line 4: Great metadata label

canyon289:
The end product overall looks great.

Currently my questions largely are around developer documentation for python script, and how to change elements of the html. Specifically please add very human readable comments in the .py file of

1. Where the hover over text comes from
2. Where the text under each card comes from
3. Where the category comes from

Codewise as I commented it would be better to even separate these string constructors into their own method on the class so its both really obvious, and if in the future we'd like to change the behavior there is a clear place to do so

image

doc/sphinxext/gallery_generator.py
sarinac: Awesome. I'll just add the rest of what Oriol recommended. I think I'll do a scan of each chart as I transition them into a category + `arviz-doc` style. That way it's easier to have critique.

sarinac:
Thanks @canyon289! 

Re: developer documentation, I added a README in the `docs/examples/` directory. I think when people add additional charts in the future, they can go here for the template/instructions.

Re: overriding text in the hover overlay, I added a `overlay_desc` metadata label, but 1) I'm not a fan of it and 2) it requires some additional work for me to get the CSS right (for the code part). Below is a screenshot of what I think you're asking for. Can't you rename the entire Plot Title as "Rank Bars Diagnostic with KDE" (having it appear in the subtitle + hover) and get rid of "Traceplot rank_bars" entirely? What's the value of having plot title different from hover desc? Maybe another example will help. 
image

Re: link the toctree title to the Category header - the practicality would be that if I clicked on the "Distributions" title in the table of contents left panel, I would go to https://arviz--2108.org.readthedocs.build/en/2108/examples/index.html#distributions . However, it's not clickable because I'm not good at sphinx D:

doc/sphinxext/gallery_generator.py
canyon289: Thank you

canyon289:
> Re: developer documentation, 

Thanks that helps. We should still make each string generator its own method on `ExampleGenerator` here. That way its clear where to make changes in the future if need be, and will make it easier to add more logic if we need to as well.

https://github.com/arviz-devs/arviz/pull/2108/files#diff-418ec90c0ae46004a8bad3fb47bc3ef633606ebf7bc2851de58d4a20c14c018bL107

> Re: overriding text in the hover overlay,

Ah the extra complexity does not seem worth it. Ignore my ask.

> Re: link the toctree title to the Category header

Lets open an issue and note this as a nice to have for the future so its not blocking this PR

From a git perspective it would be great to merge this sooner to avoid the possibility of merge conflicts given this is a wider PR. Whenever you need me or oriols opinion aggressively @ us so we can help you get this completed

OriolAbril:
> Is there even a point in adding a default figsize? It seems to get overwritten a lot, and I'm not sure if it's worth setting the figsize in all example charts to make them relatively symmetric / same size in the example gallery card.

The longer term goal is to use the same style in all the documentation pages. I would set a figsize that looks good for notebook generated pages. In the example gallery we should also probably try to make sure both figsize and plot structure are not too thin nor wide so the plot looks good as a thumbnail.

> documentation wise

I would avoid as much as possible both comments and readme. I think ideally all this would be documented in a "how to write scripts for the example gallery" (similar to the current readme) that would go in the "how-to guides" section of [contribuing]() and an "about the example gallery" page that would go in the "in depth explanations" (with why did we chose categories and why these ones, why that appearance for the cards... 

If we add this to the docs, it will be visible to anyone going to the "contributing" page, it will be easier to link from other pages and it will also be searchable. We can keep the readme for people who prefer browsing on github as a 1-2 sentence doc with a link to the relevant pages in the documentation.

> Keep the hover for example gallery (Plot Title + function)? If not, then what text should it have instead?

I would have the hover be a 1-2 sentence description of the plot, the same as the alt text in the image once we go to the example page. We'd have to check how the hover interacts with the alt text to see if the images in the example gallery should have also the same alt text or have empty alt text.

Note: If we come up with a better hover text or we prefer to use the current one for example we don't need to use that. But we still need a parameter or some way to write the alt text for each plot. The alt text can no longer be the path to the image, before that it would even be better to leave it empty.

> but I don't know how to link the toctree title to the Category header if that is possible

This is not possible with sphinx, at least for now. I'll keep an eye out. The reason for this is that the "Distribution" title is a page heading and page headings appear on the right sidebar (which we removed), the left sidebar has the file toctree/hierarchy. In https://arviz--2108.org.readthedocs.build/en/2108/contributing/index.html for example you can see how the right sidebar has welcome, before you begin, communication... and the left sidebar has all the specific pages inside the contributing section.

---

I am not sure what is going on in the latest previews of the docs but the spacing is not working properly. The content is now too thin, especially on mostly text pages. This doesn't happen in the `latest` tag and we should not merge the PR while this is happening in the previews.

doc/sphinxext/gallery_generator.py
```diff
@@ -186,6 +214,14 @@ def sphinxtag(self):
     def pagetitle(self):
         return self.docstring.strip().split("\n")[0].strip()
 
+    @property
+    def overlay_description(self):
+        return self.title + (
+            " using {apiname}".format(apiname=self.apiname)
```
OriolAbril, line 220: Important note. The `apiname` property is currently text that is rendered as a link which means that now the hover text has an unreachable link. This is not very problematic for mouse users but breaks keyboard navigation as each card has two focus elements, the card itself and an invisible extra focus that should not be there. If we kept this hover text we would have to fix this.

OriolAbril (commented):
I skimmed the code very fast and left only one comment for now. I'll try to review better soonish.

re merging fast. I am not too worried about this as none of the files are very active and I don't expect them to be edited anytime soon.

OriolAbril (commented):
Comments on categories:

### Distributions
* https://arviz--2108.org.readthedocs.build/en/2108/examples/plot_forest.html is a distribution comparison plot, it is comparing the restults of the same variable for two different models. `plot_forest` should have an example in this section but using only one model. Similarly, `plot_forest` should have an example in the mixed plots section but there for the case where rhat and ess are used.
* Most `plot_pair` examples should be here. Potential exceptions are:
  * One example using divergences in the diagnostic section
  * One example showing how to overlay plot_pairs in distribution comparison section
* The single sided violin plot example is a distribution comparison one
* `plot_density` and `plot_ecdf` should have examples here
* We should probably rethink which 2d KDE examples we want to show there

### Distribution comparison
* The joint plot chart should be in the distribution section. We are trying to make sense of a multidimensional distribution, not comparing two distributions between them.
* The double sided violin plot should go in distributions (swapped with the single sided violin plot)

### Diagnostics
* Not sure we need 4 examples of `plot_parallel`

### Regression timeseries
I am not sure this title is descriptive or correct. Both regression and time series have some simiarities in the sense that we want to plot distributions along an independent variable, be it x, time or something else, so these plots can be used for regression or for time series.

### Model comparison
I would use comparison singular as it is more common in the field.

### Model validations
If we use validation I would also use singular instead of plural, but it might be better to use `checking` or `criticism`, `validation` might hint to cross validation which is used for model checking but more commonly used for model comparison.

sarinac:
Thanks for the review @OriolAbril ! Appreciate the explanations too.

> plot_forest should have an example in this section but using only one model.

I could add this tomorrow (assuming it's just removing one of the datasets from the current `mpl_plot_forest` example)

> Similarly, plot_forest should have an example in the mixed plots section but there for the case where rhat and ess are used. 

I need help with this one... there's no example in the API reference either

> Most plot_pair examples should be here. Potential exceptions are:
One example using divergences in the diagnostic section
One example showing how to overlay plot_pairs in distribution comparison section

Can you be more specific with the existing plots? I assumed anything with the `divergences=True` argument meant that it used divergence and belonged in diagnostic section. That includes `mpl_plot_pair`, `mpl_plot_pair_kde`, `mpl_plot_pair_kde_hdi`, and `mpl_plot_pair_hex`. I'm gonna guess you mean everything in that list that's not `mpl_plot_pair` will go in Distributions... 

> plot_density and plot_ecdf should have examples here

I added 2 more charts for these. Also, it seems like `plot_ecdf` was missing from documentation entirely, so I updated that too if that's okay.

> We should probably rethink which 2d KDE examples we want to show there

I think the 2D KDE + 2D KDE (Custom style) is redundant, so can we remove the former? The latter seems more useful. Sticking to two plots seems okay with me.

> Not sure we need 4 examples of plot_parallel

Agreed. I vote for keeping regular `pair_plot` and optionally 1 of the other 3, whichever is more used.

> [Regression Timeseries] I am not sure this title is descriptive or correct. 

Okay, I made it "Regression or Time Series" because I don't know a better title. Other option is to throw it back in Miscellaneous until we have more examples.

OriolAbril:
> I could add this tomorrow (assuming it's just removing one of the datasets from the current mpl_plot_forest example)

I think that would be a good start, then we can see if we show some more variables, make it `combined=False`, things like that, but as a starting point yes.

> I need help with this one... there's no example in the API reference either

I think it boils down to `az.plot_forest(..., ess=True, rhat=True)` but it probably won't work yet due to #2039 

> Can you be more specific with the existing plots? I assumed anything with the divergences=True argument meant that it used divergence and belonged in diagnostic section. That includes mpl_plot_pair, mpl_plot_pair_kde, mpl_plot_pair_kde_hdi, and mpl_plot_pair_hex. I'm gonna guess you mean everything in that list that's not mpl_plot_pair will go in Distributions...

Yes, I think only one of the `plot_pair` cases should use `divergences=True` and go in "diagnostics", the rest should not show divergences (I think it is the default) and should go in "Distributions". 

> I added 2 more charts for these. Also, it seems like plot_ecdf was missing from documentation entirely, so I updated that too if that's okay.

Thanks!

> I think the 2D KDE + 2D KDE (Custom style) is redundant, so can we remove the former?

go for it!!

> Agreed. I vote for keeping regular plot_parallel and optionally 1 of the other 3, whichever is more used.

Let's keep only the first one and move the others to the api examples.

> Okay, I made it "Regression or Time Series" because I don't know a better title. Other option is to throw it back in Miscellaneous until we have more examples. [Update: it's currently showing up in Misc, I'll fix tomorrow]

I would keep it even if it only has 3-4 plots as it is very popular. I can also add some quick and raw examples for multiple functions that could serve as starting point for further proper examples.

sarinac:
In PR:
* Made the chart edits + recategorized  - LMK if I missed anything
* Cleaned up some CSS -> I could redo the homepage next using the cleaned up CSS
* Made the navbar fit all 7 items (instead of previously 5) from #2100 
* Added logo for dark mode

Other:
* I don't know how to resize the bokeh "images" (more like html divs + canvas with absolution position & dimensions) so large plots like [autocorr](https://arviz--2108.org.readthedocs.build/en/2108/examples/plot_autocorr.html) continue to look awkward... I haven't used Bokeh before, so I'm not sure what's a quick way to fix this.
* None of the charts besides matplotlib styles have alt text, but feel free to add later.

> I am not sure what is going on in the latest previews of the docs but the spacing is not working properly. The content is now too thin, especially on mostly text pages. This doesn't happen in the latest tag and we should not merge the PR while this is happening in the previews.

I'm still not seeing this in the non-latest versions (have no idea how to fix that either right now). @OriolAbril - Are you still seeing that, and can you link/share screenshot?

doc/sphinxext/gallery_generator.py
```diff
@@ -186,6 +215,20 @@ def sphinxtag(self):
...
 
+    @property
+    def overlay_description(self):
+        if self._alt_text != "":
+            return self._alt_text
+        return self.title + (
+            " using `{apitext}`".format(apitext=self.apitext) if self.apitext != "" else ""
+        )
+    
+    @property
```
canyon289, line 226: Nice use of property

doc/sphinxext/gallery_generator.py
```diff
@@ -186,6 +215,20 @@ def sphinxtag(self):
     def pagetitle(self):
         return self.docstring.strip().split("\n")[0].strip()
 
+    @property
+    def overlay_description(self):
+        if self._alt_text != "":
+            return self._alt_text
+        return self.title + (
```
canyon289, line 222: Make everything .format, no need for + and .format together

OriolAbril:
> I'm still not seeing this in the non-latest versions (have no idea how to fix that either right now). @OriolAbril - Are you still seeing that, and can you link/share screenshot?

current theme (from latest)

imatge

old look (from stable)

imatge

Both screenshots are full screen pictures and the current one is significantly thinner.

I can get something like this if I unset this css element:

imatge

It can be seen from the text that the 2nd and 3rd screenshots fit "All ArviZ plotting functions and some stats functions can take an optional labeller argument. By default, labels" whereas the first i cut at "By".

However, seeing how this parameter I am unsetting is new, it probably means this was an intentional change by the theme developers. Maybe it is better for accessibility to have shorter lines, so I think we can keep it for now. We probably should not change theme elements without a reason for doing so.

> I don't know how to resize the bokeh "images" (more like html divs + canvas with absolution position & dimensions) so large plots like [autocorr](https://arviz--2108.org.readthedocs.build/en/2108/examples/plot_autocorr.html) continue to look awkward... I haven't used Bokeh before, so I'm not sure what's a quick way to fix this.

no need to fix this now or in the near future.

> None of the charts besides matplotlib styles have alt text, but feel free to add later.

we can add this progressively and even organize a sprint or working session to get help from the community.

OriolAbril (approved):

OriolAbril:
If there is nothing else you want to modify in this PR I think we can merge @sarinac

sarinac:
Ready to merge if it looks good on your end :D Thanks Oriol and Ravin for the review! 

---
For container width, it looks like they purposely made it smaller and made the `.bd-article-container` class. https://github.com/pydata/pydata-sphinx-theme/blob/main/src/pydata_sphinx_theme/assets/styles/sections/_article.scss#L18-L21 

From the linked [explanation](https://www.w3.org/WAI/tutorials/page-structure/styling/#line-length):
> A line of text shouldn’t be longer than 80 characters. This helps users with certain reading or visual disabilities that have trouble keeping their place when reading long lines of text. If the width of the text container is resized, it should be allowed to scale in a way so 80 characters or less are shown. An easy way to do this is using max-width: 70em in CSS. Another way to ensure good line-length is to use a fluid layout or apply responsive web design techniques that allow the user to resize the window to find a comfortable line length.

Pydata theme likes to set font size at 15px (browser default is 16px), so I think that's why they set it at 60em instead of 70em. 

---

Also, 2 Bokeh things to flag which I didn't really look into besides matching the matplotlib example code (not sure if intentional but can create separate issues): 
* Trace plot with circular variables doesn't look circular... https://arviz--2108.org.readthedocs.build/en/2108/examples/plot_trace_circ.html 
* ELPD plot isn't showing points https://arviz--2108.org.readthedocs.build/en/2108/examples/plot_elpd.html

OriolAbril:
Can you open two issues for the bokeh things you mention?

When we added circular things bokeh didn't support polar coordinates plots, so they were implemented only in matplotlib,  we could now fix this because bokeh now does support polar coordinates.

Elpd one is clearly a bug,  I don't have any clues to why this is happening though.

sarinac:
Thanks for merging 🎉 

Opened an issue for circular trace plot, but it looks like ELPD works fine in the actual Example Gallery (strange 🤷‍♀️)

I-Mohiuddin:
Maybe this is the wrong place to post this, but  can we update the basic mpl kws  in 
 arviz/arviz/plots/backends/matplotlib/lmplot.py.

This is such a common operation in modeling and the baseline kws are so visually jarring that I honestly find it  preferable to just do the whole thing in matplotlib manually. 

For example:
y_model_mean_kwargs.setdefault("color", "y") is a  really poor choice in my opinion.  

I think it would be much more aesthetically pleasing if we changed the baseline kws to have the mean line as black 
the y_model fill and the y hat fill to follow some basic color gradient (similar to the the bayesplot color schemes in R).  

Similarly I think setting the marker face color to none, making the marker  size a bit smaller and setting the marker edge color to black would do a lot for improving the utility of the function.

ColCarroll:
Hi! Yes, I moved this to a new issue where it can be discussed.

PR #2113, opened 2022-09-11
Fix README table column width to resize images

sarinac:
After changing the example gallery image sizes (no thumbnail, change `figsize`), the README doesn't look so great right now 😅 Forgot to update while working on this https://github.com/arviz-devs/arviz/pull/2108 

----
:books: Documentation preview :books:: https://arviz--2113.org.readthedocs.build/en/2113/

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,9 +56,9 @@ python setup.py install
 
 -------------------------------------------------------------------------------
 ## [Gallery](https://python.arviz.org/en/latest/examples/index.html)
-
+<style>table#gallery td {width: 25%;}</style>
 <p>
-<table>
+<table id="gallery">
 <tr>
 
   <td>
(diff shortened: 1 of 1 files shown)


OriolAbril:
When I go to https://github.com/sarinac/arviz/tree/patch-1 this is what I see:

imatge

sarinac:
I did not realize [CSS does not work in github markdown](https://github.com/orgs/community/discussions/22728) ): 
So instead, I rearranged the plots to have similar sized images in the same row. I had to make it 3 columns because density plot is wider than the normal figsize, and the only charts with similar dimension were pair plots (felt like duplicating pair plots would be excessive). I also removed autocorrelation plot because I can't find other charts that are equally as long. 

This is what I wanted it to look like using CSS (using local editor) but alas :') 
image

OriolAbril (approved):
Now it looks good, thanks!

PR #2121, opened 2022-09-23
Allow numpyro converter to work with namedtuples

ColCarroll:
Small change to allow numpyro's converter to also grab names from a namedtuple (the current behavior will name variables of a named tuple `Param:0`, `Param:1`, ...) 

----
:books: Documentation preview :books:: https://arviz--2121.org.readthedocs.build/en/2121/


fehiepsi (approved):
Thanks, Colin!

PR #2122, opened 2022-09-26
Swapped the netCDF4 dependency to h5netcdf

varchasgopalaswamy:
## Description

Addresses https://github.com/arviz-devs/arviz/pull/2029 and https://github.com/arviz-devs/arviz/issues/2028 by swapping out the netCDF4 dependency to h5netcdf, which should be a lighter dependency, since it avoids needing the netCDF4 C binaries. 

## Checklist

----
:books: Documentation preview :books:: https://arviz--2122.org.readthedocs.build/en/2122/


ahartikainen:
I think you need to define `engine`

https://docs.xarray.dev/en/stable/user-guide/io.html

We probably should define `engine` in rcParams so that could be changed if needed?

cc @OriolAbril

requirements.txt
```diff
@@ -5,6 +5,6 @@ scipy>=1.8.0
 packaging
 pandas>=1.4.0
 xarray>=0.21.0
-netcdf4
+h5netcdf==1.0.2
```
ahartikainen, line 8:
```suggestion
h5netcdf>=1.0.2
```

ahartikainen (commented):
Couple of comments

arviz/data/inference_data.py
```diff
@@ -23,8 +23,13 @@
     Union,
     overload,
 )
+try:
+    import h5netcdf.legacyapi as nc
+    HAS_H5NETCDF = True
+except ImportError:
+    import netCDF4 as nc
+    HAS_H5NETCDF = False
```
ahartikainen, line 31: I'm thinking of this, and given that we require this, then we don't need to have this fail-safe check here.

So I think we can just remove this check and also use the main API?

CHANGELOG.md
```diff
@@ -12,7 +12,7 @@
   ([2096](https://github.com/arviz-devs/arviz/pull/2096) and [2105](https://github.com/arviz-devs/arviz/pull/2105))
 * Bokeh kde contour plots started to use `contourpy` package ([2104](https://github.com/arviz-devs/arviz/pull/2104))
 * Update default Bokeh markers for rcparams ([2104](https://github.com/arviz-devs/arviz/pull/2104))
-
+* Changed dependency on netCDF4 to h5netcdf
```
ahartikainen, line 15:
```suggestion
* Changed dependency on netCDF4 to h5netcdf  ([2122](https://github.com/arviz-devs/arviz/pull/2122))

```

arviz/data/inference_data.py
```diff
@@ -370,6 +375,8 @@ def from_netcdf(filename, group_kwargs=None, regex=False) -> "InferenceData":
                     for key, kws in group_kwargs.items():
                         if re.search(key, group):
                             group_kws = kws
+                if HAS_H5NETCDF:
```
ahartikainen, line 378: setdefault can be left, just this check can be removed?

arviz/data/inference_data.py
```diff
@@ -423,11 +430,13 @@ def to_netcdf(
...
                 if compress:
                     kwargs["encoding"] = {var_name: {"zlib": True} for var_name in data.variables}
+                if HAS_H5NETCDF:
+                    kwargs['engine'] = 'h5netcdf'
                 data.to_netcdf(filename, mode=mode, group=group, **kwargs)
                 data.close()
                 mode = "a"
         else:  # creates a netcdf file for an empty InferenceData object.
-            empty_netcdf_file = nc.Dataset(filename, mode="w", format="NETCDF4")
+            empty_netcdf_file = nc.Dataset(filename, mode="w")
```
OriolAbril, line 439: Is the format always netcdf4 with h4netcdf?

OriolAbril (commented):
Sorry about the late review. I don't know much about the differences between the two libraries. Can you develop a bit on the differences and advantages of h5netcdf over netcdf4 package? Seeing `.legacy` in the import makes me a bit nervous.

theorashid:
Hey, just to let you know the netcdf4 import is currently failing on the latest python image. So on that particular image, I can't install arviz.

netcdf4 is always a pain to install, so having a more reliable dependency would be great

arviz/data/inference_data.py
varchasgopalaswamy: I certainly don't have any objections! The docs for h5netcdf suggest that it should be a drop-in replacement for netCDF4 except in some edge cases that I don't think get used here.

arviz/data/inference_data.py
varchasgopalaswamy: Hi @OriolAbril, 

The basic reason for this is to remove the dependency on the netCDF4 C library. Since netCDF4 is a layer on top of HDF5, you could imagine using a basic HDF5 library in python to implement the netCDF4 standard, instead of calling netCDF4 C routines. `h5netcdf` does this using the `h5py` library. 

I find installing the netCDF4 C binaries to be quite painful, especially when I need to support linux, windows, intel OSX, Silicon OSX, etc. On Windows, there are some times issues with binary incompatibilities between the HDF5 libraries installed by `h5py` and `netCDF4` on `conda` Switching to using `h5py` as the backend would make the install significantly more straightforward. I think most scientific python users probably already have `h5py` installed, so practically I would imagine this wouldn't actually be a new dependency in those cases. 

I see where you're coming from with something called `legacy` showing up in an import - but I believe that the legacy API you saw allows `h5netcdf` to be used as a 'drop-in' replacement for `netCDF4` while the 'new' API looks more like `h5py`, rather than this being some kind of deprecated support issue.

> Is the format always netcdf4 with h4netcdf? 

No idea.

varchasgopalaswamy: Also, FWIW `xarray` seems to fully support `h5netcdf` as a backend. That's an appeal to authority, but it does seem to be ok for them!

arviz/data/inference_data.py
varchasgopalaswamy: Done!

OriolAbril:
I have rebased to fix the git conflicts and done some changes to allow either library as io engine. I think that makes sense as imo long term we shouldn't have either as a dependency and allow users to install whichever they one (or none if they don't use that subset of features).

Thanks for your patience, I'll merge in the following days.

OriolAbril (approved):

PR #2131, opened 2022-10-06
Save InferenceData attrs

### LINKED ISSUE #2109: Bug: to_netcdf does not save attrs
When I store meta data in `attrs` and call `to_netcdf()` and then load it subsequently, the meta-data is lost (`attrs` is empty). Is this expected? Is there a format to save this in so that they stay?

ahartikainen:
closes #2109 

## Description
Save attrs to root level

## Checklist

### DIFF
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -363,11 +363,11 @@ class InferenceData(Mapping[str, xr.Dataset]):
         InferenceData object
         """
         groups = {}
+        attrs = {}
 
         try:
             with nc.Dataset(filename, mode="r") as data:
                 data_groups = list(data.groups)
-                attrs = {}
 
             for group in data_groups:
                 group_kws = {}
@@ -378,13 +378,13 @@ class InferenceData(Mapping[str, xr.Dataset]):
                         if re.search(key, group):
                             group_kws = kws
                 with xr.open_dataset(filename, group=group, **group_kws) as data:
-                    if group == "__netcdf4_attrs":
-                        attrs.update(data.load().attrs)
+                    if rcParams["data.load"] == "eager":
+                        groups[group] = data.load()
                     else:
-                        if rcParams["data.load"] == "eager":
-                            groups[group] = data.load()
-                        else:
-                            groups[group] = data
+                        groups[group] = data
+
+            with xr.open_dataset(filename) as data:
+                    attrs.update(data.load().attrs)
 
             return InferenceData(attrs=attrs, **groups)
         except OSError as err:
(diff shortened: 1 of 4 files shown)


ahartikainen:
@OriolAbril @canyon289 you know netCDF4 better than I do, what should be the correct way to save the attrs to the main file (not in group)?

canyon289:
Is this way working? If so I assume its the correct way :D. If its working perhaps all we need is a test and were good to go

sethaxen:
Does the attributes variable name need to be netCDF4-specific? Can't it also be saved with e.g. Zarr?

ahartikainen:
@sethaxen very good question. It should.

Also, I think the main netcdf4 format includes a possibility for attr field at the root, but did not find a good/correct way to add it. There might be a way to do this in the docs, but atleast I did not find it.

I wonder if we actually want to create the dataset at the init and then just wrap attrs to point that?

And have one variable that will contain all the groups which will then be write/load normally?

And we just add special tag into our schema about it?

Or is this someting that Datatree will handle when it is released.

arviz/data/inference_data.py
```diff
@@ -412,6 +415,10 @@ def to_netcdf(
             Location of netcdf file
         """
         mode = "w"  # overwrite first, then append
+        if self._attrs:
+            xr.Dataset(attrs=self._attrs).to_netcdf(filename, mode=mode, group="__netcdf4_attrs")
```
OriolAbril, line 419:
```suggestion
            xr.Dataset(attrs=self._attrs).to_netcdf(filename, mode=mode)
```

xarray is able to both read and write datasets that only have attributes, and if no group is provided then things are saved at the root level which is what we want here.

I have double checked and I can also use netcdf4 library directly to get the saved attributes correctly with `ncattrs`, so I think it should be possible to do the same in the other languages. 

Moreover, xarray is also able to read//write datasets that only have attrs to the root level using `to_zarr` and `open_zarr`, so I think using this will also simplify the code

ahartikainen: If this works, great news.

I tried and just failed with netcdf4 lib, but I didn't dataset. I will neeed to try again.

ahartikainen:
Ok, still need to figure out the zarr.

OriolAbril:
LGTM, not sure why netcdf tests are failing though would have to test locally 🤔

ahartikainen:
They fail because `az.load_arviz_data("centered").to_netcdf(..)` fails.

Is this the object error we see even when it should not touch the compression?

OriolAbril:
> They fail because az.load_arviz_data("centered").to_netcdf(..) fails.

I can confirm this fails locally for me with `0.13.0rc1` on a new env, but works with `main`. I think tests will pass now, but there might still be some issues when using netcdf. Not sure there is anything we can do on our side though.

ahartikainen:
Yeah, text handling is always a bit hard.

I'm not sure if xarray has changed the defaults already to `S` / `U` dtypes, but then working with these values is more or less not great.

Maybe we could add some autotransform for object type strings if we can be certain it works correctly, but tbh it is a slippery slope...

OriolAbril (approved):

Issue #2079, opened 2022-07-28
RuntimeError when writing to a file with to_netcdf


plants-mom:
**Describe the bug**

Trying to write an `InferenceData` object to a file using `to_netcdf`results in an error:
`RuntimeError: NetCDF: Filter error: bad id or parameters or duplicate filter`

**To Reproduce**

```
import arviz as az
data = az.load_arviz_data("centered_eight")
data.to_netcdf("test.nc")
```
<details>
<summary> Stacktrace </summary>

``` 
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Input In [5], in <cell line: 1>()
----> 1 data.to_netcdf("test.nc")

File ~/.local/lib/python3.9/site-packages/arviz/data/inference_data.py:427, in InferenceData.to_netcdf(self, filename, compress, groups)
    425 if compress:
    426     kwargs["encoding"] = {var_name: {"zlib": True} for var_name in data.variables}
--> 427 data.to_netcdf(filename, mode=mode, group=group, **kwargs)
    428 data.close()
    429 mode = "a"

File ~/.local/lib/python3.9/site-packages/xarray/core/dataset.py:1901, in Dataset.to_netcdf(self, path, mode, format, group, engine, encoding, unlimited_dims, compute, invalid_netcdf)
   1898     encoding = {}
   1899 from ..backends.api import to_netcdf
-> 1901 return to_netcdf(
   1902     self,
   1903     path,
   1904     mode,
   1905     format=format,
   1906     group=group,
   1907     engine=engine,
   1908     encoding=encoding,
   1909     unlimited_dims=unlimited_dims,
   1910     compute=compute,
   1911     invalid_netcdf=invalid_netcdf,
   1912 )

File ~/.local/lib/python3.9/site-packages/xarray/backends/api.py:1072, in to_netcdf(dataset, path_or_file, mode, format, group, engine, encoding, unlimited_dims, compute, multifile, invalid_netcdf)
   1067 # TODO: figure out how to refactor this logic (here and in save_mfdataset)
   1068 # to avoid this mess of conditionals
   1069 try:
   1070     # TODO: allow this work (setting up the file for writing array data)
   1071     # to be parallelized with dask
-> 1072     dump_to_store(
   1073         dataset, store, writer, encoding=encoding, unlimited_dims=unlimited_dims
   1074     )
   1075     if autoclose:
   1076         store.close()

[...]

PR #2134, opened 2022-10-11
Drop object-dtyped variables and coords before saving

michaelosthege:
## Description
Variables or coordinates with `object` dtype often cause problems when saving.

In PyMC I'm refactoring stats to include a `"warning"` that that is a custom object containing detailed information about sampler divergences.
This is, of course, not ideal, but a compromise until we can settle on a data structure that would be serializeable.

For this to work, we need the `"warning"` stat to be automatically dropped when doing `.to_netcdf()`, but this is an ArviZ-level implementation.

Therefore, this PR adds helper functions to drop `object`-dtyped variables/coords before saving.

I believe this should give additional robustness against issues such as #2079.

## Checklist

----
:books: Documentation preview :books:: https://arviz--2134.org.readthedocs.build/en/2134/


arviz/data/utils.py
```diff
@@ -136,3 +143,50 @@ def extract(
...
+    coords = {}
+    coords_dropped = []
+    for k, v in data.coords.items():
+        if v.dtype.hasobject:
+            coords_dropped.append(k)
+            continue
+        coords[k] = v
+
+    ndata = Dataset(
+        data_vars=vars,
```
ahartikainen, line 167: Will this copy all data?

michaelosthege: The `Dataset` has just three kwargs and I'm passing all of them.

michaelosthege:
> I think this will also drop all text variables?

With the last push I included modified the coords dropping to have one `str` coordinate and it's not dropped.

The CI should go green now too.

...ready to review =)

cc @OriolAbril

arviz/data/utils.py
OriolAbril: I think the original question was about copying the data unnecessarly and wasting memory instead of more data than desired missing from the resulting dataset 🤔. If that were the case, it won't copy the data, xarray stores the arrays in datasets without copying them

arviz/tests/base_tests/test_data.py
```diff
@@ -14,6 +15,8 @@
 from xarray.core.options import OPTIONS
 from xarray.testing import assert_identical
 
+from arviz.data.inference_data import drop_objects_from_dataset, drop_objects_from_inferencedata
```
OriolAbril, line 18: I think this should be something like `from ...data.inference_data import`

OriolAbril (commented):
Looks good to me, I like adding helper things to inferencedata that generally have object dtype, now I won't need to delete them manually before saving.

OriolAbril (approved):

ahartikainen:
> > I think this will also drop all text variables?
> 
> With the last push I included modified the coords dropping to have one `str` coordinate and it's not dropped.
> 
> The CI should go green now too.
> 
> ...ready to review =)
> 
> cc @OriolAbril

Was the string coordinate saved as object?

e.g. if one saves one of the example datasets, will it drop coords?

ahartikainen:
I mean, that object dtype is totally valid for strings and there should not be any problems saving them to netcdf/zarr, it is just the compression that has problems with them and current main does not try to compress them.

I think we probably should have some other way to handle these things.

Ideally we should not push invalid data to InferenceData and then try to automatically fix these in our end. What if the object type used can be saved to netcdf by xarray but we just remove it. Should user now start save inference data objects manually?

michaelosthege:
I don't think `object` is valid for strings. See https://github.com/arviz-devs/arviz_example_data/pull/8 where now the `"school"` coordinate is of `<U16` dtype which has `.hasobject == False`.

And arbitrary `object`s will always cause problems, even if the underlying netcdf/zlib libraries can sometimes convert some of them..

ahartikainen:
> I don't think `object` is valid for strings. See [arviz-devs/arviz_example_data#8](https://github.com/arviz-devs/arviz_example_data/pull/8) where now the `"school"` coordinate is of `<U16` dtype which has `.hasobject == False`.
> 
> And arbitrary `object`s will always cause problems, even if the underlying netcdf/zlib libraries can sometimes convert some of them..

They are

https://github.com/pydata/xarray/issues/2059

ahartikainen:
Also to enable adding strings to a particular array of strings basically need to use object type.

michaelosthege:
> Also to enable adding strings to a particular array of strings basically need to use object type.

Unfortunately it appears you're right..
When I updated the `(non-)center_eight` `InferenceData` in https://github.com/arviz-devs/arviz_example_data/pull/8 the "school" coord had an `<U16` dtype before saving.
Loading, however, results in a `dtype('O')`.

With that it seems almost impossible to tell what constitutes an "invalid" data type before attempting to save it. 

Also, it appears that some tests are now failing on the updated example data, so this update should probably be done in a separate PR.

ahartikainen:
Part of the problem comes from numpy + string support. Not sure if there is going to be any fixes.

I'm not sure if there is a way for us to test serializing the first item for each variable (with or without compression) and then dropping invalid items.

OriolAbril:
I don't think there is any reasonable way forward given the no guarantee for round trip when it comes to dtypes. In my opinion scanning variables to check which can be saved and which not with try except is way too much work, and not the right place for such code, we save datasets as bulk so going variable per variable _and coordinate per coordinate_ is much more work and might need interfacing with netcdf directly.

I think we should close this PR and try to get logging into xarray so users know which variables and/or coordinates are the ones breaking the write process

michaelosthege:
Yes, I have made my PyMC changes independent of this, and more informative errors at the xarray level sound great

PR #2157, opened 2022-11-08
rename arviz-doc to arviz-docgrid and make arviz-doc a non-grid style

aloctavodia:
----
:books: Documentation preview :books:: https://arviz--2157.org.readthedocs.build/en/2157/

### DIFF
--- a/arviz/plots/styles/arviz-doc.mplstyle
+++ b/arviz/plots/styles/arviz-doc.mplstyle
@@ -31,7 +31,14 @@ axes.edgecolor:         .33         # axes edge color
 axes.linewidth:         0.8         # edge line width
 
 axes.grid:              False        # do not show grid
+# axes.grid.axis:         y        # which axis the grid should apply to
+axes.grid.which:        major       # grid lines at {major, minor, both} ticks
 axes.axisbelow:         True        # keep grid layer in the back
+
+grid.color:             .8          # grid color
+grid.linestyle:         -           # solid
+grid.linewidth:         0.8         # in points
+grid.alpha:             1.0         # transparency, between 0.0 and 1.0
 
 lines.solid_capstyle:   round
 
--- a/arviz/plots/styles/arviz-docgrid.mplstyle
+++ b/arviz/plots/styles/arviz-docgrid.mplstyle
@@ -30,8 +30,15 @@ axes.facecolor:         white
 axes.edgecolor:         .33         # axes edge color
 axes.linewidth:         0.8         # edge line width
 
-axes.grid:              False        # do not show grid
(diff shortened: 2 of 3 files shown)


OriolAbril:
cc @sarinac

arviz/plots/styles/arviz-doc.mplstyle
```diff
@@ -30,16 +30,9 @@ axes.facecolor:         white
 axes.edgecolor:         .33         # axes edge color
 axes.linewidth:         0.8         # edge line width
 
-axes.grid:              True        # show grid
-# axes.grid.axis:         y        # which axis the grid should apply to
-axes.grid.which:        major       # grid lines at {major, minor, both} ticks
```
sarinac, line 35: Can we keep this? Without explicitly setting `axes.grid.which` to "major", the axis will show both major and minor. I like things more simple and less clutterry, which is what I was hoping this style goes for.

I think the only differentiator between `arviz-doc` and `arviz-docgrid` is the presence of the grid. Even with arviz-doc style applied, you can override the grid visibility by applying `plt.grid()`. If the user happens to do that, then I want it to match `arviz-docgrid`'s style to a tee.

arviz/plots/styles/arviz-doc.mplstyle
```diff
@@ -30,16 +30,9 @@ axes.facecolor:         white
...
+axes.grid:              False        # do not show grid
 axes.axisbelow:         True        # keep grid layer in the back
 
-grid.color:             .8          # grid color
-grid.linestyle:         -           # solid
-grid.linewidth:         0.8         # in points
-grid.alpha:             1.0         # transparency, between 0.0 and 1.0
-
```
sarinac, line 42: Can we keep these too for the same reason above?

sarinac (changes requested):
Thanks @aloctavodia for adding! 

A few questions:
1. I made comments on keeping the doc/docgrid styles exactly the same except for `axes.grid=True`. LMK if any concerns.
2. Can you add the new `arviz-docgrid` to `doc/examples/matplotlib/mpl_styles.py`? That way users can discover it. 
3. This will remove grids for all charts generated throughout the entire documentation (homepage, example gallery, api documentation examples). Should we vary a few, change everything to `arviz-docgrid`, or leave things the way they are with the new `arviz-doc` (nothing will have grids)? My opinion is that there are some charts (e.g. violin) that look better with axes, but stuff like posterior plot (too cluttered in the examples right now) look better without.

aloctavodia:
thanks for the comments @sarinac. I updated the styles

sarinac (approved):
🎉 🎉 🎉

PR #2178, opened 2022-12-02
add ecdf comparison plot

OriolAbril:
## Description
Aiming to add a new example of ecdf comparison but the band computation seems to have a bug.

## Checklist

- [ ] Is the fix listed in the [Documentation](https://github.com/arviz-devs/arviz/blob/main/CHANGELOG.md#documentation)
      section of the changelog?

----
:books: Documentation preview :books:: https://arviz--2178.org.readthedocs.build/en/2178/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Only emit a warning for custom groups in `InferenceData` when explicitly requested ([2401](https://github.com/arviz-devs/arviz/pull/2401))
 
 ### Documentation
+- Add example of ECDF comparison plot to gallery ([2178](https://github.com/arviz-devs/arviz/pull/2178))
 
 ## v0.20.0 (2024 Sep 28)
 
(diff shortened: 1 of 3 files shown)


aloctavodia:
It seems to be a mistake with the computation of the band

sethaxen:
IIRC, we merged this quickly from the GSOC work but didn't advertise it so that we could improve it before doing so. It also uses random sampling for the bands, which is quite slow, while the paper includes a deterministic approach for generating the bands.

OriolAbril:
do you remember the main issues @sethaxen ? If so, could you open an issue?

sethaxen:
> do you remember the main issues @sethaxen ? If so, could you open an issue?

No, I don't, I'm sorry. And looking back through Slack and GitHub, I'm not seeing the discussions that I thought I remembered.

I can open an issue about performance of the confidence band computation though.

PR #2180, opened 2022-12-05
use constant gap for rank plots

### LINKED ISSUE #2177: Plot trace for rank bars need the bottoms to be aligned for hierarchical model
**Describe the bug**
For the hierarchical model,  when `compact = True`
- plot trace for rank bars needs the bottoms to be aligned
- `rank_kwargs={"bar_kwargs": {"alpha": 0.2}} is needed

image

when `compact=False`the following is showed
image

**To Reproduce**
```
import arviz as az
az.plot_trace(idata, var_names = "theta", coords = {"school": ["Choate", "Deerfield"],}, kind = "rank_bars", figsize= (20, 15), rank_kwargs={"bar_kwargs": {"alpha": 0.2}})
```

aloctavodia:
## Description

Fix #2177.  That issue is a good example of where we will benefit from decoupling computation from plotting. Currently, we compute the ranks, use the values to set the space between chains and plot that. And repeat for the next dimension of the variable.

In the meantime, this PR takes a simpler approach and set a constant gap, of two times the expected height of the uniform distribution. When the rank plot is fine this gap should be enough to make the plots looks nice. Although it will allow overlaps between bars from different chains when the rank deviates more than 2 times the expected uniform distribution. I think that's fine, because if you have such overlap then, you have bigger issues than aesthetics :-)

## Checklist

### DIFF
--- a/arviz/plots/backends/matplotlib/traceplot.py
+++ b/arviz/plots/backends/matplotlib/traceplot.py
@@ -192,6 +192,8 @@ def plot_trace(
     fill_kwargs = matplotlib_kwarg_dealiaser(fill_kwargs, "fill_between")
     rug_kwargs = matplotlib_kwarg_dealiaser(rug_kwargs, "scatter")
     rank_kwargs = matplotlib_kwarg_dealiaser(rank_kwargs, "bar")
+    if compact:
+        rank_kwargs.setdefault("bar_kwargs", {"alpha": 0.2})
 
     textsize = plot_kwargs.pop("textsize", 10)
 
--- a/arviz/plots/backends/matplotlib/rankplot.py
+++ b/arviz/plots/backends/matplotlib/rankplot.py
@@ -71,7 +71,7 @@ def plot_rank(
         all_counts = np.empty((len(ranks), len(bin_ary) - 1))
         for idx, row in enumerate(ranks):
             _, all_counts[idx], _ = histogram(row, bins=bin_ary)
-        gap = all_counts.max() * 1.05
+        gap = 2 / ranks.size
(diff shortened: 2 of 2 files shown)


arviz/plots/backends/matplotlib/traceplot.py
```diff
@@ -192,6 +192,8 @@ def plot_trace(
     fill_kwargs = matplotlib_kwarg_dealiaser(fill_kwargs, "fill_between")
     rug_kwargs = matplotlib_kwarg_dealiaser(rug_kwargs, "scatter")
     rank_kwargs = matplotlib_kwarg_dealiaser(rank_kwargs, "bar")
+    if compact:
+        rank_kwargs.setdefault("bar_kwargs", {"alpha": 0.2})
```
OriolAbril, line 196: I think this will only work in a small subset of cases and it should be something like:

```suggestion
    if compact:
        rank_kwargs.setdefault("bar_kwargs", {})
        rank_kwargs["bar_kwargs"].setdefault("alpha", 0.2)
```

OriolAbril (approved):

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@
 
 ### Maintenance and fixes
 - Fix dimension ordering for `plot_trace` with divergences ([2151](https://github.com/arviz-devs/arviz/pull/2151))
+- Fix gap for `plot_trace` with option `kind="rank_bars"` ([2180](https://github.com/arviz-devs/arviz/pull/2180))
```
OriolAbril, line 26: this ended up in the 0.14 section instead of the unreleased, other than that it is good to merge

PR #2185, opened 2022-12-21
Update docstrings of ess, autocorr and bpv plots

asael697:
## Checklist

----
:books: Documentation preview :books:: https://arviz--2185.org.readthedocs.build/en/2185/

### DIFF
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -23,37 +23,38 @@ def plot_autocorr(
     backend_kwargs=None,
     show=None,
 ):
-    """Bar plot of the autocorrelation function for a sequence of data.
+    """Bar plot of the autocorrelation function (ACF) for a sequence of data.
 
-    Useful in particular for posteriors from MCMC samples which may display correlation.
+    The ACF plots are useful as a convergence diagnostic for posteriors from MCMC 
+    samples which may display correlation. The ACF is efficiently computed using a 
+    Fast Fourier Transform for every lag for the input array.
 
     Parameters
     ----------
-    data : InferenceData
+    data : obj
         Any object that can be converted to an :class:`arviz.InferenceData` object
         refer to documentation of :func:`arviz.convert_to_dataset` for details
     var_names : list of str, optional
-        Variables to be plotted, if None all variables are plotted. Prefix the
-        variables by ``~`` when you want to exclude them from the plot. Vector-value
-        stochastics are handled automatically.
-    filter_vars : {None, "like", "regex"},  default=None
-        If `None` (default), interpret var_names as the real variables names. If "like",
-        interpret var_names as substrings of the real variables names. If "regex",
-        interpret var_names as regular expressions on the real variables names. A la
-        ``pandas.filter``.
+        Variables to be plotted. Prefix the variables by ``~`` when you want to exclude
+        them from the plot. See the :ref:`this section <common_var_names>` for usage examples.
+    filter_vars : {None, "like", "regex"},  default None
+        If `None` (default), interpret `var_names` as the real variables names. If "like",
+        interpret `var_names` as substrings of the real variables names. If "regex",
+        interpret `var_names` as regular expressions on the real variables names. See
+        the :ref:`this section <common_filter_vars>` for usage examples.
     max_lag : int, optional
-        Maximum lag to calculate autocorrelation. Defaults to 100 or num draws,
-        whichever is smaller.
-    combined : bool, default=False
+        Maximum lag to calculate autocorrelation. By Default, the plot displays the 
+        first 100 lag or the total number of draws, whichever is smaller.
+    combined : bool, default False
         Flag for combining multiple chains into a single chain. If False, chains will be
         plotted separately.
-    grid : tuple
+    grid : tuple, optional
         Number of rows and columns. Defaults to None, the rows and columns are
         automatically inferred.
-    figsize : (float, float), optional
+    figsize : tuple, optional
         Figure size. If None it will be defined automatically.
         Note this is not used if ``ax`` is supplied.
-    textsize : float
+    textsize : float, optional
         Text size scaling factor for labels, titles and lines. If None it will be autoscaled based
         on ``figsize``.
     labeller : Labeller, optional
@@ -68,8 +69,8 @@ def plot_autocorr(
         Currently specifies the bounds to use for bokeh axes. Defaults to value set in ``rcParams``.
     backend_kwargs : dict, optional
         These are kwargs specific to the backend being used, passed to
-        :func:`matplotlib.pyplot.subplots` or
-        :func:`bokeh.plotting.figure`.
+        :func:`matplotlib.pyplot.subplots` or :func:`bokeh.plotting.figure`.
+        For additional documentation check the plotting method of the backend.
     show : bool, optional
         Call backend show function.
 
--- a/arviz/plots/autocorrplot.py
+++ b/arviz/plots/autocorrplot.py
@@ -30,13 +30,13 @@ def plot_autocorr(
 
     Parameters
     ----------
-    data : obj
+    data : InferenceData
         Any object that can be converted to an :class:`arviz.InferenceData` object
         refer to documentation of :func:`arviz.convert_to_dataset` for details
     var_names : list of str, optional
         Variables to be plotted. Prefix the variables by ``~`` when you want to exclude
         them from the plot. See the :ref:`this section <common_var_names>` for usage examples.
-    filter_vars : {None, "like", "regex"},  default None
+    filter_vars : {None, "like", "regex"}, default None
         If `None` (default), interpret `var_names` as the real variables names. If "like",
         interpret `var_names` as substrings of the real variables names. If "regex",
         interpret `var_names` as regular expressions on the real variables names. See
@@ -50,12 +50,12 @@ def plot_autocorr(
     grid : tuple, optional
         Number of rows and columns. Defaults to None, the rows and columns are
         automatically inferred.
-    figsize : tuple, optional
+    figsize : (float, float), optional
         Figure size. If None it will be defined automatically.
         Note this is not used if ``ax`` is supplied.
     textsize : float, optional
         Text size scaling factor for labels, titles and lines. If None it will be autoscaled based
-        on ``figsize``.
+        on `figsize`.
     labeller : Labeller, optional
         Class providing the method ``make_label_vert`` to generate the labels in the plot titles.
         Read the :ref:`label_guide` for more details and usage examples.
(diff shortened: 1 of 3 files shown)


arviz/plots/autocorrplot.py
```diff
@@ -23,37 +23,38 @@ def plot_autocorr(
     backend_kwargs=None,
     show=None,
 ):
-    """Bar plot of the autocorrelation function for a sequence of data.
+    """Bar plot of the autocorrelation function (ACF) for a sequence of data.
 
-    Useful in particular for posteriors from MCMC samples which may display correlation.
+    The ACF plots are helpful as a convergence diagnostic for posteriors from MCMC 
+    samples which may display correlation. The ACF is efficiently computed using a 
+    Fast Fourier Transform for every lag for the input array.
```
OriolAbril, line 30:
```suggestion
    samples which display autocorrelation.
```

I would add the "The ACF is efficiently computed using a Fast Fourier Transform for every lag for the input array." part in the notes, ideally also with a reference on using fft for autocov computation (or maybe moving this to the autocov or autocorr functions instead of this one).

For the note on their use, I think this is a bit more clear? I am not aware of using autocorr plot for diagnosing posteriors with correlation between parameters

arviz/plots/autocorrplot.py
```diff
@@ -23,37 +23,38 @@ def plot_autocorr(
...
 
     Parameters
     ----------
-    data : InferenceData
+    data : obj
```
OriolAbril, line 34: I would keep this as InferenceData

arviz/plots/autocorrplot.py
```diff
@@ -23,37 +23,38 @@ def plot_autocorr(
...
+        first 100 lag or the total number of draws, whichever is smaller.
+    combined : bool, default False
         Flag for combining multiple chains into a single chain. If False, chains will be
         plotted separately.
-    grid : tuple
+    grid : tuple, optional
         Number of rows and columns. Defaults to None, the rows and columns are
         automatically inferred.
-    figsize : (float, float), optional
+    figsize : tuple, optional
```
OriolAbril, line 54: The float, float formatting is correct, no need to change it. It is a common convention that the parentheses indicate it is a tuple and this formatting allows indicating the types of each element and the number of elements. Ref: https://matplotlib.org/stable/devel/documenting_mpl.html#parameter-type-descriptions

arviz/plots/bpvplot.py
```diff
@@ -44,50 +44,50 @@ def plot_bpv(
...
+             (half the predictions below and half above the data).
+           * The ``kind="u_value"`` argument computes :math:`p_i := p(y_i* \leq y_i | y)`. i.e. like a p_value 
+             but per observation :math:`y_i`. This is also known as marginal p_value. The ideal distribution 
+             is uniform. This is similar to the LOO-pit calculation/plot, the difference is than in 
+             LOO-pit plot we compute :math:`pi = p(y_i* r \leq y_i | y_{-i} )`, where :math:`y_{-i}`, is all 
+             other data except :math:`y_i`.
+           * The ``kind="t_stat"`` argument computes :math:`:= p(T(y)* \leq T(y) | y)` where T is any test statistic. 
+             See ``t_stat`` argument below for details of available options.
+             
+    t_stat : {"mean", "median", "std"},  default "median"
```
OriolAbril, line 61: This needs to be kept as before as this is now incorrect. Allowed strings are only mean, median or std, but it is not necessarily a string, it can also be a float or a user defined function (`callable` type in python)

arviz/plots/bpvplot.py
```diff
@@ -44,50 +44,50 @@ def plot_bpv(
...
         acepted, see examples section for details.
     bpv : bool, default True
         If True add the Bayesian p_value to the legend when ``kind = t_stat``.
     plot_mean : bool, default True
         Whether or not to plot the mean test statistic.
-    reference : str, default "analytical"
-        How to compute the distributions used as reference for u_values or p_values. Allowed values
-        are "analytical" and "samples". Use `None` to do not plot any reference.
-        Defaults to "samples".
+    reference : {None,"analytical","samples"}, default "analytical"
```
OriolAbril, line 69:
```suggestion
    reference : {"analytical", "samples", None}, default "analytical"
```

Added the spaces, also enforcing the default first convention

arviz/plots/bpvplot.py
```diff
@@ -44,50 +44,50 @@ def plot_bpv(
...
-        to the LOO-pit calculation/plot, the difference is than in LOO-pit plot we compute
-        pi = p(yi* r ≤ yi | y-i ), where y-i, is all other data except yi.
-        For "t_stat" we compute := p(T(y)* ≤ T(y) | y) where T is any test statistic. See t_stat
-        argument below for details of available options.
-    t_stat : str, float, or callable, default "median"
-        Test statistics to compute from the observations and predictive distributions.
-        Allowed strings are "mean", "median" or "std".
-        Alternative a quantile can be passed as a float (or str) in the
-        interval (0, 1). Finally a user defined function is also
+    kind : {"p_value", "u_value", "t_stat"}, default "u_value"
```
OriolAbril, line 47:
```suggestion
    kind : {"u_value", "p_value", "t_stat"}, default "u_value"
```

Putting the default value first. From [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#parameters):

> When a parameter can only assume one of a fixed set of values, those values can be listed in braces, with the default appearing first:

arviz/plots/bpvplot.py
```diff
@@ -44,50 +44,50 @@ def plot_bpv(
...
-        computing u_values. Should be in the interval (0, 1]. Defaults to the
+        ``kind=u_values``. Should be in the interval (0, 1]. Defaults to the
         rcParam ``stats.hdi_prob``.
     color : str, optional
         Matplotlib color
     grid : tuple, optional
-        Number of rows and columns. Defaults to None, the rows and columns are
+        Number of rows and columns. By default, the rows and columns are
         automatically inferred.
-    figsize : (float, float), optional
```
OriolAbril, line 86: same comment as above

arviz/plots/bpvplot.py
```diff
@@ -44,50 +44,50 @@ def plot_bpv(
...
+    figsize : tuple, optional
         Figure size. If None it will be defined automatically.
     textsize : float, optional
-        Text size scaling factor for labels, titles and lines. If None it will be
-        autoscaled based on ``figsize``.
+        Text size scaling factor for labels, titles and lines. If None it will be autoscaled based
+        on ``figsize``.
```
OriolAbril, line 90: it was already wrong before, but here figsize should have only single backticks as it is referencing a parameter alone.

arviz/plots/bpvplot.py
```diff
@@ -101,14 +101,15 @@ def plot_bpv(
...
-        If `None` (default), interpret var_names as the real variables names. If "like",
-        interpret var_names as substrings of the real variables names. If "regex",
-        interpret var_names as regular expressions on the real variables names. A la
-        ``pandas.filter``.
+    var_names : list of str, optional
+        Variables to be plotted. If `None` all variable are plotted. Prefix the variables by ``~`` 
+        when you want to exclude them from the plot. See the :ref:`this section <common_var_names>` 
+        for usage examples.
+    filter_vars : {None, "like", "regex"}, optional, default None
+        If `None` (default), interpret `var_names` as the real variables names. If "like",
```
OriolAbril, line 109:
```suggestion
    filter_vars : {None, "like", "regex"}, default None
        If `None` (default), interpret `var_names` as the real variables names. If "like",
```

arviz/plots/bpvplot.py
```diff
@@ -134,14 +135,12 @@ def plot_bpv(
         Passed to :meth:`matplotlib.axes.Axes.plot` or
         :meth:`matplotlib.axes.Axes.axhspan` (when ``kind=u_value``
         and ``reference=analytical``).
-    backend_kwargs : bool, optional
+    backend_kwargs : bool, optional
```
OriolAbril, line 138: You will need to review the trailing whitespace, otherwise the auto code formatting check won't pass.

arviz/plots/bpvplot.py
```diff
@@ -134,14 +135,12 @@ def plot_bpv(
...
         These are kwargs specific to the backend being used, passed to
-        :func:`matplotlib.pyplot.subplots` or
-        :func:`bokeh.plotting.figure`. For additional documentation
-        check the plotting method of the backend.
-    group : {"prior", "posterior"}, optional
-        Specifies which InferenceData group should be plotted. Defaults to 'posterior'.
-        Other value can be 'prior'.
+        :func:`matplotlib.pyplot.subplots` or :func:`bokeh.plotting.figure`.
+        For additional documentation check the plotting method of the backend.
+    group : {"prior", "posterior"}, optional default "posterior"
```
OriolAbril, line 142:
```suggestion
    group : {"posterior", "prior"}, default "posterior"
```

arviz/plots/bpvplot.py
```diff
@@ -134,14 +135,12 @@ def plot_bpv(
...
+        Specifies which InferenceData group should be plotted.
```
OriolAbril, line 143: This is not really accurate. If `posterior`, then the values in `posterior_predictive` group are compared to the ones in `observed_data`, if `prior` then the same comparison happens, but with the values in `prior_predictive` group.

arviz/plots/essplot.py
```diff
@@ -37,56 +37,70 @@ def plot_ess(
...
         automatically inferred.
     figsize : tuple, optional
         Figure size. If None it will be defined automatically.
     textsize : float, optional
         Text size scaling factor for labels, titles and lines. If None it will be autoscaled based
-        on figsize.
-    rug : bool
-        Plot rug plot of values diverging or that reached the max tree depth.
-    rug_kind : bool
+        on ``figsize``.
```
OriolAbril, line 82: only single backtick here for figsize instead of double

arviz/plots/essplot.py
```diff
@@ -37,56 +37,70 @@ def plot_ess(
...
+    rug : bool, default False
+        Add a `rug plot <https://en.wikipedia.org/wiki/Rug_plot>` for a specific subset of values.
+    rug_kind : str, dafault "diverging"
```
OriolAbril, line 85:
```suggestion
    rug_kind : str, default "diverging"
```

arviz/plots/autocorrplot.py
asael697: I decided to explain it here because I didn't consider it specific enough to have a note, but ok. and MMM I don't think autocor is for correlation with between parameters

arviz/plots/autocorrplot.py
asael697: We used the `tuple` in plot_ess() function, lets leave this one as a standard

arviz/plots/bpvplot.py
asael697: if I got it right is it to leave a space at the end?

arviz/plots/bpvplot.py
asael697: ok

OriolAbril (approved):

PR #2210, opened 2023-02-17
Add rope to plots arguments guide

asael697:
## Description

Add the `rope` option to the "Plot's arguments guide" page; this option is available in the forest and posterior plots. Added a link to the page in both arguments' documentation.

Additionally, I removed the forest plot with ropes example from its documentation, to add other examples later

## Checklist

----
:books: Documentation preview :books:: https://arviz--2210.org.readthedocs.build/en/2210/

### DIFF
--- a/doc/source/user_guide/plots_arguments_guide.md
+++ b/doc/source/user_guide/plots_arguments_guide.md
@@ -307,19 +307,55 @@ Module {mod}`bokeh.colors`
 (common_rope)=
 ## `rope`
 
-A dictionary of tuples with lower and upper values of the Region Of Practical Equivalence. If a list with
-one interval only is provided, the `ROPE` will be displayed across the `y-axis`. If more than one interval
-is provided the length of the list should match the number of variables.
+The rope argument is used to indicate the lower and upper limits of the Region Of Practical
+Equivalence (ROPE). These limits can be expressed in two ways, depending on whether
+the ROPE is common between all variables or not:
+
+1. A single list. If a single list of two floats is given as `rope` argument,
+   these two provided values will be used as limits of the ROPE in all plotted variables.
+2. A dictionary of lists.
+
+Example of using a single list as `rope`
 
 ```{code-cell} ipython3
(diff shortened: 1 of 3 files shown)


doc/source/user_guide/plots_arguments_guide.md
```diff
@@ -304,6 +304,24 @@ Matplotlib {doc}`mpl:tutorials/colors/colors` tutorial.
 Module {mod}`bokeh.colors`
 :::
 
+(common_rope)=
+## `rope`
+
+A list with lower and upper values of the Region Of Practical Equivalence. If a list with one interval
+only is provided, the `ROPE` will be displayed across the `y-axis`. If more than one interval is
+provided the length of the list should match the number of variables.
```
OriolAbril, line 312: I think this doesn't match the actual behaviour nor the example below. From the example below it looks like the recommended usage is a dictionary of list/tuple where at the same time, each element in those list/tuples is a dictionary too. 

The first dictionary has variable names as keys, and list/tuple as values, whereas the 2nd dictionaries have coord names as keys and coord values as values for those keys and also need a `rope` key with values a (float, float) tuple which defines the rope interval for that variable/subvariable.

We should also add a single list and a list with same length of variable names (which will also serve to double check the description is correct, which might not be the case)

OriolAbril (commented):
Thanks for adding the references and moving the example of rope usage to the right section

doc/source/user_guide/plots_arguments_guide.md
asael697: I am not fully following you.

OriolAbril:
I have updated the `plots_arguments_guide` file: https://arviz--2210.org.readthedocs.build/en/2210/user_guide/plots_arguments_guide.html#rope. As suspected, the description in the docstrings was wrong. It isn't possible to give a nested list as input. @asael697 can you update the docstrings to match the actual behaviour?

OriolAbril (approved):

PR #2213, opened 2023-02-19
Post release version update

canyon289:
Post release pull request

----
:books: Documentation preview :books:: https://arviz--2213.org.readthedocs.build/en/2213/

### DIFF
--- a/arviz/__init__.py
+++ b/arviz/__init__.py
@@ -1,6 +1,6 @@
 # pylint: disable=wildcard-import,invalid-name,wrong-import-position
 """ArviZ is a library for exploratory analysis of Bayesian models."""
-__version__ = "0.16.0dev"
+__version__ = "0.16.0.dev0"
 
 import logging
 import os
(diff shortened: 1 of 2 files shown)


arviz/__init__.py
```diff
@@ -1,6 +1,6 @@
 # pylint: disable=wildcard-import,invalid-name,wrong-import-position
 """ArviZ is a library for exploratory analysis of Bayesian models."""
-__version__ = "0.15.0"
+__version__ = "0.16.0dev"
```
canyon289, line 3: I dont know if I did this right. Should it be 0.15.1dev0? How do we know there wont be a bugfix bump?


https://python.arviz.org/en/latest/contributing/how_to_release.html

OriolAbril: we can still release a patch version if necessary, but in general we start merging new features as soon as something was released, so the common situation is a minor bump. Also note the syntax to make sure it is pep compiant:

```suggestion
__version__ = "0.16.0.dev0"
```

OriolAbril (approved):

PR #2227, opened 2023-03-23
Added the ability to append an InferenceData object to an existing netCDF file 

varchasgopalaswamy:
## Checklist

----
:books: Documentation preview :books:: https://arviz--2227.org.readthedocs.build/en/2227/

### DIFF
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -23,6 +23,7 @@ from typing import (
     Union,
     overload,
 )
+import os
 
 import numpy as np
 import xarray as xr
@@ -438,6 +439,7 @@ class InferenceData(Mapping[str, xr.Dataset]):
         groups: Optional[List[str]] = None,
         engine: str = "h5netcdf",
         base_group: Optional[str] = None,
+        overwrite_existing: bool = True,
     ) -> str:
         """Write InferenceData to netcdf4 file.
 
@@ -464,7 +466,11 @@ class InferenceData(Mapping[str, xr.Dataset]):
         if base_group is None:
             base_group = '/'
 
-        mode = "w"  # overwrite first, then append
+        if os.path.exists(filename) and not overwrite_existing:
+            mode = "a"  # overwrite first, then append
+        else:
+            mode = "w"
+
         if self._attrs:
             xr.Dataset(attrs=self._attrs).to_netcdf(filename, mode=mode, engine=engine, group=base_group)
             mode = "a"
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -366,7 +366,7 @@ class InferenceData(Mapping[str, xr.Dataset]):
         regex : bool, default False
             Specifies where regex search should be used to extend the keyword arguments.
             This feature is currently experimental.
-        base_group: str, default "/"
+        base_group : str, default "/"
             The group in the netCDF file where the InferenceData is stored. By default, assumes that the file
             only contains an InferenceData object.
 
@@ -454,10 +454,10 @@ class InferenceData(Mapping[str, xr.Dataset]):
             Write only these groups to netcdf file.
         engine : {"h5netcdf", "netcdf4"}, default "h5netcdf"
             Library used to read the netcdf file.
-        base_group: str, default "/"
+        base_group : str, default "/"
             The group in the netCDF file where the InferenceData is will be stored.
             By default, will write to the root of the netCDF file
-        overwrite_existing: bool, default True
+        overwrite_existing : bool, default True
             Whether to overwrite the existing file or append to it.
 
         Returns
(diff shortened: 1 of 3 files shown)


OriolAbril (commented):
Sorry for the late review. Thanks for the 2nd PR!

It looks like the PR adds both the option to append to an existing netCDF file and to read/write InferenceData as lower level groups within a user-given base group. Is that right?

arviz/data/inference_data.py
```diff
@@ -337,7 +338,9 @@ def items(self) -> "InferenceData.InferenceDataItemsView":
 
     @staticmethod
     def from_netcdf(
-        filename, *, engine="h5netcdf", group_kwargs=None, regex=False
+        filename, *,
+        engine="h5netcdf", group_kwargs=None, regex=False,
+        base_group: Optional[str] = None,
```
OriolAbril, line 343: You'll need to run black on the file to fix the lint errors

arviz/data/inference_data.py
```diff
@@ -360,6 +363,9 @@ def from_netcdf(
         regex : bool, default False
             Specifies where regex search should be used to extend the keyword arguments.
             This feature is currently experimental.
+        base_group: str, optional
```
OriolAbril, line 366:
```suggestion
        base_group : str, optional
```

arviz/data/inference_data.py
OriolAbril, line 343:
```suggestion
        base_group: str = "/",
```

I think using this as default will be easier to follow and will simplify the code.

arviz/data/inference_data.py
```diff
@@ -438,15 +454,25 @@ def to_netcdf(
             Write only these groups to netcdf file.
         engine : {"h5netcdf", "netcdf4"}, default "h5netcdf"
             Library used to read the netcdf file.
+        base_group: str, optional
+            The group in the netCDF file where the InferenceData is will be stored.
+            By default, will write to the root of the netCDF file
```
OriolAbril, line 459: The `overwrite_existing` also needs to be documented.

varchasgopalaswamy:
> Sorry for the late review. Thanks for the 2nd PR!
> 
> It looks like the PR adds both the option to append to an existing netCDF file and to read/write InferenceData as lower level groups within a user-given base group. Is that right?

Yeah, that's the idea!

arviz/data/inference_data.py
```diff
@@ -464,10 +490,10 @@ def to_netcdf(
                         for var_name, values in data.variables.items()
                         if _compressible_dtype(values.dtype)
                     }
-                data.to_netcdf(filename, mode=mode, group=group, **kwargs)
+                data.to_netcdf(filename, mode=mode, group=f"{base_group}/{group}", **kwargs)
                 data.close()
                 mode = "a"
-        elif not self._attrs:  # creates a netcdf file for an empty InferenceData object.
+        elif not self._attrs and base_group == '/':  # creates a netcdf file for an empty InferenceData object.
```
OriolAbril, line 496: I am not sure about the logic here. If the inferencedata is empty, a file is still created, but it is an empty file. Shouldn't that happen if base_group =! "/"? I guess ideally a file with that group present as an empty dataset, otherwise probably an empty file to preserve the "to_netcdf always creates a file, even if saving an empty inferencedata object" behaviour

varchasgopalaswamy: Sounds fair enough. I've restored the previous behavior.

arviz/data/inference_data.py
```diff
@@ -360,6 +366,9 @@ def from_netcdf(
         regex : bool, default False
             Specifies where regex search should be used to extend the keyword arguments.
             This feature is currently experimental.
+        base_group : str, default "/"
+            The group in the netCDF file where the InferenceData is stored. By default, assumes that the file
+            only contains an InferenceData object.
```
OriolAbril, line 371:
```suggestion
            The group in the netCDF file where the InferenceData is stored. By default,
            assumes that the file only contains an InferenceData object.
```

this is the line being flagged as too long by pylint.

OriolAbril (commented):
I think the code is good to go now, thanks! Can you add a test too in https://github.com/arviz-devs/arviz/blob/main/arviz/tests/base_tests/test_data.py for this new arguments?

arviz/tests/base_tests/test_data.py
```diff
@@ -1294,10 +1294,11 @@ def test_io_function(self, data, eight_schools_params):
         os.remove(filepath)
         assert not os.path.exists(filepath)
 
+    @pytest.mark.parametrize("base_group", ['/','/test_group'])
```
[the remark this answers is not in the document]
varchasgopalaswamy: I don't think so... not unless the NetCDF interface library requires it.

varchasgopalaswamy:
@OriolAbril is there anything else needed to merge the PR?

OriolAbril (approved):
fixed some black complaints, should be good to go now. Sorry about the slow review cadence, thank you so much for all the PRs!

varchasgopalaswamy:
No problem - thank you for the great package! Any idea when there might be a new release? I'd like to stop pinning my software to a github ref!

PR #2233, opened 2023-04-11
Remove python 3.8 from CI

ahartikainen:
Remove 3.8 python ~and add 3.11~. Move preview mode to 3.10.

This also means we will drop CI testing for pystan 2.x. Last release where still worked normally was 3.8 and 3.9 one can still install it but it is a bit hacky. 3.10 does not support it anymore.

----
:books: Documentation preview :books:: https://arviz--2233.org.readthedocs.build/en/2233/


.readthedocs.yml
```diff
@@ -4,7 +4,7 @@ sphinx:
     configuration: doc/source/conf.py
 
 python:
-   version: "3.8"
+   version: "3.10"
    install:
```
OriolAbril, line 8: This should be updated to the improved way of defining the env on which to run rtd. We use that in https://github.com/arviz-devs/xarray-einstats/blob/main/.readthedocs.yaml

OriolAbril:
numba isn't compatible with python 3.11 yet, so the installation of optional libraries fails, not sure if there is an easy way to exclude only numba installation and tests

ahartikainen:
ok, let's wait with 3.11

OriolAbril:
The 0.57 release candidate is already available: https://pypi.org/project/numba/#history. 0.57 supports python 3.11. So it looks like we won't have to wait too much

ahartikainen:
@OriolAbril is this good to go?

PR #2241, opened 2023-05-09
Pkg resources replacement

michaelraczycki:
Replacement of deprecated pkg_resources, now arviz will use importlib.resources
which will allow to use more strict testing policy in packages importing arviz ( previously because of pkg_resources
failing test on warning was not possible, because each import of arviz was warning about deprecation of pkg_resources)

----
:books: Documentation preview :books:: https://arviz--2241.org.readthedocs.build/en/2241/

### DIFF
--- a/arviz/data/base.py
+++ b/arviz/data/base.py
@@ -1,13 +1,13 @@
 """Low level converters usually used by other functions."""
 import datetime
 import functools
+import importlib.resources
 import re
 import warnings
 from copy import deepcopy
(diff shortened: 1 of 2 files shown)


arviz/data/base.py
```diff
@@ -1,13 +1,13 @@
 """Low level converters usually used by other functions."""
 import datetime
 import functools
+import importlib.resources
```
OriolAbril, line 4:
```suggestion
import importlib
```

here we use importlib.metadata, so I think this import makes more sense

OriolAbril (approved):

PR #2280, opened 2023-09-27
Set matplotlib >=3.5 in requirements.txt

DAWells:
## Description
- set matplotlib >=3.5 in requirements.txt
Since https://github.com/arviz-devs/arviz/pull/2127 `arviz/__init__.py` requires matplotlib >= 3.5 for `mpl.colormaps.register()` otherwise `arviz` fails on import.
- add this change to changelog.md

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
 setuptools>=60.0.0
 matplotlib>=3.5
-numpy>=1.21.0,<2.0
+numpy>=1.22.0,<2.0
 scipy>=1.8.0
 packaging
-pandas>=1.3.0
+pandas>=1.4.0
 xarray>=0.21.0
 h5netcdf>=1.0.2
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 setuptools>=60.0.0
-matplotlib>=3.2
+matplotlib>=3.5
 numpy>=1.21.0,<2.0
```
OriolAbril, line 3:
```suggestion
numpy>=1.22.0,<2.0
```

OriolAbril (commented):
Thanks! I have also added a suggestion to update the numpy cap and the pandas one should also be updated so that we can roughly follow https://scientific-python.org/specs/spec-0000/. Do you mind doing that?

OriolAbril (approved):

PR #2286, opened 2023-11-01
Add kwarg to avoid alternate row shading with az.plot_forest

u7576869:
## Description
Adding a new parameter alternate_row_shading to give the user control over alternate row shading. Setting this to false will result in a more clear graph in many cases.

## Checklist

### DIFF
--- a/arviz/plots/backends/matplotlib/forestplot.py
+++ b/arviz/plots/backends/matplotlib/forestplot.py
@@ -51,7 +51,6 @@ def plot_forest(
     ess,
     r_hat,
     backend_kwargs,
-    backend_config,  # pylint: disable=unused-argument
     show,
 ):
     """Matplotlib forest plot."""
--- a/arviz/plots/forestplot.py
+++ b/arviz/plots/forestplot.py
@@ -36,7 +36,6 @@ def plot_forest(
     ax=None,
     backend=None,
     backend_config=None,
-    backend_kwargs=None,
     show=None,
 ):
     r"""Forest plot to compare HDI intervals from a number of distributions.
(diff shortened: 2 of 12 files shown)


arviz/plots/forestplot.py
```diff
@@ -36,7 +36,6 @@ def plot_forest(
     ax=None,
     backend=None,
     backend_config=None,
-    backend_kwargs=None,
```
OriolAbril, line 39: this argument should not be removed. Also, there is no argument being added, the code examples use `alternate_row_shading` so it should be added here (and documented in the docstring)

arviz/plots/forestplot.py
```diff
@@ -158,12 +157,13 @@ def plot_forest(
 
         >>> import arviz as az
         >>> non_centered_data = az.load_arviz_data('non_centered_eight')
-        >>> axes = az.plot_forest(non_centered_data,
+        >>> axes = az.plot_forest_kwargs(non_centered_data,
```
OriolAbril, line 160: The function is still called plot_forest, all occurrences of `plot_forest_kwargs` should be changed back

arviz/plots/backends/matplotlib/forestplot.py
```diff
@@ -51,7 +51,6 @@ def plot_forest(
     ess,
     r_hat,
     backend_kwargs,
-    backend_config,  # pylint: disable=unused-argument
```
OriolAbril, line 54: this also should not be deleted

OriolAbril (changes requested):
The idea is good, but it clearly needs much more work for it to be usable. Let me know if anything isn't clear enough

PR #2291, opened 2023-11-14
Make `from_dict` more flexible, and add `from_pytree`

ColCarroll:
This also makes `from_dict` work for nested dictionaries. I joined the keys with double underscores because using a `.` would break attribute access (like, it would still work, but you'd use `idata.posterior['var.x.y']`, instead of `idata.posterior.var__x__y`)

Hopefully this is a no-op for most code, but it should make arviz work automatically with namedtuples or dataclasses. 

----
:books: Documentation preview :books:: https://arviz--2291.org.readthedocs.build/en/2291/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v0.x.x Unreleased
 
 ### New features
+- Support for `pytree`s and robust to nested dictionaries. (2291)
 
 ### Maintenance and fixes
 - Fix deprecations introduced in latest pandas and xarray versions, and prepare for numpy 2.0 ones ([2315](https://github.com/arviz-devs/arviz/pull/2315)))
--- a/arviz/data/base.py
+++ b/arviz/data/base.py
@@ -8,6 +8,7 @@ from copy import deepcopy
 from typing import Any, Callable, Dict, List, Optional, Tuple, TypeVar, Union
 
 import numpy as np
+import tree
 import xarray as xr
 
 try:
@@ -65,6 +66,48 @@ class requires:  # pylint: disable=invalid-name
             return func(cls)
 
         return wrapped
+
+
+def _yield_flat_up_to(shallow_tree, input_tree, path=()):
+    """Yields (path, value) pairs of input_tree flattened up to shallow_tree.
+
+    Adapted from dm-tree (https://github.com/google-deepmind/tree) to allow
+    lists as leaves.
+
+    Args:
+        shallow_tree: Nested structure. Traverse no further than its leaf nodes.
+        input_tree: Nested structure. Return the paths and values from this tree.
+            Must have the same upper structure as shallow_tree.
+        path: Tuple. Optional argument, only used when recursing. The path from the
+            root of the original shallow_tree, down to the root of the shallow_tree
+            arg of this recursive call.
+
+    Yields:
+        Pairs of (path, value), where path the tuple path of a leaf node in
+        shallow_tree, and value is the value of the corresponding node in
+        input_tree.
+    """
+    # pylint: disable=protected-access
+    if isinstance(shallow_tree, tree._TEXT_OR_BYTES) or not (
+        isinstance(shallow_tree, tree.collections_abc.Mapping)
+        or tree._is_namedtuple(shallow_tree)
+        or tree._is_attrs(shallow_tree)
+    ):
+        yield (path, input_tree)
+    else:
+        input_tree = dict(tree._yield_sorted_items(input_tree))
+        for shallow_key, shallow_subtree in tree._yield_sorted_items(shallow_tree):
+            subpath = path + (shallow_key,)
+            input_subtree = input_tree[shallow_key]
+            for leaf_path, leaf_value in _yield_flat_up_to(
+                shallow_subtree, input_subtree, path=subpath
+            ):
+                yield (leaf_path, leaf_value)
+    # pylint: enable=protected-access
+
+
+def _flatten_with_path(structure):
+    return list(_yield_flat_up_to(structure, structure))
 
 
 def generate_dims_coords(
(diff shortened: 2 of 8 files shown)


ColCarroll:
I'll add that this adds a requirement on [dm-tree](https://github.com/google-deepmind/tree). I'd rather have used `jax.tree_utils`, but that requires all of `jax` (~1.4MB) instead of just `dm-tree` (~35kb).

ahartikainen:
Will this affect python 3.12 use? Probably not(?)

ColCarroll:
It looks like maybe: https://github.com/google-deepmind/tree/issues/109

I gave https://github.com/google-deepmind/tree/pull/111 a try.

ColCarroll:
I tried it out, and it fails when trying to install `dm-tree` on python 3.12, because it requires `CMake`. Once you `conda install CMake` and install `xcode` on OSX, you can install from source, but that's probably too big of a lift for users, and we should wait on this until `dm-tree` builds a wheel. 

Given expected workloads in `arviz`, we might just write our own `flatten_with_path` function in pure python. I'll think about that if there's no motion on the other PR.

ColCarroll:
Update: The PR was merged for `dm-tree`, but we need to wait on a pypi release.

ColCarroll:
Hey! They sneakily added PyPI wheels for 3.12 to dm-tree. I think this is now ready to go.

arviz/data/base.py
```diff
@@ -266,7 +309,7 @@ def dict_to_dataset(
     index_origin=None,
     skip_event_dims=None,
 ):
-    """Convert a dictionary of numpy arrays to an xarray.Dataset.
+    """Convert a pytree of numpy arrays to an xarray.Dataset.
```
OriolAbril, line 312:
```suggestion
    """Convert a dictionary or pytree of numpy arrays to an xarray.Dataset.
```

arviz/tests/base_tests/test_data.py
```diff
@@ -1077,6 +1077,20 @@ def test_dict_to_dataset():
     assert set(dataset.b.coords) == {"chain", "draw", "c"}
 
 
+def test_nested_dict_to_dataset():
+    datadict = {
+        "top": {"a": np.random.randn(100), "b": np.random.randn(1, 100, 10)},
+        "d": np.random.randn(100),
+    }
+    dataset = convert_to_dataset(datadict, coords={"c": np.arange(10)}, dims={("top", "b"): ["c"]})
```
OriolAbril, line 1085: I would extend the docstring, both parameter description and examples to show this. Mainly, how will flattened keys look like, and how can one specify coordinates/dimensions for nested variables. Otherwise we'd require users to read through the source code and/or tests in order to properly be able to use the function.

CHANGELOG.md
```diff
@@ -2,7 +2,6 @@
 
 ## v0.x.x Unreleased
 
-### New features
```
OriolAbril, line 5: looks like a merge/rebase error in the changelog

arviz/data/base.py
ColCarroll: All dictionaries are pytrees, only some pytrees are dictionaries. I added a little more text to help -- let me know what you think.

ColCarroll:
I think this is ready again.

ahartikainen:
if @OriolAbril approves I think we can merge

OriolAbril:
I want to make a patch release before merging, maybe a couple doc changes, then I'll merge.

ColCarroll:
Sounds good! 

`bayeux` will probably pin to the next release that has this in it, since [this example](https://github.com/jax-ml/bayeux/blob/main/docs/examples/oryx_and_bayeux.ipynb) produces a dictionary of parameters that the current `az.from_dict` fails to parse correctly.

OriolAbril:
I think it should be done now. Docs used to look like this:

Captura de pantalla de 2024-03-13 19-42-15




Now they are this: https://arviz--2291.org.readthedocs.build/en/2291/api/generated/arviz.dict_to_dataset.html

papoteur-mga:
Hello,
This is not a cool dependency for Linux distro. Building `dm-tree` from source needs to download at build time `pybind11`. However, distros prohibit any download at build time, thus the build fails. The build does not use a library provided by the distro.

ColCarroll:
Hi @papoteur-mga  -- thanks for chiming in, but I'm not sure what this means! 

- Which Linux distro? 
- Why is a linux distro building Python libraries from source? 
- Is `dm-tree` impossible to use there? 
- Is this really a `dm-tree` issue?
- Is there a different way we could achieve this functionality. I know we could depend on `jax` and use `jax.tree`, but that seems like a heavier dependency.

papoteur-mga:
> Hi @papoteur-mga -- thanks for chiming in, but I'm not sure what this means!
> 
>     * Which Linux distro?

Mageia

> 
>     * Why is a linux distro building Python libraries from source?

Python packages are included in the distro to be easily installed, in particular as dependency of any application. The most of the packages are tagged as being open source. These packages should then be built to be sure there are really open source. This is also a way to minimize the risk that faulty code is introduced.

> 
>     * Is `dm-tree` impossible to use there?
> 
>     * Is this really a `dm-tree` issue?

Yes, but I finally found a way by patching the cmake configuration files in dm-tree to use shared libraries. https://svnweb.mageia.org/packages/cauldron/python-dm-tree/current/SOURCES/0001-use-system-libraries.patch?view=markup

> 
>     * Is there a different way we could achieve this functionality. I know we could depend on `jax` and use `jax.tree`, but that seems like a heavier dependency.

I have no idea about that.

ColCarroll:
Thank you for the response! Is it right to say you're here to advocate for mageia users, and not (specifically) as a user of statistical diagnostics? 

I have a particular interest in the functionality that `dm-tree` provides, since it allows for running diagnostics in some of [my own work with bayesian neural nets](https://research.google/blog/autobnn-probabilistic-time-series-forecasting-with-compositional-bayesian-neural-networks/). 

It sounds like maybe this is all set from the patch you added? If it isn't, I can try to explore using some other library to achieve my goals (but I suspect it may introduce more/different problems!)

Issue #2059, opened 2022-06-23
Additional LOO functionality


sethaxen:
There are several useful functions that can be used for model assessment and comparison that could be added here. In some cases these are already available in [loo](https://github.com/stan-dev/loo) or soon will be:

- [ ] `loo_expectation(data, values, kwargs...)`: compute expectation of `values` wrt the leave-one-out posteriors using PSIS. e.g. if `values` are likelihoods, this computes pointwise ELPD, though less efficiently and stably than `loo` does. If `values` are posterior predictions, then the result is leave-one-out posterior predictive means. Exists in loo as [`E_loo`](https://mc-stan.org/loo/reference/E_loo.html) and was previously proposed to be added to ArviZ in https://github.com/arviz-devs/arviz/issues/1931#issuecomment-1021834871.
- [ ] `logo(data, groups, var_name, kwargs...)`: compute leave-one-group-out crossvalidation. `groups` is an array of the same shape as the log-likelihood corresponding to the var specified by `var_name`, minus sample dimensions. The unique group values define new dimensions, and log-likelihoods that share the same group are summed together. The docstring should warn that this is more likely to produce Pareto shape errors than LOO, and it will probably fail if there are group-specific parameters.
- [ ] `loo_predictive_error(data, method, kwargs...)`: use `loo_expectation` to compute point estimates of LOO posterior-predictive means, and then use the name or function `method` to compute pointwise and mean predictive error across the data points. Coming to loo: https://github.com/stan-dev/loo/pull/202
- [ ] `loo_crps(idata, scale=False, kwargs...)`: use `loo_expectation` to compute continuous ranked probability score (CRPS) or it's scaled variant SCRPS, which is another strictly proper scoring rule besides Log score (ELPD) to use for model comparison. Requires 2 posterior predictive draws for each posterior draw, so blocked by https://github.com/arviz-devs/arviz/issues/2048. Coming to loo: https://github.com/stan-dev/loo/pull/203

sethaxen:
Good point about `logo`. I agree, if we can document it, and the code only takes a few lines, then it doesn't make sense to add a `logo` implementation for this. Thanks for the example; should be straightforward to adapt.

aadya940:
[...]

PR #2301, opened 2023-12-22
Add loo_expectation

aadya940:
Related to #2059

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -44,6 +44,7 @@ __all__ = [
     "compare",
     "hdi",
     "loo",
+    "loo_expectation",
     "loo_pit",
     "psislw",
     "r2_samples",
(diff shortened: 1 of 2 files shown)


aadya940:
Hello! Can anyone please review this PR and let me know If I'm missing out something?

arviz/stats/stats.py
```diff
@@ -865,6 +866,60 @@ def loo(data, pointwise=None, var_name=None, reff=None, scale=None):
...
+def loo_expectation(data, values, pointwise=None, reff=None, **kwargs):
+    """
+    Computes the expectation of values with respect to the leave-one-out posteriors using PSIS.
+    Parameters
+    ----------
+        data: obj
+            Any object that can be converted to an :class:`arviz.InferenceData` object.
+            Refer to documentation of :func:`arviz.convert_to_dataset` for details.
+        values: ndarray
+            A vector of quantities to compute expectations for.
```
OriolAbril, line 878: We should add information about the shape of `values` here. My understanding is that is should be an array/dataarray with the same shape as the pointwise log likelihood (e.g. `chain, draw, obs_id`) which in general won't work here. I think there should be an extra check for when the input is a dataarray so that `chain, draw` dimensions get stacked into `__sample__` one, otherwise something like:

```
loo_expectation(data, data.posterior_predictive.y)
```

would not work as is.

arviz/stats/stats.py
```diff
@@ -865,6 +866,60 @@ def loo(data, pointwise=None, var_name=None, reff=None, scale=None):
...
+        pointwise: bool, optional
+            If True the pointwise predictive accuracy will be returned. Defaults to
+            ``stats.ic_pointwise`` rcParam.
```
OriolAbril, line 881: This can be removed, it is not used anywhere

arviz/stats/stats.py
```diff
@@ -865,6 +866,60 @@ def loo(data, pointwise=None, var_name=None, reff=None, scale=None):
...
+        reff: float, optional
+            Relative MCMC efficiency, ``ess / n`` i.e. number of effective samples
+            divided by the number of actual samples. Computed from trace by default.
+        **kwargs:
+            Additional keyword arguments to pass to the `psislw` function.
```
OriolAbril, line 886: psislw only takes two arguments, which are already provided explicitly, so passing any kwargs here would end up as a keyword not recognized error when calling psislw.

arviz/stats/stats.py
```diff
@@ -865,6 +866,60 @@ def loo(data, pointwise=None, var_name=None, reff=None, scale=None):
...
+    Returns
+    -------
+        expectation: float
+            The computed expectation of `values` across LOO posteriors.
```
OriolAbril, line 890: This should also have an indication of the expected output shape. From the examples in https://mc-stan.org/loo/reference/E_loo.html it looks like it should have the shape of pointwise log likelihood values minus `__sample__` dimension.

arviz/stats/stats.py
```diff
@@ -865,6 +866,60 @@ def loo(data, pointwise=None, var_name=None, reff=None, scale=None):
...
+            # this mean is over all data variables
+            reff = (
+                np.hstack([ess_p[v].values.flatten() for v in ess_p.data_vars]).mean()
+                / n_samples
+            )
+
+    log_weights, _ = psislw(-log_likelihood, reff=reff, **kwargs)
+
+    # Numerically stable Weighted sum
+    # Do computations in the log-space for numerical stability
```
OriolAbril, line 917: Right before that I would add a check for DataArrays (preferred input type) to see if they have `chain` and `draw` dimensions and if so stack them. Then sum only along the `__sample__` dimension.

arviz/stats/stats.py
```diff
@@ -865,6 +866,60 @@ def loo(data, pointwise=None, var_name=None, reff=None, scale=None):
...
+    w_exp = log_weights + np.log(np.abs(values))
+    _expectation = (np.sign(values) * np.exp(w_exp)).sum()
```
OriolAbril, line 919:
```suggestion
    expectation = (np.sign(values) * np.exp(w_exp)).sum(dim="__sample__")
```

The variable is only defined within the scope of the function, no need to add any underscore to the name.

arviz/tests/base_tests/test_stats.py
```diff
@@ -538,6 +539,13 @@ def test_loo_warning(centered_eight):
         assert loo(centered_eight, pointwise=True) is not None
     assert any("Estimated shape parameter" in str(record.message) for record in records)
 
+@pytest.mark.parametrize("reff", [None, 0.5, 1.0])
+def test_loo_expectation(centered_eight, reff):
+    log_likelihood = get_log_likelihood(centered_eight)
+    log_likelihood = log_likelihood.stack(__sample__=("chain", "draw"))
+    values = np.arange(1, log_likelihood.shape[-1] + 1)
```
OriolAbril, line 546: I think values should the the log likelihood directly, no extra processing or using its shape to create different objects.

OriolAbril (commented):
Really sorry about not reviewing until now. Let me know if still interested, if not I will finish this PR so when merged you get credit for the contribution.

cc @sethaxen I would appreciate an assist on making sure I am not messing up input/output shapes particularly

OriolAbril:
This ended up being superseeded by https://github.com/arviz-devs/arviz-stats/pull/89

Issue #2309, opened 2024-02-05
Improvements and changes to ECDF plots


sethaxen:
## Tell us about it

Here are some proposed improvements to ECDF plots:
- currently the code is quite PIT-focused. e.g. the simulated confidence bands are implemented assuming the target distribution is (discrete) uniform, but this is nowhere documented. We can generalize this method to support other distributions if the user provides both a `cdf` and an `rvs` function for the assumed distribution.
- it is the only function that takes a keyword `fpr`. A more consistent one with our API would be to specify a `prob=1-fpr`.
- we should add the optimized confidence bands from https://doi.org/10.1007/s11222-022-10090-6, which are faster and more stable than the simulated ones.
- we should consider removing `values2`. This is supposedly for ECDF comparison, but it's not as useful as the ECDF comparison plot from the paper, which we should consider making its own plot.
- we should allow the user to specify the evaluation points. The theory behind the confidence bands assumes the evaluation points are independent of the sampled values, and the below notebook shows that setting the evaluation points based on the sampled values can cause the bounds to be slightly too tight. Plus, if one wants to plot many ECDFs in the same plot, it's common one would want them to share evaluation points.
- we should allow the user to provide a pre-computed confidence band. Alternatively, the confidence band could be its own plotting function. In cases like PIT and rank plots where all subplots share the same comparison distribution and evaluation points, one would want to compute the confidence band once and use it for all subplots (even with optimization, this is much more expensive than computing the ECDF)
- When not plotting a band, we should default to using the sampled points as the evaluation points.

## Thoughts on implementation

[This notebook](https://gist.github.com/sethaxen/06c83cace937a19dd55d4b6ccedbec82) implements the bands and tests them on a few distributions. It also compares different methods of selecting the evaluation points.

sethaxen:
Also, when the evaluation points are different from the sample points, it feels a little weird using step plots. Step plots give the sense that we know what the function values are between the points, but unlike the full ECDF, we don't. Line plots aren't much better, but we're more accustomed to lines in lineplots not necessarily implying interpolation. I wonder if we should only reserve step plots for the cases where eval points and sample points are the same.

PR #2311, opened 2024-02-10
Refactor ECDF code

sethaxen:
## Description

This PR is a major refactor of the ECDF plot code to remove dead code, reduce code duplication, use more descriptive variable/function names, and make it easier to make the changes detailed in #2309. It makes no changes to the functionality of `plot_ecdf`.

All utility functions have been moved to `arviz.stats.ecdf_utils`. It might be worth exporting `ecdf_confidence_band`, but this can be left for a future PR. 

## Example

This example shows that public functionality is unchanged.

```python
import numpy as np
import scipy.stats

import arviz as az

np.random.seed(0)

dist = scipy.stats.expon(1)
x = dist.rvs(100)

fig, axs = plt.subplots(1, 3, figsize=(15, 4))
az.plot_ecdf(x, ax=axs[0])
az.plot_ecdf(x, ax=axs[1], confidence_bands=True, cdf=dist.cdf)
az.plot_ecdf(x, ax=axs[2], confidence_bands=True, cdf=dist.cdf, pit=True, difference=True)
```

**On main**
ecdf_old

**This PR**
ecdf_new

## Checklist

----
📚 Documentation preview 📚: https://arviz--2311.org.readthedocs.build/en/2311/


arviz/tests/base_tests/test_stats_ecdf_utils.py
```diff
@@ -0,0 +1,159 @@
...
+    if seed is not None:
+        random_state = np.random.RandomState(seed)
+        ecdf = _simulate_ecdf(ndraws, eval_points, rvs, random_state=random_state)
+        random_state = np.random.RandomState(seed)
+        ecdf_expected = compute_ecdf(np.sort(rvs(ndraws, random_state=random_state)), eval_points)
+    else:
+        np.random.seed(87)
+        ecdf = _simulate_ecdf(ndraws, eval_points, rvs)
+        np.random.seed(87)
+        ecdf_expected = compute_ecdf(np.sort(rvs(ndraws)), eval_points)
```
OriolAbril, line 69: I wouldn't test that, scipy will eventually stop using the global state (see https://github.com/scipy/scipy/issues/14322 for example) in which case this would start breaking for no reason.

arviz/tests/base_tests/test_stats_ecdf_utils.py
```diff
@@ -0,0 +1,159 @@
...
+@pytest.mark.parametrize("seed", [32, None])
+def test_simulate_ecdf(dist, seed):
+    """Test _simulate_ecdf."""
+    ndraws = 1000
+    eval_points = np.arange(0, 1, 0.1)
+
+    rvs = dist.rvs
+
+    if seed is not None:
+        random_state = np.random.RandomState(seed)
```
OriolAbril, line 61: following the other comments, I would swap this for np.random.default_rng(seed) (also in the other tests)

OriolAbril (commented):
This is amazing, thanks!

After this it will probably be possible to use these functions in plot_ppc(kind="cumulative") and in plot_loo_pit instead of each having their own subset of ecdf functionality

arviz/stats/ecdf_utils.py
```diff
@@ -0,0 +1,163 @@
...
+        x = np.insert(x, 0, x[0])
+        y = np.insert(y, 0, 0)
+    return x, y
+
+
+def _simulate_ecdf(
+    ndraws: int,
+    eval_points: np.ndarray,
+    rvs: Callable[[int, Optional[np.random.RandomState]], np.ndarray],
+    random_state: Optional[np.random.RandomState] = None,
```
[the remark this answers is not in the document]
sethaxen: Hmm, this makes the design a bit tricky. Because currently we require the `rvs` function takes a `random_state` keyword, which limits to functions that look like the ones in scipy.stats. But if we wanted to support e.g. `Generator.binom` or `RandomState.binom`, then we would require `rvs` is just a function that takes a sample size as argument and let the user decide how that is seeded. Does this latter option (i.e. not taking a `random_state` keyword at all) make more sense as a design?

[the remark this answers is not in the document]
OriolAbril: scipy `random_state` argument can be None, integers, RandomState or Generator. I don't think anything needs to change as it is passed as is to the scipy distribution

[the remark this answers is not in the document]
OriolAbril: Reference: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rv_continuous.rvs.html#scipy.stats.rv_continuous.rvs

sethaxen:
> After this it will probably be possible to use these functions in plot_ppc(kind="cumulative") and in plot_loo_pit instead of each having their own subset of ecdf functionality

Yeah! I also think it would be great if the plots where one could use `kind` to select the representation of some distribution (e.g. `plot_dist`) also supported `ecdf` (or `cumulative`?) and `dot` options.

I've made all requested changes.

arviz/plots/ecdfplot.py
```diff
@@ -180,75 +181,47 @@ def plot_ecdf(
...
-        lower, higher = get_lims(gamma, n, z)
-
-        ## This block is for whether to plot ECDF or ECDF-difference
-        if not difference:
-            ## We store the coordinates of our ecdf in x_coord, y_coord
-            x_coord, y_coord = get_ecdf_points(x, probs, difference)
+    if pit:
+        eval_points = np.linspace(1 / npoints, 1, npoints)
+        if cdf:
+            sample = np.apply_along_axis(cdf, 0, values)
```
OriolAbril, line 187: I take the apply_along_axis is so `cdf` can take scalar input right? Given the common case I can think of is using the `.cdf` method of scipy distributions, it should be documented, otherwise it might not be used at all. If we forced `cdf` to take array inputs, it should be equivalent to `cdf(values)` as we have already made sure `values` is a 1-dimensional array

OriolAbril (approved):
left one last documentation related comment but looks great

arviz/plots/ecdfplot.py
sethaxen: Yes, correct. The existing code already assumes `cdf` is a scalar function, and the docstring did not state otherwise, so it could be breaking to make `cdf` take a vector. Since this PR is not trying to change existing functionality, I've left it as is. But a future PR will change `cdf` to take array values and document this.

sethaxen: Aaah, okay, I misremembered. It seems it assumes both scalars and arrays are acceptable, e.g. https://github.com/arviz-devs/arviz/blob/0a91fafca89237af681980aa994576873ef03283/arviz/plots/ecdfplot.py#L215-L218. Since we'll only ever need `cdf` to take array values, I'll document this expectation and update the code accordingly.

sethaxen: Done!

arviz/tests/base_tests/test_stats_ecdf_utils.py
```diff
@@ -0,0 +1,153 @@
+import pytest
+
+import numpy as np
+import scipy.stats
+from arviz.stats.ecdf_utils import (
```
OriolAbril, line 5:
```suggestion
from ...stats.ecdf_utils import (
```

I think this is what makes CI fail

PR #2316, opened 2024-02-23
Refactor plot_ecdf arguments

sethaxen:
## Description

This PR introduces new keyword arguments to `plot_ecdf` and deprecates a few existing ones, following suggestions in #2309.

New keywords and features:
- `confidence_bands` now may take string arguments as well as boolean
- `ci_prob` specifies band probability. `stats.ci_prob` is a new rcParam.
- `eval_points` allows the user to specify the evaluation points
- `rvs`, `random_state` can be provided for simulation confidence bands

Deprecated arguments:
- `values2` is now deprecated, in favor of users passing an empirical CDF

Deprecated keywords:
- `pointwise` is now `confidence_bands="pointwise"`
- `fpr` is replaced with `1-ci_prob` for consistency with other plotting functions.
- `pit` is deprecated. We only need this for LOO-PIT, users who need it for something else will probably know how to make the plot, and if we really want to include it, it should be its own plotting function. There's now a documented example of how to plot PIT.
- `npoints`

Deprecated rcparams:
- `stats.hdi_prob` has been deprecated and replaced with `stats.ci_prob`

Additional changes:
- If `eval_points` not provided, there's a warning that in the future `eval_points` will be the unique values of the sample. This would be breaking and is saved for a future release.

None of the changes are breaking.

## Checklist

----
📚 Documentation preview 📚: https://arviz--2316.org.readthedocs.build/en/2316/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### New features
 -  Use revised Pareto k threshold ([2349](https://github.com/arviz-devs/arviz/pull/2349))
 
+-   Added arguments `eval_points`, `rvs`, and `random_state` to `plot_ecdf` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
+
 ### Maintenance and fixes
 - Ensure support with numpy 2.0 ([2321](https://github.com/arviz-devs/arviz/pull/2321))
 - Update testing strategy to include an environment without optional dependencies and
@@ -15,6 +17,8 @@
 ### Deprecation
 -  Support for arrays and DataArrays in plot_khat has been deprecated. Only ELPDdata will be supported in the future ([2349](https://github.com/arviz-devs/arviz/pull/2349))
 
+
+-   Removed arguments `values2`, `fpr`, `pointwise`, and `pit` in `plot_ecdf` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
 
 ### Documentation
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,8 @@
 ### New features
 -  Use revised Pareto k threshold ([2349](https://github.com/arviz-devs/arviz/pull/2349))
 
--   Added arguments `band_prob`, `eval_points`, `rvs`, and `random_state` to `plot_ecdf` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
--   Added rcParam `plot.band_prob` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
+-   Added arguments `ci_prob`, `eval_points`, `rvs`, and `random_state` to `plot_ecdf` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
+-   Added rcParam `stats.ci_prob` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
 
 ### Maintenance and fixes
 - Ensure support with numpy 2.0 ([2321](https://github.com/arviz-devs/arviz/pull/2321))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,12 @@
 - Fix deprecation warnings in multiple dependencies ([2329](https://github.com/arviz-devs/arviz/pull/2329),
   [2332](https://github.com/arviz-devs/arviz/pull/2332) and [2333](https://github.com/arviz-devs/arviz/pull/2333))
 
+### Deprecation
+
+-   Removed arguments `values2`, `fpr`, `pointwise`, `npoints`, and `pit` in `plot_ecdf` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
+
+### Documentation
+
 ## v0.17.1 (2024 Mar 13)
 
 ### Maintenance and fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 ### New features
 -  Use revised Pareto k threshold ([2349](https://github.com/arviz-devs/arviz/pull/2349))
 
--   Added arguments `eval_points`, `rvs`, and `random_state` to `plot_ecdf` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
+-   Added arguments `band_prob`, `eval_points`, `rvs`, and `random_state` to `plot_ecdf` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
+-   Added rcParam `plot.band_prob` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
 
 ### Maintenance and fixes
 - Ensure support with numpy 2.0 ([2321](https://github.com/arviz-devs/arviz/pull/2321))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 -  Use revised Pareto k threshold ([2349](https://github.com/arviz-devs/arviz/pull/2349))
 
 -   Added arguments `ci_prob`, `eval_points`, `rvs`, and `random_state` to `plot_ecdf` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
--   Added rcParam `stats.ci_prob` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
+-   Deprecated rcParam `stats.hdi_prob` and replaced with `stats.ci_prob` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
 
 ### Maintenance and fixes
 - Ensure support with numpy 2.0 ([2321](https://github.com/arviz-devs/arviz/pull/2321))
--- a/arviz/plots/ecdfplot.py
+++ b/arviz/plots/ecdfplot.py
@@ -22,7 +22,7 @@ def plot_ecdf(
     cdf=None,
     difference=False,
     confidence_bands=False,
-    band_prob=None,
+    ci_prob=None,
     num_trials=500,
     rvs=None,
     random_state=None,
@@ -77,7 +77,7 @@ def plot_ecdf(
 
         For simultaneous confidence bands to be correctly calibrated, provide `eval_points` that
         are not dependent on the `values`.
-    band_prob : float, default 0.95
+    ci_prob : float, default 0.95
         The probability that the true ECDF lies within the confidence band. If `confidence_bands`
         is "pointwise", this is the marginal probability instead of the joint probability.
     eval_points : array-like, optional
@@ -127,7 +127,7 @@ def plot_ecdf(
     fpr : float, optional
 
         .. deprecated:: 0.18.0
-           Instead use `band_prob=1-fpr`.
+           Instead use `ci_prob=1-fpr`.
     pit : bool, default False
         If True plots the ECDF or ECDF-diff of PIT of sample.
 
@@ -213,16 +213,16 @@ def plot_ecdf(
 
     if fpr is not None:
         warnings.warn(
-            "`fpr` has been deprecated. Use `band_prob=1-fpr` or set `rcParam['plot.band_prob']` to"
+            "`fpr` has been deprecated. Use `ci_prob=1-fpr` or set `rcParam['stats.ci_prob']` to"
             "`1-fpr`.",
             DeprecationWarning,
         )
-        if band_prob is not None:
-            raise ValueError("Cannot specify both `fpr` and `band_prob`")
-        band_prob = 1 - fpr
+        if ci_prob is not None:
+            raise ValueError("Cannot specify both `fpr` and `ci_prob`")
+        ci_prob = 1 - fpr
 
-    if band_prob is None:
-        band_prob = rcParams["plot.band_prob"]
+    if ci_prob is None:
+        ci_prob = rcParams["stats.ci_prob"]
 
     if values2 is not None:
         if cdf is not None:
@@ -294,7 +294,7 @@ def plot_ecdf(
             eval_points,
             cdf_at_eval_points,
             method=confidence_bands,
-            prob=band_prob,
+            prob=ci_prob,
             num_trials=num_trials,
             rvs=rvs,
             random_state=random_state,
--- a/arviz/plots/ecdfplot.py
+++ b/arviz/plots/ecdfplot.py
@@ -162,7 +162,7 @@ def plot_ecdf(
 
         >>> az.plot_ecdf(sample, cdf = distribution.cdf, eval_points = eval_points,
         >>>              confidence_bands = True, difference = True)
-        
+
     Plot an ECDF plot with confidence bands for the probability integral transform (PIT) of a continuous
     sample. If drawn from the reference distribution, the PIT values should be uniformly distributed.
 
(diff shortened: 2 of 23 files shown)


arviz/plots/ecdfplot.py
```diff
@@ -10,14 +12,16 @@
 def plot_ecdf(
     values,
     values2=None,
+    eval_points=None,
```
sethaxen, line 15: Is there a name we use elsewhere in the codebase that we should instead use here for consistency?

arviz/plots/ecdfplot.py
```diff
@@ -46,26 +52,41 @@ def plot_ecdf(
...
-    pointwise : bool, default False
-        If True plots pointwise confidence bands otherwise simultaneous bands.
+    confidence_bands : str or bool, optional
+        - False: No confidence bands are plotted.
+        - "pointwise": Compute the pointwise (i.e. marginal) confidence band.
+        - True or "simulated": Use Monte Carlo simulation to estimate a simultaneous confidence
+          band.
+        For simultaneous confidence bands to be correctly calibrated, provide `eval_points` that
+        are not dependent on the `values`.
+    band_prob : float, default 0.95
```
sethaxen, line 69: I'd argue that this should default to 0.94 for consistency with `hdi_prob`, but that would be breaking (since `fpr` was 0.05).

arviz/rcparams.py
```diff
@@ -262,6 +262,7 @@ def validate_iterable(value):
         "mean",
         _make_validate_choice({"mean", "median", "mode"}, allow_none=True),
     ),
+    "plot.band_prob": (0.95, _validate_probability),
```
sethaxen, line 265: Potentially this should be `stats.band_prob`, especially if we add the ECDF confidence band computation functions to the API.

arviz/stats/ecdf_utils.py
```diff
@@ -90,7 +90,7 @@ def ecdf_confidence_band(
         A function that takes an integer `ndraws` and optionally the object passed to
         `random_state` and returns an array of `ndraws` samples from the same distribution
         as the original dataset. Required if `method` is "simulated" and variable is discrete.
-    num_trials : int, default 1000
+    num_trials : int, default 500
```
sethaxen, line 93: Just changed for consistency with original behavior.

arviz/plots/ecdfplot.py
```diff
@@ -46,26 +52,41 @@ def plot_ecdf(
     values : array-like
         Values to plot from an unknown continuous or discrete distribution.
     values2 : array-like, optional
-        Values to compare to the original sample.
+        deprecated: values to compare to the original sample. Instead use
+        `cdf=scipy.stats.ecdf(values2).cdf.evaluate`.
```
sethaxen, line 56: I hope in a future PR to add an ECDF-(difference)plot option to `plot_ranks` and then recommend that here.

arviz/plots/ecdfplot.py
```diff
@@ -46,26 +52,41 @@ def plot_ecdf(
...
-        If True plots the ECDF or ECDF-diff of PIT of sample.
-    confidence_bands : bool, default None
-        If True plots the simultaneous or pointwise confidence bands with `1 - fpr`
-        confidence level.
-    pointwise : bool, default False
-        If True plots pointwise confidence bands otherwise simultaneous bands.
+    confidence_bands : str or bool, optional
+        - False: No confidence bands are plotted.
+        - "pointwise": Compute the pointwise (i.e. marginal) confidence band.
+        - True or "simulated": Use Monte Carlo simulation to estimate a simultaneous confidence
```
sethaxen, line 65: In the next PR I will add deterministic bands, which will become the default here.

sethaxen:
@OriolAbril when you get a chance, can you take a look at this?

arviz/plots/ecdfplot.py
```diff
@@ -46,26 +57,41 @@ def plot_ecdf(
...
-        The number of random ECDFs to generate for constructing simultaneous confidence bands.
-    fpr : float, default 0.05
-        The type I error rate s.t `1 - fpr` denotes the confidence level of bands.
+        The number of random ECDFs to generate for constructing simultaneous confidence bands
+        (if `confidence_bands` is "simulated").
+    random_state : {None, int, `numpy.random.Generator`,
+                    `numpy.random.RandomState`}, optional
+        If `None`, the `numpy.random.RandomState` singleton is used. If an `int`, a new
+        ``numpy.random.RandomState`` instance is used, seeded with seed. If a `RandomState` or
+        `Generator` instance, the instance is used.
```
OriolAbril, line 94: I would focus on the argument being passed to `rvs` function, all of what is described only happens for rvs functions that come from scipy distribution classes it isn't really done by any of the ArviZ functions.

OriolAbril (commented):
Sorry for the slowness, still a bit fuzzy on some of the warnings, I'll try to use the PR tomorrow to make sure I see what happens given the different combinations.

CHANGELOG.md
```diff
@@ -4,12 +4,17 @@
 
 ### New features
 
+-   Added arguments `band_prob`, `eval_points`, `rvs`, and `random_state` to `plot_ecdf` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
+-   Added rcParam `plot.band_prob` ([2316](https://github.com/arviz-devs/arviz/pull/2316))
```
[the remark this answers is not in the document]
sethaxen: Okay, so to clarify, is the idea that we replace `stats.hdi_prob` and the new `plot.band_prob` with a new `stats,ci_prob`? If so, is there a procedure for deprecating rcparams so that users can get an informative warning if they try to set `hdi_prob`?

arviz/plots/ecdfplot.py
sethaxen: Can you clarify what change you are suggesting here?

arviz/plots/ecdfplot.py
```diff
@@ -103,129 +136,172 @@ def plot_ecdf(
...
-    else:
+        eval_points = np.linspace(1 / npoints, 1, npoints)
+
+    if eval_points is None:
+        warnings.warn(
+            "In future versions, if `eval_points` is not provided, then the ECDF will be evaluated"
+            " at the unique values of the sample. To keep the current behavior, provide "
+            "`eval_points` explicitly.",
+            FutureWarning,
+        )
```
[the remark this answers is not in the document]
sethaxen: Ah, yeah I guess in the future there will be no case where `npoints` is used, so might as well deprecate it now.

arviz/plots/ecdfplot.py
```diff
@@ -46,26 +57,41 @@ def plot_ecdf(
     values : array-like
         Values to plot from an unknown continuous or discrete distribution.
     values2 : array-like, optional
-        Values to compare to the original sample.
+        deprecated: values to compare to the original sample. Instead use
+        `cdf=scipy.stats.ecdf(values2).cdf.evaluate`.
```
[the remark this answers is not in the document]
sethaxen: Looks good! (needed to change `deprecated:` to `deprecated::`)

CHANGELOG.md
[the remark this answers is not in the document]
OriolAbril: I don't think we have ever changed an rcParam key, only a couple values. We should definitely add a deprecation warning and keep both around for a bit.

A proposal could be to have ci_prob behave like current hdi_prob and hdi_prob now instead takes an extra value: `ci_prob` (new default). If the value for hdi_prob is different than that then raise a FutureWarning. I think we can achieve that with a custom validation function relatively easily.

arviz/plots/ecdfplot.py
OriolAbril: ArviZ doesn't really do most of what is described here, behaviour within ArviZ is limited to taking `random_state` and passing it as is to `rvs` function. I would limit the description to that.

Otherwise, if I were to write an rvs function from scratch, I'd make it accept only RandomState of Generator, and expect ArviZ to convert `None` and ints to RandomState before passing the argument to my custom rvs function.

All the exerpt below is scipy behaviour and will be true if `rvs` argument is the method `.rvs` of a scipy distribution but not otherwise, and `rvs` argument isn't limited to `.rvs` method of scipy distributions:

> If `None`, the `numpy.random.RandomState` singleton is used. If an `int`, a new ``numpy.random.RandomState`` instance is used, seeded with seed.

CHANGELOG.md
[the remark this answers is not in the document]
sethaxen: Here's what I did, roughly patterned after matplotlibs's own handing of deprecated rcparams: https://github.com/arviz-devs/arviz/pull/2316/commits/f6cff760dfe2850305792ea214bdeedfae1b10b1

Effectively, the `hdi_prob` rcparam is now an alias for `ci_prob`, which has a default. Anytime someone sets or gets `hdi_prob`, a deprecation warning is raised. The implementation is flexible enough to support in the future more complicated deprecations. At the moment this raises plenty of deprecation warnings in our tests/docs builds, since `hdi_prob` is regularly accessed. But before changing any of those things, It'd be nice to get feedback on this approach.

arviz/plots/ecdfplot.py
sethaxen: Okay, I modified the implementation to support the signature `rvs(ndraws)` if `random_state` is `None` (default). Then it's just documented that one can provide a `random_state` that the provided `rvs` accepts.

CHANGELOG.md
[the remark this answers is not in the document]
OriolAbril: This is great, much better than my proposal. The only comment is the emitted warnings should be `FutureWarning` (user facing) instead of `DeprecationWarning` (downstream dev facing). I don't really know how or why but this is Python convention and by default users don't even see `DeprecationWarning` unless explicitly activated. Ref https://docs.python.org/3/library/warnings.html#warning-categories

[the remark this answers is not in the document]
sethaxen: Ah, good to know, thanks! Done.

sethaxen:
@OriolAbril I've implemented all suggestions and updated the above description and changelog. This should be ready for final review.

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -1286,6 +1307,75 @@ def test_plot_ecdf_cdf():
...
+    dist = norm(0, 1)
+    data = dist.rvs(1000)
+    # base case, no deprecations
+    with does_not_warn(DeprecationWarning):
+        axes = plot_ecdf(data, cdf=dist.cdf, confidence_bands=True)
+    assert axes is not None
+
+    # values2 is deprecated
+    data2 = dist.rvs(200)
+    with pytest.deprecated_call():
```
OriolAbril, line 1357:
```suggestion
    with pytest.warns(FutureWarning):
```

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -1286,6 +1307,75 @@ def test_plot_ecdf_cdf():
...
+        axes = plot_ecdf(data, values2=data2, difference=True)
+
+    # pit is deprecated
+    with pytest.deprecated_call():
```
OriolAbril, line 1361:
```suggestion
    with pytest.warns(FutureWarning):
```

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -1286,6 +1307,75 @@ def test_plot_ecdf_cdf():
...
+        axes = plot_ecdf(data, cdf=dist.cdf, pit=True)
+    assert axes is not None
+
+    # fpr is deprecated
+    with does_not_warn(DeprecationWarning):
+        axes = plot_ecdf(data, cdf=dist.cdf, ci_prob=0.9)
+    with pytest.deprecated_call():
```
OriolAbril, line 1368:
```suggestion
    with pytest.warns(FutureWarning):
```

arviz/tests/base_tests/test_plots_matplotlib.py
```diff
@@ -1286,6 +1307,75 @@ def test_plot_ecdf_cdf():
...
+        axes = plot_ecdf(data, cdf=dist.cdf, confidence_bands=True, fpr=0.1)
+    assert axes is not None
+
+    # pointwise is deprecated
+    with does_not_warn(DeprecationWarning):
+        axes = plot_ecdf(data, cdf=dist.cdf, confidence_bands="pointwise")
+    with pytest.deprecated_call():
```
OriolAbril, line 1375:
```suggestion
    with pytest.warns(FutureWarning):
```

OriolAbril (approved):
Missed some DeprecationWarnings that are user-facing. I'll batch commit the suggestions and merge, all the changes are extremely minor

arviz/plots/ecdfplot.py
```diff
@@ -104,129 +148,172 @@ def plot_ecdf(
...
     if pit:
-        eval_points = np.linspace(1 / npoints, 1, npoints)
-        if cdf:
-            sample = cdf(values)
-        else:
-            sample = compute_ecdf(values2, values) / len(values2)
-        cdf_at_eval_points = eval_points
+        warnings.warn(
+            "`pit` has been deprecated. Specify `values=cdf(values)` instead.",
+            DeprecationWarning,
```
OriolAbril, line 258:
```suggestion
            FutureWarning,
```

OriolAbril (commented):
After seeing the warnings and trying it out I am a bit on the fence on the behaviour of `eval_points`. It is basically a required argument right now, otherwise you get a FutureWarning.

It would be nice to continue allowing plot_ecdf(samples) to work without warnings.

sethaxen:
> After seeing the warnings and trying it out I am a bit on the fence on the behaviour of `eval_points`. It is basically a required argument right now, otherwise you get a FutureWarning.
> 
> It would be nice to continue allowing plot_ecdf(samples) to work without warnings.

The reason it raises a `FutureWarning` is because in the future `plot_ecdf(samples)` will do something entirely different than it does now. The alternative is that we currently don't raise a warning and instead change the behavior in the future without warning. Personally, I think the way we do it here could be less jarring.

OriolAbril:
Maybe we could create a specific warning class for this? Something like `BehaviourChangeWarning`, I think it might signal better what is happening and also make it easier to silence (I am quite sure many users don't really care about the default as long as it works) and `plot_ecdf(samples)` will continue to work.

We could also silence it in the examples of the docstring. Now all examples use eval_points, but to illustrate how to generate confidence bands or how to make it a difference plot it doesn't matter which is the default behaviour of eval_points (and tests don't use it). So we could use the first example to describe the behaviour change, show how to maintain old behaviour and then silence the warning so following examples focus on what they want to illustrate without worrying about the warning.

What do you think?

OriolAbril:
@sethaxen I have tried out the special warning and added the filter to the docs. Now we should make sure all examples in the docstring don't trigger any warning, I have to go now, so leaving this here so when I come back later I can check the docs preview instead of locally building it myself at some point

OriolAbril:
Should be ready to merge now. There is one example in the docstring that triggers a warning, the one for 

> Plot an ECDF plot with confidence bands for comparing a given sample to a given distribution. We manually specify evaluation points independent of the values so that the confidence bands are correctly calibrated.

because `rvs` is not provided, I think this is only temporal though, is that right? Once the new default is available there won't be a warning when using that same code snippet

Issue #2313, opened 2024-02-21
Make use of numba opt-in not opt-out


ricardoV94:
Calling `forest_plot` when numba is installed ends up with cryptic errors: `ValueError: Failed in object mode pipeline (step: remove phis nodes)` This is a PITA.

Please consider making the use of numba opt-in (not simply by having numba installed).

tomicapretto:
If we wanted to do this, I think (almost?) everything is already there. We just need to make sure that

https://github.com/arviz-devs/arviz/blob/7c1637f6d4574bc8ba5757165f9e7bb02e202621/arviz/utils.py#L200

is not `True` by default when numba is installed (which is the case now)

OriolAbril:
Sounds good to me, the only change would be in https://github.com/arviz-devs/arviz/blob/7c1637f6d4574bc8ba5757165f9e7bb02e202621/arviz/utils.py#L328 make it `False` instead of the result of `numba_check`.

I'll also look into the error, it might not change much between using numba there or not if its in object mode.

drbenvincent:
+1: I'm hitting this error as well

PR #2318, opened 2024-03-06
Change `numba_flag` to `False`, #2313

drbenvincent:
#2313 outlined a bug when using `plot_forest`. In that issue @OriolAbril proposed a fix.

This PR implements this 1 line fix.

Previously I was also running into this bug, but can confirm that the bug is fixed with this change.


juanitorduz:
Thanks! I had been having this one as well 😅 !

PR #2319, opened 2024-03-11
remove conditional_jit from plot_forest label_idx

OriolAbril:
## Description
Related to #2313. This should fix the bug mentioned in that issue.
With the recent improvements to numba it has become more performant yet more restrictive
to support the performance improvements.

plot_forest had a function that was already using forceobj=True, and now due to the
combination of lists, dicts and .append, numba rejects compiling it (probably wasn't doing much
to begin with in previous versions). So I have removed the conditional_jit part.

## Checklist

- [x] Is the fix listed in the [Maintenance and fixes](https://github.com/arviz-devs/arviz/blob/main/CHANGELOG.md#maintenance-and-fixes)
      section of the changelog?

----
📚 Documentation preview 📚: https://arviz--2319.org.readthedocs.build/en/2319/


drbenvincent:
I can confirm that this fixes the bug I was encountering (#2313)

PR #2338, opened 2024-04-04
Add a close method to InferenceData

### LINKED ISSUE #2337: Add `InferenceData.close` method
## Tell us about it

When an `InferenceData` object is read from a file, `xarray` will try to lazily load the underlying `Dataset` objects keeping an open connection to the underlying file in disk. `xarray.Dataset` provide the [`close` method](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.close.html) to release any resources attached to the open dataset, but `arviz` does not have something like that for `InferenceData` objects. That makes it quite complicated to actually release resources if one might need to overwrite the contents of the file that was originally read from disk.

lucianopaz:
Closes #2337 

## Description
Adds a `close` method to `arviz.InferenceData` that simply extends from `xarray.Dataset.close`.

## Checklist

----
📚 Documentation preview 📚: https://arviz--2338.org.readthedocs.build/en/2338/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Add new example data `rugby_field` and update `rugby` example data ([2322](https://github.com/arviz-devs/arviz/pull/2322))
 - Support for `pytree`s and robust to nested dictionaries. ([2291](https://github.com/arviz-devs/arviz/pull/2291))
+- Add `.close` method to `InferenceData` ([2338](https://github.com/arviz-devs/arviz/pull/2338))
 
 
 ### Maintenance and fixes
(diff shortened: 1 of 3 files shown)


michaelosthege:
Would it help to call this in the objects destructor method automatically?

```python
def __del__(self):
    self.close()
```

lucianopaz:
> Would it help to call this in the objects destructor method automatically?
> 
> ```python
> def __del__(self):
>     self.close()
> ```

Yes, I think so. I'll add a commit that does that.

OriolAbril:
This seems to be generating an infinite loop. It's probably best to not call it at delete time. It should also be added to https://github.com/arviz-devs/arviz/blob/main/doc/source/api/inference_data.rst?plain=1#L26

OriolAbril (approved):

PR #2349, opened 2024-05-27
Use revised Pareto k threshold

aloctavodia:
The newest revision of the PSIS paper now recommends:

* "good" ---------> k < min(1 - 1/log10(S), 0.7) 
* "bad" -----------> min(1 - 1/log10(S), 0.7) <= k < 1 
* "very bad"  ---> k > 1

----
📚 Documentation preview 📚: https://arviz--2349.org.readthedocs.build/en/2349/

### DIFF
--- a/arviz/plots/backends/bokeh/khatplot.py
+++ b/arviz/plots/backends/bokeh/khatplot.py
@@ -21,7 +21,7 @@ def plot_khat(
     figsize,
     xdata,
     khats,
-    sample_size,
+    good_k,
     kwargs,
     threshold,
     coord_labels,
@@ -55,10 +55,8 @@ def plot_khat(
     if hlines_kwargs is None:
         hlines_kwargs = {}
 
-    if sample_size is None:
+    if good_k is None:
         good_k = 0.7
-    else:
-        good_k = min(1 - 1 / np.log10(sample_size), 0.7)
 
     hlines_kwargs.setdefault("hlines", [0, good_k, 1])
 
@@ -137,7 +135,7 @@ def plot_khat(
     xmax = len(khats)
 
     if show_bins:
-        bin_edges = np.array([ymin, threshold, 1, ymax])
+        bin_edges = np.array([ymin, good_k, 1, ymax])
         bin_edges = bin_edges[(bin_edges >= ymin) & (bin_edges <= ymax)]
         hist, _, _ = histogram(khats, bin_edges)
         for idx, count in enumerate(hist):
--- a/arviz/plots/backends/bokeh/khatplot.py
+++ b/arviz/plots/backends/bokeh/khatplot.py
@@ -21,6 +21,7 @@ def plot_khat(
     figsize,
     xdata,
     khats,
+    sample_size,
     kwargs,
     threshold,
     coord_labels,
@@ -53,7 +54,13 @@ def plot_khat(
 
     if hlines_kwargs is None:
         hlines_kwargs = {}
-    hlines_kwargs.setdefault("hlines", [0, 0.5, 0.7, 1])
+
+    if sample_size is None:
+        good_k = 0.7
+    else:
+        good_k = min(1 - 1 / np.log10(sample_size), 0.7)
+
+    hlines_kwargs.setdefault("hlines", [0, good_k, 1])
 
     cmap = None
     if isinstance(color, str):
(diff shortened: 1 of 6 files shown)


arviz/plots/backends/bokeh/khatplot.py
```diff
@@ -130,7 +137,7 @@ def plot_khat(
     xmax = len(khats)
 
     if show_bins:
-        bin_edges = np.array([ymin, 0.5, 0.7, 1, ymax])
+        bin_edges = np.array([ymin, threshold, 1, ymax])
```
OriolAbril, line 140:
```suggestion
        bin_edges = np.array([ymin, good_k, 1, ymax])
```

The goal for this is to count how many values lie within each region visually separated with hlines, so it should match the horizontal lines being plotted. It should also be modified in the matplotlib version

arviz/plots/khatplot.py
```diff
@@ -165,12 +163,19 @@ def plot_khat(
         color = "C0"
 
     if isinstance(khats, np.ndarray):
+        _log.warning(
```
OriolAbril, line 166: I think this should be `FutureWarning` instead

aloctavodia:
"good_k" value is computed once and stored in the ELPDData. I am open to suggestions for a better name.

arviz/plots/khatplot.py
```diff
@@ -165,12 +164,20 @@ def plot_khat(
...
+        )
         khats = khats.flatten()
         xlabels = False
         legend = False
         dims = []
+        good_k = None
     else:
         if isinstance(khats, ELPDData):
+            good_k = khats.good_k
             khats = khats.pareto_k
```
OriolAbril, line 181:
```python
        if isinstance(khats, ELPDData):
            good_k = khats.good_k
            khats = khats.pareto_k
        else:
            good_k = None
            warnings.warn()
```

This should be something like this instead. Right now, dataarrays are also valid input as they are array-like, but we have more info than we do for numpy arrays, so they are treated more similarly to elpddata input. I think this is a reason for some of the test failures.

Note: also rebase on main to avoid unrelated failures that have been fixed already.

arviz/stats/stats.py
```diff
@@ -785,13 +786,15 @@ def loo(data, pointwise=None, var_name=None, reff=None, scale=None):
...
+    good_k = min(1 - 1 / np.log10(n_samples), 0.7)
+
+    if np.any(pareto_shape > good_k):
         warnings.warn(
-            "Estimated shape parameter of Pareto distribution is greater than 0.7 for "
-            "one or more samples. You should consider using a more robust model, this is because "
-            "importance sampling is less likely to work well if the marginal posterior and "
-            "LOO posterior are very different. This is more likely to happen with a non-robust "
-            "model and highly influential observations."
+            f"Estimated shape parameter of Pareto distribution is greater than {good_k:.1f} "
```
OriolAbril, line 793: I'd probably show 2 digits at least, otherwise it might be confusing to people when checking the plot so see why 0.68 is counted as bad if the printed value is 0.7 (but the good_k value is actually 0.66). The plot generally allows for more than one digit precision lookouts, all plots in https://python.arviz.org/en/stable/api/generated/arviz.plot_khat.html for example do.

arviz/stats/stats_utils.py
```diff
@@ -454,10 +454,9 @@ def get_log_likelihood(idata, var_name=None, single_var=True):
 
 Pareto k diagnostic values:
                          {{0:>{0}}} {{1:>6}}
-(-Inf, 0.5]   (good)     {{2:{0}d}} {{6:6.1f}}%
- (0.5, 0.7]   (ok)       {{3:{0}d}} {{7:6.1f}}%
-   (0.7, 1]   (bad)      {{4:{0}d}} {{8:6.1f}}%
-   (1, Inf)   (very bad) {{5:{0}d}} {{9:6.1f}}%
+(-Inf, {{8:.1f}}]   (good)     {{2:{0}d}} {{5:6.1f}}%
```
OriolAbril, line 457: same comment here

arviz/plots/khatplot.py
```diff
@@ -165,15 +164,29 @@ def plot_khat(
...
         if not isinstance(khats, DataArray):
-            raise ValueError("Incorrect khat data input. Check the documentation")
```
OriolAbril, line 176: The valueerror is a different if altogether, not the else branch of the one above. It is reached if the input isn't one of numpy array, dataarray or elpddata, or if the elpddata for some reason doesn't store the khat data as a dataarray

OriolAbril (approved):
I think it's good now, we can merge once test pass. If you can it would be nice to add a line to the changelog too

PR #2368, opened 2024-08-19
Add optimized simultaneous ECDF bands

sethaxen:
## Description

This pull request adds the optimization method from https://doi.org/10.1007/s11222-022-10090-6 for computing simultaneous ECDF bands and ~switches to using it as the default approach~ uses it as the default approach sometimes based on a heuristic. 

Part of #2309. The changes are non-breaking, since the docstring of `plot_ecdf` intentionally does not document which method is the default.

## Example

```python
import numpy as np
import scipy.stats

import arviz as az

np.random.seed(0)

dist = scipy.stats.expon()
x = dist.rvs(1000)

eval_points = np.linspace(0, 10, 100)

fig, axs = plt.subplots(2, 1, figsize=(8, 5))
%time az.plot_ecdf(x, ax=axs[0], eval_points=eval_points, confidence_bands=True, cdf=dist.cdf, difference=True)
%time az.plot_ecdf(x, ax=axs[1], eval_points=eval_points, confidence_bands='simulated', cdf=dist.cdf, difference=True)
axs[0].set_title('Optimized confidence bands')
axs[1].set_title('Simulated confidence bands')
plt.tight_layout()
```

```
CPU times: user 261 ms, sys: 116 μs, total: 261 ms
Wall time: 261 ms
CPU times: user 96.2 ms, sys: 32 μs, total: 96.3 ms
Wall time: 96.3 ms
```

tmp

Note that while optimized confidence bands are more stable than simulated, with this implementation they are not necessarily faster. Perhaps that can be improved using Numba.

## Checklist

      PR format?
      section of the changelog?

----
📚 Documentation preview 📚: https://arviz--2368.org.readthedocs.build/en/2368/

### DIFF
--- a/arviz/stats/ecdf_utils.py
+++ b/arviz/stats/ecdf_utils.py
@@ -12,10 +12,10 @@ try:
     from numba import jit, vectorize
 except ImportError:
 
-    def jit(*args, **kwargs):
+    def jit(*args, **kwargs):  # pylint: disable=unused-argument
         return lambda f: f
 
-    def vectorize(*args, **kwargs):
+    def vectorize(*args, **kwargs):  # pylint: disable=unused-argument
         return lambda f: f
 
 
--- a/arviz/stats/ecdf_utils.py
+++ b/arviz/stats/ecdf_utils.py
@@ -1,11 +1,25 @@
 """Functions for evaluating ECDFs and their confidence bands."""
 
+import math
 from typing import Any, Callable, Optional, Tuple
 import warnings
 
 import numpy as np
 from scipy.stats import uniform, binom
 from scipy.optimize import minimize_scalar
+
+try:
+    from numba import jit, vectorize
+except ImportError:
+
+    def jit(*args, **kwargs):
+        return lambda f: f
+
+    def vectorize(*args, **kwargs):
+        return lambda f: f
+
+
+from ..utils import Numba
 
 
 def compute_ecdf(sample: np.ndarray, eval_points: np.ndarray) -> np.ndarray:
@@ -141,8 +155,7 @@ def _update_ecdf_band_interior_probabilities(
     prob_left: np.ndarray,
     interval_left: np.ndarray,
     interval_right: np.ndarray,
-    cdf_left: float,
-    cdf_right: float,
+    p: float,
     ndraws: int,
 ) -> np.ndarray:
     """Update the probability that an ECDF has been within the envelope including at the current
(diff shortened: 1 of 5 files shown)


arviz/stats/ecdf_utils.py
```diff
@@ -129,6 +137,90 @@ def ecdf_confidence_band(
...
+        previous points are in the interior.
+    """
+    z_tilde = (cdf_right - cdf_left) / (1 - cdf_left)
+    interval_left = interval_left[:, np.newaxis]
+    prob_conditional = binom.pmf(interval_right, ndraws - interval_left, z_tilde, loc=interval_left)
+    prob_right = prob_left.dot(prob_conditional)
+    return prob_right
+
+
+def _ecdf_band_optimization_objective(
```
sethaxen, line 180: This should ideally be jitted with numba if available for a likely significant speed-up. Currently the problem is that the functions it calls use `scipy.stats.binom.interval` and `scipy.stats.binom.pmf`, which are not JIT-able. If we reimplement these functions here, then we could probably get a significant speed-up.

sethaxen: `interval` calls `ppf`, which defers to Boost's implementation, which in turn uses TOMS Algorithm 748 to find a root in an interval: https://doi.org/10.1145/210089.210111, so reimplementing that to be JIT-able would be the bottleneck. Which seems like it could be a lot of work.

sethaxen: Ah wait, since `interval` is called outside the for loop and only `pmf` is called within, this could be simple.

`@conditional_jit` doesn't seem to like it if I call another conditionally-jitted function within the first conditionally-jitted function. Is there a way to make this work?

sethaxen:
> I'll need to think more about whether this would work for arbitrary ECDFs or only for cases where the CDF at eval points is a uniform grid between 0 and 1 (as is the case for uniform eval points for uniform distribution).

Yeah, that's the only context where this makes sense, which is the same context as PIT plots. Ergo, I'll leave interpolation for a separate PR, and in this PR I'll use a benchmark to heuristically identify when we should switch from optimized bands to simulated as the default in `plot_ecdf`.

OriolAbril (approved):
changes look good, not sure if you still need help with numba, if so it wasn't clear to me what wasn't working. Do you have more changes planned for ecdf functionality? I wanted to test porting it to `arviz-stats` too

Issue #2369, opened 2024-08-20
Issue in the documentation (https://python.arviz.org/en/latest/contributing/index.html)


varuntotakura:
**Description**
In the documentation [Contributions](https://python.arviz.org/en/latest/contributing/index.html) page, when I tried to open the pull request hyperlink in the **"Write documentation"** and **"Make changes in the code"** sections, it is redirecting to [https://www.pymc.io/projects/docs/en/stable/contributing/pr_tutorial.html#pr-tutorial](https://www.pymc.io/projects/docs/en/stable/contributing/pr_tutorial.html#pr-tutorial) page, but in actual it should redirect to [https://python.arviz.org/en/latest/contributing/pr_tutorial.html#pr-tutorial](https://python.arviz.org/en/latest/contributing/pr_tutorial.html#pr-tutorial).

**To Reproduce**
Open the pull request hyperlink in the **"Write documentation"** and **"Make changes in the code"** sections, it is redirecting to [https://www.pymc.io/projects/docs/en/stable/contributing/pr_tutorial.html#pr-tutorial](https://www.pymc.io/projects/docs/en/stable/contributing/pr_tutorial.html#pr-tutorial) page

**Expected behavior**
It should redirect to [https://python.arviz.org/en/latest/contributing/pr_tutorial.html#pr-tutorial](https://python.arviz.org/en/latest/contributing/pr_tutorial.html#pr-tutorial).

PR #2370, opened 2024-08-20
Fix the issue in the Contribution References Documentation

varuntotakura:
## Description
In the documentation [Contributions](https://python.arviz.org/en/latest/contributing/index.html) page, when I tried to open the pull request hyperlink in the **"Write documentation"** and **"Make changes in the code"** sections, it is redirecting to [https://www.pymc.io/projects/docs/en/stable/contributing/pr_tutorial.html#pr-tutorial](https://www.pymc.io/projects/docs/en/stable/contributing/pr_tutorial.html#pr-tutorial) page, but in actual it should redirect to [https://python.arviz.org/en/latest/contributing/pr_tutorial.html#pr-tutorial](https://python.arviz.org/en/latest/contributing/pr_tutorial.html#pr-tutorial)

Link to issue: [Issue in the documentation](https://github.com/arviz-devs/arviz/issues/2369)

## Checklist

- [X] New features are properly documented

----
📚 Documentation preview 📚: https://arviz--2370.org.readthedocs.build/en/2370/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Deprecation
 
 ### Documentation
+- Fixed the issue in the [Contribution References Documentation](https://python.arviz.org/en/latest/contributing/index.html) ([2369](https://github.com/arviz-devs/arviz/issues/2369))
 
 ## v0.19.0 (2024 Jul 19)
 
--- a/doc/source/contributing/index.md
+++ b/doc/source/contributing/index.md
@@ -1,7 +1,7 @@
 (contributing_guide)=
 # Contributing to ArviZ
 ## Welcome
-Welcome to the [ArviZ](https://www.arviz.org/en/latest/index.html) project!
+Welcome to the {doc}`ArviZ <arviz_org:index>` project!
 If you’re reading this guide, that probably means you’d like to get involved
 and start contributing to the project.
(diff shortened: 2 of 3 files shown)


OriolAbril (changes requested):
Thanks for catching the wrong link! It was an unwanted side effect of intersphinx looking at all the documentation sites available via `intersphinx_mapping`. As the requested target didn't exist in the ArviZ docs but it did in the PyMC ones, it ended up linking to the pymc docs.

However, we don't want to stop using cross references, to fix the issue we need to fix that particular references using the non-existent `pr_tutorial` target. The reason for that is that if we used urls instead of cross references we would not be able to have multi-versioned docs, every time a link that should be internal is clicked we'd always end up in `latest` version. But if we are browsing the docs for version 0.17 and we click on internal links we want to get to the linked page, but still for the same version we were browsing.

The ideal fix for the issue would be to add `(pr_tutorial)=` right above https://github.com/arviz-devs/arviz/blob/main/doc/source/contributing/pr_tutorial.md?plain=1#L1 so the `pr_tutorial` target exists within our own docs and the cross reference gets resolved as a link to this page.

varuntotakura:
@OriolAbril I have made the modifications as suggested. Please check it and let me know if any more changes are needed

OriolAbril (approved):
Thanks!

PR #2372, opened 2024-08-24
Added extensions for virtual environments in .gitignore

varuntotakura:
## Description
[.gitigore](https://github.com/arviz-devs/arviz/blob/main/.gitignore) doesn't have anything about environments in its extensions.

When we try to use a virtual environment to run the whole package locally, if we try to add the files to the repository, the environment files are also being added.

To fix that I have included extensions of the virtual environments

## Checklist
- [X] New features are properly documented (with an example if appropriate)

----
📚 Documentation preview 📚: https://arviz--2372.org.readthedocs.build/en/2372/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,13 @@ var/
 .ipynb_checkpoints
 .virtual_documents
 
+# Environments
+.env
+.venv
+env/
+venv/
+ENV/
+
 # PyInstaller
 #  Usually these files are written by a python script from a template
 #  before PyInstaller builds the exe, so as to inject date/other infos into it.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Deprecation
 
 ### Documentation
-- Fixed the issue in the [Contribution References Documentation](https://python.arviz.org/en/latest/contributing/index.html) ([2369](https://github.com/arviz-devs/arviz/issues/2369))
+- Added extensions of virtual environments in [.gitignore](https://github.com/arviz-devs/arviz/blob/main/.gitignore) ([2371](https://github.com/arviz-devs/arviz/issues/2371))
 
 ## v0.19.0 (2024 Jul 19)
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -28,6 +28,13 @@ var/
...
 .virtual_documents
 
+# Environments
+.env
+.venv
+env/
+venv/
+ENV/
+
 # PyInstaller
```
tomicapretto, line 38: Other options that come to my mind are 

* `hatch.toml`
* `pdm.lock`
* `Pipfile.lock`
* `Pipfile`
* `poetry`.lock`

I'm not completely sure if people should be building ArviZ following an approach different from the recommended one, but people can do whatever they want so it's probably good to have these files/directories listed.

varuntotakura: Yes, I believe it's important to use the recommended approaches. However, when installing dependencies, some may prefer to use a virtual environment. While it’s important to track the core dependency and configuration files like Pipfile, Pipfile.lock, poetry.lock, pdm.lock, and hatch.toml in version control, it’s equally important to ignore files and directories that are specific to individual developers' environments, such as virtual environments. These files are not relevant to other people's builds and can clutter the repository.

amaloney:
If you're looking to

> ...ignore files and directories that are specific to individual developers' environments...

then you may want to add items you want git to ignore from your personal setup in the file `.git/info/exclude`. This way you can ignore what ever you want, and we would not have to keep track of the most recent flavor of things to ignore when building an environment in the `.gitignore` file.

I'm not opposed to the addition of the `.env` as this is typically where people will store credentials for an app they are using, and use something like `python-dotenv` to read those credentials, but the others are not necessary in my opinion. Especially when there are mechanisms for local ignores when using git, and you can cherry pick new/updated files via the command line without running a `git add *`, which can be dangerous if you have dotfiles with credentials stored in them.

OriolAbril (approved):

PR #2379, opened 2024-09-11
make dm-tree optional

OriolAbril:
## Description
`dm-tree` dependency is only used to support pytree objects like nested dicts which 
is not very common and is also very localized in the codebase, being only used via
either `dict_to_dataset` or `from_dict` (both also having the dict->pytree alternative naming).

I would like to keep dependencies for base functionality to something more basic so
we can for example use arviz via pyodide. Also in preparation to `arviz-xyz` refactoring
where anything not required throughout the library is an optional dependency.

## Checklist

----
📚 Documentation preview 📚: https://arviz--2379.org.readthedocs.build/en/2379/

### DIFF
--- a/arviz/data/base.py
+++ b/arviz/data/base.py
@@ -9,8 +9,12 @@ from copy import deepcopy
 from typing import Any, Callable, Dict, List, Optional, Tuple, TypeVar, Union
 
 import numpy as np
-import tree
 import xarray as xr
+
+try:
+    import tree
+except ImportError:
+    tree = None
 
 try:
     import ujson as json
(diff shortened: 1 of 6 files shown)


ColCarroll (approved):
Looks good to me! 

I picked `dm-tree` as less weird of a dependency than `jax`, but the latter is probably way more common (especially if the user _has_ a pytree or nested dict). I might open an issue (or just make the PR) to allow `tree` to be either `dm-tree` or `jax.tree`.

OriolAbril:
> to allow tree to be either dm-tree or jax.tree.

That would be nice

PR #2383, opened 2024-09-16
2kde use hdi_probs by default

### LINKED ISSUE #2382: 2D Kde showing "extra" contour
### Discussed in https://github.com/arviz-devs/arviz/discussions/2380

<div type='discussions-op-text'>

<sup>Originally posted by **jecampagne** September 14, 2024</sup>
Hello
Looking at [KDE 2D example demo]( https://python.arviz.org/en/stable/examples/plot_kde_2d_hdi.html)
image
I wander how to not display the very non-smooth contour located far from the 3-sigma contour.

Thanks
</div>

aloctavodia:
Closes #2382

An "extra" contour is plotted by default, when passing hdi_probs or when passing an integer to `levels`.  This little change "fix" the issue. Not entirely happy with the solution

----
📚 Documentation preview 📚: https://arviz--2383.org.readthedocs.build/en/2383/

### DIFF
--- a/arviz/plots/backends/matplotlib/kdeplot.py
+++ b/arviz/plots/backends/matplotlib/kdeplot.py
@@ -1,5 +1,4 @@
 """Matplotlib kdeplot."""
-
 import numpy as np
 from matplotlib import pyplot as plt
 from matplotlib import _pylab_helpers
--- a/arviz/plots/backends/matplotlib/kdeplot.py
+++ b/arviz/plots/backends/matplotlib/kdeplot.py
@@ -1,4 +1,5 @@
 """Matplotlib kdeplot."""
+
 import numpy as np
 from matplotlib import pyplot as plt
 from matplotlib import _pylab_helpers
--- a/arviz/plots/kdeplot.py
+++ b/arviz/plots/kdeplot.py
@@ -72,7 +72,7 @@ def plot_kde(
         If True plot the 2D KDE using contours, otherwise plot a smooth 2D KDE.
     hdi_probs : list, optional
         Plots highest density credibility regions for the provided probabilities for a 2D KDE.
-        Defaults to matplotlib chosen levels with no fixed probability associated.
+        Defaults to [0.5, 0.8, 0.94].
     fill_last : bool, default False
         If True fill the last contour of the 2D KDE plot.
     figsize : (float, float), optional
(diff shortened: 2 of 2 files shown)


OriolAbril:
I agree with changing the behaviour with `hdi_probs`, I'd keep the same behaviour with int levels though to follow the "path of less surprise". I think users expect `contour_kwargs` and `contourf_kwargs` to be passed as is to matplotlib after adding some defaults, this would change that.

I remembered `hdi_probs` were converted into `levels` so it means we can change the hdi_probs->levels translation to fix the issue (if we restrict ourselves to that). Here https://github.com/arviz-devs/arviz/blob/main/arviz/plots/kdeplot.py#L282 both `0` and the maximum are added to the `levels` which with the new behaviour I think we only want for `contourf` but not for `contour`, so changing this to put a slice of the variable instead should fix the issue: https://github.com/arviz-devs/arviz/blob/main/arviz/plots/kdeplot.py#L292

aloctavodia:
What you suggest will fix the issue when passing `hdi_probs`, actually that was my first thought too. But to my surprise if you specify `countour_kwargs={"levels":x}, countourf_kwargs={"levels":x}``, where `x` is an integer, or if you just do the default plot, you will still get the "extra" contour.

OriolAbril:
> if you specify countour_kwargs={"levels":x}, countourf_kwargs={"levels":x}, where x is an integer, or if you just do the default plot, you will still get the "extra" contour.

The thing is this new "extra" contour is matplotlib new default behaviour. It might be a bit surprising at first because it has been added recently, but I think this is a temporal effect which will soon be reversed. People will have done/seen contourplots with matplotlib which show this outer contour delimiting where there is data but using the same parameters they won't get it if using ArviZ, in fact, depending on what we do it might even be impossible to get this behaviour at all if using ArviZ.

With `hdi_probs` the expected behaviour is clearer and unless someone asks for 0 as hdi_prob to be plotted that outer contour should not be there. But with integer value passed to `levels` the matplotlib docs say "If an int n, use [MaxNLocator](https://matplotlib.org/stable/api/ticker_api.html#matplotlib.ticker.MaxNLocator), which tries to automatically choose no more than n+1 "nice" contour levels", plus the fact that it is/was expected contour(levels=x) and contourf(levels=x) basically produce the same plot also makes things quite challenging.

I have gone quite a bit off-topic, but I think it mostly boils down to "do we actively dislike this outer contour and want to diverge from matplotlib behaviour?"

aloctavodia:
I see your point, but my visceral dislike for this outer contour is greater, haha.  What if we default to some values of hdi_probs other than None like [0.5, 0.8, 0.94]?

OriolAbril:
> What if we default to some values of hdi_probs other than None like [0.5, 0.8, 0.94]?

Sounds good

aloctavodia:
@OriolAbril I think this is ready to merge.

OriolAbril (approved):

Issue #2391, opened 2024-10-07
`from_numpyro` doesn't work for contributed NestedSampler


nstarman:
**Describe the bug**

Arvin's `from_numpyro` function works on many model objects in numpyro, but not the officially supported bridge with JaxNS — `numpyro.contrib.nested_sampling` — because `NestedSampler.get_samples` requires PRNG and num_samples args.

**To Reproduce**

```python
>>> import arviz
>>> from jax import random
>>> import jax.numpy as jnp
>>> import numpyro
>>> import numpyro.distributions as dist
>>> from numpyro.contrib.nested_sampling import NestedSampler

>>> true_coefs = jnp.array([1., 2., 3.])
>>> data = random.normal(random.PRNGKey(0), (2000, 3))
>>> labels = dist.Bernoulli(logits=(true_coefs * data).sum(-1)).sample(random.PRNGKey(1))

>>> def model(data, labels):
...     coefs = numpyro.sample('coefs', dist.Normal(0, 1).expand([3]))
...     intercept = numpyro.sample('intercept', dist.Normal(0., 10.))
...     return numpyro.sample('y', dist.Bernoulli(logits=(coefs * data + intercept).sum(-1)),
...                           obs=labels)

>>> ns = NestedSampler(model)
>>> ns.run(random.PRNGKey(2), data, labels)

>>> arviz.from_numpyro(ns)
TypeError: NestedSampler.get_samples() missing 2 required positional arguments: 'rng_key' and 'num_samples'
```

**Expected behavior**

Arviz dispatches on the numpyro model type to pass a prng & num_samples args to the NestedSampler.get_samples. Maybe `from_numpyro` adds an `args` kwarg for specifying these values.

PR #2400, opened 2024-12-05
Add NestedToMCMCAdapter to enable compatibility with MCMC (#2391)

SamuelSonoiki:
## Description

This PR addresses issue #2391 by adding the `NestedToMCMCAdapter` class to convert `NestedSampler` objects into an MCMC-compatible interface. This change ensures compatibility with tools like ArviZ for posterior analysis and visualization.

### Key Changes:
- Implemented `NestedToMCMCAdapter` to reshape posterior samples from `NestedSampler` into a chain-and-draw structure expected by MCMC workflows.
- Added `get_samples` method to provide reshaped or flattened posterior samples.
- Added `get_extra_fields` method to simulate dummy sampling statistics such as acceptance probabilities, step sizes, and number of steps.
- Supports passing model arguments (e.g., `data`, `labels`) for models requiring input parameters.

This update allows seamless integration of `NestedSampler` results into ArviZ workflows and enhances usability for downstream analysis.

## Checklist

- [x] New features are properly documented (with an example if appropriate)


OriolAbril:
sorry about the lack of feedback so far @SamuelSonoiki. @kylejcaron do you think this would complement https://github.com/arviz-devs/arviz-base/pull/95 if ported over to arviz-base?

kylejcaron:
> sorry about the lack of feedback so far @SamuelSonoiki. @kylejcaron do you think this would complement [arviz-devs/arviz-base#95](https://github.com/arviz-devs/arviz-base/pull/95) if ported over to arviz-base?

yeah absolutely, this is a very similar pattern to the proposed [SVIWrapper](https://github.com/arviz-devs/arviz-base/pull/95/files#:~:text=posterior%20%3D%20SVIWrapper( ). The MCMC sampler in numpyro stores training data, while many other samplers don't. The approach @SamuelSonoiki and I both took is to have an adapter class so that the NumPyroConverter class requires less changes. 

### Why its necessary
Compared to MCMC...

SVI additionally requires
 * The SVI instance (callable or purse function)
 * The SVI result that stores fitted VI parameters
 * The training data
 * an rng_key
 * Number of samples to generate (and then it needs to actually sample as opposed to MCMC)
 * **Missing** thinning, chains, draws

while NestedSampler requires
 * The Nested Sampler instance
 * The training data
 * an rng_key
 * Number of samples to generate (and then it needs to actually sample as opposed to MCMC)
 * **Missing** thinning, chains, draws

### How we could approach this
Taking a second look at NumPyroConverter, the strategy I've taken so far is to try and change as little as possible from the original code so that the NumPyroConverter is "MCMC first"

What are your thoughts on using a Strategy Pattern with different Adapter classes for each sampler (MCMCAdapter, SVIAdapter, NestedSamplingAdapter, etc) to try and help generalize the `NumPyroConverter`? These would be inherited from an abstract base class that matches what NumPyroConverter expects. This would all be done under the hood so that people could still call az.from_numpyro(), az.from_numpyro_sampler(...), etc

I think if we pursued this I'd like to finish the az.from_numpyro_svi PR first and then pursue a refactor in case I run into unexpected issues. This nested sampler could be bundled in with the refactor, and this would help make future samplers easier to adapt. Seems like there's been alot of innovation with samplers and approximate bayesian inference lately

OriolAbril:
> What are your thoughts on using a Strategy Pattern with different Adapter classes for each sampler (MCMCAdapter, SVIAdapter, NestedSamplingAdapter, etc) to try and help generalize the NumPyroConverter? 

Sounds good, but also there is nothing inherent about `NumPyroCoverter` class itself that requires minimizing changes to it. If subclassing the converter itself helps also feel free to do that. Go with whichever seems more suited when tackling the refactor, no constraints on our side towards either alternative.

> I think if we pursued this I'd like to finish the az.from_numpyro_svi PR first and then pursue a refactor in case I run into unexpected issues. This nested sampler could be bundled in with the refactor, and this would help make future samplers easier to adapt. Seems like there's been alot of innovation with samplers and approximate bayesian inference lately

Looks like a plan 🚀 , thanks for helping with this.

OriolAbril:
closing the PR for now, moving the discussion to https://github.com/arviz-devs/arviz-base/issues/106 for nested sampling support

PR #2435, opened 2025-03-07
changes after release

aloctavodia:
----
📚 Documentation preview 📚: https://arviz--2435.org.readthedocs.build/en/2435/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## Unreleased
+
+### New features
+
+### Maintenance and fixes
+
+### Documentation
+
 ## v0.21.0 (2025 Mar 06)
 
 ### New features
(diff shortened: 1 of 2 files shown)


tomicapretto:
Does it make sense to use setuptools SCM to handle versions automatically? 

See

https://github.com/bambinos/bambi/blob/a798c89be873987eadf30dba021b5f030cecf011/pyproject.toml#L2-L3
https://github.com/bambinos/bambi/blob/a798c89be873987eadf30dba021b5f030cecf011/pyproject.toml#L13
https://github.com/bambinos/bambi/blob/a798c89be873987eadf30dba021b5f030cecf011/bambi/__init__.py#L16

aloctavodia:
I would do it for new arviz. There changelogs are also handled automatically.

Issue #2022, opened 2022-04-28
InferenceData coords for NumPyro plates


ColdTeapot273K:
## Tell us about it

When creating `InferenceData` using
```python
az.from_numpyro(...)
```
the resulting autogenerated coordinates are note very telling:
![image](https://user-images.githubusercontent.com/54956904/165839956-668eee15-47a6-4194-9bec-0f473d92613e.png)

Now consider the NumPyro model below that produced these samples.
These coords with autogenerated names are in fact [plate](https://en.wikipedia.org/wiki/Plate_notation) dimensions. And plates have names.
```python

import numpy as np
import pandas as pd

# INFO: PPL specific imports

import jax.numpy as jnp
from jax import random

import numpyro
import numpyro.distributions as dist
import numpyro.optim as optim
from numpyro.infer import SVI, Trace_ELBO, Predictive
from numpyro.infer import MCMC, NUTS
from numpyro.infer.autoguide import AutoLaplaceApproximation, AutoNormal

from jax import lax, random
from jax.scipy.special import expit

import arviz as az

# %%

data_uri = "https://raw.githubusercontent.com/rmcelreath/rethinking/master/data/NWOGrants.csv"
df_dev = pd.read_csv(data_uri, sep=";")
df_dev.head()

df_dev["gender"] = df_dev["gender"] == "m"
df_dev["gender"] = df_dev["gender"].astype(int)
df_dev["discipline"] = df_dev["discipline"].astype("category").cat.codes

# %%


def model(data: pd.DataFrame, observed=True):
    applications = data["applications"].values
    awards = data["awards"].values

    discipline = data["discipline"].values
    discipline_card = np.unique(discipline).shape[0]
    gender = data["gender"].values
    gender_card = np.unique(gender).shape[0]

    observations_card = data.shape[0]

    # INFO: good plate version
    with numpyro.plate("plate_gender", gender_card):
        with numpyro.plate("plate_discipline", discipline_card):
            alpha_gender_discipline = numpyro.sample("alpha_gender_discipline", dist.Normal(-1, 1))

    assert alpha_gender_discipline.shape == (9, 2)

    link_p = numpyro.deterministic("link_p", alpha_gender_discipline[discipline, gender])

    with numpyro.plate("plate_observations", observations_card):
        numpyro.sample(
            "awards", dist.Binomial(total_count=applications, logits=link_p), obs=awards if observed else None
        )


kernel = NUTS(model)
mcmc = MCMC(
    kernel,
[...]

Issue #1489, opened 2021-01-08
Write conversion guides


OriolAbril:
## Tell us about it

Convert the cookbook into a collection of notebooks (one per PPL) with extended documentation on the respective converter function `from_xyz`. PPL tracker:

* [ ] PyMC3 (if the examples below are good enough it's only a matter of putting the notebook together)
* [ ] PyStan
* [ ] CmdStanPy
* [ ] Pyro
* [ ] NumPyro
* [ ] emcee (there are extensive examples in the [from_emcee](https://arviz-devs.github.io/arviz/api/generated/arviz.from_emcee.html#arviz.from_emcee) api docs that can help write this)
* [ ] TensorFlow Probability

## Thoughts on implementation

I think it would be great to define a set of tasks of increasing difficulty and then (if possible) show how to tackle them in each of the PPLs and how to convert the data to InferenceData. The current example in the [cookbook](https://arviz-devs.github.io/arviz/getting_started/InferenceDataCookbook.html) is a possible starting point. The [examples](https://github.com/OriolAbril/arviz-probprog-2020#inferencedata-creation) (click on the QR codes) for my ProbProg poster are a more complete example of InferenceData creation. And my pymc3-arviz [blogpost](https://oriolabril.github.io/oriol_unraveled/python/arviz/pymc3/xarray/2020/09/22/pymc3-arviz.html) could be used as intermediate examples.

OriolAbril:
Tagging as Discussion for now until we agree on what kind of conversion tasks and models we want to use for our ArviZ rosetta stone

OriolAbril:
Extra pro of doing this (which can be done once the examples in from_emcee docstring are moved to a notebook). Doc generation won't need any inference library so we can have the doc job depend only on basetests and remove the externaltests dependency to speed-up CI time.

PR #2441, opened 2025-03-31
Infer NumPyro InferenceData dims automatically

kylejcaron:
## Description

Fixes: #2022

 If dims are not provided, `az.from_numpyro` now infers dims automatically based on model plates, sample sites, and provided `coords`. 

For event_dims...
 * for numpyro.sample sites, infer={"event_dims":["dim1"]} can be provided
 * for deterministic sites, there is an `extra_event_dims` field in az.from_numpyro to manually provide them.

`pred_dims` are also now inferred by default if they arent provided.

## Checklist

----
📚 Documentation preview 📚: https://arviz--2441.org.readthedocs.build/en/2441/

### DIFF
--- a/arviz/data/io_numpyro.py
+++ b/arviz/data/io_numpyro.py
@@ -14,25 +14,28 @@ from .inference_data import InferenceData
 _log = logging.getLogger(__name__)
 
 
-def _add_dims(dimsA: Dict[str, List[str]], dimsB: Dict[str, List[str]]):
+def _add_dims(dims_a: Dict[str, List[str]], dims_b: Dict[str, List[str]]):
     merged = defaultdict(list)
 
-    for k, v in dimsA.items():
+    for k, v in dims_a.items():
         merged[k].extend(v)
 
-    for k, v in dimsB.items():
+    for k, v in dims_b.items():
         merged[k].extend(v)
 
     # Convert back to a regular dict
     return dict(merged)
 
 
-def infer_dims(model, model_args=(), model_kwargs={}):
+def infer_dims(model, model_args=None, model_kwargs=None):
 
     from numpyro import handlers, distributions as dist
     from numpyro.ops.pytree import PytreeTrace
     from numpyro.infer.initialization import init_to_sample
     import jax
+
+    model_args = tuple() if model_args is None else model_args
+    model_kwargs = dict() if model_args is None else model_kwargs
 
     def _get_dist_name(fn):
         if isinstance(fn, (dist.Independent, dist.ExpandedDistribution, dist.MaskedDistribution)):
--- a/arviz/data/io_numpyro.py
+++ b/arviz/data/io_numpyro.py
@@ -66,7 +66,7 @@ def infer_dims(model, model_args=None, model_kwargs=None):
     named_dims = {}
 
     for name, site in trace.items():
-        batch_dims = [frame.name for frame in site["cond_indep_stack"]]
+        batch_dims = [frame.name for frame in sorted(site["cond_indep_stack"], key=lambda x: x.dim)]
         event_dims = list(site.get("infer", {}).get("event_dims", []))
         if site["type"] in ["sample", "deterministic"] and (batch_dims or event_dims):
             named_dims[name] = batch_dims + event_dims
--- a/arviz/data/io_numpyro.py
+++ b/arviz/data/io_numpyro.py
@@ -2,7 +2,7 @@
 
 from collections import defaultdict
 import logging
-from typing import Callable, Optional, Dict, List
+from typing import Any, Callable, Optional, Dict, List, Tuple
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


arviz/tests/external_tests/test_data_numpyro.py
```diff
@@ -282,3 +290,61 @@ def model():
...
+        import numpyro.distributions as dist
+        from numpyro.infer import MCMC, NUTS
+        import jax.numpy as jnp
+
+        def model():
+            alpha = numpyro.sample("alpha", dist.Normal(0, 1))
+            sigma = numpyro.sample("sigma", dist.HalfNormal(1))
+            with numpyro.plate("obs_idx", 3):
+                # mu is plated by obs_idx, but isnt broadcasted to the plate shape
+                mu = numpyro.deterministic("mu", alpha)
```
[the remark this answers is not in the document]
kylejcaron: I decided to proceed with the least assumptions, which was option 3. In this example, the dims for alpha simply dont match the plate (even though it could be broadcasted). changed the test to xfail

kylejcaron:
This is ready to be reviewed! 

There are some CI failures that seem unrelated to these changes (Im seeing the same issue on other PRs)

OriolAbril (commented):
Thanks for the PR! This will be a very nice addition.

CI should be fixed now, sorry for the slow response, we do what we can.

I looked a bit over the code and if I understand correctly, dimension names will only be inferred when the `coords` argument is explicitly provided but I think it is not necessary to require `coords` (both quickly going over the code and conceptually).

In xarray, dimensions are always named, even if it is a `var_name_id` name, but coordinate values are completely optional to the point that even when no coord values are present `.sel` still works without problem and simply behaves like `.isel`. Consequently, in my opinion, for a plated model where no coords are provided it is preferable to end up with properly named dimensions without coord values than it is to get nothing at all.

In practice I think this will only mean removing the `@requires("coords")` decorators and updating some of the tests to cover that case.

---

Lastly, specially now that the converter has more functionality, I think it would be great to have a page like https://python.arviz.org/en/stable/getting_started/ConversionGuideEmcee.html for numpyro too. This is related to https://github.com/arviz-devs/arviz/issues/1489 and it is not necessary to add the notebook in this PR but it would be great if you could then create a dedicated issue for the numpyro conversion guide with a proposed outline as you probably have a better idea of what to include than me for example. I think this will make the issue much more approachable to other less experienced contributors.

arviz/tests/external_tests/test_data_numpyro.py
```diff
@@ -85,6 +85,7 @@ def test_inference_data_namedtuple(self, data):
         data.obj.get_samples = lambda *args, **kwargs: data_namedtuple
         inference_data = from_numpyro(
             posterior=data.obj,
+            dims={},  # This mock test needs to turn off autodims like so or mock group_by_chain
```
kylejcaron, line 88: Calling out this change to a pre-existing test. Looks like my infer_dims logic broke this test, however I found out that its also due to the fact that the mock object for get_samples() didnt have any `group_by_chain` logic. This was causing the test to have the wrong posterior shape of (chain*draw, J) instead of (chain, draw, J). 

the infer_dims logic was catching this and failing, so i chose to turn off the infer_dims logic and let the test work as originally intended (I'm not sure what the original purpose of this test actually is and which edge case its covering)

arviz/data/io_numpyro.py
```diff
@@ -410,14 +409,9 @@ def infer_dims(self) -> Dict[str, List[str]]:
 
     @requires("posterior")
     @requires("model")
-    @requires("coords")
     @requires("predictions")
     def infer_pred_dims(self) -> Dict[str, List[str]]:
-        if self.predictions_constant_data is not None:
-            dims = infer_dims(self.model, model_kwargs=self.predictions_constant_data)
```
kylejcaron: Assumption: removed this logic and chose to infer dims with the original input data in all cases - otherwise the `test_inference_data_constant_data` test was failing since it passed in `x2` as prediction constant data, even though x2 isnt a model input

kylejcaron:
Thanks for taking a look! 

> Lastly, specially now that the converter has more functionality, I think it would be great to have a page like https://python.arviz.org/en/stable/getting_started/ConversionGuideEmcee.html for numpyro too. This is related to #1489 and it is not necessary to add the notebook in this PR but it would be great if you could then create a dedicated issue for the numpyro conversion guide with a proposed outline as you probably have a better idea of what to include than me for example. I think this will make the issue much more approachable to other less experienced contributors.

Happy to do so, I'll get started on making the issue (and add in the conversion guide itself if I have time!). 

As part of this, what are your thoughts on also adding numpyro converter logic for variational inference? I know PyMC has one (where chains=1), but I also know its slightly misleading for less experienced users since many of ArviZ's diagnostics are MCMC based. I personally still use inferencedata objects for numpyro SVI results so it would atleast be helpful for myself


> In practice I think this will only mean removing the `@requires("coords")` decorators and updating some of the tests to cover that case.

good callout, I removed that requirement. This lead me to making two bigger assumptions in my logic to get tests passing, I called them out with comments if you want to take a look

arviz/tests/external_tests/test_data_numpyro.py
OriolAbril: Also not sure about the goal of the test, it might even predate `group_bu_chain` logic and never gotten updated. It is ok with me to leave like this

OriolAbril (approved):

Issue #2306, opened 2024-01-13
HDI naming


sboukortt:
Hi,

I’m opening this because it’s not really clear why, in ArviZ’s output, the bounds for, say, a 94% HDI are named `hdi_3%` and `hdi_97%`. Such names seem to imply that they correspond respectively to the 3rd and 97th percentiles of the posterior distribution, which would make them equal-tailed credible intervals if that were the case, but it isn’t necessarily. (HDIs and equal-tailed credible intervals don’t have to coincide.) Indeed, the following program shows a counterexample:

```python
import arviz as az
import pymc as pm
import scipy.stats

with pm.Model() as model:
    pm.Exponential('expon', lam=.01)
    trace = pm.sample(10000)

summary = az.summary(trace, hdi_prob=.94)
print(summary)

hdi_lower = summary['hdi_3%'].expon
hdi_upper = summary['hdi_97%'].expon

dist = scipy.stats.expon(scale=100)

print(f"\nThe 94% HDI ([{hdi_lower:.3f}, {hdi_upper:.3f}]) contains {dist.cdf(hdi_upper) - dist.cdf(hdi_lower):.3%} of the probability.")
print(f"{dist.cdf(hdi_lower):.3%} of the probability is to the left of hdi_3%.")
print(f"{dist.sf(hdi_upper):.3%} of the probability is to the right of hdi_97%. ({dist.cdf(hdi_upper):.3%} to its left.)\n")

print(f"The equal-tailed 94% credible interval is [{dist.ppf(0.03):.3f}, {dist.ppf(0.97):.3f}].")
```

Here is its output:

```
          mean       sd  hdi_3%  hdi_97%  mcse_mean  mcse_sd  ess_bulk  ess_tail  r_hat
expon  100.309  100.502   0.001  281.632       0.79    0.559   10051.0    8522.0    1.0

The 94% HDI ([0.001, 281.632]) contains 94.016% of the probability.
0.001% of the probability is to the left of hdi_3%.
5.983% of the probability is to the right of hdi_97%. (94.017% to its left.)

The equal-tailed 94% credible interval is [3.046, 350.656].
```

Is there maybe another reason for the naming? Or should it perhaps be reconsidered?

Thanks.

aloctavodia:
Thanks for reporting this. Probably a better name would be hdi_lb (lower bound) and hdi_up (upper bound). Do you want to submit a PR?

sethaxen:
[...]

PR #2454, opened 2025-04-26
Addressed issue titled "HDI naming" by changing the lower and upper hdi bound names in stats.py summary() function to hdi_{hdi_prob}_lb and hdi_{hdi_prob}_ub

YunwooC:
## Description
Addresed issue titled "HDI naming", ID: #2306. Improved naming for lower and higher hdi bounds in stats.py summary() function by renaming them to hdi_{hdi_prob}_ lb and hdi_{hdi_prob}_ ub.

## Checklist


OriolAbril (changes requested):
Agreed with @aloctavodia that it would be best to match arviz-stats. Also, note that the PR adds the _`l/ub` part between the probability and the % sign: 

imatge

so it can't be merged as is even if matching arviz-stats weren't an issue

PR #2485, opened 2025-10-07
Add daily warning message about ArviZ's refactoring

aloctavodia:
----
📚 Documentation preview 📚: https://arviz--2485.org.readthedocs.build/en/2485/

### DIFF
--- a/arviz/__init__.py
+++ b/arviz/__init__.py
@@ -14,7 +14,7 @@ import matplotlib as mpl
 from packaging import version
 
 
-def warn_once_per_day():
+def _warn_once_per_day():
     warning_dir = Path.home() / "arviz_data"
     warning_dir.mkdir(exist_ok=True)
 
@@ -32,13 +32,14 @@ def warn_once_per_day():
             "while maintaining a user-friendly interface."
             "\nSome upcoming changes may be backward incompatible."
             "\nFor details and migration guidance, visit: "
-            "https://python.arviz.org/en/stable/user_guide/migration_guide.html",
+            "https://python.arviz.org/en/latest/user_guide/migration_guide.html",
             FutureWarning,
         )
 
         stamp_file.write_text(today.isoformat())
 
-warn_once_per_day()
+
+_warn_once_per_day()
 
 
 class Logger(logging.Logger):


OriolAbril:
Now that I am seeing this. Do we want to include this in the next release or first make a release without warnings yet but including the fix for i/o, attrs in idata/datatree conversion... then make a follow up release to add the warning?

I think the next release is the one that will allow proper migration through the `arviz.preview` and I think it would be nice to have a version where users can actively work on the migration without warnings yet. The migration is bound to take some time.

On the other hand, if we add a pin for the "loud" version of arviz in downstream dependencies like pymc then downstream libraries will know to migrate but users won't which is also not a good development

arviz/__init__.py
```diff
@@ -4,13 +4,43 @@
...
 from matplotlib.pyplot import style
 import matplotlib as mpl
 from packaging import version
 
 
+def warn_once_per_day():
+    warning_dir = Path.home() / "arviz_data"
+    warning_dir.mkdir(exist_ok=True)
+
+    stamp_file = warning_dir / f"daily_warning"
```
OriolAbril:
```suggestion
    stamp_file = warning_dir / "daily_warning"
```

code-wise this is the only comment, and I didn't actually see at first, it was pylint that flagged it

arviz/__init__.py
```diff
@@ -4,13 +4,43 @@
...
+    today = datetime.date.today()
+
+    if stamp_file.exists():
+        last_date = datetime.date.fromisoformat(stamp_file.read_text().strip())
```
tomicapretto: TIL about `read_text` and `write_text`!

tomicapretto (approved):
@aloctavodia approved after you apply @OriolAbril's suggestion

aloctavodia:
We should have a migration plan. That may be something like 

1) Do a new release of arviz-xyz libraries.
2) Do a new release of arviz, including the warning message.
3) Help PyMC/Bambi (and potentially others) with the transition. For instance, we could start adapting examples (but probably also changing internal where arviz is used).

We may want to have some tentative dates for these steps, even if vague and subject to change. We may want to add more steps, too.

PR #2487, opened 2025-10-14
Fix typo in error message for zarr 3

pranavkp71:
This PR updates the zarr version check and error message in `InferenceData.to_zarr()`.

The goal is to improve clarity for users now that zarr 3.0.0 has been officially released. The error message was updated to remove references to development builds and to clearly state the supported version range (>=2.5.0,<3). No functional support for zarr 3 is added in this PR.

### DIFF
--- a/arviz/data/inference_data.py
+++ b/arviz/data/inference_data.py
@@ -872,7 +872,7 @@ class InferenceData(Mapping[str, xr.Dataset]):
         if version.parse(zarr.__version__) >= version.parse("3.0.0.dev0"):
             raise ImportError(
                 "Found zarr>=3, which is not supported by ArviZ. Instead, you can use "
-                "'dt = InferenceData.to_datatree' followed by 'dt.to_zarr()' "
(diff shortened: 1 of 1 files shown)


pranavkp71:
Thanks for the detailed clarification

My goal with this PR was only to improve the version check and error messaging not to add full zarr 3 support yet.   Based on your feedback, I’ll update the PR accordingly and remove the “closes” reference to avoid implying the issue is resolved.

Regarding full zarr 3 compatibility, I’m interested in exploring it gradually on the side to understand what changes would be needed. I’m not committing to implementing full support right now, but if I make progress, I can follow up with a separate PR.

Let me know if that approach works, and I’ll start by adjusting this PR to focus only on the messaging.

OriolAbril:
Sounds good!

pranavkp71:
Hi @OriolAbril, I've applied the requested change to restore the workaround wording.  
Please let me know if anything else needs adjusting.

Issue #2359, opened 2024-07-17
az.compare stacking weights do not sum to one


hschuett:
Dear Arviz team,

**Issue**
when using arviz.compare() to compute stacking weights for model averaging, we got a) weights that do not sum to one and b) got two different sets of weights in two different runs with the same data. (literally, all other numbers in the compare output are the same, except for the weights. Below are the two compare results dfs we got. 

**Documentation**
First run (on a grid batch job):
rank | elpd_loo | p_loo | elpd_diff | weight | se | dse | warning | scale
-- | -- | -- | -- | -- | -- | -- | -- | --
0 | -299161.07 | 20338.32 | 0.00 | 0.93 | 757.80 | 0.0000 | True | log
1 | -315073.54 | 29909.27 | 15912.47 | 0.39 | 741.13 | 238.65 | True | log
2 | -318882.55 | 805.33 | 19721.49 | 0.00 | 726.20 | 272.98 | True | log
3 | -328994.76 | 617.96 | 29833.70 | 0.00 | 717.08 | 320.34 | True | log
4 | -343025.50 | 89.02 | 43864.43 | 0.00 | 694.47 | 381.01 | True | log

Arviz version: 0.18.0
Pandas version: 2.0.3
Numpy version: 1.24.3
Python version: 3.11.5 (main, Sep 11 2023, 13:54:46) [GCC 11.2.0]
OS info: posix, Linux, 4.19.0-25-amd64

Second run (on Jupyter server using the same resources):   

rank | elpd_loo | p_loo | elpd_diff | weight | se | dse | warning | scale
-- | -- | -- | -- | -- | -- | -- | -- | --
0 | -299161.0654 | 20338.3236 | 0.0000 | 1.0000 | 757.7960 | 0.0000 | True | log
1 | -315073.5328 | 29909.2672 | 15912.4675 | 0.0019 | 741.1303 | 238.6455 | True | log
2 | -318882.5520 | 805.3308 | 19721.4866 | 0.0011 | 726.2029 | 272.9801 | True | log
3 | -328994.7609 | 617.9605 | 29833.6956 | 0.0005 | 717.0755 | 320.3400 | True | log
4 | -343025.4954 | 89.0185 | 43864.4300 | 0.0000 | 694.4689 | 381.0094 | True | log

Arviz version: 0.18.0
Pandas version: 2.0.3
Numpy version: 1.24.3
[...]

PR #2492, opened 2025-10-17
Fix az.compare stacking weights not sum up to one

littleTT0704:
## Description

While using `az.compare(models, method="stacking")`, I encountered the same problem as mentioned in #2359, where the `weights` column in the returned dataframe sums up to greater than 1.

My assumption is that `scipy.optimize.minimize` [converts our dict-type constraints to a list of `NonLinearConstraint`](https://github.com/scipy/scipy/blob/main/scipy/optimize/_constraints.py#L607), which over-complicates the problem in `az.compare`, since the sum of weights is a linear combination.  
Moreover, by removing the `np.concatenate` operation in `w_fuller` and using vectorized operations, we can increase the efficiency of the optimizer.  
In my extreme use case of comparing 54 models, I was able to reduce computation time by 14.3% (543s -> 465s) while getting the correct weight sum (3.636 -> 1.000).

However, I was not able to synthesize some test data to capture this behavior. Although we already have [some test cases that asserts sum of weights should be 1](https://github.com/arviz-devs/arviz/blob/main/arviz/tests/base_tests/test_stats.py#L203), the models are too simple to produce a failure.

## Checklist

----
📚 Documentation preview 📚: https://arviz--2492.org.readthedocs.build/en/2492/

### DIFF
--- a/arviz/stats/stats.py
+++ b/arviz/stats/stats.py
@@ -1,7 +1,6 @@
 # pylint: disable=too-many-lines
 """Statistical functions in ArviZ."""
 
-import itertools
 import warnings
 from copy import deepcopy
 from typing import List, Optional, Tuple, Union, Mapping, cast, Callable


OriolAbril:
also requesting more reviews to also see how to port that to [arviz-stats](https://github.com/arviz-devs/arviz-stats) too.

aloctavodia (approved):
LGTM! I will add it to arviz-stats. Thanks @littleTT0704

PR #2494, opened 2025-11-11
Convert arviz library to metapackage and global docs

OriolAbril:
Prepare for 1.0 release with `arviz.preview` becoming the "main package".
The main repo will then be a metapackage that installs the 3 arvizverse libraries
and exposes their functions through `arviz.xyz`.
Consequently, tests are only about aliases coming from the 3 other libraries.

I think we should keep some global docs here. I can look into making the docs of the
other 3 libraries be subprojects of the main one. That means they would get
`python.arviz.org/projects/arviz-base/...` as URLs, and we can look at doing something
similar to what we do with PyMC to allow searching across all the different websites.

I have also found a couple minor issues with imports and targets for cross-references in the docs.
I will open PRs to the other repos soon.

----
📚 Documentation preview 📚: https://arviz--2494.org.readthedocs.build/en/2494/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "flit_core.buildapi"
 [project]
 name = "arviz"
 readme = "README.md"
-requires-python = ">=3.11"
+requires-python = ">=3.12"
 license = {file = "LICENSE"}
 authors = [
     {name = "ArviZ team", email = "arvizdevs@gmail.com"}
@@ -19,9 +19,9 @@ classifiers = [
     "Operating System :: OS Independent",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
 ]
 dynamic = ["version", "description"]
 dependencies = [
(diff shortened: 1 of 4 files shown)


OriolAbril:
I started copying the `preview.py` file into `__init__.py` but I have now realized unless someone does `pip install arviz --no-deps` all 3 libraries will be installed, so we can probably get rid of the try except blocks

pyproject.toml
```diff
@@ -1,15 +1,92 @@
...
+    "myst-nb",
+    "pydata-sphinx-theme>=0.13",
+    "sphinx>=5",
+    "sphinx-copybutton",
+    "sphinx-design",
+    "sphinx-notfound-page",
+    "sphinxcontrib-youtube",
+]
+test = ["pytest"]
+check = ["build", "pre-commit"]
```
tomicapretto: Are these used for testing/checking purposes? I don't see them in the GA files... Does tox use them under the hood? Or, are those for developers to do:

```shell
pip install -e .[test]
```

?

tests/test_namespace.py
tomicapretto: Looks great to me. 

I wonder if there could be a case where arviz imports arviz_(base|stats|plots) from a location and the tests import it from a different one, and then they're not equal in identity.... but I guess that's a complex edge case that's not up to the test to handle.

PR #2502, opened 2025-12-20
Proofreading docs

aloctavodia:


OriolAbril (approved):
I think we can merge and continue iterating as we see fit

PR #2504, opened 2025-12-23
Improve error message for mismatched ArviZ subpackage versions

ShivamRajSri:
This PR improves the error message raised when incompatible minor versions of
arviz-base, arviz-stats, and arviz-plots are detected, clearly reporting which
subpackage has which version while preserving existing behavior.

### DIFF
--- a/src/arviz/__init__.py
+++ b/src/arviz/__init__.py
@@ -61,11 +61,16 @@ info = f"Status information for ArviZ {__version__}\n\n{info}"
 
 pat = re.compile(r"arviz_(base|stats|plots)\s([0-9]+\.[0-9]+)")
 matches = pat.findall(info)
-if any(matches[0][1] != match[1] for match in matches[1:]):
+
+versions = {name: version for name, version in matches}
+unique_versions = set(versions.values())
+
+if len(unique_versions) > 1:
     raise ImportError(
-        "Versions of arviz-xyz packages don't match to the minor version. "
-        f"The versions found are: {matches}"
+        "Incompatible ArviZ subpackage versions detected. "
+        "All arviz-* packages must share the same minor version. "
+        f"Found versions: {versions}"
     )
 
 # clean namespace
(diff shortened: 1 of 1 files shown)


ShivamRajSri:
Thanks for the clarification — good point.
Below is a concrete before/after example of the error message to illustrate the change this PR introduces:
Before (main branch):
ImportError: Versions of arviz-xyz packages don't match to the minor version.
The versions found are: [('base', '0.7'), ('stats', '0.6'), ('plots', '0.7')]
After (this PR):
ImportError: Incompatible ArviZ subpackage versions detected:
- arviz_base: 0.7
- arviz_stats: 0.6
- arviz_plots: 0.7

All ArviZ subpackages must share the same minor version.
I’m currently away from my laptop, but I can move this comparison into the PR description or add a small test once I’m back if you think that would be better

OriolAbril:
It's fine to keep here as part of the conversation, no need to update the PR description. Adding a test for the error message with a monkeypatch of the arviz subpackage versions would be great yes.

I see now the main difference is changing the `matches/versions` object printed in the error from a list of tuples to a dictionary. I am still a bit on the fence with purposely multiline error messages but we are already adding them in several places and this looks like a good fit so I'd say go for it. However, the current formatting can't be doing that as it is only doing `{versions}` with `versions` being a dictionary

ShivamRajSri:
Thanks for the detailed feedback, that makes sense 👍
I agree on the test — I’ll add a small test that monkeypatches the arviz_base, arviz_stats, and arviz_plots versions to assert the error message content and formatting.
Good catch on the formatting as well. I’ll adjust the implementation so the multiline message is explicitly constructed (instead of relying on {versions} stringification), ensuring each subpackage/version is rendered on its own line as intended.
I’ll push an update with the test and formatting fix shortly.

ShivamRajSri:
Closing this PR in favor of #2506, which includes the updated error formatting and the added test requested in review.

PR #2506, opened 2026-01-01
Improve error message for incompatible ArviZ subpackage versions

ShivamRajSri:
This PR improves the error raised when incompatible minor versions of arviz_base, arviz_stats, and arviz_plots are detected.

What’s changed

.Explicitly constructs a clear, multiline error message that lists each ArviZ subpackage and its detected version, instead of relying on dictionary stringification.

.Adds a regression test using monkeypatch to validate the error message when subpackage versions do not match, covering the case where arviz_stats may not define __version__.

Why

.Makes the version mismatch error easier to read and understand for users.

.Ensures the improved message is protected against future regressions by tests.

All tests and pre-commit checks pass locally.

### DIFF
--- a/src/arviz/__init__.py
+++ b/src/arviz/__init__.py
@@ -62,14 +62,14 @@ info = f"Status information for ArviZ {__version__}\n\n{info}"
 pat = re.compile(r"arviz_(base|stats|plots)\s([0-9]+\.[0-9]+)")
 matches = pat.findall(info)
 
-versions = {name: version for name, version in matches}
+versions = dict(pat.findall(info))
 unique_versions = set(versions.values())
 
 if len(unique_versions) > 1:
     lines = ["Incompatible ArviZ subpackage versions detected:"]
 
-    for package in sorted(versions):
-        lines.append(f"- arviz_{package}: {versions[package]}")
+    for package, version in sorted(versions.items()):
+        lines.append(f"- arviz_{package}: {version}")
 
     lines.append("All ArviZ subpackages must share the same minor version.")
 


ShivamRajSri:
This PR supersedes the previous one and incorporates the review feedback: the error message is now explicitly constructed as a multiline message, and a test using monkeypatch has been added to cover incompatible subpackage versions. All tests and pre-commit checks pass locally.

OriolAbril (approved):
Thanks!

Approving for now but waiting a bit to merge to see if someone else has extra feedback

ShivamRajSri:
Thanks a lot for the review and approval!
Happy to wait and address any additional feedback if it comes up.

src/arviz/__init__.py
```diff
@@ -61,11 +61,20 @@
 
 pat = re.compile(r"arviz_(base|stats|plots)\s([0-9]+\.[0-9]+)")
 matches = pat.findall(info)
-if any(matches[0][1] != match[1] for match in matches[1:]):
-    raise ImportError(
-        "Versions of arviz-xyz packages don't match to the minor version. "
-        f"The versions found are: {matches}"
-    )
+
+versions = {name: version for name, version in matches}
```
aloctavodia:
```suggestion
versions = dict(pat.findall(info))
```

src/arviz/__init__.py
```diff
@@ -61,11 +61,20 @@
...
+unique_versions = set(versions.values())
+
+if len(unique_versions) > 1:
+    lines = ["Incompatible ArviZ subpackage versions detected:"]
+
+    for package in sorted(versions):
+        lines.append(f"- arviz_{package}: {versions[package]}")
```
aloctavodia:
```suggestion
```

ShivamRajSri:
Thanks a lot for the suggestions!
I’ve applied the proposed changes, updated the iteration over the versions dictionary, and simplified its construction.
Tests and pre-commit hooks are passing locally now. Please let me know if anything else should be adjusted.

PR #2509, opened 2026-01-02
extend objects.inv with aliases

OriolAbril:
I still have to do some extra tests but as I added in one of the docstrings,
the goal is this to both prevent broken links but also simplify downstream documentation.
That is, we can use `` :func:`arviz.plot_forest` `` in pymc-examples and the cross-reference
will work (not be redirected but work directly). There is a bit of a nuance on the intersphinx
mapping so it uses only `python.arviz.org` as base URL instead of `python.arviz.org/en/<version>/`

I will probably also add simple html only pages in `/api/generated/xyz` to redirect to
the new locations/functions for those using URLs instead of sphinx-like cross-references.

Consequently, we could also manually add entries to `objects.inv` pointing to those redirects,
but that feels a bit weird. To me, having a redirect from
`python.arviz.org/en/<version>/api/generated/plot_posterior.html` to
`python.arviz.org/projects/plots/en/<version>/api/generated/plot_dist.html`,
feels much different than having `` :func:`arviz.plot_posterior` `` in the list of documented functions.

Regarding the redirects themselves, some will be to xarray docs directly too.


OriolAbril:
whoever comes across this PR, please go over a couple notebooks in https://pymcio--832.org.readthedocs.build/projects/examples/en/832/gallery.html and check all the cross-referencing is correct. The link is for the pymc example gallery so there are some references scattered through the text, but the code itself also contains links. The css highlight doesn't work properly though, but they are there. Therefore, the best way to check cross-references are extracted correctly is to go through the code blocks and over over uses of `az.summary` and other arviz objects to check it does link to [the respective page in arviz-stats](https://python.arviz.org/projects/stats/en/stable/api/generated/arviz_stats.summary.html#arviz.summary).

aloctavodia:
I tested a few notebooks, and I only get links when hovering over code for the notebooks using arviz.preview, but not ArviZ's functions, but for other packages including pymc, numpy, and pandas. Is this expected?

OriolAbril:
<p>true, opening an issue for this</p>

---

 <a href='https://app.reviewnb.com/arviz-devs/arviz/pull/2509/discussion/#comment-3845867218'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

PR #2510, opened 2026-01-03
Add test for compatible ArviZ subpackage versions

ShivamRajSri:
This PR adds a test ensuring that ArviZ does not raise an ImportError when
subpackage minor versions are compatible.

This complements the existing test for incompatible versions and improves
test coverage around version validation logic.

### DIFF
--- a/tests/test_namespace.py
+++ b/tests/test_namespace.py
@@ -52,7 +52,10 @@ def test_incompatible_subpackage_versions(monkeypatch):
 
 
 def test_compatible_subpackage_versions(monkeypatch):
-    """Compatible minor versions should not raise an ImportError."""
+    """
+    Compatible minor versions should not raise an ImportError and should be
+    reported correctly.
+    """
 
     # Same minor version (0.7.x) → should be allowed
     monkeypatch.setattr(arviz_base, "__version__", "0.7.0", raising=False)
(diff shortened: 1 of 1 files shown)


ShivamRajSri:
@OriolAbril This adds a small follow-up test to complement the version-mismatch work.

ShivamRajSri:
Fixed the compatible-version test to handle missing __version__ attributes and resolved formatting issues.
All tests pass locally.

tests/test_namespace.py
```diff
@@ -49,3 +49,15 @@ def test_incompatible_subpackage_versions(monkeypatch):
...
+def test_compatible_subpackage_versions(monkeypatch):
+    """Compatible minor versions should not raise an ImportError."""
+
+    # Same minor version (0.7.x) → should be allowed
+    monkeypatch.setattr(arviz_base, "__version__", "0.7.0", raising=False)
+    monkeypatch.setattr(arviz_stats, "__version__", "0.7.1", raising=False)
+    monkeypatch.setattr(arviz_plots, "__version__", "0.7.2", raising=False)
+
+    # Should not raise
+    importlib.reload(az)
```
OriolAbril: Can you add some asserts on `az.info` to check the monkeypatched versions with the different patch numbers are correctly being added?

Something similar to the asserts in the previous tests would be good enough:

```python
assert "arviz_base 0.7.0 available" in az.info
...
```

PR #2513, opened 2026-01-08
REF: centralize arviz_stats version handling

ShivamRajSri:
This PR introduces an internal helper to centralize ArviZ subpackage import and
version handling.

As a first step, the arviz_stats import logic in arviz/__init__.py is refactored
to use this helper, preserving existing behavior, logging, and error messages.

Dedicated tests are added for the helper. All formatting hooks were run via
pre-commit, and tests pass locally.

### DIFF
--- a/tests/test_versioning.py
+++ b/tests/test_versioning.py
@@ -1,4 +1,5 @@
 import pytest
+
 from arviz._versioning import import_arviz_subpackage
 
 


ShivamRajSri:
The CI failure on Python 3.13 was caused by formatting issues caught by pre-commit (EOF fixer + ruff).

I’ve fixed the formatting, rerun pre-commit run --all-files locally, and confirmed all tox environments now pass (3.12, 3.13, 3.14).

No functional changes were made in this update — only style/formatting fixes.

ShivamRajSri:
Thanks for the clarification, that makes sense.

I understand that the fallback logic was only intended as a temporary workaround for the release candidate, and that for the 1.0 stable release the plan is to rely on strict __version__ requirements instead of carrying compatibility code.

I’m happy to drop this refactor and keep things aligned with the long-term direction of ArviZ. Please let me know whether you’d prefer me to close this PR or make any adjustments.

aloctavodia:
Thanks, @ShivamRajSri, for your enthusiasm and willingness to help. There are many other places where you can help. Remember also to check issues for arviz-plots, arviz-stat,s and arviz-base repos.

PR #2514, opened 2026-01-16
update requirements

OriolAbril:
Release 1.8 of h5netcdf which is our default library to read netcdf files no longer installs h5py
by default to allow the option of python only environments with h5netcdf and pyfive.
Given users will most probably not be aware of that and expect an arviz install to be able
to read netcdf files out of the box, I am adding h5py explicitly to the requirements.

Don't merge yet, I still have to double check how to set things up in order to be able to release
from this branch because we have now activated trusted publishing on pypi so the old workflow
won't work anymore, I need to adapt the updated workflow to this branch.


OriolAbril:
All test failures seem to be HTTPErrors so I think we are good to go for the patch release. The 0.23.1 docs won't build, but I will remove the version from the version switcher and leave only 0.23 whose content is exactly the same.

PR #2523, opened 2026-01-27
Use system cache directory for timestamp file

schmoelder:
This PR moves the timestamp file to the system cache directory using `platformdirs` and handles empty/corrupt files gracefully.

Fixes #2520

PR #2525, opened 2026-02-04
Create parent directories for cache

schmoelder:
Hi,

I'm really sorry, but my recent PR (#2523) seems to have broken ArviZ since the parent directories appear to be missing on Windows (see [log](https://github.com/fau-advanced-separations/CADET-Process/actions/runs/21588907871/job/62463288624#step:8:89). This PR aims to fix this by adding the `parents=True` flag.

Again, I'm really sorry for causing this. :see_no_evil:


OriolAbril:
Released! No worries, I did test the changes in several different scenarios and environments in the previous PR but I only have linux computers 😅

Issue #2475, opened 2025-09-11
Maintenance: Apache license headers


maresb:
I'm doing some maintenance work, auditing the license situation and such. For Apache, [it's recommended](https://github.com/arviz-devs/arviz/blob/01c8b9454349247eed2145a27b03f9231acb412f/LICENSE#L178-L201) (but not strictly required) to add a header to all source code files. We should consider doing that here.

Note that @lucianopaz maintains [a pre-commit hook](https://github.com/lucianopaz/head_of_apache) for this purpose.

OriolAbril:
I got some resources about copyright and years. Not sure if out of scope but it seems the current best practice is not including any years in it

aloctavodia:
Not including the years sounds reasonable to me.

ellatso:
Hi! I’ll work on #2475. I’ll make sure shebang/encoding lines stay at the very top and the header is inserted right after them.
After this PR, I can follow up with `tests/` (or other directories) if you’d like.

PR #2534, opened 2026-02-20
MAINT: add Apache license headers

ellatso:
This is an infrastructure/maintenance change for the `arviz` repository.

- Add an Apache-2.0 license header template (no years) in `tools/APACHE_HEADER.txt`.
- Add a pre-commit `insert-license` hook and apply it to Python files under `src/` and `tests/`.

part of #2475.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,3 +23,15 @@ repos:
   rev: 0.4.1
   hooks:
     - id: no-print-statements
+- repo: https://github.com/Lucas-C/pre-commit-hooks
+  rev: v1.5.5
+  hooks:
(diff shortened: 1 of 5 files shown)


ellatso:
Hi! Since this PR is from a fork, it looks like the GitHub Actions workflow needs maintainer approval to run.
Could a maintainer please approve the workflow so CI can start? Thanks!

OriolAbril:
Thanks! The only comment is I updated the metadata so the issue won't be closed when merging the PR. We use issues here also as global for all python arviz libraries so we should extend this to the other repos at some point too (the other repos are larger and more active so we might need to be a bit more careful to avoid too many git conflicts)

PR #2546, opened 2026-03-03
fix: handle race condition in daily warning v0.x

### LINKED ISSUE #2530: Race condition for daily warning?
### Describe the bug

```
tests/tools/_coda/test_sccoda.py:3: in <module>
    import arviz as az
../../../.local/share/hatch/env/virtual/pertpy/xQQBcIEI/hatch-test.py3.12/lib/python3.12/site-packages/arviz/__init__.py:61: in <module>
    _warn_once_per_day()
../../../.local/share/hatch/env/virtual/pertpy/xQQBcIEI/hatch-test.py3.12/lib/python3.12/site-packages/arviz/__init__.py:58: in _warn_once_per_day
    _atomic_write_text(stamp_file, today.isoformat())
../../../.local/share/hatch/env/virtual/pertpy/xQQBcIEI/hatch-test.py3.12/lib/python3.12/site-packages/arviz/__init__.py:36: in _atomic_write_text
    tmp.replace(path)
/usr/lib/python3.12/pathlib.py:1378: in replace
    os.replace(self, target)
E   FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/.cache/arviz/daily_warning.tmp' -> '/home/runner/.cache/arviz/daily_warning'
```

### Expected behaviour

I would expect such a required folder to be silently created

### Steps to reproduce

I have a CI setup where I run pytest with parallel workers. Up until recently, this was all fine but as of a few days or very few weeks ago, my CI regularly fails with the above message. My best guess is that there's a race condition going on.

I think the folder should always be properly created or something else is likely wrong with how the warning works.

### Additional context

_No response_

player3zxcvbnm:
Hi

I’ve implemented a fix for the race condition in _warn_once_per_day as discussed in #2530.

The Problem:
On Windows, parallel execution (e.g., pytest -n) causes a FileNotFoundError. This is due to multiple processes clashing over the temporary file creation and renaming process.

The Fix:

- Replaced the "temporary file + rename" logic with an atomic lock-file pattern.
- Uses open(path, "x") to ensure only one process creates the stamp file.
- Catches FileExistsError so secondary processes skip the update rather than crashing.
- Preserves the parents=True directory creation from #2525.

Testing:
Verified on Windows 11. Parallel tests that previously failed now pass consistently.

Closes #2530

### DIFF
--- a/arviz/__init__.py
+++ b/arviz/__init__.py
@@ -1,6 +1,6 @@
 # pylint: disable=wildcard-import,invalid-name,wrong-import-position
 """ArviZ is a library for exploratory analysis of Bayesian models."""
-__version__ = "0.23.3"
+__version__ = "0.23.4"
 
 import logging
 import os


yogadityag:
Changes:

Encoding: Added encoding='utf-8' to file reads/writes.
Linting: Cleaned up unused imports and whitespace.

CI Failures:
They are all TypeErrors in the mcse diagnostics (specifically the sd method). These appear to be pre-existing environment issues.

Let me know if there's anything else I should adjust on my end.

arviz/__init__.py
```diff
@@ -47,15 +47,24 @@ def _atomic_write_text(path: Path, text: str) -> None:
...
-            "while maintaining a user-friendly interface."
-            "\nSome upcoming changes may be backward incompatible."
-            "\nFor details and migration guidance, visit: "
-            "https://python.arviz.org/en/latest/user_guide/migration_guide.html",
-            FutureWarning,
-        )
-        _atomic_write_text(stamp_file, today.isoformat())
+        try:
+            # Use a 'lock' file with today's date to ensure only one process warns
+            lock_file = stamp_file.with_suffix(f".{today}.lock")
```
OriolAbril: I am not sure accumulating these lockfiles over time is a very good idea. We don't want the warning to be too overbearing but if it shows multiple times when importing arviz in parallel it would be perfectly fine too. If it works it might be fine though, this is a backport fix on a frozen version so no need to overthink it too much

yogadityag:
Applied the suggestion for the lockfile and bumped the version to 0.23.4. Ready for testing! My apologies for the repo confusion earlier—I'm still getting oriented with the ArviZ ecosystem.

PR #2551, opened 2026-03-09
Exclude plotly templates and pio from namespace alias test

SpandanBhoiIITM:
## Description

`test_aliases` currently fails because several objects exposed through
`arviz_plots` originate from `plotly` instead of `arviz`.

These include:
- plotly templates (`arviz_*_template`)
- `pio` (`plotly.io`)

Since these objects are intentionally exposed through the ArviZ namespace,
this PR excludes them from the namespace origin check in `test_aliases`.

## Changes
- Added plotly templates and `pio` to the allowed alias list in `test_namespace.py`

## Tests
All tests pass locally:

pytest → 6 passed

### DIFF
--- a/tests/test_namespace.py
+++ b/tests/test_namespace.py
@@ -33,7 +33,16 @@ def test_info_attr():
 
 def test_aliases():
     # These are xarray aliases exposed for user convenience
-    xarray_aliases = {"from_netcdf", "from_zarr"}
+    xarray_aliases = {
+    "from_netcdf",
+    "from_zarr",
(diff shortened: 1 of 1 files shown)


OriolAbril:
https://github.com/arviz-devs/arviz/actions/runs/23090508176 was executed yesterday and ran this same tests successfully. I am not sure what you mean by "test_aliases currently fails"

aloctavodia:
Closing as not reproducible and not clear what error this is intended to fix.

PR #2552, opened 2026-03-15
Add NumPyro and ArviZ integration guide

adi-0077:
This PR adds a basic NumPyro → ArviZ conversion guide to the user guide section.

The guide includes:
- a minimal working NumPyro example
- converting results to ArviZ `InferenceData`
- basic diagnostics and visualization using ArviZ

The goal is to help NumPyro users understand how to analyze their sampling results using ArviZ tools.

PR #2553, opened 2026-03-15
Add numpyro_conversion to user guide index

adi-0077:

### DIFF
--- a/docs/source/user_guide/index.rst
+++ b/docs/source/user_guide/index.rst
@@ -9,4 +9,5 @@ User Guide
   getting_started
   migration_guide
   label_guide
+  numpyro_conversion
   ../schema/schema


OriolAbril:
same as https://github.com/arviz-devs/arviz/pull/2552. As extra note here, this should have been part of the other PR, you can commit to the same branch you opened a PR from to include multiple commits as part of the same PR. Otherwise reviewing PRs and making sure review comments are addressed is a nightmare

adi-0077:
Hi @OriolAbril, apologies for the delayed response and thanks for the clarification.

I understand now that this change should have been added to the existing PR instead of opening a separate one. I didn’t realize that it’s better to keep related changes within the same branch and PR to make the review process smoother.

I appreciate the guidance.

PR #2556, opened 2026-03-17
Add simple example to label guide for clarity

adi-0077:
This PR adds a minimal example to the label guide to help users better understand how ArviZ organizes chains, draws, and variables.

The goal is to make the documentation more beginner-friendly by providing a simple, practical example.

### DIFF
--- a/docs/source/user_guide/label_guide.rst
+++ b/docs/source/user_guide/label_guide.rst
@@ -254,3 +254,34 @@ mixtures and the like.
 
 Custom labellers
 ----------------
+Example: Understanding labels with a simple dataset
+
+Here is a minimal example showing how ArviZ organizes posterior samples:
+
+.. code-block:: python
+
+    import arviz as az
+    import numpy as np
+
+    # Create fake posterior samples
+    data = {
+        "theta": np.random.randn(4, 100)  # 4 chains, 100 draws
+    }
+
+    idata = az.from_dict(posterior=data)
+
+    print(idata)
+
+In this example:
+
+- ``chain`` represents independent MCMC chains
+- ``draw`` represents samples within each chain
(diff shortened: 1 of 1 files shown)


adi-0077:
Hi! Just adding a bit more context here.

While going through the label guide, I felt that concepts like chains and draws can be a bit abstract for new users. So I added a small example to make it more concrete and easier to understand.

The idea was to keep it simple while showing how dimensions appear in practice using `InferenceData`.

Happy to refine or expand this further based on your feedback.

aloctavodia:
I don't think this is a good example.

OriolAbril:
I am closing this PR. The content being added has nothing to do with ArviZ's labeling capabilities, is is being added in a different format which is also not executable by sphinx at build time, and it is being added at the end of the doc, in the "custom labellers" section when if an intro to xarray's dims and coords were needed, it should go at the very beginning of the page.

The label guide does need updating so I have opened https://github.com/arviz-devs/arviz/issues/2558. If you are interested in working on this please comment there. That being said, I don't think it is a very good issue to start for someone who doesn't yet know how to use ArviZ.

adi-0077:
Hi @OriolAbril and @aloctavodia, thanks for the detailed feedback.

I understand now that my example was not aligned with the purpose of the label guide, both in terms of content and placement. I also realize I didn’t follow the expected documentation structure and added something that wouldn’t integrate properly with the existing docs.

I approached this without fully understanding ArviZ’s labeling concepts and the documentation system, and I see why the change wasn’t useful.
I’ll take some time to better understand the library and how documentation is structured before contributing again, and focus on making more meaningful, well-aligned contributions going forward.

Thanks again for the guidance, I appreciate it.

PR #2557, opened 2026-03-17
Enhance API documentation with usage guidelines

adi-0077:
This PR enhances the API documentation by adding a short section explaining how users typically interact with ArviZ through the top-level `arviz` namespace.

While going through the docs, I noticed that the API structure (arviz_base, arviz_stats, arviz_plots) might be slightly confusing for new users. This addition aims to clarify that most functionality is accessed via `import arviz as az`, along with a simple example.

The goal is to improve clarity and make the documentation more beginner-friendly without changing existing content.

Happy to refine the wording or placement based on feedback.

### DIFF
--- a/docs/source/api.md
+++ b/docs/source/api.md
@@ -9,3 +9,19 @@ The complete API is described at the following pages:
 * {doc}`arviz-base API <arviz_base:api/index>`
 * {doc}`arviz-stats API <arviz_stats:api/index>`
 * {doc}`arviz-plots API <arviz_plots:api/index>`
+
+## How to use the API effectively
+
+Most users will interact with ArviZ through the top-level `arviz` namespace.
+
+For example:
+
+```python
+import arviz as az
+
+summary = az.summary(idata)
+az.plot_trace(idata)
+```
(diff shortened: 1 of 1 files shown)


adi-0077:
Hi! Adding a bit more context to this change.

While exploring the API docs, I felt that the separation into different submodules might be a bit confusing for new users at first glance. So I added a small section to clarify how most functionality is accessed through the top-level `arviz` namespace, along with a simple example.

The intention was to make the structure easier to understand without adding too much detail.

Happy to adjust the content, placement, or level of explanation based on your suggestions.

aloctavodia:
OK, this just look as spam. If you really want to contribute my recomendation is that you close these PRs, you spend some more time trying to better understand what arviz does (and how thing are done in new arviz vs the legacy arviz), then check if you can find one opened issue. Better to submit a single PR that's actually helful than several very superficial ones.

adi-0077:
Hi @aloctavodia,thanks for the honest feedback, and apologies for the delayed response as well as the noise from my PRs.

I realize I approached this the wrong way,I focused more on making multiple contributions instead of taking the time to properly understand ArviZ and ensure the changes were meaningful and aligned with the project.
I’m still new to contributing here, and I see now that I should spend more time going through the documentation and existing discussions before opening PRs.

Thanks again for the guidance, I really appreciate it.

PR #2575, opened 2026-04-20
Emit prominent `MigrationWarning` instead of breaking `InferenceData` import

michaelosthege:
The 1.0 release is not even a week old and I already have colleagues running into import errors :(

In today's case it's with cases where we have PyMC versions pinnend and the import in PyMC at the time was not yet ArviZ 1.0 compatible.

Please consider this PR which will still bombard users with prominent deprecation warnings upon import, but give library maintainers and dev-ops people some time to fix all the places that are not directly under the control of the end user.

### DIFF
--- a/src/arviz/__init__.py
+++ b/src/arviz/__init__.py
@@ -27,7 +27,7 @@ _log = logging.getLogger(__name__)
 info = ""
 
 
-class MigrationWarning(FutureWarning):
+class MigrationWarning(UserWarning, FutureWarning):
     """Warning raised when a legacy name is accessed on the ``arviz`` namespace."""
 
 
--- a/src/arviz/__init__.py
+++ b/src/arviz/__init__.py
@@ -27,8 +27,8 @@ _log = logging.getLogger(__name__)
 info = ""
 
 
-class MigrationError(RuntimeError):
-    """Error raised when a legacy name is accessed on the ``arviz`` namespace."""
+class MigrationWarning(DeprecationWarning):
+    """Warning raised when a legacy name is accessed on the ``arviz`` namespace."""
 
 
 try:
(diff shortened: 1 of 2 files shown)


src/arviz/__init__.py
```diff
@@ -27,8 +27,8 @@
 info = ""
 
 
-class MigrationError(RuntimeError):
-    """Error raised when a legacy name is accessed on the ``arviz`` namespace."""
+class MigrationWarning(DeprecationWarning):
```
aloctavodia:
```suggestion
class MigrationWarning(FutureWarning):
```
DeprecationWarning is intended for devs and is less visible than FutureWarning.

src/arviz/__init__.py
```diff
@@ -104,11 +104,17 @@ class MigrationError(RuntimeError):
...
     if name == "InferenceData":
-        raise MigrationError(
+        import warnings
+        from xarray import DataTree
+
+        warnings.warn(
             "arviz.InferenceData is no longer available on the "
             "arviz package; ArviZ now uses xarray's DataTree for the same "
-            f"role. See the migration guide: {_MIGRATION_GUIDE_URL}"
+            f"role. See the migration guide: {_MIGRATION_GUIDE_URL}",
```
aloctavodia, line 113: Should we mention here that there are many other potential breaking changes? I will update the migration guide to have a summary of changes at the top.

aloctavodia:
The 1.0 release was more than one month ago. But yeah, we can consider this PR.

aloctavodia:
The error with the docs is unrelated to this PR and it has been addressed elsewhere.

src/arviz/__init__.py
OriolAbril: I would actually argue it should be a subclass of UserWarning. It is very much a present issue and the vast majority of InferenceData use will be broken now. Only type hints, and things like isinstance(...) or calling a handful of methods without keyword arguments will continue to work.

michaelosthege: I subclassed it from both warning types; `UserWarning` first

aloctavodia (approved):

zaxtax:
I also support this PR!

PR #2578, opened 2026-04-23
Pin GitHub Actions by SHA

maresb:
Pinning the SHAs is considered best security practice against supply-chain attacks, and they will be auto-updated by dependabot PRs.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -19,11 +19,11 @@ jobs:
       attestations: write
       id-token: write
     steps:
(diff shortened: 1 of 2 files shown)


maresb:
```
TypeError: Passing a Dataset as data_vars to the Dataset constructor is not supported. Use ds.copy() to create a copy of a Dataset.
```

aloctavodia:
That error should be fixed now.

OriolAbril (approved):

PR #2592, opened 2026-06-18
Add ELLIS Finland as institutional sponsor

aloctavodia:


docs/source/index.md
```diff
@@ -234,8 +234,13 @@ We thank these institutions for generously supporting the development and mainte
 :::{grid-item}
 [![fcai_logo](_static/sponsor_fcai.png)](https://fcai.fi/)
 :::
+:::{grid-item}
+[![ellis_finland_logo](_static/sponsor_ellis_finland.png)](https://www.ellisinstitute.fi/)
```
OriolAbril, line 238: The image is not being added with the PR so it doesn't render properly

OriolAbril (commented):
There is probably a leftover trailing whitespace or something like that but other than that I think we can merge once the image is added and `tox -e check` runs

PR #2609, opened 2026-08-13
Refactor minor-version compatibility check to avoid parsing formatted…

### LINKED ISSUE #2607: Refactor minor-version compatibility check to avoid parsing formatted status output
### Describe the bug

The ArviZ metapackage currently checks compatibility between arviz_base, arviz_stats, and arviz_plots by constructing a formatted status string and then parsing that string with a regular expression.

In src/arviz/__init__.py, the compatibility check currently uses:

python
pat = re.compile(r"arviz_(base|stats|plots)\s([0-9]+\.[0-9]+)")
matches = pat.findall(info)
versions = dict(pat.findall(info))
unique_versions = set(versions.values())
The check therefore depends on the formatting of the generated status messages rather than directly inspecting the version metadata of the three packages.

This works with the current version format, but it makes the compatibility logic tightly coupled to an intermediate human-readable string. Changes to that formatting could unintentionally break the compatibility check.

Furthermore, there is no explicit handling of how version metadata containing pre-release or other PEP-440 version information should be treated by this compatibility check (the regex implicitly discards everything after major.minor).

I think the compatibility check would be more robust if it operated directly on the installed package versions rather than parsing the formatted status output.

### Expected behaviour

The compatibility check should determine whether arviz_base, arviz_stats, and arviz_plots satisfy ArviZ's intended version-compatibility policy using their version metadata directly.

The existing requirement that the ArviZ subpackages share the same minor-version line should be preserved.

The implementation should also make the intended behaviour for edge-case versions (such as release candidates or pre-releases) explicit, rather than relying on the regular expression to implicitly discard them.

### Steps to reproduce

Steps to reproduce
Note: This is an implementation robustness issue rather than a currently demonstrated runtime failure.

The behaviour can be inspected in src/arviz/__init__.py, where the versions of arviz_base, arviz_stats, and arviz_plots are first inserted into formatted status strings and are subsequently extracted using:

python
pat = re.compile(r"arviz_(base|stats|plots)\s([0-9]+\.[0-9]+)")
The extracted major.minor values are then compared. A concrete user-facing failure would only occur if there is a change in the status-string formatting or a version format that is not represented correctly by the current regular expression.

Additional context
Current main defines the ArviZ metapackage version as 1.3.0 and correctly declares the constraints in pyproject.toml:

arviz_base>=1.3.0,<1.4.0
arviz_stats[xarray]>=1.3.0,<1.4.0
arviz_plots>=1.3.0,<1.4.0
I am not proposing a specific implementation (e.g., using packaging.version) at this stage because the desired semantics for pre-release versions should be clarified first. However, a direct version-metadata comparison using an appropriate PEP-440-aware mechanism may be a more robust direction.

If the maintainers agree that this is worth addressing, I would be happy to investigate the existing tests and submit a PR for this refactor

### Additional context

_No response_

Dikshant-Neupane:
… status output

closes https://github.com/arviz-devs/arviz/issues/2607

### DIFF
--- a/tests/test_namespace.py
+++ b/tests/test_namespace.py
@@ -126,3 +126,23 @@ def test_compatible_package_versions(monkeypatch):
     assert "arviz_base 0.7.0 available" in info
     assert "arviz_stats 0.7.1 available" in info
     assert "arviz_plots 0.7.2 available" in info
+
+
+def test_compatible_package_versions_with_prerelease(monkeypatch):
+    """
+    Compatible minor versions with pre-releases should not raise an ImportError.
+    """
+
+    monkeypatch.setattr(arviz_base, "__version__", "1.3.0rc1", raising=False)
+    monkeypatch.setattr(arviz_stats, "__version__", "1.3.0", raising=False)
+    monkeypatch.setattr(arviz_plots, "__version__", "1.3.0.dev0", raising=False)
+
+    # Should not raise
+    importlib.reload(az)
+
+    # Assert info reports the patched versions correctly
+    info = az.info
+
+    assert "arviz_base 1.3.0rc1 available" in info
+    assert "arviz_stats 1.3.0 available" in info
+    assert "arviz_plots 1.3.0.dev0 available" in info
--- a/src/arviz/__init__.py
+++ b/src/arviz/__init__.py
@@ -82,17 +82,21 @@ __version__ = "1.3.0"
 
 info = f"Status information for ArviZ {__version__}\n\n{info}"
 
-pat = re.compile(r"arviz_(base|stats|plots)\s([0-9]+\.[0-9]+)")
-matches = pat.findall(info)
+from packaging.version import parse as parse_version
+import sys
 
-versions = dict(pat.findall(info))
-unique_versions = set(versions.values())
+package_versions = {
+    "base": parse_version(sys.modules["arviz_base"].__version__).release[:2],
+    "stats": parse_version(getattr(sys.modules["arviz_stats"], "__version__", "0.7.0")).release[:2],
+    "plots": parse_version(sys.modules["arviz_plots"].__version__).release[:2],
+}
 
-if len(unique_versions) > 1:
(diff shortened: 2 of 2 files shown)


tests/test_namespace.py
```diff
@@ -126,3 +126,23 @@ def test_compatible_package_versions(monkeypatch):
     assert "arviz_base 0.7.0 available" in info
     assert "arviz_stats 0.7.1 available" in info
     assert "arviz_plots 0.7.2 available" in info
+
+
+def test_compatible_package_versions_with_prerelease(monkeypatch):
```
OriolAbril, line 131: I think this test will pass on `main` so I don't see how it is protecting against a potential regression

OriolAbril (commented):
I have added the related issue to the description and gone over it, and most of the points raised there I don't think are reasons for changing this.

> The check therefore depends on the formatting of the generated status messages rather than directly inspecting the version metadata of the three packages.

Those status messages are generated lines above by **directly** checking the version attribute. In fact, the proposed changes here go as far as copying the getattr patch for the 0.7 version when accessing the version attribute of arviz_stats despite it having a todo comment about it needing to be removed

> Furthermore, there is no explicit handling of how version metadata containing pre-release or other PEP-440 version information should be treated by this compatibility check (the regex implicitly discards everything after major.minor).

The regex is explicitly only checking for the first `X.Y` part of the version, so if present, any patch indicator, release candidate info or extra info is ignored on purpose.

Also, please review what you post so comments represent what you are doing and what you mean as well as possible. Only 20 minutes passed between the issue saying "I am not proposing a specific implementation (e.g., using packaging.version) at this stage because the desired semantics for pre-release versions should be clarified first." and this PR was open which seems a bit incoherent.

---

The main (or only) improvement would be adding a test so we can't inadvertently break this check. And I think the added test can be added by iself as it will pass on main? Are there other reasons for using packaging.version if we only care about major/minor part? Are you ok with adding the test only in the PR?

PR #2610, opened 2026-08-13
chore: expand Dependabot to cover pip and pre-commit ecosystems

Dikshant-Neupane:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,17 @@ updates:
     actions:
       patterns:
         - "*"
+
+- package-ecosystem: "pip"
+  directory: "/"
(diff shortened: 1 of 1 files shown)


OriolAbril:
Looks like you mixed the PR descriptions and opened an issue with its content? Please go over the https://python.arviz.org/en/stable/contributing/pr_checklist.html and be careful with what you post. Opening the issue is perfectly ok and a good place to discuss the approach before working on a PR but its content should be adapted and relevant, and PR addressing an issue should reference that issue in their description as mentioned in the PR checklist.

I saw https://github.com/arviz-devs/arviz/pull/2611 before this one so please go over the comments there too

PR #2614, opened 2026-09-04
Reorganize HDI API change in migration guide

jordandeklerk:
Seems odd to have an entire section dedicated to the HDI API change. I think the guide flows better if we just call this out explictly in the guide itself as a core change. 

Open to suggestions about where this should live as well. Seemed appropriate to have it in the credible intervals section.


aloctavodia (approved):
I like it!

OriolAbril:
The PR that added that section was paired with this one in ArviZ-stats: https://github.com/arviz-devs/arviz-stats/pull/392. I do completely agree that a standalone section at the bottom of the page (thus outside the arviz-stats section) is something to fix, but we now have an error message in arviz-stats that will contain a broken link once we make a new release. I'll open a PR to try and keep the link active

jordandeklerk:
Oh shoot, sorry I forgot this was tied to the other PR with that error message!

PR #2615, opened 2026-09-09
Preserve explicit link from error to hdi section

OriolAbril:
Follow up to https://github.com/arviz-devs/arviz/pull/2614


OriolAbril:
The explicit target seems to be working so people clicking on the error link will go directly to the box about the hdi changes:

image

jordandeklerk (approved):
LGTM, thanks!

